PR #13, opened 2026-03-01
Build pipeline hardening (fixes 1-4)

jasonLadd:
## Summary

- **Lock files**: Generate pinned `.lock` files via `pip-compile` so Docker builds skip dependency resolution (~30s → <2s on cache hit). Dockerfiles install from single `.lock` file instead of both `requirements.txt` + variant.
- **Source hash cache busting**: `ARG SOURCE_HASH` (git short SHA) inserted before source COPY layers in both Dockerfiles. Ensures Docker never serves stale source code while preserving dependency layer cache.
- **`.dockerignore` tightening**: Exclude `sips/`, `.venv/`, `.claude/`, `.ruff_cache/`, `.benchmarks/`, `data/`, `dist/`, `node_modules/`, `secrets/`, `*.egg-info/` from build context.
- **Console failure isolation**: Console and runtime-api build failures no longer abort the pipeline — agents still get rebuilt and restarted. Failures are reported at script end with rebuild commands, and the script exits 1.

Implements fixes 1-4 from `docs/plans/build-pipeline-hardening-plan.md`. Fix 5 (skip unchanged services) is deferred.

## Test plan

- [x] Regression suite: 2474 passed, 1 skipped, 0 failures
- [ ] `docker-compose build --no-cache neo` — confirm it installs from `requirements-agent.lock`
- [ ] Build twice with source change between — dep layers should be CACHED, source layers should rebuild
- [ ] `curl -s http://localhost:8001/health` — version should match after rebuild
- [ ] Temporarily break console build, run `rebuild_and_deploy.sh all` — agents should still deploy, script should exit 1


jasonLadd (commented):
verified with a cycle

PR #72, opened 2026-04-30
feat(sip-0092 m1.1): implementation plan rename + typed acceptance schema/parser

jasonLadd:
## Summary

PR 1.1 of 3 for SIP-0092 Stage M1. Adds typed acceptance schema and parser-side validation; doesn't touch the validator yet (that's PR 1.3) or evaluator framework (PR 1.2).

Two commits, intentionally split for review legibility:

- **`a8dd313` Commit A — rename sweep (no behavior change).** `build_manifest.py → implementation_plan.py`, `BuildTaskManifest → ImplementationPlan`, `ManifestTask → PlanTask`, `ManifestSummary → PlanSummary`, config key `build_manifest → build_plan`, artifact_type `control_manifest → control_implementation_plan`, four test files renamed, all 13 importers updated. 3455 tests stay green — verifiable in isolation.
- **`1cf9ca4` Commit B — typed acceptance (the actual M1.1 work).** Adds `TypedCheck` dataclass, the `acceptance_check_spec.py` registry as single source of truth, and parser-side rejection of malformed typed criteria at plan-parse time per RC-11. 19 new tests, 3474 total green.

## What's in scope (per plan doc M1.1)

- `TypedCheck(check, params, severity, description)` frozen dataclass; canonical internal form
- `acceptance_check_spec.py` with `CheckSpec` + `CHECK_SPECS` registry — Rev 1 vocabulary entries: `endpoint_defined`, `import_present`, `field_present`, `regex_match`, `count_at_least`, `command_exit_zero`. The same registry is consumed by both the parser (today) and the M1.2 evaluator framework (next), preventing drift.
- `PlanTask.acceptance_criteria` widened to `list[str | TypedCheck]`. Prose strings still pass through unchanged.
- `from_yaml()` parses mixed prose+typed lists with the flat-YAML normalization rule (params = entry minus `{check, severity, description}`)
- Authoring-time rejections (RC-11): unknown check name, missing required param, unknown param, wrong param type, unknown severity, absolute path / `..` traversal in path-typed params, malformed entries
- `to_dict()` serializes typed checks as flat authored form so `from_yaml → to_dict → safe_dump → from_yaml` is stable identity (load-bearing for M3's canonical hashing per §6.3.6)

## What's NOT in scope (deferred)

- **Evaluator framework / static checks / safety** — PR 1.2 (`acceptance_checks.py`, the actual `endpoint_defined` etc. evaluators, command safelist, path chrooting at evaluation time)
[...]

PR #73, opened 2026-04-30
docs(sip): propose SIP-Multi-Role-Plan-Authoring; wire into SIP-0092 M1→M2 gate

jasonLadd:
## Summary

- Adds **`sips/proposed/SIP-Multi-Role-Plan-Authoring.md`** — a proposed sibling SIP that argues both SIP-0092 M1 (Max combined author/reviewer) and M2 (Neo authors, Max reviews) share a sole-broker structural property, and proposes a propose-merge model: each contributing role authors plan tasks for its domain and Max merges into the canonical plan.
- Wires the proposal into the SIP-0092 plan doc at **Gate M1 → M2** so the gate evaluation has both authoring models on the table — three paths: M2 as written, the multi-role alternative, or neither. The eval doc must record which path was chosen and why; a non-M2 winner makes M2 the spun-out SIP rather than this alternative.

## Why now

Drafted while implementing SIP-0092 M1 on #72. The window to put the alternative on the table is *before* M2 implementation begins, so the gate evaluation can choose evidence-driven rather than retroactively re-litigating after M2 ships. No code changes; no SIP number assigned. The M1 → M2 gate evaluation (`docs/plans/SIP-0092-gate-M1-evaluation.md`) is the decision point.

## Scope

- Design proposal only — no implementation, no number, no migration.
- Independent of #72 (M1.1 typed acceptance) and the upcoming framing-rename / evaluator-framework PRs.

## Test plan

- [x] `git diff main` shows only the new SIP file and the targeted plan-doc paragraph.
- [x] Plan-doc edit is in the Milestone Gates → Gate M1 → M2 section with a relative link to the proposed SIP.
- [x] Proposed SIP's self-reference to the plan doc matches the actual section name (no spurious "Plan Rev 4").
- [ ] Maintainer design review of the propose-merge model and Open Questions before any implementation work.

### DIFF
--- a/docs/plans/SIP-0092-implementation-plan-improvement-plan.md
+++ b/docs/plans/SIP-0092-implementation-plan-improvement-plan.md
@@ -580,6 +580,8 @@ Proceed to Stage M2 only when **all** of the following hold across a tracking wi
 
 If any criterion fails, **M2 is spun out as a separate proposed SIP** (`SIP-Implementation-Plan-Reviewer-Separation`) that re-litigates the design with the evidence-in-hand.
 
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
approved

jasonLadd (commented):
approved

PR #74, opened 2026-04-30
refactor(sip-0092): rename WorkloadType.PLANNING → FRAMING

jasonLadd:
## Summary

Brings the persisted workload-type vocabulary in line with the **framing vs planning** discipline established during SIP-0092 implementation: the pre-build sequence (`data.research → strategy.frame → … → governance.assess_readiness → GATE`) is the **framing** sequence, not the planning sequence. "Planning" was overloaded across the framing sequence, the SIP-0092 implementation plan artifact, and the executor task plan; "framing" disambiguates.

## Changes

**Production code (~30 lines):**
- `WorkloadType.PLANNING` → `WorkloadType.FRAMING` (value: `"planning"` → `"framing"`) in `src/squadops/cycles/models.py`
- `PLANNING_TASK_STEPS` → `FRAMING_TASK_STEPS` in `src/squadops/cycles/task_plan.py` and call sites
- `adapters/cycles/distributed_flow_executor.py`: `wt == "planning"` comparison updated
- CRP profile `workload_sequence` entries: `type: planning` → `type: framing`
- `profiles/planning.yaml` → `profiles/framing.yaml` (file rename + `name:` + `description:`)
- Module-level docstring in `capabilities/handlers/planning_tasks.py` updated

**Tests:** mechanical fixture-string update across 16 test files.

**No migration.** Initial draft included `1000_workload_type_planning_to_framing.sql`; dropped after verifying the dev DB. All 12 persisted `workload_type='planning'` runs are terminal (11 completed, 1 cancelled). Rewriting terminal records is cosmetic only and adds SQL-bug risk at runtime-api startup. Old runs keep their historical label; new runs and new CRP profiles emit `framing`.

## Out of scope (intentional)

- `NextCycleRecommendation.PLANNING` (different enum)
- `src/squadops/capabilities/handlers/planning_tasks.py` filename (broad import surface)
- Test method names containing `planning`
- Pulse-suite IDs `planning_scope_guard` / `planning_completeness` and `planning_artifact.md`

## Test plan

- [x] `./scripts/dev/run_regression_tests.sh` — 3474 passed, 1 skipped
- [x] Verified no in-flight cycles affected: all 12 persisted `workload_type='planning'` runs in dev DB are terminal; migration unnecessary.

## Operator-facing impact

Breaking change: `--request-profile planning` no longer resolves; users invoke `--request-profile framing` instead. Existing completed cycles continue to display as `planning` in run history (preserved as historical labels).

[...]

PR #75, opened 2026-05-01
feat(sip-0092 m1.2): typed-acceptance evaluator framework + six checks

jasonLadd:
## Summary

SIP-0092 Stage M1, Phase 1.2 — typed-acceptance evaluator framework. Builds on the M1.1 `CHECK_SPECS` registry by supplying the runtime evaluator that actually runs typed checks against the workspace. Two new files only; nothing in the runtime path consumes this yet (M1.3 wires it into `_validate_output_focused`).

## What lands

**`src/squadops/cycles/acceptance_checks.py`** (production):
- `CheckOutcome` dataclass with `passed/failed/skipped/error` factories.
- `BaseCheck` abstract + `@register_check(name)` decorator that binds evaluator classes to `CHECK_SPECS` entries.
- `assert_registry_complete()` runs at module import — any `CHECK_SPECS` entry without a paired evaluator raises immediately at deploy, not mid-cycle.
- Six concrete evaluators:
  - `endpoint_defined` — FastAPI route AST walk; tolerant of trailing slash; rejects malformed `methods_paths` tokens with `error` rather than silent miss.
  - `import_present` — Python AST for `.py`; `.ts/.js` returns `skipped` reason `frontend_acceptance_checks_disabled` (gated behind a follow-up flag); other extensions return `skipped` reason `unsupported_file_extension`.
  - `field_present` — Python class AST covering dataclasses and Pydantic v2 (annotated assignments + plain assignments to a `Name` target).
  - `regex_match` — size-bounded against ReDoS surface (pattern cap, content cap); no implicit `re.MULTILINE` (authors opt in via `(?m)`).
  - `count_at_least` — workspace-chrooted glob with 10k match cap; absolute and `..` globs rejected upfront.
  - `command_exit_zero` — argv-only; pattern safelist matched against the full argv (not just `argv[0]`); clean restricted env; per-command timeout clamped to ≤60s.

**Safety (RC-10 / RC-10a):**
- `_safe_resolve` rejects absolute paths, `..` traversal, and symlinks whose targets escape `workspace_root`.
- Command safelist: `python -c`, `python -m pip`, `python -m <unlisted-module>`, shell strings, and any unmatched argv produce `error` reason `command_not_in_safelist` per RC-9a — *not* `skipped`. The plan asked for something we explicitly will not run; treat as a check failure rather than silent pass.
- Restricted env strips `LD_PRELOAD` / `PYTHONPATH` / `LD_LIBRARY_PATH`; keeps `PATH`/`HOME`/locale/`TZ`.

**RC-12a stack-context handling:** checks with `requires_stack_context=True` return `skipped` reason `unsupported_stack_or_syntax` when stack is unset (not `error`). Stack is passed as a keyword to `evaluate()`; M1.3 will populate from the resolved profile or plan metadata.

**`tests/unit/cycles/test_acceptance_checks.py`** (72 new tests):
- Per-check `passed/failed/skipped/error` matrix.
- 18 parametrized command-safelist cases — allowed and rejected argv shapes per RC-10a.
- Path safety boundary across every path-taking check: `..` traversal and absolute paths both yield `error` `path_escapes_workspace`. Symlink-out-of-workspace rejected; symlink-inside allowed.
- Glob match cap and regex input cap both produce `error` outcomes.
- Registry invariants: every spec has an evaluator; `get_check` raises on unknown names.

## Out of scope (deferred to M1.3)

- Wiring into `_validate_output_focused`'s FC3 — typed-check evaluation does not yet steer self-eval or correction.
- Authoring prompt updates (telling Neo/Max about the new typed-check vocabulary).
- Stack-context plumbing — the evaluator accepts a `stack` keyword today; M1.3 is the phase that decides where it comes from (resolved profile vs plan metadata, per RC-12a).

## Why now

PR #74 (workload_type rename) merged. M1.2 is additive; it doesn't touch any runtime paths and ships under the same M1 → M2 gate as M1.1 and M1.3 (the plan doc explicitly: "Each stage's PR sequence ships under one stage's gate, not three sub-gates"). No regression-cycle requirement before merge.

## Test plan

- [x] `./scripts/dev/run_regression_tests.sh` — 3546 passed, 1 skipped (was 3474 before this PR; +72 new tests).
- [x] Module import smoke-test: `_CHECK_IMPLS` and `CHECK_SPECS` keys match exactly; `_argv_matches_safelist` rejects `python -c`, accepts `python -m py_compile <file>`, etc.
- [ ] No runtime-path integration to verify — by design. M1.3 is where typed-check outcomes start affecting cycle decisions.

### DIFF
--- a/src/squadops/cycles/acceptance_checks.py
+++ b/src/squadops/cycles/acceptance_checks.py
@@ -1,0 +1,690 @@
+"""Typed acceptance check evaluators (SIP-0092 M1.2).
+
+Builds on the M1.1 ``CHECK_SPECS`` registry in ``acceptance_check_spec.py``.
+Each spec there declares the contract; this module supplies the runtime
+evaluator. The pairing is enforced at module import: any ``CHECK_SPECS``
+entry without a matching ``_CHECK_IMPLS`` registration will raise.
+
+Outcomes follow RC-9a: ``error`` is reserved for evaluator failures
+(unsafe path, command not in safelist, regex pathological input, etc.) —
+not for application gaps. ``skipped`` is reserved for stack-context-unset
+or syntax-not-supported cases that authoring-time validation deliberately
+allowed through (RC-12 / RC-12a).
+
+This module is import-clean: nothing in the runtime path consumes it yet.
+M1.3 wires it into ``_validate_output_focused``.
+"""
+
+from __future__ import annotations
+
+import ast
+import asyncio
+import logging
+import re
+from collections.abc import Callable
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any
+
+from squadops.cycles.acceptance_check_spec import CHECK_SPECS, CheckSpec
+
+logger = logging.getLogger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# Outcome
+# ---------------------------------------------------------------------------
+
+
+@dataclass(frozen=True)
+class CheckOutcome:
+    """Result of evaluating a single typed check.
+
+    ``status`` ∈ {passed, failed, skipped, error}; see RC-9a / RC-12 for
+    the semantic distinction between ``failed`` (app gap) and ``error``
+    (evaluator failure) and ``skipped`` (intentionally not run).
+    """
+
+    status: str
+    actual: dict[str, Any] = field(default_factory=dict)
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
approved

Issue #84, opened 2026-05-02
data.analyze_failure produces too-shallow diagnosis to drive correction (plan_delta changes are wrong)


jasonLadd:
## Summary

The SIP-0086 correction loop's `data.analyze_failure` handler is producing diagnoses too thin for the downstream `governance.correction_decision` handler to author useful plan deltas. Result: corrections fire but prescribe changes that wouldn't actually fix the failure.

## Live evidence (2026-05-02, cycle `cyc_4cac11018af7`, run `run_e2933f952352`)

Bob's `builder.assemble` failed with the legacy `qa_handoff.md missing required sections` error. Correction loop fired:

- `data: data.analyze_failure` ran in **3 seconds**, generated only **79 completion tokens** (model `qwen2.5:3b-instruct` — a stack-aware downgrade, but still suspicious)
- `lead: governance.correction_decision` ran in 15 seconds, produced this plan delta:

```json
{
  "delta_id": "3ebccae8d6204670856e518d24f3b53a",
  "correction_path": "rewind",
  "trigger": "task_failure:builder.assemble",
  "failure_classification": "unknown",
  "analysis_summary": "N/A",
  "decision_rationale": "The failure analysis indicates that the qa_handoff.md document is missing required sections. To recover from this issue, it's best to rewind to the last checkpoint and retry from that point.",
  "changes": ["backend_tests", "frontend_tests", "integration_testing"],
  "affected_task_types": ["backend_tests", "frontend_tests", "integration_testing"]
}
```

## What's broken

1. **`failure_classification: "unknown"`** — Data's analyzer couldn't categorize the failure even though the rationale text plainly identifies it (missing markdown sections in qa_handoff).
2. **`analysis_summary: "N/A"`** — No structured analysis was passed forward to the lead. The lead had to derive everything from the raw failure trigger string.
3. **`changes` and `affected_task_types` are completely unrelated to the actual problem.** The failure was in `builder.assemble` (Bob produced an incomplete qa_handoff). The correction targets `backend_tests`, `frontend_tests`, `integration_testing` — none of which would fix the missing markdown sections, and none of which are even in Max's manifest as task types.
4. **The decision_rationale text correctly identifies the symptom** but the structured output doesn't reflect that understanding. The handler's outputs are *internally inconsistent*.

## Root cause hypothesis (needs investigation)

Three candidates, in priority order:

[...]

PR #85, opened 2026-05-02
chore(squad-profiles): spark-squad-with-builder all roles → qwen3.6:27b

jasonLadd:
## Summary

Uniform 27b across all 6 roles in `spark-squad-with-builder` (was: 27b for max/neo/nat/bob, 14b for eve, 7b for data). Eliminates Ollama model swapping under load and gives Eve + Data the same reasoning depth as the rest of the squad.

## Why

User-requested upgrade following observation that `data.analyze_failure` on the small model produced too-shallow analyses to drive useful corrections (see #84). Uniform 27b also means a single loaded model in Ollama for the whole cycle — no swap latency.

Trade-off: Eve + Data tasks slower per token (10.9 t/s vs 22-41 t/s prior). Acceptable since their dispatches are typically short.

## Test plan

- [ ] Restart agent containers and verify `qwen3.6:27b` is the active model in Eve + Data containers
- [ ] Run a short focused-mode cycle and confirm Eve + Data complete their tasks
- [ ] Confirm `governance.correction_decision` plan_delta diagnoses are improved when paired with the schema fix from #84

### DIFF
--- a/config/squad-profiles.yaml
+++ b/config/squad-profiles.yaml
@@ -35,14 +35,14 @@ profiles:
 
   - profile_id: spark-squad-with-builder
     name: "Spark Squad with Builder"
-    description: "6 agents optimized for DGX Spark — qwen3.6:27b for lead/dev/strat/builder, 14b for qa, 7b for data"
-    version: 3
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
approved

PR #89, opened 2026-05-02
fix(cycle-reply): recover lost agent replies via long-block consume + cache invalidation

jasonLadd:
## Summary

- Fixes the bug where `dev[3]` in `cyc_c9ca088599c0`/`run_edc1d0dc7bf4` (group_run, 2026-05-02) was marked FAILED even though Neo published its reply 21 minutes earlier — the orchestrator's reply-queue poll loop never delivered it. Three reply messages remain durably stuck in `cycle_results_run_edc1d0dc7bf4` as concrete proof.
- Tactical: replaces the per-0.5s short-iterator churn in `_publish_and_await` with `consume_blocking()` (one consumer per ~30s chunk), adds queue-cache invalidation so a stale handle bound to a closed RobustChannel is re-declared, and adds try/except recovery so a single transient `Channel was not opened` no longer kills a 30-min wait.
- Structural follow-up proposed in `sips/proposed/SIP-Reply-Channel-Subscription-Model.md`: replace chunked polling entirely with a `subscribe(queue, until, timeout)` primitive, add reply-queue TTL/cleanup (production has 18+ orphan `cycle_results_*` queues), and drain-before-retry to avoid duplicate LLM runs on lost-reply timeouts.

## Root cause (short)

`RabbitMQAdapter.consume()` opens a fresh `queue.iterator(no_ack=False)` per call with a 1.0 s timeout. Called every ~0.5 s by the orchestrator's poll loop, this churns ~3600 consumer subscriptions on one channel per 30-min wait. Per AMQP semantics a message dispatched to a tag that's about to be canceled is supposed to be redelivered, but with the orchestrator's pattern there is no other consumer ready — only a future iterator that hasn't been created yet. Replies vanish into the gap and the channel eventually wedges.

## What changed

- `src/squadops/ports/comms/queue.py` — `consume_blocking()` and `invalidate_queue()` added to the port (with default impls so existing adapters keep working).
- `adapters/comms/rabbitmq.py` — native `consume_blocking()` with one long-lived iterator per call; `_get_queue()` now detects channel-swap and re-declares; `invalidate_queue()` drops the cached handle.
- `adapters/cycles/distributed_flow_executor.py:_publish_and_await` — rewritten to call `consume_blocking()` in chunks, with try/except + invalidate-on-error.
- 6 `mock_queue` test fixtures updated to wire `consume_blocking` through `consume.side_effect` so all 30+ existing tests keep passing without modification.
- New unit tests: cache invalidation across channel-swap, defensive invalidation of unknown queues, error wrapping in `consume_blocking`, and two new `_dispatch_task` tests for transient-error recovery and use-of-blocking-consume.

## Test plan

- [x] All 3579 existing regression tests pass (`./scripts/dev/run_regression_tests.sh`).
- [x] 4 new unit tests in `tests/unit/comms/test_rabbitmq_adapter.py` — pass.
- [x] 2 new unit tests in `tests/unit/cycles/test_distributed_flow_executor.py::TestDispatchTask` — pass.
- [x] `ruff check`/`ruff format` clean (one pre-existing C901 on `execute_run` not from this PR).
- [ ] **Manual verification**: rebuild runtime-api + agent images, re-run a `group_run` cycle, confirm dev[3]-equivalent task succeeds without channel-staleness errors. (Memory: `feedback_rebuild_before_sip_test.md` — restart alone won't pick up the new code.)
- [ ] After deploy, confirm orphan messages in `cycle_results_run_edc1d0dc7bf4` get drained on the next run touching the queue, OR delete it explicitly.

## Follow-ups (in proposed SIP, not this PR)

- Replace chunked polling with `subscribe()` primitive for true single-consumer waits.
- Reply-queue TTL (`x-expires`) + explicit run-completion `delete_queue()`.
- Drain-before-retry to recover late-arriving replies instead of paying for a duplicate LLM run.
- Same anti-pattern in `entrypoint.py:_consume_messages` and `aci_executor.py:140` is unchanged here — both tolerable because their queues are rarely empty for long, but worth fixing for consistency.

### DIFF
--- a/adapters/comms/rabbitmq.py
+++ b/adapters/comms/rabbitmq.py
@@ -64,14 +64,35 @@ class RabbitMQAdapter(QueuePort):
         return queue_name
 
     async def _get_queue(self, queue_name: str) -> Queue:
-        """Get or create a queue, applying namespace if configured."""
+        """Get or create a queue, applying namespace if configured.
+
+        Cached Queue handles are bound to the channel they were declared on.
+        If the underlying channel reconnects (RobustChannel) the cached
+        handle can become stale and operations against it raise
+        ``Channel was not opened``. We track which channel each cached
+        queue is bound to and re-declare on mismatch.
+        """
         full_name = self._apply_namespace(queue_name)
-        if full_name not in self._queues:
-            await self._ensure_connection()
-            queue = await self._channel.declare_queue(full_name, durable=True)
-            self._queues[full_name] = queue
-            logger.debug(f"Declared queue: {full_name}")
-        return self._queues[full_name]
+        await self._ensure_connection()
+
+        cached = self._queues.get(full_name)
+        if cached is not None and getattr(cached, "channel", None) is self._channel:
+            return cached
+
+        queue = await self._channel.declare_queue(full_name, durable=True)
+        self._queues[full_name] = queue
+        logger.debug(f"Declared queue: {full_name}")
+        return queue
+
+    async def invalidate_queue(self, queue_name: str) -> None:
+        """Drop the cached Queue handle for ``queue_name``.
+
+        Called by long-running consume loops after a transient failure
+        (e.g. ``Channel was not opened``) so the next call re-declares
+        against the current channel.
+        """
+        full_name = self._apply_namespace(queue_name)
+        self._queues.pop(full_name, None)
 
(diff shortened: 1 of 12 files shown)


jasonLadd (commented):
approved

PR #90, opened 2026-05-02
sip(rev2): per-agent reply queues replace per-run cycle_results scoping

jasonLadd:
## Summary

Reshapes proposed SIP-Reply-Channel-Subscription-Model.md based on the design walk-through after PR #89 merged. Per-run reply queues (`cycle_results_{run_id}`) are not just leaky — they were the wrong partition key in the first place. Replacing with per-agent reply queues (`{agent_id}_results`, parallel to the existing `{agent_id}_comms`) eliminates the orphan-queue leakage class structurally instead of mitigating it with TTL + run-completion cleanup.

## What changed in the SIP

- **§1 abstract / §3 goals**: reframed around per-agent queues as the primary structural change; long-lived subscriptions become the consumer pattern that fits.
- **§5.1 (new)**: per-agent reply queue design, parallel to the dispatch queue table.
- **§5.2 `subscribe()` primitive**: shape changed from "wait for one matching message with timeout" to "register an `on_message` callback for the consumer's lifetime" — the right shape for an orchestrator-startup router.
- **§5.3 (new) `ReplyRouter`**: in-process futures keyed by `task_id`, started at runtime-api boot.
- **§5.4 executor wiring**: `_publish_and_await` becomes `register-future, publish, await future`. No while loop, no sleep, no error-recovery branch.
- **§5.5 migration**: ordered cutover — agents declare `_results` queues first (no behavior change), then runtime-api switches the dispatch metadata's `reply_queue` field.
- **§4 non-goals**: explicitly defers `amq.rabbitmq.reply-to` (cleaner still, but doesn't slot into the existing `{agent_id}_X` pattern as cleanly).
- **§7 risks**: replaced "auto-delete races" and "drain false positives" with the new ones (cross-run task_id collision, late-reply leakage, startup ordering).
- **§9 rollout**: drop drain-before-retry; add explicit step for one-shot `cycle_results_*` cleanup after soak.
- Plan doc `docs/plans/1-0-x-build-reliability-hardening-plan.md` updated to rev 4 to reflect the reshape.

## Why per-agent over per-run

| | per-run (current) | per-agent (proposed) |
|---|---|---|
| Queues over time | one per run, accumulates | fixed: one per agent (≤7) |
| Cleanup needed | yes (TTL + explicit delete) | none |
| Correlation key | task_id (already present) | task_id (already present) |
| Symmetric to dispatch design | no | yes (`{agent_id}_X`) |
| Stale-reply isolation | by queue identity | by task_id matching in router (with cross-run UUID guarantee) |

The per-run partition was giving us no isolation that `task_id` matching wasn't already providing — only lifecycle work that nobody was doing.

## Test plan

- [x] Plan doc rev bump applied.
- [x] Memory `project_consume_iterator_churn.md` updated to reflect rev 2.
- [ ] When this lands, accept the SIP and start an implementation branch off main per the standard SIP workflow.

### DIFF
--- a/docs/plans/1-0-x-build-reliability-hardening-plan.md
+++ b/docs/plans/1-0-x-build-reliability-hardening-plan.md
@@ -1,7 +1,7 @@
 # 1.0.x Build Reliability Hardening Plan
 
 **Created:** 2026-04-27
-**Updated:** 2026-05-02 (rev 3 — added runtime substrate precondition)
+**Updated:** 2026-05-02 (rev 4 — S1 reshaped to per-agent reply queues)
 **Status:** Active — #1 drafted, #2 next; substrate SIP S1 proposed
 **Scope:** SquadOps 1.0.x patch series (Spark lane); orthogonal to v1.1 work (SIP-0088+)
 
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
approved

PR #91, opened 2026-05-03
fix(parser): tolerate common LLM fenced-block format drift

jasonLadd:
## Summary

The strict \`\`\`<lang>:<path>\` fence parser silently dropped any block whose header didn't match exactly. On cyc_b7cf604aed46 this turned **8821 tokens of Bob's builder output and 4877 tokens of Eve's qa output into \"No valid fenced code blocks found\" failures** — both classified as content bugs but actually format bugs. The cycle then spent ~25 GPU minutes in correction loops chasing the wrong problem.

This PR adds three permissive resolution paths after the strict format, plus a worked-example block in the builder.assemble request template so models have a concrete pattern to imitate.

## Changes

**Parser (\`src/squadops/capabilities/handlers/fenced_parser.py\`)** — strict format stays primary; new fallbacks tried in priority order:

1. **Special filename in language slot** — \`\`\`\`Dockerfile\`\`\`, \`\`\`\`Makefile\`\`\` etc. for files whose name *is* the canonical language tag.
2. **Filename heading immediately preceding the fence** — \`### path/to/file.py\` then a plain \`\`\`\`python\`\`\` fence. Each heading labels exactly one fence (tracked via \`last_used_heading_pos\`) so subsequent unlabelled fences don't get silently attributed to a stale heading.
3. **First-line filename comment inside fence body** — \`# path/to/file.py\` or \`// path/to/file.py\` as line 1; comment line is stripped from the extracted content.

Path security checks (no traversal, no absolute, no colons) apply to all paths uniformly.

**Builder request template (\`src/squadops/prompts/request_templates/request.builder_assemble.build_assemble.md\`)** — adds a worked example showing the exact \`<language>:<filepath>\` shape with a real Dockerfile + qa_handoff.md, plus an explicit \"output that does NOT use this header will be rejected\" warning. Models are imitative; one good example beats two paragraphs of rules.

## Tests

- All 19 existing parser tests pass unchanged (strict format behavior preserved).
- 18 new tests covering: special filename in language slot, heading-before-fence (including distance limits and one-heading-per-fence), first-line comment, strict-wins-over-permissive, and a real-world reproduction of the Bob fullstack-assemble pattern.
- \`./scripts/dev/run_regression_tests.sh\` — 3597 passed, 1 skipped.
- \`ruff check\` + \`ruff format\` clean.

## What this does NOT fix

Two related issues observed on cyc_b7cf604aed46 are out of scope for this PR:

1. **Three-way conflict between request template, profile system prompt, and validator \`required_files\`** for \`fullstack_fastapi_react\` (request says \`__main__.py\` + \`Dockerfile\` + \`requirements.txt\`; profile system prompt says \`Dockerfile\` + \`docker-compose.yaml\` + \`start.sh\`; validator only requires \`Dockerfile\` + \`qa_handoff.md\`). Fix in a separate PR scoped to profile reconciliation.
2. **\`qa: validate_repair\` failed in 13ms with no LLM call** — likely a routing/inputs precondition bug separate from the parser. Worth a focused investigation.

## Test plan

- [x] 37/37 parser unit tests pass
- [x] 3597/3598 regression tests pass (1 skipped)
- [x] Lint + format clean
- [ ] After merge: rebuild runtime-api + agent images; re-run \`group_run\` with the \`build\` profile; verify Bob and Eve now produce extractable output

### DIFF
--- a/src/squadops/capabilities/handlers/fenced_parser.py
+++ b/src/squadops/capabilities/handlers/fenced_parser.py
@@ -4,6 +4,22 @@ Extracts tagged fenced code blocks from LLM responses into structured
 file records. Used by DevelopmentDevelopHandler, BuilderAssembleHandler,
 and QATestHandler to parse multi-file output.
 
+Recognized formats, in priority order (most explicit first):
+
+1. Strict tagged fence — ``` ``` <language>:<filepath>``` ```
+2. Filename in language slot for special names — ``` ``` Dockerfile``` ```,
+   ``` ``` Makefile``` ```, etc.
+3. Filename heading immediately preceding the fence — ``#`` through ``######``
+   ending in a path-like token (``models.py``, ``backend/main.py``,
+   ``Dockerfile``).
+4. First-line filename comment inside the fence — ``# path/to/file.py`` or
+   ``// path/to/file.py`` as the first line of the body. The comment line
+   is stripped from the extracted content.
+
+Models naturally drift between these formats; the parser tolerates all of
+them so a one-character format slip doesn't drop the whole task on the floor.
+The strict format is still the recommended one and round-trips losslessly.
+
 Part of SIP-Enhanced-Agent-Build-Capabilities, Phase 1.
 """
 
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
approved

PR #98, opened 2026-05-03
fix(builder): make required_files/optional_files the single source of truth (#92)

jasonLadd:
## Summary

Fixes #92 — Bob's prompt, build profile system prompt, and validator no longer disagree about what files he must produce. The validator's \`required_files\` / \`optional_files\` tuples are now the single source of truth, and the system prompt the LLM sees is composed from them via \`BuildProfile.full_system_prompt\`. The narrative \`system_prompt_template\` is forbidden from enumerating filenames; six new parametrized invariants enforce this and prevent re-introduction of drift.

This is gate-readiness block-list item **B2** in #97.

## Changes

- **`build_profiles.py`**: added \`BuildProfile.full_system_prompt\` property; cleaned each profile's narrative template to drop filenames; added \`qa_handoff.md\` to every profile's \`required_files\` (the validator already demanded it universally — the divergence was just unstated); removed duplicate \`requirements.txt\` from \`python_cli_builder.optional_files\`.
- **\`cycle_tasks.py\`**: \`BuilderAssembleHandler\` now passes \`profile.full_system_prompt\` to the LLM (was \`profile.system_prompt_template\`); the fallback inline prompt in \`_build_assembly_prompt\` no longer enumerates filenames.
- **\`request.builder_assemble.build_assemble.md\`** (v1 → v2): dropped the hard-coded file list. The build profile's \`full_system_prompt\` owns it.
- **\`test_build_profiles.py\`**: six new parametrized invariants in \`TestProfileSourceOfTruthInvariants\` (24 tests across 4 profiles).

## Live evidence this is the dominant remaining failure mode

Cycle \`cyc_000a71ca883b\` / run \`run_1eba35363a38\` (group_run, 2026-05-03, post-PR #91 / #96): Bob produced **zero** of the four files the request template asked for. Validator tripped immediately on missing \`qa_handoff.md\`. The correction loop fired, then qa.test failed (separately tracked as #93), and the run terminated.

Comparison vs. the prior failed cycle (pre-PR #91): PR #91 fixed the parser layer (Neo's 10 frontend files now land correctly). But Bob's three-sources-of-truth issue means he was being asked for one set in three slightly different ways and chose none of them consistently. This PR removes that ambiguity.

## Test plan

- [x] Updated \`test_build_profiles.py\` with 24 new parametrized invariants covering all four profiles
[...]

PR #96, opened 2026-05-03
fix(correction-loop): preserve analyze_failure outputs through to PlanDelta (#95)

jasonLadd:
## Summary

- Fixes #95 — the correction protocol's `_run_correction_protocol` reused a single `correction_outputs` variable across both `CORRECTION_TASK_STEPS` iterations, so the analyzer's `classification`/`analysis_summary` were silently overwritten by the lead's correction-decision outputs (which don't carry those fields). Every plan_delta in production had defaulted `failure_classification: "unknown"` and `analysis_summary: "N/A"`, even after PR #88's Pydantic validation made the analyzer reliable.
- Each correction step's outputs are now captured in their own named variable (`analysis_outputs`, `decision_outputs`); each PlanDelta field is read from the handler that owns it.
- Patch-path repair dispatch now passes `decision_outputs` as `correction_decision` (semantically what it always meant — same value as before, just renamed).

## Live evidence

Cycle `cyc_000a71ca883b` / run `run_1eba35363a38` (group_run, 2026-05-03). Both plan_deltas:

```
delta_0: failure_classification="unknown", analysis_summary="N/A"
delta_1: failure_classification="unknown", analysis_summary="N/A"
```

Despite `decision_rationale` text in both deltas explicitly referencing the analyzer's diagnosis ("the failure analysis indicates…"), confirming `data.analyze_failure` ran successfully and the lead consumed its output — the executor just dropped it before storing the delta.

## Test plan

- [x] Updated `tests/unit/cycles/test_correction_protocol.py::TestPlanDelta::test_plan_delta_stored` to mirror prod by removing `classification`/`analysis_summary` from the mocked `governance.correction_decision` outputs, and to assert on delta CONTENT not just storage.
- [x] Verified the new test fails on `main` with the exact assertion: `'unknown' == 'work_product'` — confirms it's a real regression test.
- [x] Full regression suite passes: 3597 passed, 1 pre-existing skip.

## References

- Issue: #95
- Related already-merged: PR #88 (issue #84) — Pydantic validation; PR #91 — parser tolerance
- Code touched: `adapters/cycles/distributed_flow_executor.py:1955-2070`, `tests/unit/cycles/test_correction_protocol.py`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #95, opened 2026-05-03
Correction loop drops classification/analysis_summary before PlanDelta is built — re-introduces 'unknown'/'N/A' even after #84


jasonLadd:
## Summary

Even after PR #88 (issue #84) added Pydantic validation to `data.analyze_failure`, plan_deltas in real cycles still show `failure_classification: "unknown"` and `analysis_summary: "N/A"`. The analyzer is producing valid output — the executor throws it away before the PlanDelta is constructed.

## Live evidence (post-PR #88, post-PR #91)

Cycle `cyc_000a71ca883b` / run `run_1eba35363a38` (group_run, 2026-05-03), both plan_deltas:

```json
// delta_0 (patch correction after builder.assemble failure)
{
  "failure_classification": "unknown",
  "analysis_summary": "N/A",
  "decision_rationale": "The failure is due to the missing `qa_handoff.md` file...",
  "changes": ["QA", "Documentation"],
  ...
}

// delta_1 (rewind correction after qa.test failure)
{
  "failure_classification": "unknown",
  "analysis_summary": "N/A",
  "decision_rationale": "The failure analysis indicates that the team needs to replan...",
  ...
}
```

Note: `decision_rationale` references "the failure analysis" — meaning `data.analyze_failure` did run successfully and the lead saw its output. But the structured fields don't survive into the delta.

## Root cause

`adapters/cycles/distributed_flow_executor.py:_run_correction_protocol` reuses a single `correction_outputs` variable across both steps of `CORRECTION_TASK_STEPS`:

```python
# adapters/cycles/distributed_flow_executor.py:1955
correction_outputs: dict[str, Any] = {}

for step_idx, (task_type, role) in enumerate(CORRECTION_TASK_STEPS):
    # ... dispatch task_type (analyze_failure, then correction_decision) ...

    # adapters/cycles/distributed_flow_executor.py:2039-2041
    correction_outputs = {
        k: v for k, v in (corr_result.outputs or {}).items() if k != "artifacts"
    }
```

After iteration 0 (`data.analyze_failure`), `correction_outputs` contains:
- `classification`, `analysis_summary`, `contributing_factors` (per `analyze_failure.py:201-216`)

After iteration 1 (`governance.correction_decision`), it is **reassigned** to:
- `correction_path`, `decision_rationale`, `affected_task_types` (per `correction_decision.py:140-155`)

The analyzer's fields are gone. Then at line 2064-2065:

```python
delta = PlanDelta(
[...]

Issue #92, opened 2026-05-03
Reconcile fullstack_fastapi_react: three sources of truth for required files conflict


jasonLadd:
## Background

Surfaced during cyc_b7cf604aed46 (group_run, 2026-05-02), the diagnostic for builder[4] / qa[6] failures that led to PR #91. PR #91 fixes the *parser* side. This issue tracks the *prompt/validator* side that PR #91 explicitly leaves out of scope.

## Problem

For \`build_profile=fullstack_fastapi_react\`, three different files independently declare what Bob (\`builder.assemble\`) must produce, and they disagree:

| Source | Says builder must produce |
|---|---|
| \`src/squadops/prompts/request_templates/request.builder_assemble.build_assemble.md\` | \`__main__.py\`, \`Dockerfile\`, \`requirements.txt\`, \`qa_handoff.md\` |
| \`src/squadops/capabilities/handlers/build_profiles.py\` system prompt for \`fullstack_fastapi_react\` | Multi-stage \`Dockerfile\`, \`docker-compose.yaml\`, \`start.sh\`, CORS notes, \`qa_handoff.md\` |
| \`src/squadops/capabilities/handlers/build_profiles.py\` \`required_files\` (validator) | \`Dockerfile\`, \`qa_handoff.md\` only |

Bob receives **both prompt sources concurrently** (request template stacked with the profile-specific system prompt), so the model sees three conflicting expectations. Even if it produces a coherent set, that set may not satisfy the strictest expectation in the stack.

The validator is the loosest of the three (only \`Dockerfile\` + \`qa_handoff.md\`), which means a lot of the prompt instructions are aspirational rather than enforced. Inverse case: \`python_cli_builder\` validator requires \`__main__.py\` which the fullstack profile prompt doesn't ask for.

## Why this matters

- Fullstack assemble runs are 8K–12K-token LLM calls (~14 min on qwen3.6:27b on GB10). A misaligned prompt costs real GPU minutes per cycle.
- The validator's \`required_files\` is the contract Bob is actually graded against. Telling the model to produce things that aren't graded — and not telling it to produce the things that are — wastes tokens and increases failure variance.
- The same shape exists for \`qa.test\` handler (validator vs. \`request.qa_test.test_validate.md\` vs. system prompt) — should be checked in the same pass.

## Scope

For each \`BuildProfile\` in \`build_profiles.py\`:

[...]

PR #99, opened 2026-05-03
fix(bootstrap): register QAValidateRepairHandler — closes #93 (qa.validate_repair 13ms failure)

### LINKED ISSUE #93: qa.validate_repair task fails in 13ms with no LLM call — pre-LLM precondition bug
## Background

Surfaced during cyc_b7cf604aed46 (group_run, 2026-05-02). The implementation run hit builder[4] (a content failure), the correction loop fired (data.analyze_failure → governance.correction_decision → development.repair → qa.validate_repair), and qa.validate_repair failed in **13 milliseconds** without ever issuing an LLM call.

## Evidence

From \`task_run\` for the failed task (\`922ccef3-c2d3-42e4-8dbb-9e0f1f53f6b0\`):

\`\`\`
state_type: FAILED
run_count:  1
start_time: 2026-05-03 01:09:32.079192+00
end_time:   2026-05-03 01:09:32.092361+00
total_run_time: 00:00:00.013169
\`\`\`

State transitions: \`PENDING → RUNNING → FAILED\`, all within 13 ms.

From \`log\` for that task_run_id:

\`\`\`
20  task_submitted
20  executing_capability
20  Dispatched task repair-run_6502a379-00-qa.validate_repair (qa.validate_repair) to eve_comms, awaiting reply on cycle_results_run_6502a3793efd
\`\`\`

That is the entirety of the log. No \`task_heartbeat\`, no \`LLM ... completed\`, no \`handler_failed\` or \`handler_succeeded\`. The dispatch fired but no result came back, and yet the task transitioned to FAILED before the first 30s heartbeat would have ticked.

This rules out:

- Channel/queue issues (PR #89 fix is live and zero \`Channel was not opened\` warnings appeared in this whole 90-min run)
- Eve agent failure (Eve never got a chance to process — 13 ms is too short for the dispatch to even reach the comms queue and round-trip)
- LLM timeout (no LLM call was made)

## Likely causes

This is a *pre-LLM* failure. Most likely candidates:

1. **Capability routing failure** — \`qa.validate_repair\` may not be a registered capability_id, or its handler isn't wired up. The dispatch line shows it was sent to \`eve_comms\` but maybe the orchestrator immediately gave up.
2. **Inputs validation precondition** — the handler's \`validate_inputs()\` may have failed synchronously before the LLM call, returning a typed error that the executor treats as immediate failure.
3. **Correction-loop wiring bug** — the correction protocol may be constructing an envelope with a missing required field that the qa.validate_repair handler bails on.
4. **TaskOutcome auto-classification** — if data.analyze_failure or governance.correction_decision returned a verdict that maps to \"don't bother validating,\" the qa step might be short-circuited.

## What to investigate

1. \`grep -rn 'qa.validate_repair' src/ adapters/\` to find handler registration and routing.
2. Inspect the correction-loop control flow in \`adapters/cycles/distributed_flow_executor.py\` (\`_handle_task_outcome\` and surrounding) — find where qa.validate_repair is dispatched and what could make it fail before reaching the agent.
3. Check whether the correction loop intentionally *skips* validation when \`development.repair\` returns \"no change needed\" (that 25-second \`development.repair\` completion in the same cycle is suspicious and may be related — see compounding observation below).

## Compounding observation

In the same cycle, \`dev: development.repair\` completed in 25 seconds. That's barely enough wall time for an LLM call to complete on qwen3.6:27b (10 t/s would give only ~250 tokens). Either:

- The repair handler correctly diagnosed \"no code changes needed\" and returned a small \"nothing to do\" payload (plausible — the failure was Bob's output format, not the source code), OR
- The repair handler returned a stub/skip without running the LLM.

Either way, if repair returned \"nothing changed,\" then qa.validate_repair has nothing to validate — and the 13ms failure may be the validator correctly detecting \"no repair occurred, cannot validate.\" If so the bug is in how the correction loop responds to a no-op repair, not in qa.validate_repair itself.

## Reproducer

Trigger any builder.assemble or qa.test failure where the failure is a *format* failure rather than a code failure. The correction loop will fire and qa.validate_repair will likely fail the same way.

## References

- Failing cycle: \`cyc_b7cf604aed46\` / \`run_6502a3793efd\`, 2026-05-02
- Failed task_run id: \`922ccef3-c2d3-42e4-8dbb-9e0f1f53f6b0\`
- Related: parser tolerance fix PR #91 (would have prevented the upstream builder[4] failure that triggered this code path in the first place)

jasonLadd:
## Summary

Fixes #93. Root cause confirmed: \`QAValidateRepairHandler\` (in \`src/squadops/capabilities/handlers/impl/repair_handlers.py\`) was never imported or added to \`HANDLER_CONFIGS\` in \`src/squadops/bootstrap/handlers.py\`. When the correction protocol's REPAIR_TASK_STEPS dispatched \`qa.validate_repair\` to eve, eve's HandlerExecutor hit \`HandlerNotFoundError\` at \`handler_executor.py:100\` and returned a FAILED TaskResult in ~13 ms with no LLM call — exactly the symptom #93 reported.

Of the three root-cause candidates listed in #93, this is **candidate 1: capability routing failure (handler not registered)**.

This is gate-readiness block-list item **B3** in #97.

## Changes

- **\`bootstrap/handlers.py\`**: import \`QAValidateRepairHandler\` from \`impl/repair_handlers\` and add \`(QAValidateRepairHandler, ("qa",))\` to \`HANDLER_CONFIGS\`. A comment flags the parallel \`DevelopmentRepairHandler\` duplicate (see Follow-ups below).
- **\`tests/unit/agent_foundation/bootstrap/test_bootstrap.py\`**: two new tests
  - \`test_correction_protocol_capabilities_registered\` iterates \`CORRECTION_TASK_STEPS\` + \`REPAIR_TASK_STEPS\` + \`WRAPUP_TASK_STEPS\` and asserts every dispatched capability is in the registry. This generalizes the regression so the same wiring gap can't recur for any future step.
  - \`test_qa_validate_repair_registered_for_qa_role\` asserts the capability is reachable specifically via the qa role.

## Verified

- Pre-fix the new tests fail with a perfectly diagnostic message:

  > AssertionError: Capabilities dispatched by the executor are not registered in bootstrap.handlers: ['qa.validate_repair']. Add them to HANDLER_CONFIGS or remove them from the corresponding *_TASK_STEPS list.

- Full regression suite passes: 3623 passed (+2 new), 1 pre-existing skip.
- \`ruff check\` clean for changed files.

## Follow-ups (NOT in this PR)

There are **two** \`DevelopmentRepairHandler\` classes that share \`_capability_id = "development.repair"\`:

| File | Parent | Currently registered? |
|---|---|---|
| \`handlers/repair_tasks.py\` | \`_RepairTaskHandler\` (SIP-0070 pulse-check) | Yes — wins in registry |
| \`handlers/impl/repair_handlers.py\` | \`_CycleTaskHandler\` (SIP-0079 correction-loop) | No |

This explains the \"compounding observation\" in #93: \`development.repair\` completed in 25 s during cyc_b7cf604aed46, suspiciously fast for an LLM repair on qwen3.6:27b. The correction loop dispatched it, but the pulse-check handler ran (it expects \`verification_context\` from the pulse loop and falls back to a minimal prompt without it). I left a comment in \`HANDLER_CONFIGS\` flagging this; deciding which class should own \`development.repair\` for the correction loop (and removing the duplicate) belongs in a separate, more involved PR rather than expanding this one.

## References

- Issue: #93
- Meta tracking: #97 (B3 of the gate-readiness block list — only B4 \"validation profile\" remains after this lands)
- Related already-merged: PR #91 (parser tolerance), PR #96 (correction-loop output preservation, #95), PR #98 (builder source-of-truth, #92)
- Code touched: \`src/squadops/bootstrap/handlers.py\`, \`tests/unit/agent_foundation/bootstrap/test_bootstrap.py\`

### DIFF
--- a/src/squadops/bootstrap/handlers.py
+++ b/src/squadops/bootstrap/handlers.py
@@ -44,6 +44,9 @@ from squadops.capabilities.handlers.impl.correction_decision import (
 from squadops.capabilities.handlers.impl.establish_contract import (
     GovernanceEstablishContractHandler,
 )
+from squadops.capabilities.handlers.impl.repair_handlers import (
+    QAValidateRepairHandler,
+)
 from squadops.capabilities.handlers.planning_tasks import (
     DataResearchContextHandler,
     DevelopmentDesignPlanHandler,
@@ -119,6 +122,15 @@ HANDLER_CONFIGS: list[tuple[type[CapabilityHandler], tuple[str, ...]]] = [
     (GovernanceEstablishContractHandler, ("lead",)),
     (DataAnalyzeFailureHandler, ("data",)),
     (GovernanceCorrectionDecisionHandler, ("lead",)),
+    # Correction-loop repair validator (SIP-0079 §7.7).
+    # NOTE: `development.repair` is registered above via the SIP-0070
+    # pulse-check version (handlers.repair_tasks.DevelopmentRepairHandler)
+    # rather than the cycle-task version in impl/repair_handlers.py.
+    # Both classes share `_capability_id = "development.repair"`; the
+    # pulse-check one wins because it is imported and registered above.
+    # Issue #93 follow-up: decide which implementation should own
+    # `development.repair` for the correction loop and remove the duplicate.
+    (QAValidateRepairHandler, ("qa",)),
     # Planning handlers (SIP-0078: Planning Workload Protocol)
     (DataResearchContextHandler, ("data",)),
     (StrategyFrameObjectiveHandler, ("strat",)),
--- a/tests/unit/agent_foundation/bootstrap/test_bootstrap.py
+++ b/tests/unit/agent_foundation/bootstrap/test_bootstrap.py
@@ -261,6 +261,41 @@ class TestHandlerBootstrap:
         qa_caps = registry.list_by_role("qa")
         assert "qa.test_execution" in qa_caps
 
+    def test_correction_protocol_capabilities_registered(self):
+        """Issue #93: every capability the executor's correction protocol
+        dispatches MUST be in the bootstrap registry. An unregistered
+        capability fails fast with HandlerNotFoundError (~13ms, no LLM
+        call), which masquerades as a queue/agent issue. Catch the
+        wiring gap at startup-test time, not in production cycles.
+        """
+        from squadops.cycles.task_plan import (
+            CORRECTION_TASK_STEPS,
+            REPAIR_TASK_STEPS,
+            WRAPUP_TASK_STEPS,
+        )
+
+        registry = create_handler_registry()
+        capabilities = set(registry.list_capabilities())
+
+        all_dispatched: list[tuple[str, str]] = (
+            CORRECTION_TASK_STEPS + REPAIR_TASK_STEPS + WRAPUP_TASK_STEPS
+        )
+        missing = [cap for cap, _role in all_dispatched if cap not in capabilities]
+        assert not missing, (
+            f"Capabilities dispatched by the executor are not registered "
+            f"in bootstrap.handlers: {missing}. Add them to HANDLER_CONFIGS "
+            f"or remove them from the corresponding *_TASK_STEPS list."
+        )
+
+    def test_qa_validate_repair_registered_for_qa_role(self):
+        """Issue #93 regression: qa.validate_repair must be available to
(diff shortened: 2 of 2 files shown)


jasonLadd (commented):
approved

PR #103, opened 2026-05-03
fix(builder-prompt): mark qa_handoff section list non-negotiable + worked skeleton

jasonLadd:
## Summary

Cycle 1 of the SIP-0092 M1 gate batch (cyc_11367982fd06, 2026-05-03) failed because Bob's qa_handoff.md was missing one of the validator's three required sections (\`## Expected Behavior\`). Inspection showed the failure wasn't model laxity — it was a contract drift: the plan author (Neo) wrote a typed acceptance criterion asking for sections \`## How to Run|## How to Test|## Implemented Scope|## Known Limitations\` (no Expected Behavior), and Bob followed the more specific user-prompt task description over the system prompt's section list. Validator rejected, lead chose rewind, run terminated after one correction.

Tighten the system prompt so the validator's section list cannot be quietly overridden:

- Mark the section list **NON-NEGOTIABLE** in the heading
- State that headings must appear exactly as written, with the leading \`## \` and exact casing
- Note that the validator does literal substring matching; paraphrased headings will be rejected
- Explicitly permit task-requested additional sections, on top of the required ones
- Show a worked qa_handoff.md skeleton with the exact heading text in a copy-pasteable fenced block

## Why this and not validator tuning

The validator's looseness isn't the bug; the prompt drift is. Loosening the validator (e.g. accepting \`## Run\` for \`## How to Run\`) would hide future drift. Tightening Bob's prompt forces the LLM to satisfy the contract that already exists.

## Why not the deeper architectural fix instead

The cleanest end-state is to pass the build profile's \`qa_handoff_expectations\` into the planning handler so Neo's typed checks always include the validator's sections — eliminating the drift instead of patching around it. That's a bigger plumbing change (touches the planning handler's input). This PR is the faster first lever; if cycle 2 still fails on a related issue, the architectural fix becomes the next move.

## Test plan

- [x] 4 new parametrized invariants (\`test_qa_handoff_sections_marked_non_negotiable\`, one per profile) assert on:
  - \`NON-NEGOTIABLE\` marker in the heading
  - \`mandatory\` framing language (lowercased substring check)
  - \`additional sections\` clause permitting on-top extras (lowercased substring check)
  - presence of the \`\`\`markdown:qa_handoff.md\` worked skeleton example
- [x] Existing \`test_narrative_template_does_not_enumerate_filenames\` still passes — the worked skeleton uses backticked code-block tokens that the existing invariant strips before scanning
- [x] All 66 build_profile tests pass (was 62)
[...]

PR #104, opened 2026-05-03
fix(correction-loop): persist repair source files + route by failed task type

jasonLadd:
## Summary

Two substrate bugs surfaced in `cyc_4178f25a0dff` (the second SIP-0092 validation cycle, 2026-05-03) that confounded the M1 → M2 gate signal by silently losing repair work. Both fixed here.

- **Repair handlers swallowed source code into a single markdown doc.** `DevelopmentCorrectionRepairHandler` did not override `handle()`, so the base wrapped the entire LLM response as `repair_output.md` regardless of fenced source files. The repair LLM ran twice in the failing cycle (44.5 t/s, 803 tokens; 42.5 t/s, 1263 tokens) but no `backend/main.py` or `frontend/src/pages/*` ever landed in artifact storage — repaired files dropped on the floor inside an inert markdown artifact.
- **Repair routing ignored the failed task's type.** The correction loop hardcoded `for ... in REPAIR_TASK_STEPS` (always the dev pair). The LLM-emitted `affected_task_types: [\"QA Handoff\"]` was free-text and unvalidated, so a failed `builder.assemble` silently routed to Neo (dev), who had no context for qa_handoff/packaging output and ignored the assignment. Bob never received a repair task.

## What changed

- `_CycleTaskHandler` gains a `_build_artifacts_from_content` hook (default keeps the existing single-document wrap so narrative handlers are unaffected).
- `DevelopmentCorrectionRepairHandler` and the **new** `BuilderAssembleRepairHandler` override the hook to extract per-file artifacts via `extract_fenced_files`, with a markdown fallback so narrative-only repairs (\"no code change needed, root cause was X\") are still preserved.
- New `repair_steps_for(failed_task_type)` lookup in `cycles/task_plan.py`. The executor now calls it with `envelope.task_type` (authoritative) instead of looping the static `REPAIR_TASK_STEPS`. `builder.assemble` → `builder.assemble_repair` + `qa.validate_repair`. Unknown task types fall back to the existing dev pair.
- `BuilderAssembleRepairHandler` registered in `bootstrap/handlers.py` for the `builder` role so Bob picks up `builder.assemble_repair` messages.

## Why this matters for SIP-0092 gate evidence

These two bugs were inflating the apparent failure rate of the correction loop in the validation-profile cycles — repaired files weren't landing, and the wrong agent was answering the page. Without these fixes, the M1 → M2 gate batch is measuring substrate noise rather than M1's typed-acceptance behavior. PR #103's qa_handoff prompt hardening can only help if the resulting repair work actually reaches Bob and the resulting files actually persist.

## Test plan

- [x] New: `test_dev_repair_extracts_fenced_code_into_per_file_artifacts` — regression for the markdown-swallow bug
- [x] New: `test_dev_repair_falls_back_to_markdown_when_no_fenced_blocks` — fallback preserves narrative-only output
- [x] New: `test_builder_assemble_repair_extracts_fenced_files` — covers the new handler end-to-end
- [x] New: `TestRepairStepsFor` — three tests covering dev / builder / unknown failed types
- [x] Full regression suite: **3645 passed, 1 skipped, 0 failures**
- [ ] Manual verification: rebuild agent + runtime-api images, kick off a fresh validation cycle on `group_run`, confirm a forced builder.assemble failure routes to Bob and a forced dev failure persists fenced files into artifact storage

### DIFF
--- a/adapters/cycles/distributed_flow_executor.py
+++ b/adapters/cycles/distributed_flow_executor.py
@@ -1929,7 +1929,7 @@ class DistributedFlowExecutor(FlowExecutionPort):
         """
         from uuid import uuid4
 
-        from squadops.cycles.task_plan import CORRECTION_TASK_STEPS, REPAIR_TASK_STEPS
+        from squadops.cycles.task_plan import CORRECTION_TASK_STEPS, repair_steps_for
 
         # 1. Emit CORRECTION_INITIATED
         self._cycle_event_bus.emit(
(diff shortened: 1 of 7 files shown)


jasonLadd (commented):
approved

PR #106, opened 2026-05-03
fix(correction-loop): enrich failure_evidence with validation context (#84)

### LINKED ISSUE #84: data.analyze_failure produces too-shallow diagnosis to drive correction (plan_delta changes are wrong)

jasonLadd:
Closes #84.

## Summary

The structural half of #84 (Pydantic-validated FailureAnalysis output) landed in 93c1694, but Data was still reasoning from impoverished inputs. This PR is the data-flow half: hand Data the structured signals the failed handler already produces, so the analysis is concrete enough for downstream correction-decision to pick patch vs rewind correctly.

**Live evidence motivating this PR:** \`cyc_d1c1a259c983\` (cycle 3 of the SIP-0092 gate batch) produced a structurally clean analysis (\`'missing qa_handoff sections'\`) and the correction LLM *still* picked rewind over patch — because \`failure_evidence\` carried only \`error\` + \`outcome_class\`. No way for Data (or downstream Max) to see that this was a content-fix-suffices situation rather than a scope-too-wrong one.

## What changed

- Pull through the failed handler's structured outputs into \`failure_evidence\`:
  - \`validation_result\` (passed/summary/missing_components/checks) — already populated by handlers when \`output_validation\` is on (\`cycle_tasks.py:1615\`); executor was just dropping it
  - \`preliminary_failure_classification\` — handler's own call (e.g. \`WORK_PRODUCT\` for typed-check rejections); Data may corroborate or override
  - \`rejected_artifacts\` — name/type/size + first 1500 chars of each emitted artifact, so Data can compare what was produced against what the failing checks expected
- Update analyze_failure system prompt to explicitly direct the LLM at these fields and to distinguish patchable content failures (\`WORK_PRODUCT\`) from structural failures (\`ALIGNMENT\`) — that's the distinction downstream correction-decision uses to choose patch vs rewind
- Extract \`_build_failure_evidence\` as a static helper so the dict construction is unit-testable without mocking the whole executor

## Out of scope (deliberate)

- **Bumping the data role's model.** The full-squad profile uses qwen2.5:3b for data; the spark-squad-with-builder profile already uses qwen3.6:27b. Issue #84's suggested fix #1 was a config change, not a code change — it can be done in \`config/squad-profiles.yaml\` independently. This PR keeps to the data-flow fix that benefits all squad profiles.

## Test plan

- [x] New: \`test_enriched_failure_evidence_appears_in_user_prompt\` — verifies enriched fields reach the LLM as identifiable text (validation_result keys, check names, snippet content)
- [x] New: \`TestBuildFailureEvidence\` (3 tests) — validation_result passthrough, snippet truncation at 1500 chars, empty-outputs safety
- [x] Full regression: **3649 passed, 1 skipped, 0 failures**
- [x] Ruff clean (1 pre-existing C901 on \`execute_run\` unrelated)
- [ ] Manual cycle verification: rebuild + run cycle 4 of the gate batch, confirm correction-decision picks \`patch\` (not \`rewind\`) on a builder.assemble qa_handoff failure — and confirm PR #104's \`builder.assemble_repair\` routing is finally exercised end-to-end

### DIFF
--- a/adapters/cycles/distributed_flow_executor.py
+++ b/adapters/cycles/distributed_flow_executor.py
@@ -1906,6 +1906,62 @@ class DistributedFlowExecutor(FlowExecutionPort):
     # Correction protocol
     # ------------------------------------------------------------------
 
+    @staticmethod
+    def _build_failure_evidence(
+        envelope: TaskEnvelope,
+        result: TaskResult,
+        *,
+        prior_plan_deltas_count: int,
+    ) -> dict[str, Any]:
+        """Assemble the failure-evidence payload handed to data.analyze_failure.
+
+        Issue #84 follow-up: Data was previously handed only `error` +
+        `outcome_class` and had to guess at the failure shape; downstream
+        correction-decision then picked rewind on content failures because
+        it had no indication that a patch would suffice. Pull through the
+        failed handler's structured `validation_result` + preliminary
+        `failure_classification` + rejected artifact summaries so the
+        analyzer reasons about concrete checks instead of free-text
+        error strings. Each rejected-artifact content snippet is capped
+        at 1500 chars so a multi-file failure doesn't bloat the prompt.
+        """
+        result_outputs = result.outputs or {}
+        validation_result = result_outputs.get("validation_result") or {}
+        rejected_artifacts: list[dict[str, Any]] = []
+        for art in result_outputs.get("artifacts", []) or []:
+            content = art.get("content", "")
+            if isinstance(content, str):
+                size = len(content)
+                snippet = content[:1500]
+            else:
+                size = 0
+                snippet = ""
+            rejected_artifacts.append(
+                {
+                    "name": art.get("name", ""),
+                    "type": art.get("type", ""),
+                    "size": size,
+                    "content_snippet": snippet,
+                }
+            )
+        return {
+            "failed_task_id": envelope.task_id,
+            "failed_task_type": envelope.task_type,
+            "error": result.error or "",
+            "outcome_class": result_outputs.get("outcome_class", ""),
+            "preliminary_failure_classification": result_outputs.get(
+                "failure_classification", ""
+            ),
+            "validation_result": {
+                "passed": validation_result.get("passed"),
+                "summary": validation_result.get("summary", ""),
+                "missing_components": validation_result.get("missing_components", []),
+                "checks": validation_result.get("checks", []),
+            },
+            "rejected_artifacts": rejected_artifacts,
+            "prior_plan_deltas_count": prior_plan_deltas_count,
+        }
+
     async def _run_correction_protocol(
         self,
         run_id: str,
@@ -1945,12 +2001,9 @@ class DistributedFlowExecutor(FlowExecutionPort):
         )
 
         # 2. Build correction task envelopes (deterministic IDs)
-        failure_evidence = {
-            "failed_task_id": envelope.task_id,
-            "failed_task_type": envelope.task_type,
-            "error": result.error or "",
-            "outcome_class": (result.outputs or {}).get("outcome_class", ""),
-        }
+        failure_evidence = self._build_failure_evidence(
+            envelope, result, prior_plan_deltas_count=len(plan_delta_refs)
+        )
 
         # Issue #95: capture each correction step's outputs in its own variable
         # so the analyzer's classification/analysis_summary survive past the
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
approved

PR #108, opened 2026-05-03
fix(builder): scope required_files by task expected_artifacts (#107)

jasonLadd:
Closes #107.

## Summary

When framing decomposes builder work into multiple tasks (one for manifests, one for qa_handoff documentation), the profile-level \`required_files\` used to override the framing decomposition: every builder task was forced to redundantly emit the full set, exceeding the per-call token budget. This PR makes the framing decomposition load-bearing — the active task's \`expected_artifacts\` is the source of truth, with profile defaults as the fallback for tasks framing didn't decompose.

## Live evidence

\`cyc_d1c1a259c983\` task 8 (\"Dependency manifests, Vite proxy & start scripts\"): the \`fullstack_fastapi_react\` profile requires \`qa_handoff.md\`, so Bob tried to emit manifests AND a full qa_handoff in one shot, hit the 8192 clamp at 8435 completion tokens, qa_handoff was incomplete, validator rejected, run failed.

The same coupling caused the qa_handoff regex rejection across cycles 1, 2, and 3 of the SIP-0092 gate batch — masked as M1 typed-acceptance signal but really substrate. Until this lands, the gate criterion #2 tally (typed-acceptance changed an outcome) is over-credited from substrate noise.

## What changed

- \`BuildProfile.system_prompt_for_files(scope)\` produces a prompt scoped to the active task's required files. The qa_handoff NON-NEGOTIABLE section block appears only when \`qa_handoff.md\` is in scope. \`full_system_prompt\` becomes a thin wrapper passing \`None\` so both paths stay in sync.
- \`_validate_builder_output\` accepts an optional \`task_required_files\` parameter; when provided, validates against those instead of \`profile.required_files\` and skips the qa_handoff section check when qa_handoff.md is out of scope. Empty/\`None\` falls back to profile defaults — preserves single-task builder behavior.
- \`BuilderAssembleHandler.handle\` reads \`expected_artifacts\` from the task envelope (already plumbed through by \`task_plan.py\` since SIP-0086), derives \`task_required_files\` as basenames, and threads it through to both prompt assembly and validation.

## Verification path

After this lands, a fullstack validation cycle should let framing's multi-builder-task decomposition land cleanly:
- task 8 emits only manifests, task 9 emits only qa_handoff
- Both fit under the 8192 clamp
- Bob's failure mode shifts away from \"qa_handoff missing required sections\" — the dominant failure pattern across cycles 1–3 is unblocked
- PR #104's \`builder.assemble_repair\` routing finally gets a real shot at exercising end-to-end if any builder failure does occur (the rewind/patch trigger goes away)

## Test plan

- [x] Six new tests on \`BuildProfile.system_prompt_for_files\` (scoped subset, qa_handoff inclusion/exclusion, None/empty fallback equivalence)
- [x] Three new tests on \`BuilderAssembleHandler\` (manifest-only task success, qa_handoff-task validation still biting, empty-expected-artifacts profile fallback)
- [x] Full regression: **3658 passed, 1 skipped, 0 failures**
- [x] Ruff clean (2 pre-existing C901 on \`handle\` methods unrelated)
- [ ] Manual cycle verification: rebuild + run cycle 4 of the gate batch with the same overrides, confirm task 8 (manifests) no longer requires qa_handoff and the run progresses past it

## Related

- PR #103 (qa_handoff prompt hardening) — necessary but not sufficient: ensured Bob couldn't quietly produce a malformed qa_handoff; this PR ensures Bob isn't forced to produce qa_handoff in tasks framing didn't route to him for it
- PR #104 (repair-handler routing) — the routing fix needs a builder failure that picks correction_path: patch; #107 + #106 together unblock that scenario
- PR #105 (qwen3.6:27b model registry) — the 8192 clamp is what surfaced #107; both fixes ship as a pair, with the clamp encoding pressure to decompose and #107 letting decomposition actually take effect
- PR #106 (failure_evidence enrichment for #84) — orthogonal correction-decision quality fix
- SIP-0092 gate eval doc (forthcoming) — should reference this issue: criterion #2's tally needs revisiting since 3 of 3 typed-acceptance hits across cycles 1–3 were the same coupling defect, not three independent M1 trips

### DIFF
--- a/src/squadops/capabilities/handlers/build_profiles.py
+++ b/src/squadops/capabilities/handlers/build_profiles.py
@@ -80,10 +80,20 @@ class BuildProfile:
 
     @property
     def full_system_prompt(self) -> str:
-        """Compose the narrative template with the canonical file list block.
+        """Profile-level prompt — every file the profile requires, qa_handoff section block always present.
 
-        The file list is derived from `required_files`/`optional_files` so
-        the prompt stays in lockstep with what the validator enforces.
+        Used when the executing task has no `expected_artifacts` to scope
+        the requirements down (legacy single-task builder flows). Prefer
+        :meth:`system_prompt_for_files` when the framing step decomposed
+        builder work and the active task only owns a subset of the
+        profile's required files.
+        """
+        return self.system_prompt_for_files(None)
+
+    def system_prompt_for_files(
+        self, task_required_files: tuple[str, ...] | list[str] | None
+    ) -> str:
+        """Compose the system prompt scoped to the active task's required files.
 
         Cycle-1 evidence (cyc_11367982fd06, 2026-05-03): the build profile
         validator and the plan author can disagree about which qa_handoff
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
approved

jasonLadd (commented):
approved

PR #113, opened 2026-05-04
fix(framing): require PRD↔acceptance coverage in implementation_plan (#112)

jasonLadd:
## Summary

- Tightens `GovernanceReviewHandler._MANIFEST_PROMPT_EXTENSION` to require an explicit PRD ↔ acceptance_criteria coverage pass before the implementation_plan is emitted, with sub-requirement enumeration per deliverable and a worked example pinned to the recurring `qa_handoff` section-coverage defect.
- Adds 7 regression tests pinning the new prompt contract (section header, shape coverage, typed-check mapping, worked example, anti-pattern warning, audit-trail requirement, end-to-end propagation into the LLM call).
- Includes the SIP-0092 M1→M2 gate evaluation doc that motivates the fix — defers M2 dev pending more validation-profile cycles on the now-stable substrate.

Closes #112.

## Why this fix, M1-compatible

There is no separate plan-review step today — `GovernanceReviewHandler` is one LLM call producing both the review document and the implementation_plan via `_MANIFEST_PROMPT_EXTENSION`. M2 will split this (proposer/judge separation per SIP-0092), but until M2 ships, prompt-tightening is the smallest viable patch that addresses the recurring class of defect we see in 4 of 5 SIP-0092 gate-batch cycles.

## Test plan

- [x] `pytest tests/unit/capabilities/handlers/test_governance_review_plan.py` — 16/16 pass (9 existing + 7 new)
- [x] `pytest tests/unit/capabilities/ tests/unit/cycles/` — 2058 pass
- [x] `ruff check` / `ruff format` — clean on touched files (2 pre-existing C901 complexity errors on unrelated handlers, unchanged)
- [ ] **Verification cycle on `group_run` validation profile** (post-merge, after rebuild): confirm the qa_handoff section-coverage plan_delta no longer fires
- [ ] Run a second verification cycle to confirm the first wasn't a fluke

Per memory `feedback_rebuild_before_sip_test`: rebuild agent + runtime-api images before the verification cycle, otherwise the new prompt won't reach Max.

## Out of scope (explicitly)

- Run-contract enforcement against `required_artifacts` at run completion
- Builder-side pre-completion artifact-existence check
- Structured `plan_review.yaml` with typed `coverage_concern` objects (that is M2's design surface)
- Evaluator instrumentation (`typed_check_evaluation.json` per-cycle artifact + extended plan_delta `trigger` field) — separate hardening item, needed to formally close C1/C2 on the M1→M2 gate

### DIFF
--- a/docs/plans/SIP-0092-gate-M1-evaluation.md
+++ b/docs/plans/SIP-0092-gate-M1-evaluation.md
@@ -1,0 +1,136 @@
+# SIP-0092 Gate M1 → M2 Evaluation
+
+**Status:** Deferred (insufficient sample; directional evidence supports M2)
+**Signed:** Jason Ladd
+**Date:** 2026-05-04
+**SIP:** [SIP-0092 — Implementation Plan Improvement](../../sips/accepted/SIP-0092-Implementation-Plan-Improvement.md)
+**Plan:** [SIP-0092 Implementation Plan](./SIP-0092-implementation-plan-improvement-plan.md)
+
+## TL;DR
+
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
approved

PR #117, opened 2026-05-04
docs(SIP-0092): revise M1→M2 gate eval to proceed (cycles 5-6 evidence)

jasonLadd:
## Summary

Updates the SIP-0092 M1→M2 gate evaluation from this morning's deferral to **proceed to M2**, based on evidence from cycles 5 and 6.

## What changed in the verdict

**This morning** (initial deferral at 5 cycles): C3 qualitatively positive, C1/C2 unmeasurable, recommended 3-5 more cycles + minimal instrumentation (PR #115) before re-call.

**Now** (7 cycles in window): proceed to M2.

- **Cycle 5** (`cyc_7febd710e565`) cancelled at gate after triage caught PR #113 patching the wrong handler; PR #116 was the real fix.
- **Cycle 6** (`cyc_897459e05965`) post-PR-#116 produced a clean framing plan with six explicit `regex_match` section-header checks for `qa_handoff.md` plus a PRD coverage audit-trail comment block. Bob shipped `qa_handoff.md` missing two of the required sections anyway.
- That's the strongest C3 signal yet: even when M1 prompt discipline produces a defect-free plan at the proposer level, the generator ignores the typed `acceptance_criteria`. The typed checks are downstream-validation data, not upstream-generation guidance — the proposer-judge collapse pattern SIP-0092 §6.1.3 / SIP-0086 §6.1.3 named explicitly, observed under load.
- PR #115's instrumentation didn't fire on cycle 6 (zero `typed_check_evaluation_*.json` artifacts, zero `typed_acceptance_summary` log lines despite the code being in the running container). Closing the C1/C2 measurement gap on this substrate requires more debugging cycles whose ROI is poor.

## Decision

Proceed to M2 dev. C1/C2 gap acknowledged and not gating — M2 produces a structured `plan_review.yaml` directly, providing equivalent C2-class signal for the M2→M3 gate independent of M1's evaluator instrumentation.

## What this does NOT do

- Doesn't close the qa_handoff defect class. M2 ships behind `split_implementation_planning: false`. The defect rate won't change until the flag flips, which is a separate decision after M2 default-flip metrics bake (SIP §6.2.4).
- Doesn't commit to M3. M2→M3 has its own gate.
- Doesn't abandon `SIP-Multi-Role-Plan-Authoring`. That comparison happens in M2.1 design review, not in this gate eval.

## Test plan

- [x] Document review only — no code changes
- [ ] Reviewer signs off on the proceed-to-M2 decision

[...]

PR #116, opened 2026-05-04
fix(framing): wire PRD coverage discipline into the manifest prompt actually used (#112 real fix)

jasonLadd:
## Summary

PR #113 patched the wrong handler. Cycle 5 (`cyc_7febd710e565`) proved it: the framing flow's `implementation_plan.yaml` reproduced the exact qa_handoff acceptance-coverage defect class. This PR puts the prompt change in the prompt path the framing flow actually invokes.

Reopens #112.

## Root cause of the miss

Two manifest-producing prompt paths exist in this codebase:

| Path | File | When invoked |
|---|---|---|
| `GovernanceReviewHandler._MANIFEST_PROMPT_EXTENSION` | `cycle_tasks.py:610` | `governance.review` task — not used by current framing |
| `GovernanceReviewAssessReadinessHandler._produce_manifest` (`manifest_prompt`) | `planning_tasks.py:573` | `governance.assess_readiness` task — **what the framing flow actually runs** |

PR #113 patched the first. The framing uses the second. My audit caught tautological-test concerns but missed the bigger structural issue.

## What this PR does

- **Extracted** the PRD-coverage discipline section to a module-level constant `_PRD_COVERAGE_DISCIPLINE_SECTION` in `cycle_tasks.py` — single source of truth.
- **Wired** it into `planning_tasks.py:manifest_prompt` (the prompt the framing LLM actually sees) by importing and inserting it between `typed_acceptance_section` and the schema output.
- **Kept** the `cycle_tasks.py` reference (defense-in-depth — any present-or-future caller of `GovernanceReviewHandler` still gets the discipline).
- **Added 3 behavioral tests** that exercise `GovernanceAssessReadinessHandler.handle()` with a mocked LLM and assert the discipline text lands in the SECOND LLM call (the manifest producer). This is the regression guard the original PR lacked.

## Test plan

- [x] `pytest tests/unit/capabilities/test_planning_handlers.py::TestPRDCoverageDisciplineReachesManifestPrompt -v` — 3/3 pass
- [x] `pytest tests/unit/capabilities/test_planning_handlers.py tests/unit/capabilities/handlers/test_governance_review_plan.py` — 201 pass (16 existing on the constant + 3 new behavioral + the rest of the planning_handlers suite)
- [x] `./scripts/dev/run_regression_tests.sh` — 3687 pass, 1 skipped (+3 new vs main)
- [x] `ruff check` / `ruff format` — clean (3 pre-existing C901 complexity warnings on unrelated handle methods, unchanged)
[...]

PR #118, opened 2026-05-04
refactor(SIP-0092): align code names with the SIP-0092 plan vocabulary (M2.1 prep)

jasonLadd:
## Summary

Carries out the renames the SIP-0092 plan committed to in Rev 3 but were never actually applied to code:

| Old (still in code on main) | New (already in SIP-0092 plan doc) |
|---|---|
| `governance.assess_readiness` (capability ID) | `governance.review_plan` |
| `GovernanceAssessReadinessHandler` (class) | `GovernanceReviewPlanHandler` |
| `_produce_manifest` (function) | `_produce_plan` |
| `FRAMING_TASK_STEPS` (constant) | `PLANNING_TASK_STEPS` |
| `task_type.governance.assess_readiness.md` (fragment) | `task_type.governance.review_plan.md` |

## Why now

The plan-doc/code drift caused real damage during M1 hardening: PR #113's issue body referenced `governance.review_plan` based on the plan doc and patched `cycle_tasks.py:GovernanceReviewHandler` instead of the framing handler — wrong handler entirely. Cycle 5 caught it; PR #116 fixed it with another PR. **This rename removes the same trap from M2.1's surface** so the M2.1 implementer sees plan and code agreeing.

Per PR #117 (gate eval doc proceeding to M2), M2.1 starts after this lands.

## What this is and isn't

**Is**: pure string-substitution rename across code, prompt fragments, profile config, and tests. No semantic changes.

**Is not**:
- Not a migration. The capability_id flows through `task_type` fields in postgres TaskRun records, but those are read-only history of completed runs — nothing routes off historical task_type values, so old records keep their strings without breaking anything.
- Not a behavior change. Same handlers, same dispatch, same flow — just renamed identifiers.
- Not changing M2's design. M2.1 still has the same split structure described in the SIP-0092 plan.

## Test plan

- [x] `./scripts/dev/run_regression_tests.sh` — 3687 pass, 1 skipped (identical to pre-rename count)
- [x] `ruff check` / `ruff format` — clean on changed files (2 pre-existing C901 complexity warnings on unrelated `handle` methods, unchanged)
- [ ] **Rebuild + smoke test** before merge — capability_id changes need fresh agent images (`./scripts/dev/ops/rebuild_and_deploy.sh all`); a quick smoke cycle confirms the rename didn't break dispatch end-to-end
- [ ] **Don't merge while a cycle is in flight** — in-flight envelopes carrying `task_type: governance.assess_readiness` would not match the renamed handler; safe to merge with no active cycles (cycle 6 completed, cycle 7 not yet kicked off)

## Files touched

- `src/squadops/bootstrap/handlers.py`
- `src/squadops/capabilities/handlers/cycle_tasks.py`
- `src/squadops/capabilities/handlers/planning_tasks.py`
- `src/squadops/cycles/task_plan.py`
- `src/squadops/contracts/cycle_request_profiles/profiles/framing.yaml`
- `src/squadops/prompts/fragments/manifest.yaml`
- `src/squadops/prompts/fragments/shared/task_type/task_type.governance.review_plan.md` (renamed from `task_type.governance.assess_readiness.md`)
- 6 test files

### DIFF
--- a/src/squadops/bootstrap/handlers.py
+++ b/src/squadops/bootstrap/handlers.py
@@ -52,8 +52,8 @@ from squadops.capabilities.handlers.impl.repair_handlers import (
 from squadops.capabilities.handlers.planning_tasks import (
     DataResearchContextHandler,
     DevelopmentDesignPlanHandler,
-    GovernanceAssessReadinessHandler,
     GovernanceIncorporateFeedbackHandler,
+    GovernanceReviewPlanHandler,
     QADefineTestStrategyHandler,
     QAValidateRefinementHandler,
     StrategyFrameObjectiveHandler,
@@ -136,7 +136,7 @@ HANDLER_CONFIGS: list[tuple[type[CapabilityHandler], tuple[str, ...]]] = [
     (StrategyFrameObjectiveHandler, ("strat",)),
     (DevelopmentDesignPlanHandler, ("dev",)),
     (QADefineTestStrategyHandler, ("qa",)),
-    (GovernanceAssessReadinessHandler, ("lead",)),
+    (GovernanceReviewPlanHandler, ("lead",)),
     # Refinement handlers (SIP-0078: Planning Workload Protocol)
     (GovernanceIncorporateFeedbackHandler, ("lead",)),
     (QAValidateRefinementHandler, ("qa",)),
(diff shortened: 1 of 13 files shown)


jasonLadd (commented):
approved

PR #119, opened 2026-05-04
fix(executor): persist correction-task and repair-task output artifacts

jasonLadd:
## Summary

Closes the silent artifact-drop pattern observed across SIP-0092 gate-batch cycles 4b and 6: `builder.assemble_repair` runs, produces a `qa_handoff.md` in its outputs, the executor checkpoints completion, and the `qa_handoff.md` never reaches the artifact registry. Run marks 'completed' while violating its own contract.

## Root cause

Both correction-task and repair-task success branches in `_run_correction_protocol` called `_checkpoint_correction_task` directly. That helper only snapshots existing `all_artifact_refs` into a checkpoint — it never persists new artifacts. Compare to `_collect_artifacts_and_checkpoint` (used by regular tasks) which iterates `result.outputs[\"artifacts\"]` and stores each via `_store_artifact` before checkpointing. The repair path was missing that loop entirely.

## Fix

New helper `_store_correction_task_artifacts(result, envelope, cycle, run_id, all_artifact_refs, stored_artifacts)` mirrors the storage loop from `_collect_artifacts_and_checkpoint`. Persists each artifact via `_store_artifact` with `producing_task_type` metadata pinned so triage can distinguish repair output from original task output. Both success branches now call it before `_checkpoint_correction_task`.

The post-branch logic that strips `artifacts` from `prior_outputs` is preserved (artifacts shouldn't chain as context between tasks). The fix is additive: artifacts are now STORED in the registry and excluded from prior_outputs, instead of being silently dropped from both.

## What this affects

- **Future correction/repair cycles**: artifacts produced by `data.analyze_failure`, `governance.correction_decision`, `builder.assemble_repair`, `qa.validate_repair`, `development.correction_repair` will now reach the registry.
- **Historical records**: unchanged — old postgres rows reflect the old behavior.
- **Contract enforcement**: still a separate hardening item. This PR removes one mechanism by which contracts were being silently violated, but doesn't enforce them at run completion.

## Test plan

- [x] `pytest tests/unit/cycles/test_correction_protocol.py::TestCorrectionTaskArtifactStorage -v` — 3/3 pass (unit helper test, no-artifacts no-op, end-to-end patch-flow with qa_handoff.md asserted in vault)
- [x] `./scripts/dev/run_regression_tests.sh` — 3690 pass, 1 skipped (+3 new)
- [x] `ruff check` / `ruff format` clean on touched files (1 pre-existing C901 complexity warning on `execute_run`, unchanged)
- [ ] **Verification cycle on `group_run` validation profile** post-merge: confirm `qa_handoff.md` (or whatever the repair produces) appears in the artifact list when a correction loop fires

## Sequence

This PR can merge ahead of, in parallel with, or after #117 (gate eval doc) and #118 (rename refactor). Independent code path — no conflicts expected.

### DIFF
--- a/adapters/cycles/distributed_flow_executor.py
+++ b/adapters/cycles/distributed_flow_executor.py
@@ -1878,6 +1878,46 @@ class DistributedFlowExecutor(FlowExecutionPort):
                     return agent.agent_id, model, overrides
         return role, None, {}
 
+    async def _store_correction_task_artifacts(
+        self,
+        result: TaskResult,
+        envelope: TaskEnvelope,
+        cycle: Cycle,
+        run_id: str,
+        all_artifact_refs: list[str],
+        stored_artifacts: list[tuple[str, ArtifactRef]],
+    ) -> None:
+        """Persist a correction-task or repair-task's output artifacts.
+
+        Mirrors the artifact-storage loop in
+        ``_collect_artifacts_and_checkpoint`` but is split out so the
+        correction/repair success branches can call it before
+        ``_checkpoint_correction_task`` — which only snapshots existing
+        ``all_artifact_refs`` into a checkpoint and does not itself
+        persist new artifacts. Without this call, repaired deliverables
+        (e.g. the ``qa_handoff.md`` produced by ``builder.assemble_repair``
+        or the ``correction_decision.md`` from the correction protocol)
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
approved

Issue #78, opened 2026-05-02
GovernanceReviewHandler asks LLM to fill identifiers it can't know (project_id, cycle_id, prd_hash) — Max fabricates them


jasonLadd:
## Summary

The `GovernanceReviewHandler` prompt template asks the LLM to emit `project_id`, `cycle_id`, and `prd_hash` in the `implementation_plan.yaml` it produces — but these are facts the system knows, not values the LLM has access to. The template passes literal `<placeholder>` strings to the LLM, which then fabricates values that pass syntax validation but are incorrect and could break downstream lookups.

## Reproduction

Run any cycle that produces an `implementation_plan.yaml` (e.g., `--request-profile build`). Inspect the output, e.g.:

```yaml
# data/artifacts/<project>/<cycle>/<run>/art_*/implementation_plan.yaml
version: 1
project_id: group_run_mvp           # FAKE — actual is `group_run`
cycle_id: cycle_v03                  # FAKE — actual is `cyc_4cac11018af7`
prd_hash: a1b2c3d4e5f6               # FAKE — clearly fabricated
```

Live examples:
- Cycle `cyc_4cac11018af7` (2026-05-02): `project_id: group_run_mvp`, `cycle_id: cycle_v03`, `prd_hash: a1b2c3d4e5f6`
- Cycle `cyc_01e2280207cd` (2026-05-02, prior run on stale code): `project_id: group_run_mvp_v03`, `cycle_id: cycle_001`, `prd_hash: e8a3f9c2d1b4567890abcdef1234567890abcdef1234567890abcdef1234567890` (`1234567890abcdef` repeating — a near-canonical "I'm fabricating" signature)

## Root cause

`src/squadops/capabilities/handlers/cycle_tasks.py:561-567` — the `_MANIFEST_PROMPT_EXTENSION` constant contains:

```python
"```yaml:implementation_plan.yaml\n"
"version: 1\n"
"project_id: <project_id>\n"
"cycle_id: <cycle_id>\n"
"prd_hash: <sha256 of PRD>\n"
```

These are passed to the LLM as literal placeholder text. The handler has the real values in scope (cycle.project_id, cycle.cycle_id, and could compute `sha256(prd_content)`), but doesn't substitute them before rendering.

## Expected

The system should either (a) substitute authoritative values into the template at render time so the LLM never has to invent them, or (b) post-process the manifest server-side to overwrite these fields with the correct values, or (c) treat divergence as a rejection rather than a warning.

[...]

PR #122, opened 2026-05-05
fix(governance): make Max's structured outputs reliable enough for M2 (#109)

jasonLadd:
## Summary

Two Max-output reliability failure modes that silently degraded structured outputs would, left in place, mask the SIP-0092 M2 \`plan_review.yaml\` rubric the same way they masked \`assess_readiness\`. Fixing them before M2 dev starts so the new artifact lands on a trustworthy substrate.

**1. Frontmatter omission (assess_readiness / governance.review_plan).** The handler used to silently prepend a default \`readiness=revise / sufficiency_score=3\` block when the LLM omitted YAML frontmatter — every downstream consumer saw a value Max never authored. Now retries Max once with a corrective instruction; if the retry still omits frontmatter, the task fails so the cycle's correction loop can fire instead of papering over it.

**2. Identifier fabrication (governance.review / governance.review_plan).** Both manifest-producing prompts asked the LLM to fill \`<project_id>\` / \`<cycle_id>\` / \`<sha256 of PRD>\` placeholders. Small models fabricated plausible-looking values (\`group_run_mvp\`, \`cycle_v03\`, \`a1b2c3d4e5f6\`) that pass syntax validation but break downstream lookups. Now: the prompt is pre-filled with the cycle's authoritative identifiers, and the parsed manifest is rewritten server-side as defense in depth.

Closes #109. Subsumes #78 (already closed as duplicate).

## Mechanism

- \`ExecutionContext\` gains \`project_id\` alongside the existing \`cycle_id\`; \`HandlerExecutor\` wires it from \`envelope.project_id\`.
- \`_MANIFEST_PROMPT_EXTENSION\` uses \`.format()\` placeholders; new \`_render_manifest_extension\` substitutes from context at prompt-build time.
- New module-scope helper \`_rewrite_manifest_identifiers\` overwrites \`project_id\` / \`cycle_id\` / \`prd_hash\` on the parsed YAML and logs the diff when the LLM disagreed (visibility, not silence).
- \`GovernanceReviewPlanHandler\` gets a single corrective retry on missing frontmatter via \`_retry_without_frontmatter\`, then fails the task on continued omission.

## Test plan

- [x] \`pytest tests/unit/capabilities/test_planning_handlers.py tests/unit/capabilities/handlers/test_governance_review_plan.py tests/unit/prompts/test_migration_validation.py\` (227 passed)
- [x] \`./scripts/dev/run_regression_tests.sh\` (3724 passed, 1 skipped — up 6 from 3718 baseline)
- [x] \`ruff format\` clean on changed files
- [ ] First M2-dev cycle: confirm \`implementation_plan.yaml\` carries the real \`project_id\` / \`cycle_id\` / \`prd_hash\` (was the cyc_4cac11018af7 failure mode in #109)
- [ ] First M2-dev cycle that tickles framing without frontmatter: confirm the corrective retry fires and either recovers or fails visibly (was the cyc_d1c1a259c983 silent-default mode)

## New test coverage

- \`TestManifestIdentifierRewrite\` × 4 — fabricated-id rewrite, identity no-op, prompt pre-fill verification, fallback when context identifiers are empty.
- \`TestProduceManifestIdentifierRewrite\` × 2 — same coverage on the framing-time \`_produce_plan\` path.
- Replaced silent-synthesis tests with retry-then-fail contract: \`test_missing_frontmatter_retries_then_fails\`, \`test_missing_frontmatter_retry_recovers_when_llm_complies\`, \`test_retry_corrective_message_targets_frontmatter\`, \`test_evidence_preserved_when_frontmatter_retry_fails\`.

### DIFF
--- a/src/squadops/capabilities/handlers/context.py
+++ b/src/squadops/capabilities/handlers/context.py
@@ -70,6 +70,7 @@ class ExecutionContext:
         role_id: Role of the agent
         task_id: Current task ID
         cycle_id: Current cycle ID
+        project_id: Current project ID
         ports: PortsBundle for port access
         skill_registry: Registry for skill execution
     """
@@ -80,6 +81,7 @@ class ExecutionContext:
     cycle_id: str
     ports: PortsBundle
     skill_registry: SkillRegistry
+    project_id: str = ""
     correlation_context: CorrelationContext | None = None
     _skill_executions: list[SkillExecutionRecord] = field(default_factory=list)
 
@@ -92,6 +94,7 @@ class ExecutionContext:
         cycle_id: str,
         ports: PortsBundle,
         skill_registry: SkillRegistry,
+        project_id: str = "",
         correlation_context: CorrelationContext | None = None,
     ) -> ExecutionContext:
         """Factory method for creating execution context.
@@ -103,6 +106,10 @@ class ExecutionContext:
             cycle_id: Current cycle ID
             ports: PortsBundle from agent
             skill_registry: SkillRegistry with available skills
+            project_id: Current project ID (issue #109 — handlers that
+                emit implementation_plan.yaml need authoritative
+                project_id alongside cycle_id so they don't ask the LLM
+                to invent it)
             correlation_context: Optional LangFuse correlation context
 
         Returns:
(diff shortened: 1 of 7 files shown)


jasonLadd (commented):
approved

PR #126, opened 2026-05-05
refactor: externalize SIP-0079 impl handler system prompts to task_type fragments

jasonLadd:
## Summary

The three SIP-0079 implementation handlers — \`correction_decision\`, \`analyze_failure\`, \`establish_contract\` — hardcoded their system prompts as Python string constants (\`_DECISION_SYSTEM_PROMPT\`, \`_ANALYSIS_SYSTEM_PROMPT\`, \`_CONTRACT_SYSTEM_PROMPT\`). That bypassed the PromptService / LangFuse versioning the \`planning_tasks.py\` handlers already use, and meant prompt edits couldn't be tracked or A/B'd.

Move each system prompt into a \`task_type\` fragment file under \`src/squadops/prompts/fragments/shared/task_type/\`. Handlers now call \`prompt_service.assemble(role, hook=\"agent_start\", task_type=capability_id)\` — the same plumbing the planning handlers use. Composes role identity + task_type fragment automatically.

## What ships

| File | Purpose |
|---|---|
| \`fragments/shared/task_type/task_type.governance.correction_decision.md\` | Was \`_DECISION_SYSTEM_PROMPT\` |
| \`fragments/shared/task_type/task_type.data.analyze_failure.md\` | Was \`_ANALYSIS_SYSTEM_PROMPT\` (with FailureClassification enum interpolation resolved to literals) |
| \`fragments/shared/task_type/task_type.governance.establish_contract.md\` | Was \`_CONTRACT_SYSTEM_PROMPT\` |
| \`impl/correction_decision.py\` / \`impl/analyze_failure.py\` / \`impl/establish_contract.py\` | Constants deleted; handlers call \`prompt_service.assemble(...)\` |
| \`tests/unit/capabilities/test_impl_handlers.py\` | New \`TestImplHandlerSystemPromptExternalization\` class — pinned contract |
| \`tests/unit/prompts/test_planning_fragments.py\` | Fragment-count assertion bumped 12 → 15 |

## Behavior change

\`assemble()\` composes the role identity fragment (\`fragments/roles/{role}/identity.md\`) + the new task_type fragment, where previously the hardcoded constant was passed alone. So the system prompt now also includes the role identity content. This is consistent with how every planning handler works today — the impl handlers were the outliers.

I'd expect quality to be neutral-to-slightly-better (richer role context primes the LLM more consistently), but it's a measurable change. If the next gate cycle shows worse correction-decision quality, easy revert: pull the role-identity layer out of \`assemble()\` for these specific task_types.

[...]

PR #127, opened 2026-05-05
fix(governance): repair merge-conflict resolution between #124 and #126

jasonLadd:
## Summary

**Main is broken.** The conflict resolution when #124 and #126 both merged dropped two pieces:

1. **\`_VALID_PLAN_CHANGE_CANDIDATES\` tuple lost** from \`correction_decision.py\`. The handler still references it on line 130 (in the diagnostic-field validation block from #124), so every code path that exercises \`governance.correction_decision\` raises \`NameError\` at runtime. **16 tests fail on main today.**

2. **Diagnostic-question content didn't migrate to the fragment.** PR #124 added the diagnostic-question section + the 5-field JSON schema to the (now-deleted) \`_DECISION_SYSTEM_PROMPT\` constant. PR #126 deleted the constant in favor of the new fragment file. The conflict resolution kept the new fragment but didn't carry over #124's additions — so the LLM is never asked to emit \`structural_plan_change_candidate\`, every \`plan_delta\` would silently default it to \`\"none\"\`, and the SIP-0092 M3 gate evidence we wanted to start collecting is invisible.

## Fix

- **Restore \`_VALID_PLAN_CHANGE_CANDIDATES\`** in \`correction_decision.py\` with the same allowed values from #124 (\`none\` / \`add_task\` / \`tighten_acceptance\` / \`other\`).
- **Move the diagnostic-question section + the 5-field JSON output schema** into \`task_type.governance.correction_decision.md\` (the fragment that #126 introduced as the new authoritative source). Bumped fragment version 1.0.0 → 1.1.0.
- **Remove the now-redundant \"Diagnostic question\" section from the request template** — was added in #124 as a fallback when only the hardcoded constant existed; now duplicated by the fragment. Single source of truth in the fragment. Bumped template version 2 → 3.

## Test plan

- [x] \`./scripts/dev/run_regression_tests.sh\` — was 3743 passed / 16 failed against current main; **3759 passed / 0 failed after this fix**.
- [x] \`pytest tests/unit/capabilities/test_impl_handlers.py::TestCorrectionDecision\` — 14 tests now pass (were all failing).
- [x] \`pytest tests/unit/cycles/test_correction_protocol.py::TestPlanDelta\` — both tests now pass (were failing).
- [ ] Validate end-to-end on next cycle that fires correction: confirm \`plan_delta_*.json\` carries a non-default \`structural_plan_change_candidate\` when the LLM picks one. (The previous validation cycle would have silently emitted \`\"none\"\` regardless of what the LLM thought, so the next cycle is the real first measurement.)

## Why this slipped past CI

#124 and #126 each passed their own regression suite at PR-author time — but each was authored against \`main\` independently and didn't see the other's deletes. The conflict resolution at merge time needed to take #124's diagnostic-field additions and rewrite them onto #126's fragment file (instead of the deleted constant). That manual lift didn't happen.

Trivially preventable in retrospect: when a follow-up PR (#126) deletes the surface a sibling PR (#124) is modifying, the rebase should be done before merge, not at it. Worth flagging as a soft rule for future stacked-PR situations.

### DIFF
--- a/src/squadops/capabilities/handlers/impl/correction_decision.py
+++ b/src/squadops/capabilities/handlers/impl/correction_decision.py
@@ -26,6 +26,15 @@ logger = logging.getLogger(__name__)
 
 _VALID_CORRECTION_PATHS = ("continue", "patch", "rewind", "abort")
 
+# SIP-0092 M2 → M3 gate diagnostic. The correction protocol can today
+# only choose continue/patch/rewind/abort — it cannot mutate the
+# implementation plan. M3 will add `decision: plan_change` with two
+# operations (add_task, tighten_acceptance). To know whether M3 is
+# worth shipping, we capture which structural plan change the lead
+# would have chosen if it were available — the field is non-operative
+# and exists only to drive the M3 justification gate.
+_VALID_PLAN_CHANGE_CANDIDATES = ("none", "add_task", "tighten_acceptance", "other")
+
 
 class GovernanceCorrectionDecisionHandler(_CycleTaskHandler):
     """Decide the correction path after a failure analysis."""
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
approved

PR #128, opened 2026-05-05
fix(impl): tolerant JSON extraction + raw-response logging in SIP-0079 handlers

jasonLadd:
## Summary

Cycle \`cyc_a4e6dc3afe7a\` (2026-05-05) failed implementation in ~4 minutes when \`governance.establish_contract\` returned a 2521-token response that the handler parser rejected with \`Expecting value: line 1 column 1 (char 0)\`. The handler never persisted the raw response, so root cause had to be inferred from token counts alone.

This PR is the load-bearing fix regardless of root cause: the strip-leading-fence-then-\`json.loads\` pattern was always one model-behavior change away from this same failure. PR #126's prompt assembly change may have been the trigger; the brittleness was always the root.

## Honest scope

I called PR #126 the root cause earlier and the user pushed back — fairly. I was inferring from token counts and correlation; I hadn't seen the actual response. This PR ships defense regardless: it would have prevented the cycle-2 failure whether the actual issue was:

- PR #126's role-identity prepend biasing prose-preamble output
- Qwen3 \`<think>\` block emission that wasn't there in cycle 1
- Pure LLM nondeterminism on a single sample

The new logging path means **next time we will see the raw response** and won't have to guess.

### New: \`src/squadops/capabilities/handlers/impl/_json_extraction.py\`

Shared helper \`extract_first_json_object(text)\` that:

- Strips \`<think>...</think>\` blocks anywhere in the text (Qwen3-family).
- Strips markdown code fences anywhere, not just at the start.
- Finds the first balanced \`{...}\` substring with a string-aware brace matcher (ignores \`{\`/\`}\` inside double-quoted strings, handles escaped quotes).
- Raises \`JSONExtractionError\` with a \`raw_excerpt\` attribute so callers can log the truncated response without re-truncating.

### Modified: all three SIP-0079 impl handlers

\`establish_contract.py\`, \`correction_decision.py\`, \`analyze_failure.py\` now call the shared helper. On parse failure, each emits a structured warning carrying the raw response — \`%s: failed to parse ... | raw[:500]=%r\`.

### Tests

- 18 unit tests for the extractor: pure JSON, all real LLM output shapes (think-block, prose preamble, json-in-fence, json-then-prose, combined), string-aware brace matching (3 cases), 7 error paths.
- 2 handler-level regression tests on \`GovernanceEstablishContractHandler\` pinning the cycle-2 failure modes (\`<think>\` block + prose preamble before fenced JSON).

## Test plan

- [x] \`pytest tests/unit/capabilities/handlers/test_json_extraction.py tests/unit/capabilities/test_impl_handlers.py\` — 70 passed
- [x] \`./scripts/dev/run_regression_tests.sh\` — 3779 passed (+20), 1 skipped (was 3759)
- [x] \`ruff format\` clean
- [ ] Validate end-to-end on a re-cycled run: \`establish_contract\` no longer dies on first task; if a parse failure does occur somewhere, the raw response is in the logs for triage.

## What we learn from the next cycle

Even if this fix prevents the visible failure, the new logging tells us **what** the LLM was actually emitting. If it was a \`<think>\` block, that confirms Qwen3 thinking mode is firing under #126's longer prompt; we may want to disable thinking mode at the model layer or strip it earlier (the extractor handles it, but the wasted tokens are a real cost). If it was prose-preamble, we get the actual prose for prompt-tuning.

### DIFF
--- a/src/squadops/capabilities/handlers/impl/_json_extraction.py
+++ b/src/squadops/capabilities/handlers/impl/_json_extraction.py
@@ -1,0 +1,183 @@
+"""Robust JSON extraction for SIP-0079 impl handlers.
+
+The three SIP-0079 implementation handlers (establish_contract,
+correction_decision, analyze_failure) ask the LLM to return a JSON
+object as the entire response. Strict-from-start parsing is brittle —
+LLMs commonly emit:
+
+- ``<think>...</think>`` blocks (Qwen3-family thinking mode) before
+  the actual content.
+- Prose preamble like ``"Here is the contract:"`` before a fenced
+  JSON block.
+- Trailing prose after the JSON.
+- Code fences (``` json ... ```) anywhere in the response, not just
+  at the start.
+
+Any of those produce ``json.loads("<think>...") -> "char 0"`` errors
+even when the actual JSON the LLM emitted is well-formed.
+
+This module provides ``extract_first_json_object(text)`` — a tolerant
+parser that finds the first balanced ``{...}`` after stripping
+fences and brace-matching, returning the parsed object or raising
+``JSONExtractionError`` with the truncated raw response so callers
+can log it for triage.
+
+The brace-matcher is string-aware: it ignores ``{`` and ``}`` inside
+double-quoted strings (handling escaped quotes) so a JSON value like
+``"description": "Use {} for placeholders"`` doesn't false-match.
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
approved

PR #129, opened 2026-05-06
fix(impl): suppress role-identity prepend on JSON-emitting impl handlers

jasonLadd:
## Summary

PR #128's raw-response logging caught the actual root cause of the cycle 2 / cycle 4 failures. **Direct evidence**, captured from cycle \`cyc_a867cbf02205\` (2026-05-05):

**Max (governance.correction_decision):**
\`\`\`
### 🔹 Initialization Verification
1. **Role Configuration**: \`LeadAgent_SquadOps\` loaded ✅
2. **Resource Access**: ...
3. **Pending Tasks**: ...
4. **Ready Status**: \`READY_FOR_CYCLE_START\`
### 🔹 Replan Strategy ...
\`\`\`

**Data (data.analyze_failure):**
\`\`\`
# Data Agent Initialization Cycle
**Status:** Ready
**Role:** Data Agent (SquadOps)
**Context:** Recovery from failed \`development.develop\` task...
## 1. Initialization Check
- **Role Configuration:** Loaded...
\`\`\`

Both handlers asked for JSON; both received role-play markdown narratives instead. The cause: PR #126 externalized impl handler system prompts to fragments and switched assembly to \`PromptService.assemble()\`, which by design layers identity + constraints + lifecycle + task_type. The lead/data identity fragments (\"You are the Lead Agent in the SquadOps framework... Orchestration / Decision Making / Communication / Maintain a strategic perspective...\") prime smaller spark-squad models (qwen3.6:27b) to introduce themselves before doing the requested task.

Pre-#126 these handlers used hardcoded \`_*_SYSTEM_PROMPT\` constants containing JUST the task instructions — no role-identity preamble. That's why cycles 1+2's \`establish_contract\` succeeded.

## Honest accounting

I called PR #126 the root cause earlier. You pushed back fairly because I was inferring from token counts. PR #128's logging finally surfaced the raw responses. **You were right to challenge the premature claim, and the evidence now confirms it.**

## Fix

| Layer | Change |
|---|---|
| **Port** \`squadops/ports/prompts/service.py\` | New abstract method \`assemble_task_only(role, task_type)\` |
| **Adapter** \`squadops/prompts/assembler.py\` | Concrete impl: resolves the task_type fragment via \`_resolve_fragment\`, verifies hash, composes alone (no identity / constraints / lifecycle) |
| **Handlers** \`impl/correction_decision.py\` / \`analyze_failure.py\` / \`establish_contract.py\` | Switched from \`assemble()\` → \`assemble_task_only()\`. Externalized fragments stay as the single source of truth; LangFuse version tracking still applies. |

The three task_type fragment files (added in #126) are unchanged and stay authoritative. Only the assembly path differs.

## What this preserves vs the old hardcoded constants

| | Pre-#126 hardcoded | Post-this-PR |
|---|---|---|
| LLM-visible content shape | task instructions only | task instructions only — identical |
| LangFuse version tracking | ❌ | ✅ |
| Prompt edits as reviewable diffs | ❌ (\`git blame\` on string constants) | ✅ (fragment file diffs) |
| Triggers role-play priming | ❌ | ❌ |

## Test plan

- [x] \`pytest tests/unit/prompts/test_assembler.py tests/unit/capabilities/test_impl_handlers.py\` — 69 passed
- [x] \`./scripts/dev/run_regression_tests.sh\` — 3782 passed (+3), 1 skipped
- [x] \`ruff format\` clean
- [ ] **Validate end-to-end on next cycle:** governance.correction_decision and data.analyze_failure should produce JSON, not role-play narratives. PR #128's \`raw[:500]=\` logging will capture any residual failure mode.

## New tests

- 3 assembler tests for \`assemble_task_only\`: returns task fragment alone (identity/constraints content NOT present), raises \`MandatoryLayerMissingError\` when fragment missing, honors role-specific override precedence.
- Existing \`TestImplHandlerSystemPromptExternalization\` updated: asserts \`assemble_task_only\` is called and the legacy \`assemble\` is NOT.

## Follow-up filed separately

Neo's \`development.develop\` fence-drop pattern (cycle 1 + cycle 4 both showed Neo emitting unparseable code without fenced blocks) is a different brittleness in a different handler — issue filed for separate triage.

### DIFF
--- a/src/squadops/capabilities/handlers/impl/analyze_failure.py
+++ b/src/squadops/capabilities/handlers/impl/analyze_failure.py
@@ -108,13 +108,16 @@ class DataAnalyzeFailureHandler(_CycleTaskHandler):
                 user_parts.append(f"\n\n## Failure Evidence\n\n{evidence_json}")
             user_prompt = "\n".join(user_parts)
 
-        # System prompt assembled from role + task_type fragments
-        # (fragments/roles/data + fragments/shared/task_type/
-        # task_type.data.analyze_failure.md). PromptService tracks
-        # the version and surfaces it to LangFuse.
-        assembled = context.ports.prompt_service.assemble(
+        # System prompt is the task_type fragment ALONE — no role
+        # identity prepend. Same regression as governance.correction_decision
+        # captured in cycle cyc_a867cbf02205 (2026-05-05): Data wrote
+        # "# Data Agent Initialization Cycle / Role: Data Agent
+        # (SquadOps) / ## 1. Initialization Check" instead of the
+        # required JSON FailureAnalysis schema. Role-identity layer
+        # primes the role-play response. Suppress it for this
+        # JSON-emitting handler.
+        assembled = context.ports.prompt_service.assemble_task_only(
             role=self._role,
-            hook="agent_start",
             task_type=self._capability_id,
         )
         messages = [
--- a/src/squadops/capabilities/handlers/impl/correction_decision.py
+++ b/src/squadops/capabilities/handlers/impl/correction_decision.py
@@ -80,15 +80,18 @@ class GovernanceCorrectionDecisionHandler(_CycleTaskHandler):
                 )
             user_prompt = "\n".join(user_parts)
 
-        # System prompt assembled from role + task_type fragments
-        # (fragments/roles/lead + fragments/shared/task_type/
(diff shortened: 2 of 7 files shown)


jasonLadd (commented):
approved

PR #131, opened 2026-05-06
fix(prompts): register SIP-0079 task_type fragments missing from manifest

jasonLadd:
## Summary

- PR #126 created three task_type fragment files on disk but never registered them in `fragments/manifest.yaml`. The shared-fragment lookup goes through the manifest, so unregistered fragments are invisible.
- Latent until PR #129 made the task_type layer mandatory via `assemble_task_only()`. Every cycle now fails at its first impl task (Max / `governance.establish_contract`) with `MandatoryLayerMissingError`.
- Adds the missing manifest entries, bumps manifest version, and adds regression tests so this can't happen again.

## Repro / impact

Cycle `cyc_8c27dc7f3260` / run `run_1a627a5d7fb1` (group_run, 2026-05-06): framing completed cleanly, impl run failed in 0.6 seconds on the first dispatch — `governance.establish_contract` (Max) raised `MandatoryLayerMissingError: Mandatory layer 'task_type' missing for role 'lead'` from `prompts/assembler.py:248` (`assemble_task_only`).

Same failure mode would hit `governance.correction_decision` (Max) and `data.analyze_failure` (Data) when SIP-0086 corrections kick in.

## Root cause

`FileSystemPromptRepository._resolve_path` for shared fragments calls `manifest.get_fragment_meta(fragment_id)` and falls back to None when the entry is missing. Three SIP-0079 impl handler fragments were missing from the manifest:

| Fragment | File on disk | In manifest? |
|---|---|---|
| `task_type.governance.establish_contract` | yes | no |
| `task_type.governance.correction_decision` | yes | no |
| `task_type.data.analyze_failure` | yes | no |

Before PR #129 these were optional via `assemble()`; PR #129 made them mandatory via `assemble_task_only()` which exposed the gap.

## Changes

- `src/squadops/prompts/fragments/manifest.yaml`: add the three entries with computed sha256, bump version `0.9.18 -> 0.9.19`, refresh `updated_at`, recompute `manifest_hash`.
- `tests/unit/prompts/test_manifest_completeness.py` (new): three regression tests
  - every `*.md` fragment file on disk is registered in the manifest
  - every manifest entry points to an existing file
  - explicit guard naming the three SIP-0079 impl-handler fragments

## Test plan

- [x] `pytest tests/unit/prompts/test_manifest_completeness.py -v` — 3 passed
- [x] `pytest tests/unit/prompts/ -v` — 305 passed
- [x] `./scripts/dev/run_regression_tests.sh` — 3785 passed, 1 skipped, 0 failed
- [x] Round-trip verification: `assemble_task_only(role, task_type)` succeeds for all three new entries
- [ ] Rebuild agent + runtime-api images on Spark and relaunch a `group_run` validation cycle to confirm impl phase gets past `governance.establish_contract`

## Notes

While running the manifest-completeness check I noticed two unrelated preexisting hash mismatches between the manifest and `roles/comms/identity.md` / `shared/identity/identity.md`. Those are out of scope for this fix; will track separately.

### DIFF
--- a/src/squadops/prompts/fragments/manifest.yaml
+++ b/src/squadops/prompts/fragments/manifest.yaml
@@ -1,5 +1,5 @@
-version: 0.9.18
-updated_at: '2026-03-13T14:40:23.887991Z'
+version: 0.9.19
+updated_at: '2026-05-06T13:45:00.000000Z'
 fragments:
 - fragment_id: identity
   path: roles/comms/identity.md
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
approved

PR #162, opened 2026-06-13
fix(tests): restore green regression baseline on main

jasonLadd:
## Why
While scoping SIP-0089 Phase 2, found the merge gate (`run_regression_tests.sh`) was **red on main** — it halted at collection — and the broader `pytest tests/unit` had 21 failures. Distinct from #148 (httpx 0.28 drift, now resolved). Fixes #160.

## Root causes & fixes
| # | Cluster | Verdict | Fix |
|---|---------|---------|-----|
| 1 | `domain_cycles` marker unregistered | **Gate-blocking** — strict-markers collection error interrupts the whole suite | Register the marker in `pyproject.toml` |
| 2 | `_safe_resolve` symlink-escape (8 typed-acceptance tests) | **Real code bug** | Stop the symlink walk at `workspace_root`; trusted symlinked ancestors (`/var → /private/var` on macOS, symlinked mounts on Linux) no longer trip `path_escapes_workspace`. Escape detection for absolute / `..` / in-path symlinks preserved |
| 3 | a2a `_build_send_payload` (3 tests) | **Test stale** | Rename to `_build_payload` (the actual method name since the initial commit) |
| 4 | keycloak realm-lint fixture (2 tests) | **Test stale** | Add `loginTheme` + `squadops-console` `post.logout.redirect.uris` (lint rules added in 0497720; the fixture predated them) |

Triaged each cluster code-vs-test via git history before fixing — only #2 was a code change; #3/#4 were stale tests updated to match intended behavior (not weakened).

## Verification
- `./scripts/dev/run_regression_tests.sh`: **green, 3987 passed** (was halting at collection).
- `pytest tests/unit`: **21 → 8 failures**. The remaining 8 are telemetry **cumulative test-pollution** (pass in isolation and in the gate; only fail in full `tests/unit`) — not gate-blocking, tracked separately in **#161**.
- `ruff check` clean on changed lines; `ruff format` applied to touched files.

## Not in scope
- #161 — telemetry cumulative pollution (test-infra; pairs naturally with #149 CI).
- #148 — httpx drift, already resolved; this is the residual.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -128,6 +128,7 @@ markers = [
     "domain_events: Cycle lifecycle event system tests",
     # Runtime Modes (SIP-0089)
     "domain_runtime: Agent runtime state, modes, and coordinator tests",
+    "domain_cycles: Cycle/run/gate execution domain tests",
 ]
 
 ###########################
--- a/src/squadops/cycles/acceptance_checks.py
+++ b/src/squadops/cycles/acceptance_checks.py
@@ -110,19 +110,22 @@ def _safe_resolve(path_str: str, workspace_root: Path) -> Path:
     except ValueError as exc:
         raise _SafetyError("path_escapes_workspace") from exc
 
-    # Symlink escape: any symlink in the chain pointing outside the workspace.
+    # Symlink escape: any symlink WITHIN the user-supplied path (at or below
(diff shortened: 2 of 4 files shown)


jasonLadd (commented):
reviewed

PR #163, opened 2026-06-13
fix(tests): use asyncio.run in telemetry health tests (#161)

jasonLadd:
## Why
The 8 telemetry failures tracked in #161 — they failed **only** in a full `pytest tests/unit` run, passing in isolation and in the regression gate.

## Root cause (not what #161 first suspected)
Not OTel/langfuse global-state pollution. The real error was `RuntimeError: There is no current event loop in thread 'MainThread'` on every `health()` test. These **sync** tests called the async `health()` via `asyncio.get_event_loop().run_until_complete(...)`. In isolation a current loop exists; in the full suite an earlier async test leaves the main-thread loop closed/cleared, and Python 3.11's `get_event_loop()` raises instead of auto-creating one. The adapters' `health()` is correct — only the test invocation was fragile.

## Fix
Replace all 8 occurrences with `asyncio.run(...)` (creates/tears down its own loop, immune to prior loop-state). All assertions preserved; surgical 8-line diff across the 3 telemetry test files.

## Verification
- Full `pytest tests/unit`: **4480 passed, 0 failed** (was 8 failed), deterministic across repeated runs.
- Telemetry in isolation: still 139 passed.
- `ruff check` clean.

Fixes #161.

### DIFF
--- a/tests/unit/telemetry/test_adapters.py
+++ b/tests/unit/telemetry/test_adapters.py
@@ -220,7 +220,7 @@ class TestNoOpLLMObservabilityAdapter:
         import asyncio
 
         adapter = NoOpLLMObservabilityAdapter()
-        result = asyncio.get_event_loop().run_until_complete(adapter.health())
+        result = asyncio.run(adapter.health())
         assert result["status"] == "ok"
         assert result["backend"] == "noop"
         assert result["details"] == {}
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #174, opened 2026-06-14
feat(cycles): lite profile + filter builder-only task types by squad capability

jasonLadd:
## Summary
Fixes the root cause of the `cyc_0024e1a0b6b5` implementation failure (full-squad aborted at task 2/8 with `No handler for capability: builder.assemble`) and adds a cheap builder-equipped profile for end-to-end framework validation.

## Changes
**1. Planner capability filter (root cause)**
`produce_plan` now offers the LLM only task types the squad can execute. `builder.assemble` is dropped from the allowed list when the squad has no `builder` role, so the planner can't author a task no agent can handle — packaging/scaffolding routes to `development.develop` instead, and **builder-less squads build successfully**. New `planner_build_task_types()` in `implementation_plan.py` is the single source (co-located with `_KNOWN_BUILD_TASK_TYPES`). The plan *validator* still accepts the full set — it validates plans regardless of authoring squad.

**2. New `lite` squad profile**
All 6 roles incl. builder on `qwen2.5:7b` — a lightweight, builder-equipped squad for cheap end-to-end framework validation (exercises the full build path, including `builder.assemble`, without large models). First profile under the no-`-with-builder`-suffix convention (see #173).

## Tests
- `planner_build_task_types` unit tests: both branches (exact sets) + mutation-safety (returns a fresh set).
- `produce_plan` behavioral test: `builder.assemble` is offered **iff** the squad has a builder role.
- Updated `test_allowed_task_types_injected_into_prompt` — its old assertion encoded the bug (builder.assemble always offered); now asserts the capability-filtered behavior.
- Regression: **3999 passed, 1 skipped, 0 failed**; test-quality lint clean.

## Scope / follow-ups (deliberately not here)
- **#172** — gate-time plan↔squad capability pre-flight (defense-in-depth; `ImplementationPlan.validate_against_profile` may already cover most of it — needs wiring).
- **#173** — profile-name consolidation (drop `-with-builder`, collapse `full`/`spark`, reconcile `smoke`/`lite`/`full`). Breaking; needs a DB migration + doc sweep.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -111,7 +111,7 @@ squadops artifacts list <run-id>           # List artifacts for run
   - `secrets/` - env, file, docker_secret providers
   - `comms/` - RabbitMQ adapter
   - `persistence/` - PostgreSQL runtime with connection pooling
-  - `cycles/` - DistributedFlowExecutor, MemoryCycleRegistry, PostgresCycleRegistry, factory
+  - `cycles/` - DispatchedFlowExecutor, MemoryCycleRegistry, PostgresCycleRegistry, factory
   - `telemetry/` - LangFuse adapter (buffered, with redaction) and factory
   - `auth/` - Keycloak adapter, JWT middleware
   - `llm/` - Ollama adapter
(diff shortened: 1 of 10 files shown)


jasonLadd (commented):
approved

PR #179, opened 2026-06-20
fix(tests): retarget builder-squad tests to spark-squad-with-builder (green the regression suite)

jasonLadd:
## Problem
PR #175 removed the `full-squad-with-builder` squad profile but left **four** test files asserting it loads from `config/squad-profiles.yaml`. The regression suite has been red — **13 failures** — for ~6 days, which degrades the "2900+ always pass" signal every subsequent change relies on.

Affected files:
- `tests/unit/cycles/test_squad_profile_builder.py`
- `tests/unit/cycles/test_builder_e2e_validation.py`
- `tests/unit/cycles/test_builder_bootstrap_verification.py`
- `tests/unit/cycles/test_task_plan_builder.py` (cosmetic inline label only)

## Fix
Re-point the config-loading tests at the surviving builder profile, `spark-squad-with-builder` (identical 6-agent shape: lead/dev/strat/builder/qa/data, bob=builder). This **preserves the SIP-0071 builder-profile coverage** rather than deleting it. Inline cosmetic `profile_id` labels updated for consistency; the removed profile is now only referenced in an explanatory docstring.

## Verification
Regression: **3999 passed, 1 skipped, 0 failed** (was 13 failed).

## Notes
- Fallout from #175; broader squad-profile name consolidation is tracked in #173 — this PR only re-points stale assertions, it does not rename anything.
- Process note: #175 pruned a profile without updating its tests and shipped a red suite — worth catching at review next time.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #144, opened 2026-05-16
feat(SIP-0093 PR 93.2): role proposer handlers + RC-23 failure records

jasonLadd:
## Summary

Third implementation PR for SIP-0093 (Multi-Role Plan Authoring). Adds the three role proposers (development, qa, strategy) and the structured-failure record they emit on retry exhaustion. **Registered in bootstrap; NOT wired into `PLANNING_TASK_STEPS`** — that's the cutover PR (93.3). Cycle behavior unchanged at runtime.

Authored after issue #140's prompt-registry cleanup lesson — every LLM call routes through the registered template + task-type fragment from day one. No inline prompts.

## Handlers (`src/squadops/capabilities/handlers/planning_tasks.py`)

All three subclass a new `_ProposeBaseHandler` that drives the retry-with-corrective-feedback LLM loop via `retry_yaml_call` (`_plan_authoring` helper), enforces `source_brief_id` matching against the upstream brief (RC-22), and emits a `ProposalFailure` artifact on exhaustion rather than raising (RC-23). `HandlerResult.success` stays `True` on failure — the cycle keeps moving, the merger reads the failure artifact as "this role's proposal is missing."

- **`DevelopmentProposePlanTasksHandler`** (`development.propose_plan_tasks`) — reads brief from `prior_outputs["artifact_contents"]`; emits `proposed_plan_tasks.yaml` parsed by `ProposedRoleTasks.from_yaml`.
- **`QaProposePlanTasksHandler`** (`qa.propose_plan_tasks`) — same shape, qa-scoped. SIP-0093 design: QA is the gap-catching pen — proposes tasks Development omitted.
- **`StrategyProposePlanGuidanceHandler`** (`strategy.propose_plan_guidance`) — emits `plan_guidance.yaml` parsed by `PlanGuidance.from_yaml`. Strategy proposes cross-cutting overlay (priority, ordering, time-budget guidance), not tasks. Per SIP-0093 §5.3.

Rev 1 ships sequential (per plan-doc Spark amendment from #143): parallel fan-out wouldn't deliver wall-clock savings on bandwidth-bound hardware.

## Prompt registry (issue #140 discipline applied from day one)

- **3 task-type fragments** registered: `task_type.development.propose_plan_tasks`, `task_type.qa.propose_plan_tasks`, `task_type.strategy.propose_plan_guidance`. Each role-scoped, each pinning brief immutability (RC-22), domain boundaries (no cross-domain task proposals), and output-shape discipline.
[...]

PR #180, opened 2026-06-20
fix(sip-0093): enable multi-role plan authoring via plan_authoring_contributors

jasonLadd:
## Problem
SIP-0093's per-role proposer handlers + `governance.merge_plan` merger merged (#144/#145/#165), but the **activation key** — `plan_authoring_contributors` — was never registered as an allowed CRP default. So no request profile could load with it set, and the merger has been running **sole-author in every cycle**. The multi-role authoring code is effectively dormant: SIP-0093-B has never actually run in multi-role mode.

The key already survives the full request path unchanged — `applied_defaults` → JSONB → `resolved_config = {**applied_defaults, **execution_overrides}` (task_plan.py) → `build_planning_steps` — exactly like `build_tasks`. Only the client-side CRP allow-list entry was missing.

## Change
- **schema**: add `plan_authoring_contributors` to `_APPLIED_DEFAULTS_EXTRA_KEYS`.
- **profiles**: add `validation-multirole` — the first request profile that activates the path (framing → `progress_plan_review` gate → implementation; contributors = development/qa/strategy).
- **tests**: contributors-configured framing sequence (proposer steps in canonical order), unknown-contributor `CycleError`, and profile-load guards.

## Verification
New tests pass; ruff clean. Activates the path that lets a validation cycle finally exercise multi-role authoring + merge_decisions.yaml end-to-end.

## Merge order
Branched off main **before** #179 (the regression-suite green-up). Rebase onto main after #179 merges so this PR sits on a green baseline. Files are disjoint from #179 — no conflict.

Part of closing the SIP-0093 partial (multi-role plan authoring core → validation).

### DIFF
--- a/src/squadops/contracts/cycle_request_profiles/profiles/validation-multirole.yaml
+++ b/src/squadops/contracts/cycle_request_profiles/profiles/validation-multirole.yaml
@@ -1,0 +1,49 @@
+name: validation-multirole
+description: >
+  SIP-0093 multi-role plan authoring validation. Same framing → gate →
+  implementation shape as `validation`, but turns ON the multi-role authoring
+  path by setting `plan_authoring_contributors`, so development, qa, and
+  strategy each emit `*.propose_plan_tasks` and `governance.merge_plan`
+  reconciles them into the canonical `implementation_plan.yaml` +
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #181, opened 2026-06-20
docs(sip): reply-channel subscription model — Rev 3 (lazy subscription)

jasonLadd:
## Summary
Revises the proposed `SIP-Reply-Channel-Subscription-Model` (S1) from Rev 2 → Rev 3. Design doc only (in `sips/proposed/`), no code.

## What changed
The Rev 2 `ReplyRouter` assumed a fixed agent roster known at orchestrator boot (`start()` subscribed to every agent in the active squad profile). But squad profiles resolve **per-run** (`SquadProfilePort.resolve_snapshot(profile_id)`) and are overridable per run — there is no fixed agent set at startup. Rev 3 switches to **lazy subscription**: the executor calls `ensure_subscribed(agent_id)` before each dispatch; the first call for an agent opens its `{agent_id}_results` subscription, subsequent calls are no-ops.

Threaded the consistency edits through §1, §5.1, §5.3, §5.4, §7 (the startup-ordering risk largely dissolves under lazy subscription). Added `cancel()`/`stop()` to the `ReplyRouter` sketch to match the §8 test plan.

## Context
S1 is the structural fix for the cycle-reply-loss failure class (tactical patch already merged as PR #89). It's been pulled forward because it's the upstream dependency for the MacBook lane's SIP-0089 Phase 2 reserve-buffer guard (both touch `_publish_and_await`). This PR just gets the spec to accept-ready; acceptance (numbering) + implementation are separate steps.

### DIFF
--- a/sips/proposed/SIP-Reply-Channel-Subscription-Model.md
+++ b/sips/proposed/SIP-Reply-Channel-Subscription-Model.md
@@ -3,11 +3,13 @@
 **Status:** Proposed
 **Authors:** SquadOps Architecture
 **Created:** 2026-05-02
-**Revision:** 2
+**Revision:** 3
+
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #183, opened 2026-06-20
fix(#182): generate proposer typed-acceptance vocabulary from CHECK_SPECS

jasonLadd:
Fixes the root-cause prompt gap behind #182.

## Problem
SIP-0093 proposers (`development.propose_plan_tasks`, `qa.propose_plan_tasks`) are told to use the SIP-0092 typed-acceptance vocabulary, but they only ever saw check **names** (in the task-type fragments) — never the param schema. The `{{typed_acceptance_vocabulary}}` placeholder that both proposer templates already render was being fed the **empty string** (`planning_tasks.py:800`, with a comment that wrongly claimed "the fragment carries the vocabulary"). Models guessed param names; `count_at_least` (the only check with two required params, `glob`+`min_count`) failed validation, and because the parser is all-or-nothing, one bad criterion dropped the whole proposal.

Observed in `cyc_05d25747c7d2`: both dev and qa proposers failed identically → merged `implementation_plan.yaml` had `tasks: []`.

## Fix
Generate the vocabulary from `CHECK_SPECS` (the existing single source of truth) and inject it into the slot that was empty. Each check now renders with its exact **required + optional params (with types)** and a **parser-valid flat-YAML example**:

```yaml
### `count_at_least`
- Required: `glob` (string), `min_count` (integer)
- Example:
  - check: count_at_least
    glob: "tests/test_*.py"
    min_count: 3
    severity: error
```

Because it's generated from the registry, it fixes **every** check at once and **cannot drift** from the parser contract — no hand-maintained param prose duplicated in prompt files.

## Changes
- `cycles/acceptance_check_spec.py` — add `example` to `CheckSpec` (populated for all 6 checks) + `render_typed_acceptance_vocabulary()`.
- `capabilities/handlers/planning_tasks.py` — feed the renderer into `{{typed_acceptance_vocabulary}}` (was `""`).
- `tests/unit/cycles/test_typed_acceptance_vocabulary.py` — new:
  - every check's example round-trips through the real `_parse_acceptance_criteria` (guarantees we never show models malformed YAML),
  - every check + its required params appear in the rendered block,
  - `count_at_least` shows both `glob` and `min_count` (direct #182 regression),
  - non-empty guard (against regressing to `""`).

## Verification
Regression: **4023 passed, 1 skipped, 0 failed.**

## Out of scope (tracked elsewhere)
- Parser leniency (skip a malformed criterion instead of dropping the whole proposal) — belongs to **SIP-0093 B′** (the revision loop should re-prompt a failed proposer), not authoring-time validation weakening.
- The empirical acceptance gate — a `validation-multirole` cycle producing a non-empty merged plan — is the close criterion for #182; this PR is the fix, re-validation confirms it.

### DIFF
--- a/src/squadops/capabilities/handlers/planning_tasks.py
+++ b/src/squadops/capabilities/handlers/planning_tasks.py
@@ -30,6 +30,7 @@ from squadops.capabilities.handlers.base import (
 from squadops.capabilities.handlers.cycle_tasks import (
     _CycleTaskHandler,
 )
+from squadops.cycles.acceptance_check_spec import render_typed_acceptance_vocabulary
 from squadops.llm.exceptions import LLMError
 from squadops.llm.models import ChatMessage
 
@@ -797,7 +798,10 @@ class _ProposeBaseHandler(_PlanningTaskHandler):
             "prd": prd,
             "roles_section": roles_section,
             "builder_section": builder_section,
-            "typed_acceptance_vocabulary": "",  # task-type fragment carries the vocabulary
+            # Generated from CHECK_SPECS so the proposer sees exact param names
+            # + a parser-valid example per check (issue #182 — was "", which let
+            # models guess param names and fail count_at_least validation).
+            "typed_acceptance_vocabulary": render_typed_acceptance_vocabulary(),
         }
 
     async def handle(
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #184, opened 2026-06-20
fix: single-quote proposer acceptance-check examples so regex patterns parse

jasonLadd:
## Problem

The SIP-0093 B re-validation cycle `cyc_82c9b3f5a2c1` (the re-run after #182) was meant to produce a non-empty multi-role plan. The **development proposer failed again** — but for a *different* reason than #182:

```
Malformed proposal YAML: while scanning a double-quoted scalar
  pattern: "participant|participants?.*length|\.length"
                                               ^
found unknown escape character '.'
```

Result: `proposal_completeness: partial`, dev proposal dropped, merged plan was **QA-only (0 dev tasks)** with 8 dangling `dev:*` dependencies — unbuildable.

## Root cause

The typed-acceptance vocabulary handed to proposers renders every string example with **double quotes** via `_format_example_value` (`acceptance_check_spec.py`). Double-quoted YAML scalars interpret backslash escapes, but regex patterns are made of them (`\.`, `\w`, `\d`, `\b`). Proposers copy the exemplar's quote style, so the moment one writes a realistic regex, `yaml.safe_load` raises *unknown escape character* and the whole proposal is discarded.

Ironically the **same #182 PR** that fixed the `count_at_least` empty-param bug introduced this renderer and picked the one quote style that breaks regexes. The `retry_yaml_call` path can't recover, because the bad exemplar steers the model right back to double quotes on every retry.

Trace:
1. `acceptance_check_spec.py` `_format_example_value` → `return f'"{value}"'`
2. `render_typed_acceptance_vocabulary()` renders the `regex_match` example double-quoted
3. injected into the proposer prompt at `planning_tasks.py:804` (`typed_acceptance_vocabulary`)
4. proposer LLM (`planning_tasks.py:185`) copies the style for its own regex
5. merge `yaml.safe_load` (`proposed_role_tasks.py:168-170`) raises `Malformed proposal YAML`

## Fix

- **`_format_example_value` → single quotes.** Single-quoted YAML scalars don't process backslash escapes, so `'\.length'` round-trips literally. Embedded `'` escaped as `''`.
- **`regex_match` example now carries a real backslash** (`def test_\w+`), so the rendered exemplar actively teaches the safe style instead of a backslash-free `def test_`.
- **Fixed the duplicate hardcoded exemplar** in `_plan_authoring_service.py` (the merge-author path) the same way — same bug class, second proposer-facing prompt.

## Tests

`tests/unit/cycles/test_typed_acceptance_vocabulary.py`:
- backslash-bearing regex round-trips through `yaml.safe_load` (the exact failure)
- embedded single-quote escaping (`''`)
- every fenced `\`\`\`yaml` example in the rendered vocabulary parses via `yaml.safe_load`
- the regex exemplar keeps its backslash **and** single quotes

243 passing across the affected planning/proposer/merge suites.

## What this validates / leaves open

- #182's `count_at_least` fix is **confirmed** (QA proposed cleanly in `cyc_82c9b3f5a2c1`); this PR removes the *second* blocker.
- A full multi-role (dev+qa+strategy) merge still needs a clean re-run to demonstrate B closure.

## Follow-up (not in this PR)

`_plan_authoring_service.py` hand-maintains a **second full copy** of the typed-acceptance vocabulary that duplicates `CHECK_SPECS`. It should be single-sourced onto `render_typed_acceptance_vocabulary()` so exemplars can't drift again.

### DIFF
--- a/src/squadops/capabilities/handlers/_plan_authoring_service.py
+++ b/src/squadops/capabilities/handlers/_plan_authoring_service.py
@@ -147,7 +147,9 @@ async def produce_plan(
         "- check: regex_match\n"
         '  description: "At least three test functions exist"\n'
         "  file: tests/test_users.py\n"
-        '  pattern: "def test_"\n'
+        # Single-quote regex patterns: double quotes interpret backslash escapes
+        # (\w, \.) and break yaml.safe_load. See acceptance_check_spec.CHECK_SPECS.
+        "  pattern: 'def test_\\w+'\n"
         "  count_min: 3\n"
         "\n"
         "# count_at_least — glob match count under workspace (stack-agnostic)\n"
--- a/src/squadops/cycles/acceptance_check_spec.py
+++ b/src/squadops/cycles/acceptance_check_spec.py
@@ -120,7 +120,10 @@ CHECK_SPECS: dict[str, CheckSpec] = {
         supported_stacks=frozenset(),
         requires_stack_context=False,
         path_params=frozenset({"file"}),
-        example={"file": "tests/test_runs.py", "pattern": "def test_", "count_min": 5},
+        # pattern carries a backslash escape on purpose: the rendered example
+        # must teach proposers the single-quote style for real regexes, since
+        # double-quoting \w / \. is exactly what broke YAML parsing in #182's wake.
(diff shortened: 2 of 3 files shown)


jasonLadd (commented):
reviewed

PR #188, opened 2026-06-20
fix(#187): tolerate malformed optional sections in proposer parser

jasonLadd:
Closes #187.

## Problem

`ProposedRoleTasks.from_yaml` (`src/squadops/cycles/proposed_role_tasks.py`) was **all-or-nothing**: any `ValueError` while parsing an *advisory* section aborted the whole proposal, discarding every valid task. The SIP-0093 multi-role dev proposer hit this **three cycles running**, each on a different field:

| Cycle | Dev proposer failure | Fix |
|---|---|---|
| — | `count_at_least` params | #182 / #183 |
| `cyc_82c9b3f5a2c1` | regex backslash in `"..."` YAML | #184 |
| `cyc_ee7a5dfcdb16` | `brief_conflicts[0]` missing `proposed_change` | **this PR** |

Each time the dev proposal was discarded → QA-only, unbuildable plan (`total_dev_tasks: 0`). `brief_conflicts` is *optional/advisory* (the QA proposer left it `[]`); one fumbled sub-field shouldn't nuke real work.

## Fix (structural — option B from #187)

Optional sections degrade instead of raising; `tasks` and required header fields stay strict.

- **`brief_conflicts`: per-entry tolerance** — a malformed conflict is skipped, valid siblings preserved (extracted `_parse_one_brief_conflict`).
- **str-list sections** (`assumptions` / `risks` / `gaps_not_covered` / `source_artifact_refs`): a wrong-typed section is dropped whole.
- Dropped sections are `logger.warning`'d **and recorded** in a new `ProposedRoleTasks.degraded_sections` field — not silent, so the merger can surface them.
- **`tasks` + required header fields unchanged** — a problem there still raises and the merger drops the proposal (load-bearing stays strict).

## Tests

- Rewrote the two brief_conflict "rejected" tests to the new resilient contract (skipped + recorded, not fatal).
- Added: malformed entry **preserves valid sibling + the load-bearing task**; str-list wrong-type dropped + only-offending-section dropped.
- 2341 unit tests pass; ruff clean.

## Follow-up (not in this PR)

#187 sub-item **(A)**: document the `brief_conflicts` schema in the proposer prompt (à la #182's `render_typed_acceptance_vocabulary`) to reduce how often it's emitted malformed. **(B)** here makes it non-fatal regardless, so this is frequency-reduction, not correctness.

### DIFF
--- a/src/squadops/cycles/proposed_role_tasks.py
+++ b/src/squadops/cycles/proposed_role_tasks.py
@@ -22,6 +22,7 @@ fails, the merger falls back to sole-broker authoring per SIP-0093 §5.4.
 
 from __future__ import annotations
 
+import logging
 import re
 from dataclasses import dataclass, field
 
@@ -31,6 +32,8 @@ from squadops.cycles.implementation_plan import (
     TypedCheck,
     _parse_acceptance_criteria,
 )
+
+logger = logging.getLogger(__name__)
 
 _FOCUS_NORMALIZE_RE = re.compile(r"\s+")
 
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #190, opened 2026-06-20
fix(#189): normalize cross-proposal dependency keys (role display-name vs role-id)

jasonLadd:
## Problem

In SIP-0093 multi-role plan authoring, the merger built each produced task's canonical key with `focus_key()` (which lowercases the focus) but compared the **raw** `depends_on_focus` reference string against that index. So a proposer that wrote the role **display name** (`development:Backend models`) never matched the produced **role-id** key (`dev:backend models`) — the prefix never collapsed.

Result: every such dependency edge was flagged as a dangling "gap_fill candidate" in `merge_decisions.yaml` and the merged tasks landed with `depends_on: []`. This hit on every multi-role cycle once dev proposals started surviving the merge (`cyc_1bbe424095fc`), and it dropped even dev→dev *intra*-proposal edges. The dependency graph was silently discarded; plans only ran because sequential mode masks it.

The root cause is narrower than the issue first framed it: case was never the problem (`focus_key` already lowercases both sides) — it was purely the **role-token prefix** (`development` vs `dev`) plus a normalization **asymmetry** (produce side normalized, resolve side raw).

## Fix

One shared normalization, used by both the produce and resolve sides so they can't drift:

- **`role_to_id()` + `ROLE_NAME_TO_ID`** (`proposed_role_tasks.py`) collapse display-names (`development`→`dev`, `strategy`→`strat`) to role-ids; `qa`/`builder`/`data` pass through unchanged. `focus_key()` now runs the role through it.
- **`canonicalize_dep_ref()`** normalizes a raw `"role:focus"` reference the same way; `_plan_merger` resolves through it (extracted into `_resolve_dependency_edges`, which also keeps `merge_proposals` under the C901 complexity bound).
- **Class-2 (hallucinated targets)** — references to a focus no task produced (e.g. qa's `dev:run api endpoints`) still surface as gap_fill candidates, but the operator note now reports the **normalized** key so trivially-recoverable misses don't drown the real ones.
- **Single-sourcing:** `task_plan._PLAN_AUTHORING_PROPOSER_STEPS` derives its role-id column from `role_to_id`, so the proposer-step vocabulary can't drift from the merger's normalization.
- **Prompt fragments (dev + qa):** pin the dependency prefix to role-ids (`dev:`, never `development:`) to reduce malformed emissions at the source. Defense-in-depth — the code fix is load-bearing regardless of prompt.

## Tests

- Helper unit tests: display-name/case collapse, `canonicalize_dep_ref` idempotence on already-canonical input, bare-string-no-prefix edge.
- Merger regression tests: intra-dev display-name self-reference resolves (the exact `cyc_1bbe424095fc` shape), cross-role display-name resolves, and a hallucinated focus still gap_fills **with the normalized key**.
- Full `cycles` + `capabilities` suites: **2351 pass**, ruff clean.

## Deploy

- **runtime-api** rebuild for the merger fix (the executor/merger runs there).
- **agent-image** rebuild to pick up the prompt-fragment change.

Closes #189

### DIFF
--- a/src/squadops/capabilities/handlers/_plan_merger.py
+++ b/src/squadops/capabilities/handlers/_plan_merger.py
@@ -42,6 +42,7 @@ from squadops.cycles.plan_guidance import PlanGuidance
 from squadops.cycles.proposed_role_tasks import (
     ProposedRoleTasks,
     ProposedTask,
+    canonicalize_dep_ref,
     focus_key,
 )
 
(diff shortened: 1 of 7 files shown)


jasonLadd (commented):
reviewed

Issue #94, opened 2026-05-03
Prefect task names: per-role numbering + consistent format across framing and implementation runs


jasonLadd:
## Problem

Prefect UI task names are inconsistent between framing and implementation runs, and the implementation numbering doesn't communicate progress per-role. Three concrete issues observed on cyc_b7cf604aed46 and earlier cycles.

## Current behavior

Framing run shows tasks with no index, no title:
\`\`\`
data: data.research_context
strat: strategy.frame_objective
dev: development.design_plan
qa: qa.define_test_strategy
lead: governance.assess_readiness
\`\`\`

Implementation run shows tasks with global index and title (better, but flawed):
\`\`\`
dev[0]: Backend Pydantic models & in-memory repository
dev[1]: FastAPI endpoints & validation logic
dev[2]: Frontend Vite config, routing shell & API proxy
dev[3]: React view components & fetch state logic
builder[4]: Packaging, startup scripts & qa_handoff artifact
qa[5]: Backend pytest suite & validation checks
qa[6]: Integration smoke test & cycle validation
\`\`\`

## What's wrong

1. **Framing tasks lack a numeric position and title.** Two cycles down the road I can't tell at a glance whether \`strat: strategy.frame_objective\` is the second or fourth framing task.
2. **Implementation indexes are global across the whole manifest, not per-role.** \`builder[4]\` looks like the fifth builder task; it's the only builder task. \`qa[5]\` looks like the sixth qa task; it's actually the first qa task. The numbers misrepresent position-within-role.
3. **Indexes start at 0.** Operationally clear in code; visually confusing in a UI where humans expect 1-based.

## Proposed shape

Both run types should use \`{role}[{n_in_role}/{total_for_role}]: {title}\`:

**Framing run:**
\`\`\`
data[1/1]: Research context
strat[1/1]: Frame objective
dev[1/1]: Design plan
qa[1/1]: Define test strategy
lead[1/1]: Assess readiness
\`\`\`

**Implementation run (the manifest from cyc_b7cf604aed46):**
\`\`\`
dev[1/4]: Backend Pydantic models & in-memory repository
dev[2/4]: FastAPI endpoints & validation logic
dev[3/4]: Frontend Vite config, routing shell & API proxy
dev[4/4]: React view components & fetch state logic
builder[1/1]: Packaging, startup scripts & qa_handoff artifact
qa[1/2]: Backend pytest suite & validation checks
qa[2/2]: Integration smoke test & cycle validation
[...]

PR #192, opened 2026-06-20
fix(#185): prefix Prefect task labels with agent name; extract task_naming module

jasonLadd:
## Problem

The Prefect graph/Gantt task label prefixed with the **role**, which duplicates the `task_type`'s own domain segment and hides which agent actually ran the task:

```
qa: qa.define_test_strategy          # role repeats the "qa." already in task_type
```

The agent name is strictly more informative and removes the redundancy:

```
eve: qa.define_test_strategy
```

`envelope.agent_id` is already on every dispatched envelope (routing keys on it — `queue_name = f"{envelope.agent_id}_comms"`), and it comes from the squad profile, so the label stays profile-configurable. No new resolution/plumbing.

## What changed

- **New `adapters/cycles/task_naming.py::build_task_name(envelope)`** — a pure function that derives the prefix as `envelope.agent_id or role` (role fallback for correction/repair and gate-boundary envelopes that leave `agent_id` empty; `"unknown"` when neither is present), preserving the focused-subtask `prefix[idx]: focus` formatting.
- **Resolved a triple-duplication.** The label logic existed in **three** sites that drifted independently — the canonical `_build_task_name` staticmethod plus inlined copies in `_execute_sequential` **and `_execute_fan_out`** (the issue only flagged two). All three now route through the one helper. The staticmethod is deleted.

## Why this shape

This is the natural **first strangler slice of #186** (executor decomposition): the naming logic was already a pure staticmethod, so extracting it proves the pattern at near-zero risk. The module is shaped so **#94**'s `{n/total}` numbering (`{agent_id}[n/total]: {title}`) lands as edits to `task_naming.py` rather than re-rewriting the executor.

## Tests

- Moved the 5 naming tests to `tests/unit/cycles/test_task_naming.py`, re-pointed at the helper with agent-name expectations.
- Added cases: agent-name wins over role, `agent_id`-empty falls back to role, no-agent-and-no-role falls back to `"unknown"`.
- Updated the Prefect lifecycle integration assertion to the agent-name label (proves the executor wires the helper end-to-end).
- **4451 unit tests pass**, ruff clean.

> Note: the run excludes 5 pre-existing collection errors (missing `sqlalchemy`/`python-jose` in `tests/requirements.txt`) that reproduce on clean `main` — filed separately as #191, not caused by this change.

## Blast radius

Label-only — routing already uses `agent_id`, so execution is untouched. Deploy = rebuild **runtime-api** only (the executor runs there, not in agent images).

Closes #185

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -23,6 +23,7 @@ from hashlib import sha256
 from typing import TYPE_CHECKING, Any
 from uuid import uuid4
 
+from adapters.cycles.task_naming import build_task_name
 from squadops.cycles.checkpoint import RunCheckpoint
 from squadops.cycles.models import ArtifactRef, Cycle, GateDecisionValue, Run, RunStatus
 from squadops.cycles.plan_delta import PlanDelta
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #199, opened 2026-06-21
feat(SIP-0094 PR 94.1): declare per-agent {agent_id}_results queue at startup (inert, deploy-first)

jasonLadd:
## What & why

**PR 94.1 of the SIP-0094 sequence** (S1 — Per-Agent Reply Queues). This is the deploy-first, **inert** rollout-safety step (decision **D9**): each agent declares its own `{agent_id}_results` reply queue at startup, so the orchestrator never addresses a queue the agent hasn't created.

Agents only ever *publish* to `{agent_id}_results` (never consume from it), so the lazy declaration that covers `{agent_id}_comms` never creates it — it must be declared explicitly.

**Nothing consumes from the new queue yet.** The reply path doesn't change until the 94.3 cutover. The only observable effect of this PR is one extra (idle) queue per agent.

- Plan: `docs/plans/SIP-0094-per-agent-reply-queues-plan.md`
- SIP: `sips/accepted/SIP-0094-Per-Agent-Reply-Queues-Long-Lived.md`
- Motivating incident: `cyc_c9ca088599c0` (reply lost in the poll-loop gap, orphan `cycle_results_*` queue leaked)

## Changes

- `ports/comms/queue.py` — `QueuePort.ensure_queue()` (concrete no-op default) + single-sourced `REPLY_QUEUE_DECLARE_ARGS` constant (**D3**: declare args must be identical everywhere or a durable redeclare hits `PRECONDITION_FAILED`).
- `adapters/comms/rabbitmq.py` — `ensure_queue()` as a thin wrapper over `_get_queue()`; the sole `declare_queue` call routed through `REPLY_QUEUE_DECLARE_ARGS` so comms + results declare byte-identically.
- `agents/entrypoint.py` — `ensure_queue("{agent_id}_results")` once before the consume loop.
- Tests: `ensure_queue` declares with shared args; results/comms args identical (D3); re-declares after channel swap; entrypoint declares the results queue exactly once.
- Docs: hardening-plan S1 row repointed to the accepted SIP + plan.

## Verification

- `tests/unit/comms/ tests/unit/agents/`: **79 passed**.
- Full regression suite: **4044 passed, 0 failed**.

> ⚠️ **CI coverage note:** `scripts/dev/run_regression_tests.sh` (and therefore the new CI gate) does **not** include `tests/unit/comms/` or `tests/unit/agents/`, so this PR's tests are verified locally but are **not** exercised by CI. Worth evaluating adding those dirs to the gate (separately — they may surface pre-existing issues, as `console/` did under newer deps).

## Deploy & rollback

- **Deploy-first:** rebuild + redeploy agents (`rebuild_and_deploy.sh agents`) so `{agent_id}_results` exists before the 94.3 runtime-api cutover. Idempotent declare (D9) means the cutover never assumes the agent pre-declared it.
- **Rollback:** inert — safe to leave deployed; fully compatible with the current `cycle_results_{run_id}` path.

Part of S1 / #146. Followed by 94.2 (subscribe primitive), 94.3 (cutover), 94.4 (soak + cleanup).

### DIFF
--- a/adapters/comms/rabbitmq.py
+++ b/adapters/comms/rabbitmq.py
@@ -10,7 +10,7 @@ import aio_pika
 from aio_pika import Connection, Message, Queue
 
 from squadops.comms.queue_message import QueueMessage
-from squadops.ports.comms.queue import QueuePort
+from squadops.ports.comms.queue import REPLY_QUEUE_DECLARE_ARGS, QueuePort
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #201, opened 2026-06-21
test(ci): gate tests/unit/comms + tests/unit/agents in the regression suite (closes #200)

jasonLadd:
## What

Adds `tests/unit/comms/` and `tests/unit/agents/` to `REGRESSION_DIRS` in `scripts/dev/run_regression_tests.sh`, so the regression suite and the CI gate actually exercise the comms-transport and agent-runtime domains.

## Why

These domains were silently ungated. SIP-0094 PR 94.1 (#199) adds tests in both dirs — they pass locally but the CI on #199 went green **without running them**. The whole S1 substrate track is comms/agents work, so this needs closing for CI to mean anything there.

## Safety check

Both dirs are pure mock-based unit tests:
- test-quality lint: clean
- no `@pytest.mark.{rabbitmq,database,redis,docker,integration}` markers
- no live connections (only a fake `amqp://test` URL against mocked channels)
- 75 tests, 0 skips

Local regression after adding them: **4119 passed, 0 failed**. CI (x86_64, serviceless) is the authoritative confirmation — see the check on this PR.

## Out of scope

Other excluded unit dirs (`auth`, `memory`, `adapters`, `runtime`, `config`, `core`, `observability`, `ports`, `architecture`, `bootstrap`) remain excluded — several need live services or are legacy. Evaluating them is a separate follow-up.

Closes #200

### DIFF
--- a/scripts/dev/run_regression_tests.sh
+++ b/scripts/dev/run_regression_tests.sh
@@ -20,6 +20,8 @@ REGRESSION_DIRS=(
     "tests/unit/prompts/"
     "tests/unit/tools/"
     "tests/unit/agent_foundation/"
+    "tests/unit/agents/"
+    "tests/unit/comms/"
     "tests/unit/capabilities/"
     "tests/unit/cycles/"
     "tests/unit/events/"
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

Issue #198, opened 2026-06-21
Console app + unit tests break on FastAPI/Starlette >=0.136 (router double-include); deps unpinned so CI drifts to it


jasonLadd:
## Summary

26 `tests/unit/console/` tests fail with:

> `AssertionError: Cannot include an APIRouter instance that already includes this router. Did you mean to include a different router?`

(plus `test_app_includes_auth_router` / `test_app_includes_continuum_router` finding no routes, downstream of the same import-time failure).

**Green locally, red in CI.** Local dev (DGX Spark) has `fastapi 0.135.1 / starlette 0.52.1`; CI installed `fastapi 0.138.0` because `tests/requirements.txt` pins `fastapi>=0.104.0` (unbounded). Starlette ≥0.136 added a strict check that rejects including a router that already (transitively) includes its parent — the console app's router graph (`console/app/main.py`, `app.include_router(continuum_api_router)` + `auth_bff_router`) trips it at import time.

First surfaced by the new CI workflow (PR #197 / #149): https://github.com/backspring-labs/squad-ops/actions/runs/27909083358

## Two real problems

1. **Console router wiring is cyclic under newer Starlette.** `console/app/main.py` includes a router that already includes the app/parent (a diamond/cycle). Needs the include graph flattened so a router is mounted exactly once. (The unit-test continuum mock may also contribute — see `tests/unit/console/test_console_main.py:28-29`.)
2. **Dependencies are unpinned**, so CI silently tests against versions the project hasn't adopted. This is why a latent incompat became a hard failure. The project runs 0.135; 0.138 is not adopted.

## Fix

- **Short term (unblock CI):** bound `fastapi` in `tests/requirements.txt` to the supported range (`>=0.104.0,<0.136`) so CI tests the version the project actually runs. (Being done as part of #195's CI to get a green gate.)
- **Real fix:** flatten the console app's router includes so no router is included twice / cyclically; confirm against fastapi ≥0.136; then lift the version cap.
- **Hygiene:** consider a pinned/locked CI requirements set so dep drift can't silently change what's tested (recurring theme — cf. the bogus `opentelemetry-instrumentation-aiohttp` line, also unpinned).

## Reproduce

```bash
pip install 'fastapi>=0.136'   # e.g. 0.138.0
pytest tests/unit/console/ -q   # 26 failed
```

jasonLadd:
### Status split (Spark CI-trust arc)

[...]

PR #203, opened 2026-06-21
ci: pin CI dependencies via ci-constraints.txt for reproducible installs (closes #202)

jasonLadd:
## Why

CI installed unbounded `>=` deps, so every run resolved whatever PyPI shipped latest — **three silent breaks this week** from version drift: the bogus `opentelemetry-instrumentation-aiohttp` (#196 area), fastapi 0.138 vs the console tests (#198), and a2a-sdk 0.4 dropping `new_agent_text_message` (#200). Each was capped one-at-a-time; this fixes the class.

## What

- **`ci-constraints.txt`** — the exact x86_64 / py3.12 resolution of `tests/requirements.txt`, captured from a clean install on the CI target itself (so every pin is guaranteed installable there).
- **CI install** now uses `pip install -r tests/requirements.txt -c ci-constraints.txt`, and the pip cache key includes the lock.
- Local dev is unchanged — `tests/requirements.txt` keeps its flexible `>=` ranges; the lock applies only in CI.

The CI run on this PR is the proof: it installs the pinned set and runs the full regression. Green here = the lock is both installable and test-passing on x86_64.

## Follow-ups (noted, not blocking)

- The scattered `<x` caps (fastapi, a2a-sdk) are now redundant in CI but kept as local-dev guards / supported-range docs; they can fold into the lock-bump process later.
- Renovate/Dependabot to propose deliberate lock bumps (each proven by this CI) is a natural next step.

Closes #202


jasonLadd (commented):
reviewed

PR #204, opened 2026-06-21
refactor(SIP-0094): rename per-agent reply queue _results → _replies

jasonLadd:
## Why

The per-agent reply queue was named `{agent_id}_results` — but that names the queue by *today's* payload (a `TaskResult`), not by *what it is*: the agent's outbound reply channel.

`{agent_id}_replies` is better:
- **Channel, not payload** — parallels the inbound `{agent_id}_comms`.
- **Consistent with the SIP's own vocabulary** — `ReplyRouter`, `reply_queue`, `_publish_and_await`, `REPLY_QUEUE_DECLARE_ARGS`. `_results` was the odd term out.
- **Mode-agnostic** — under the SIP-0088+ duty/ambient/cycle run modes, an agent's outbound channel carries more than task results; `_results` would read wrong the moment a non-task reply flows.

## Why now

94.1 is **inert**: the `_results` queues are empty, have no consumers, and nothing downstream (94.2/94.3) references the name yet. So this is a near-free rename — versus painful once the cutover wires real traffic and code to it.

## Scope

- **Code:** `entrypoint.py` declaration, the D3 constant comment + docstrings in `ports/comms/queue.py` and `adapters/comms/rabbitmq.py`, and the two 94.1 tests.
- **Docs:** SIP body, implementation plan, and the hardening-plan S1 row. Earlier revision-history entries intentionally keep the `_results` name they shipped with.
- `cycle_results_*` (the legacy per-run queue) is **deliberately untouched** — verified by grep.

## Ops after merge

1. Redeploy agents (`rebuild_and_deploy.sh agents`) → they declare `_replies`.
2. Drop the orphan empty `_results` queues (created by the 94.1 deploy; inert, empty).

## Verification

- `tests/unit/comms/ tests/unit/agents/`: **79 passed**. ruff clean.
- Repo-wide grep confirms no agent-reply `_results` remain and `cycle_results_*` is intact.

Follows SIP-0094 94.1 (merged). Precedes 94.2.

### DIFF
--- a/adapters/comms/rabbitmq.py
+++ b/adapters/comms/rabbitmq.py
@@ -90,7 +90,7 @@ class RabbitMQAdapter(QueuePort):
 
         Thin wrapper over :meth:`_get_queue`, so the reply queue is declared
         with byte-identical args to the agent comms queue. Agents call this at
-        startup for their ``{agent_id}_results`` queue — which they only ever
+        startup for their ``{agent_id}_replies`` queue — which they only ever
         publish to, never consume from — so the lazy declaration on
         :meth:`consume` never creates it.
         """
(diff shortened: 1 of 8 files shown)


jasonLadd (commented):
reviewed

PR #208, opened 2026-06-21
SIP-0094 94.2a: QueuePort.subscribe() primitive + SubscriptionHandle (inert)

jasonLadd:
## What

Adds the long-lived subscription primitive for SIP-0094 — the dormant building block the cutover (94.3) routes per-agent replies through. **Inert: nothing subscribes yet.**

- **`SubscriptionHandle`** — lifecycle wrapper over the background delivery task; idempotent `async cancel()`, an `active` flag, never exposes the raw task.
- **`QueuePort.subscribe(queue_name, *, on_message) -> SubscriptionHandle`** — concrete default (**D6**): a background loop over `consume_blocking()` → `on_message`, acking each delivery.
  - Declares the queue first (**D9**) — a subscriber never assumes a peer pre-declared it.
  - Callback exceptions isolated (**D12**): a raising `on_message` is logged, the message is **still acked** (reply messages are never redelivered — a failed callback is a waiter-side logic error, not a transport fault), and the loop keeps consuming.
  - Supports **sync and async** callbacks (the 94.3 cutover resolves a `Future` synchronously).
- **`NoOpQueuePort`** inherits the default and never fires the callback (its `consume` returns nothing).

The native RabbitMQ override (real long-lived consumer + channel-close resubscribe, SIP §7) is **PR 94.2b** — kept separate so the risky adapter code is isolated.

## Tests

`tests/unit/comms/test_queue_subscribe.py` (5 tests): delivery+ack+declare-first (D9), sync-callback support, **callback-exception-doesn't-stop-loop + failed msg still acked** (D12), cancel-stops-delivery + idempotent cancel, NoOp-never-delivers-but-cancels-clean (D6).

Placed in the **already-gated** `tests/unit/comms/` dir (pure comms-domain, no adapter imports) rather than the ungated `tests/unit/adapters/` — see Notes.

## Notes — two pre-existing defects surfaced (filed, not fixed here)

- **#206** — `sqlalchemy` is an **undeclared dependency** (imported by `ports/db.py` + postgres runtime; absent from pyproject + `ci-constraints.txt`). Breaks clean install of postgres persistence; makes 3 adapter test files fail collection in lean envs.
- **#207** — `tests/unit/adapters/` is **not in the CI regression gate** (blocked by #206). That's why 94.2a's tests went into the gated `comms/` dir instead.

## Verification

- `pytest tests/unit/comms/` → 19 passed
- `ruff check` + `ruff format --check` clean on both files
- `lint_test_quality.py tests/unit/comms/` → clean

### DIFF
--- a/src/squadops/ports/comms/queue.py
+++ b/src/squadops/ports/comms/queue.py
@@ -3,10 +3,17 @@ Port interface for queue providers.
 Defines the contract that any queue transport provider must satisfy.
 """
 
+import asyncio
+import inspect
+import logging
+import time
 from abc import ABC, abstractmethod
+from collections.abc import Awaitable, Callable
 from typing import Any
 
 from squadops.comms.queue_message import QueueMessage
+
+logger = logging.getLogger(__name__)
 
 # SIP-0094 D3: canonical declaration args for agent comms/reply queues
 # (`{agent_id}_comms`, `{agent_id}_replies`). Every declaration of these queues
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #210, opened 2026-06-21
SIP-0094 94.2b: native RabbitMQ subscribe() + channel-close resubscribe

jasonLadd:
## What

The native long-lived consumer that the reply-router cutover (94.3) will route per-agent replies through — **THE SIP-0094 fix** (the riskiest part, SIP §7). Still **inert**: nothing subscribes yet.

Overrides the default poll-based `subscribe()` (from 94.2a) with a native consumer that holds **one** `queue.iterator` for the whole subscription, instead of opening a fresh short-lived consumer per reply wait. The per-call path races arriving replies against consumer-tag teardown — that's the reply loss SIP-0094 eliminates.

### Channel-close resubscribe (the hard part)

The shared `RobustChannel` can be swapped under a long-lived consumer on reconnect, stranding it on a dead channel. The reconnect loop treats **any** iterator termination (a channel-close exception, or a clean end) as a re-establish signal:
1. invalidate the stale cached `Queue` handle,
2. re-declare against the live channel (the `_get_queue` channel-swap path),
3. re-open the consumer,
4. bump `_resubscribe_total` (**D7**), surfaced via `health()` so a flapping reply channel is observable.

**Design note:** I deliberately did *not* add a separate `add_close_callback` re-consume. That would race aio_pika's own RobustChannel consumer restoration and double-subscribe. Letting the iterator surface the failure keeps a single consumer and is directly testable.

### Other behavior
- **D12** callback isolation: a raising `on_message` is logged, the message is acked anyway (reply messages are never redelivered), the consumer keeps running.
- Queue declared before consuming (**D9**); sync + async callbacks supported.
- Killed copy-paste: extracted the `QueueMessage` construction shared by `consume`/`consume_blocking` into `_to_queue_message` (now also used by `subscribe`).

## Tests

**Unit** (mocked, no broker — gated `tests/unit/comms/`): happy-path delivery+ack · D12 callback survives + both acked · **channel-close → transparent resubscribe** (`resubscribe_total==1`, stale handle invalidated, post-reconnect message delivered) · cancel tears down the iterator/consumer.

**Integration** (live broker, `tests/integration/adapters/`): **100 replies → one held subscription → zero loss, exactly-once** · **two reply queues, interleaved publishes → no cross-agent delivery, exactly-once**. Both pass locally against the running `squadops-rabbitmq`. They need a broker, so they don't run in the current CI gate (unit-only) — validated locally.

## Verification

- `pytest tests/unit/comms/` → 23 passed; `ruff check`/`format --check` clean; test-quality lint clean.
- `pytest tests/integration/adapters/...TestNativeSubscribeIntegration` → 2 passed (live broker).

## Defect surfaced (filed, not fixed here)

- **#209** — `tests/integration/test_config.env` credentials drift from the deployed broker (`squadops123` vs `squadops-dev`) **and** shadow env overrides, so all rabbitmq/postgres integration tests fail auth by default. (Validated 94.2b's tests by temporarily correcting the password; the file was restored unchanged.)

## Next

**94.3 — CUTOVER**: `ReplyRouter` + rewrite `_publish_and_await` onto `{agent_id}_replies` + DI wiring. Behavior changes there; this PR stays dormant until then.

### DIFF
--- a/adapters/comms/rabbitmq.py
+++ b/adapters/comms/rabbitmq.py
@@ -3,6 +3,7 @@ RabbitMQ adapter implementation for QueuePort.
 """
 
 import asyncio
+import inspect
 import logging
 from typing import Any
 
@@ -10,9 +11,18 @@ import aio_pika
 from aio_pika import Connection, Message, Queue
 
 from squadops.comms.queue_message import QueueMessage
-from squadops.ports.comms.queue import REPLY_QUEUE_DECLARE_ARGS, QueuePort
+from squadops.ports.comms.queue import (
+    REPLY_QUEUE_DECLARE_ARGS,
+    QueuePort,
+    SubscriptionCallback,
+    SubscriptionHandle,
+)
 
 logger = logging.getLogger(__name__)
+
+# SIP-0094 D6/§7: pause before re-establishing a dropped long-lived
+# subscription, so a flapping channel doesn't spin a tight reconnect loop.
+_SUBSCRIBE_RECONNECT_BACKOFF = 0.5
 
 
 class QueueError(Exception):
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #212, opened 2026-06-21
SIP-0094 94.3: ReplyRouter cutover — _publish_and_await onto per-agent reply queues

jasonLadd:
## What

**The behavior-changing cutover.** The orchestrator no longer polls a per-run `cycle_results_{run_id}` reply queue (thousands of short-lived consumers per long task, losing replies in the gaps, leaking an orphan queue per run). It now routes replies through the long-lived per-agent subscription built in 94.1/94.2.

`_publish_and_await` becomes: `ensure_subscribed(agent_id) → register(task_id) → publish to {agent_id}_comms → await the router future`. The reply address in dispatch metadata is now `{agent_id}_replies`.

## Pieces

- **`adapters/cycles/reply_router.py`** — `ReplyRouter` (SIP §5.3) holding one `{agent_id}_replies` subscription per agent, opened lazily on first dispatch, resolving an `asyncio.Future` keyed by `task_id`:
  - **D4** lock-guarded + double-checked `ensure_subscribed` (concurrent first-dispatch → one subscription).
  - **D9** declares the queue via `subscribe` before consuming.
  - **D10** `register` raises `DuplicateRegistration` rather than overwriting a pending future.
  - **D11** `_handle_reply` never strands the consumer — malformed/late replies counted+dropped, a `from_dict` failure fails only that future.
  - **D13** `stop()` sets stopped state (ensure_subscribed/register fail fast), cancels subscriptions, fails pending futures with `ReplyRouterStopped`.
  - **Ack reconciliation:** the 94.2b `subscribe` primitive auto-acks every delivery, so `_handle_reply` must **not** ack — the SIP §5.3 sketch's `await ack` would double-ack (an AMQP error). The router never acks; the test double enforces this (double-ack raises).
- **`TaskResult.from_dict`** hardened to drop unknown keys (**D8**) — a newer agent's reply payload can't crash `_handle_reply`.
- **`dispatched_flow_executor.py`** — `_publish_and_await` rewritten; `reply_router` constructor arg; every exit path (success, timeout, publish failure, router failure, cancellation) drops the pending future so `task_id` never leaks (**#9/#10**).
- **`api/runtime/main.py`** — construct `ReplyRouter(queue_adapter)`, inject into the executor, `stop()` it on shutdown (**D13**). **`factory.py`** threads it through.

## Tests

- **`test_reply_router.py`** (8): register/resolve, duplicate-register (D10), late/malformed drops + survival (D11), `from_dict` failure fails only that future, `stop()` (D13), concurrency-safe subscribe (D4), fail-fast once stopped.
- **`TestPublishAndAwaitInvariants`** (5, new): subscribe+register before publish (D14/#2), publish-failure removes pending future (#10), timeout leaves no pending future (#9), concurrent same-agent dispatch both resolve (#13), cross-run task_id isolation (D14).
- **Full executor-test migration (~12 files).** Replacing the reply mechanism breaks every test that drove a task to completion via the old `queue.consume` poll. Single-sourced a `FakeReplyRouter` double + `reply_router` fixture in `tests/unit/conftest.py`; `bind(mock_queue)` auto-delivers the agent reply on publish; replies configured via `responder`/`results`/`suppress`. Migrated **faithfully**: the real `register→publish→await` path is exercised, scripted per-dispatch sequences (outcome-routing's 7 scenarios, correction-protocol's continue/patch/rewind/abort) carried over **verbatim** via stateful responders, **no assertions changed**. Two tests deleted (their sole purpose was the deleted polling mechanism; coverage moved to 94.2b + `test_reply_router.py`); two would-be **vacuous passes** caught and fixed by re-supplying specific reply content.

## Verification

- `cycles` + `events` gated suites: **1483 passed**.
- Full regression gate (`run_regression_tests.sh`: `ruff check .` + test-quality lint + ~2900 tests): **green (exit 0)**. Prompt-fragment manifest in sync.

## Defect surfaced (filed, not fixed here)

- **#211** — pre-existing: `integration/test_pulse_check_e2e.py` seeds via `loop.run_until_complete()` inside an async test ("event loop is already running"). Confirmed failing on the pre-cutover file; it's in the un-gated integration tree. Its 94.3 reply-router migration is done; only the orthogonal seed bug remains.

## Next

**94.4** — soak on a real long-cycle build (no new `cycle_results_*`, `{agent_id}_replies` drain to zero, late-drop/malformed counters zero), guarded cleanup of the ~95 orphan `cycle_results_*` queues, promote SIP → implemented.

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -50,6 +50,7 @@ from squadops.telemetry.context import use_correlation_context, use_run_ids
 from squadops.telemetry.models import CorrelationContext
 
 if TYPE_CHECKING:
+    from adapters.cycles.reply_router import ReplyRouter
     from squadops.capabilities.acceptance import AcceptanceCheckEngine
     from squadops.capabilities.models import AcceptanceContext
     from squadops.cycles.models import GateDecision, SquadProfile
@@ -82,8 +83,10 @@ class DispatchedFlowExecutor(FlowExecutionPort):
     """Flow executor that dispatches tasks to agent containers via RabbitMQ.
 
     Uses a request/reply pattern: publishes a ``comms.task`` message to
-    ``{agent_id}_comms``, then consumes the ``TaskResult`` from a per-run
-    reply queue (``cycle_results_{run_id}``).
+    ``{agent_id}_comms``, then awaits the ``TaskResult`` via the
+    :class:`ReplyRouter`, which holds a long-lived subscription on the agent's
+    per-agent reply queue (``{agent_id}_replies``) and resolves a future keyed
+    by ``task_id`` (SIP-0094).
     """
 
     def __init__(
@@ -97,10 +100,12 @@ class DispatchedFlowExecutor(FlowExecutionPort):
         llm_observability: LLMObservabilityPort | None = None,
         workflow_tracker: WorkflowTrackerPort | None = None,
         event_bus: CycleEventBusPort | None = None,
+        reply_router: ReplyRouter | None = None,
     ) -> None:
         self._cycle_registry = cycle_registry
         self._artifact_vault = artifact_vault
         self._queue = queue
+        self._reply_router = reply_router
         self._squad_profile = squad_profile
         self._project_registry = project_registry
         self._task_timeout = task_timeout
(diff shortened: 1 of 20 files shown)


jasonLadd (commented):
reviewed

PR #213, opened 2026-06-21
sip: promote SIP-0094 → implemented (per-agent reply queues)

jasonLadd:
Promotes **SIP-0094 (Per-Agent Reply Queues + Long-Lived Subscription Model)** from accepted → implemented. S1 of the 1.0.x build-reliability hardening is complete.

## What shipped
The orchestrator's request/reply path now runs on long-lived per-agent `{agent_id}_replies` subscriptions resolved by an in-process `ReplyRouter`, replacing the per-run `cycle_results_{run_id}` polling loop. This eliminates the two structural failure classes the SIP targeted:
- **Reply loss** from consumer-tag churn (thousands of short-lived consumers per long task).
- **Orphan-queue leak** (one durable `cycle_results_*` left behind per run).

Landed across: #199 (94.1 declare reply queues), #208 (94.2a `subscribe()` primitive), #210 (94.2b native RabbitMQ subscribe + channel-close resubscribe), #212 (94.3 ReplyRouter cutover).

## Soak validation (cyc_2bba7d0729dd, play_game full build)
- ✅ Completed in ~30 min with **6 artifacts across all roles** → every dispatch got its reply (no losses → no timeouts).
- ✅ **Zero `cycle_results_*` created** (count steady at 95 throughout).
- ✅ `_replies` subscriptions held **live (`consumers=1`)** and drained to zero post-run.
- ✅ Router error counters **zero** (malformed / unknown-late / from_dict).

## Orphan cleanup
93 empty orphan `cycle_results_*` queues guard-deleted (`--if-empty --if-unused`). 2 non-empty relics holding stuck pre-cutover replies were preserved by the guard (harmless dead data; the exact failure the SIP fixes).

## Changes
- `sips/accepted/…SIP-0094…` → `sips/implemented/…`
- frontmatter + body Status → Implemented
- `sips/registry.yaml` status → implemented

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #214, opened 2026-06-21
chore(release): bump to v1.0.6 + document SIP-0094

jasonLadd:
## What

Closes out the SIP-0094 (per-agent reply queues) release. The promote-to-implemented PR (#213) merged **without** the version bump + ROADMAP entry that every prior implemented SIP carries, so the framework was left pinned at 1.0.5 with a stale roadmap.

## Changes

- **Version bump → 1.0.6** (`pyproject.toml`). This is the single source — `squadops.__version__` (what agents and runtime-api report) derives from package metadata, so there's nothing else to stamp. Patch bump, consistent with how every prior implemented SIP was released (0084→1.0.1, 0085→1.0.2, 0086→1.0.4, 0087→1.0.5).
- **ROADMAP** — add the **v1.0.6 (SIP-0094)** entry and backfill the two timeline entries that were also missed: **v1.0.5 (SIP-0087)** and **v1.0.4 (SIP-0086)**. Refresh the *Accepted (Next Up)* table (SIP-0086 was listed there but is implemented; the real accepted set is SIP-0088–0093) and the *Stats* block (version + 57 implemented / 6 accepted / 8 proposed from the registry).
- **`version_cli.py`** — drop the dead `__init__.py` rewrite. `__version__` is now derived dynamically, so the regex never matched and every bump emitted a misleading `version pattern not found` warning. pyproject is the only place the version is written.

## Verification

- `version_cli.py bump 1.0.6` updates pyproject cleanly (no warnings after the cleanup).
- After `pip install -e .`, both `version_cli.py version` and `python -c "import squadops; print(squadops.__version__)"` report **1.0.6**.
- `ruff check scripts/maintainer/version_cli.py` passes.

## Follow-up

Deployed agent + runtime-api images report 1.0.5 until rebuilt — a `rebuild_and_deploy.sh all` on merge will pick up 1.0.6.

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -4,7 +4,22 @@ Living document tracking the implementation progression from initial prototype t
 
 ## Release Timeline
 
-### v1.0.3 (2026-04) — Current
+### v1.0.6 (2026-06-21) — Current — Per-Agent Reply Queues
+- **SIP-0094** Per-Agent Reply Queues + Long-Lived Subscription Model
+  - Replaces the leaky per-run `cycle_results_{run_id}` reply queues — which lost replies in the consumer-tag churn window and leaked one orphan queue per run — with durable per-agent `{agent_id}_replies` queues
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

Issue #219, opened 2026-06-27
Normalize unversioned chat routes (/api/chat, /api/agents) onto /api/v1


jasonLadd:
## Problem

SIP-0085 introduced the console messaging routes under **unversioned** prefixes — `/api/chat/*` and `/api/agents/messaging` — diverging from the versioned `/api/v1/*` resource surface used by every other authenticated resource. This was done in a single commit (`57f6305`) with no recorded rationale, and is the one cheaply-fixable deviation in the API-prefix inconsistency captured in the companion standard issue.

Current routes (`src/squadops/api/routes/chat/routes.py:35-36`):
- `POST /api/chat/{agent_id}`
- `GET  /api/chat/sessions/{session_id}/messages`
- `GET  /api/chat/{agent_id}/sessions`
- `GET  /api/agents/messaging`

## Blast radius — low (this is why it's worth doing)

Single browser consumer flowing browser → Caddy → console BFF → runtime-api:
- `console/.../ChatDrawer.svelte` (the only plugin that calls it)
- console BFF proxy handlers (`console/app/main.py` — 2 handlers)
- Caddy rules (`console/Caddyfile` — 2 rules)
- runtime-api router prefixes (`chat/routes.py`)
- chat tests

CLI does not touch it; agents do not touch it.

## Options

1. **Move to `/api/v1`** — `/api/v1/chat/*` + `/api/v1/agents/messaging`. Brings chat onto the resource lane.
2. **Formally exempt** — document a deliberate rationale for chat being unversioned (e.g. SSE/streaming surface treated separately) so it's an intentional decision, not accidental drift.

## Acceptance criteria

- [ ] Decision (move vs. exempt) recorded under the API standard (companion governance issue).
- [ ] If moved: all consumers updated in lockstep — Svelte plugin, BFF handlers, Caddy rules, runtime-api router prefixes, and tests — with no orphaned paths.
- [ ] If exempt: rationale documented alongside the standard.

## Related

- Companion: API prefix/versioning standard issue (governance).
- Surfaced during the SIP-0089 §2.7 API-consistency assessment.

---
_Companion governance issue: #218._

Issue #218, opened 2026-06-27
Define and document the runtime-api URL prefix & versioning standard


jasonLadd:
## Problem

The runtime-api HTTP surface has accreted **four different URL-prefix conventions with no owning standard**. Each was chosen per-SIP in isolation; nothing governs the surface as a whole, and one doc even claims (incorrectly) that it's already consistent.

### Current state

| Lane | Prefix | Auth | What's on it | Owner SIP |
|---|---|---|---|---|
| Resource API | `/api/v1/*` | required (middleware) | projects, cycles, runs, gates, artifacts, squad-profiles, cycle-request-profiles, models | SIP-0064/74/75 |
| Chat | `/api/chat`, `/api/agents` | required | console messaging (SSE) | SIP-0085 |
| Operational | `/health/*` | **none** (allowlisted) | infra probes, agent status + heartbeat ingest, SIP-0089 runtime-state | SIP-0069/0089 |
| Identity | `/auth/userinfo` | required | userinfo (rest of `/auth/*` is the console BFF) | SIP-0062 |

### Why it matters now

- **No SIP/doc/test owns the prefix or versioning policy.** SIP-0064 asserts `/api/v1` for *its* routes only; SIP-0085 introduced unversioned `/api/chat`+`/api/agents` in one commit with no rationale.
- `docs/ideas/pre_1_0_hardening.md:65` **falsely claims** `/api/v1/` is "consistent."
- No exported OpenAPI artifact and no schema-pinning contract test, so drift is uncaught (FastAPI's declared OpenAPI `version` is the framework version, not an API contract version).
- This surfaced while adding the SIP-0089 `assignment` CLI/API surface — we needed a rule for where new resources go and there wasn't one.

## Proposed standard

Classify by **nature of the endpoint**, not by which SIP added it:

| Lane | Prefix | Auth | Rule |
|---|---|---|---|
| Resource API | `/api/v1/<resource>` | required | All authenticated, managed REST resources (CRUD). **Default home for anything new.** |
| Operational probes | `/health/*` | none | Read-only liveness/status + heartbeat ingest **only**. Never a writable business resource. |
| Identity | `/auth/*` | mixed | userinfo / token surface. |

Rules:
- **Don't add `/api/v2`** — extend v1 (aligns with the proposed `SIP-API-Contract-Hardening` "get v1 right before it's stable" stance).
- New writable business resources → `/api/v1/`.
[...]

PR #221, opened 2026-06-27
SIP-0089 Phase 2 — Agent Runtime State (assignments, scheduler, coordinator, reserve guard)

jasonLadd:
## SIP-0089 Phase 2 — Agent Runtime State

Implements Phase 2 of SIP-0089: durable **Assignments** with duty windows, the scheduler/coordinator that act on them, and the cycle reserve-buffer guard — plus the operator CLI/REST surface to inspect and (experimentally) create them. Targets v1.1.0 (no version bump until Phase 4 per the plan).

Branch starts from the accepted SIP on `main`; 11 incremental commits, one per phase section.

### What's included (§2.0–§2.8)

| § | Commit | Summary |
|---|---|---|
| 2.0 | `59fd93c` | Spike: recruitment seam vs. the post-SIP-0094 dispatch path |
| 2.1 | `f653b52` | `Assignment`/`DutyWindow` models + `window_state()` 6-state classifier |
| 2.2 | `041bd5c` | `agent_assignments` migration (1110) + `agent_id` holder on `Assignment` |
| 2.3 | `31b5c35` | `AssignmentPort` + `PostgresAssignment` adapter (time filtering pushed into SQL) |
| 2.6 | `ef964fc` | `RuntimeCoordinator` — the D16 sole mode-transition authority (idempotent per D12, event/reason distinct per D18) |
| 2.4 | `61aeb39` | `DutyScheduler` — polling claimant (D21), opens/closes windows via coordinator, full missed-window policy table; inert until `start()` |
| 2.5 | `45d8d48` | Reserve-buffer recruitment guard — an upcoming **hard** duty window defers cycle recruitment (`RUN_PAUSED` + typed payload `reason=upcoming_hard_duty_window`; no new event type minted) |
| 2.7 | `00b6399` | `squadops assignment list/show/create` CLI + `/api/v1` REST surface |
| 2.8 | `f133f78` | Close-out: gate the new test dirs in CI + cardinality acceptance test |

### API surface + a prefix-standard decision

§2.7 needed an HTTP route because the CLI is forbidden from importing Postgres adapters directly (D26). Choosing where to put it surfaced that the runtime-api had drifted into **four** URL-prefix conventions with no owning standard. Rather than add a fifth, we adopted a lane standard:

- **`/api/v1/<resource>`** — authenticated, managed REST resources (where assignments live).
- **`/health/*`** — read-only, *unauthenticated* operational probes only (so a writable `POST` must not live here — it'd be unauthenticated).
- **`/auth/*`** — identity. **No `/api/v2`** — extend v1.

Assignments therefore sit on `/api/v1`: `GET /api/v1/agents/{id}/assignments`, `GET /api/v1/assignments/{id}`, `POST /api/v1/assignments` (experimental/internal). The standard is enforced going forward by a new **CLAUDE.md API-conventions rule** (`f67e98e`); the rest of the standardization (an executable architecture test, chat-route normalization, error-shape envelope) is tracked in **#218** / **#219**.

### Testing

- **DTO-purity:** serialization lives in the API layer (timedeltas → integer seconds, `DutyWindow` nested, modes tuple→list); round-trip preserves every field.
- **Reserve defaults (§11.4/D7):** 15m hard / 0 soft applied at create; explicit-zero is honored, not overridden.
- **CI gate (#220):** added `tests/unit/runtime/` + `tests/unit/architecture/` to `REGRESSION_DIRS` — previously these (the whole SIP-0089 suite + the D26 forbidden-imports test) did **not** run on PRs. Both dirs are green (78 tests) and pass the test-quality linter.
- **Acceptance audit:** 14/15 §2.8 items were already covered; the one gap (cardinality — an agent may hold multiple assignments) is now tested.

### Note on the full regression suite

The local Python 3.11 venv cannot produce a clean full-suite green — `TestClient`-based route tests break under the drifted deps, and there's a 3.11-specific asyncio hang in `tests/unit/cycles/test_correction_protocol.py`. Both are pre-existing and tracked in **#217**; neither is caused by this work. CI (pinned deps) is the authoritative gate, and the newly-gated dirs were verified green locally without `TestClient`.

### Linked issues

- **Closes #220** — the CI gate for `tests/unit/runtime/` + `tests/unit/architecture/` is fully contained in this PR.
- Follow-up (track:spark, not closed here): #218 (API prefix standard + enforcement), #219 (chat-route normalization).
- Context: #217 (local↔CI env reconciliation), #186 (executor decomposition — follow-up, not in scope).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -205,12 +205,19 @@ python scripts/maintainer/update_sip_status.py sips/accepted/SIP-0067-My-Feature
 **Proactive guidance**: If you observe a workflow or code best practice being bypassed, call it out early — don't wait to be asked. Examples:
 - Workflow: developing on main instead of a feature branch, skipping tests, hardcoding secrets
 - Code structure: copy-pasted logic that should be a shared helper, inconsistent patterns across similar modules, missing registry updates when adding new entries, constants duplicated across files instead of single-sourced
+- Cross-cutting surface drift: adding a new HTTP route prefix, API error shape, event type, or `SQUADOPS__*` config-var convention that diverges from the existing standard — flag it and conform; never justify a new variant by "it doesn't collide" or "it's easy" (this is exactly how the runtime-api drifted into 4 URL conventions — see API Conventions below, #218)
 
 ## Repository Rules
 
 **Read-Only Areas**:
 - Never modify `dist/` or generated metadata (`manifest.json`, `agent_info.json`)
 - Version bumps via `scripts/maintainer/version_cli.py` only
+
+**API Conventions** (runtime-api HTTP surface):
+- Before adding or moving any route, read the **whole** existing surface — do not reason only about the neighborhood. Conform to the lane standard; if no standard covers your case, surface the gap and propose it **before** adding (#218).
+- **Lanes:** authenticated, managed REST resources → `/api/v1/<resource>` (default home for anything new). `/health/*` = read-only, unauthenticated operational probes/heartbeats **only** — never a writable business resource (it's the only no-auth lane). `/auth/*` = identity. **Do not add `/api/v2`** — extend v1.
+- A new prefix/variant is a deliberate, justified decision, never a default. "It doesn't collide" is not a justification.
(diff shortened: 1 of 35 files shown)


jasonLadd (commented):
reviewed

PR #227, opened 2026-06-27
SIP-0089 §2.4: activate the in-process duty scheduler (+ close-sweep fix, #226)

jasonLadd:
## Summary

Activates the SIP-0089 §2.4 **duty scheduler** — the dormant `DutyScheduler` + `RuntimeCoordinator` (built and unit-tested in #221 but constructed nowhere outside tests) are now wired into the runtime-api process so duty windows drive **live ambient↔duty mode transitions**. Validated end-to-end on the dev stack.

Live activation also exposed a latent bug in the merged §2.4 scheduler — agents stranded in `duty` when `reserve_after_window=0` (the hard-duty default) — fixed here as a second commit.

**Closes #226.**

## What's in it

**Activation (`9ec1ddc`)**
- `runtime.scheduler.{enabled,poll_interval_seconds}` config (`SchedulerConfig`/`RuntimeConfig`). **Opt-in** — `enabled` defaults `false` so a deployment activates duty mode deliberately. Verified the new nested `runtime` model doesn't collide with the flat `runtime_api_url` env path.
- `LoggingRuntimeEventPublisher` (`adapters/events/`) — records runtime-state events (`mode_transition`, `window.skipped/review_required`) as structured logs. Best-effort per D22 (never raises). Deliberately **not** bridged onto the locked cycle-event taxonomy (separate vocabulary).
- `create_duty_scheduler(config, pool)` (`api/runtime/scheduler_bootstrap.py`) — the composition-root activation gate (returns `None` when disabled or pool-less). `main.py` starts the loop in `startup` (background asyncio task, same pattern as the health reconciliation loop) and `stop()`s it first in `shutdown`.

**Close-sweep fix — #226 (`0b5f2a1`)**
- **Bug:** with `reserve_after_window=0`, an opened agent never closed. `tick()` only iterated `list_active_assignments(now)`, whose upper bound is `now < window_end + reserve_after`; at `reserve_after=0` the assignment leaves the active set the instant `window_end` passes — before any tick observes `window_state == in_reserve_after` — so the close branch was unreachable. Unit tests missed it (they drove `tick()` with the assignment already in the mocked active set).
- **Fix:** a disjoint **pass-2 close-sweep**. New `AssignmentPort.list_assignments_to_close(now)` (PostgresAssignment JOINs `agent_runtime_state` for `mode='duty'` + `now >= window_end + reserve_after`); `DutyScheduler.tick()` closes each via the coordinator (D21 — requests, never writes). Complementary to the active-set upper bound, so the two passes never overlap. Also acts as a catch-up if a `reserve_after>0` close was missed during downtime.

**Enable in dev stack (`f70d82e`)**
- `docker-compose.yml` runtime-api env: `SQUADOPS__RUNTIME__SCHEDULER__ENABLED=true`, 10s poll. runtime-api is the **single central writer** of agent mode (D16).

## Live validation (dev stack, real agents)
- `ambient → duty` on window open ✓
- `duty → ambient` on close — both *with* a reserve buffer **and** with the `reserve_after=0` default (the #226 case) ✓
- Other agents stay `ambient`; heartbeats don't revert mode (D17) ✓

## Tests
- 14 new tests: config gate + `ge=1` poll constraint; publisher structured-fields + best-effort contract; activation gate (disabled/no-pool/inert-on-build); close-sweep (fires `duty→ambient` from an empty active set; empty sweep is a no-op); adapter JOIN/predicate/binding.
- 87 runtime + architecture tests green locally; D26 forbidden-imports + test-quality linter clean. Full regression runs in CI (local 3.11 can't, per #217).

## Single-writer constraint (deferred)
The coordinator is the sole writer of `mode` (D16) — safe with **one** runtime-api instance (today's deployment). Scaling to multiple replicas would need leader election (a Postgres advisory lock). Deferred while single-instance and gated behind the `enabled` flag; documented at the bootstrap call site and `.env.example`.

## Follow-ups
- Closes #226.
- **#222** — scheduler auto-resume of duty-deferred PAUSED runs on window close is a natural close-sweep follow-up (not in this PR).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -46,6 +46,13 @@ SQUADOPS_PROFILE=dev  # Options: dev, local, lab, cloud
 # Runtime API
 # SQUADOPS__RUNTIME__API__URL=http://runtime-api:8001
 
+# Duty-transition scheduler (SIP-0089 §2.4) — opt-in. When enabled, the
+# runtime-api process runs an in-process scheduler that opens/closes duty
+# windows by writing agent mode (ambient<->duty). Run exactly one runtime-api
+# instance with this enabled (single central mode-writer; no leader election).
+# SQUADOPS__RUNTIME__SCHEDULER__ENABLED=false
+# SQUADOPS__RUNTIME__SCHEDULER__POLL_INTERVAL_SECONDS=30
+
 # Prefect Configuration
 # SQUADOPS__PREFECT__API__URL=http://prefect-server:4200/api
 
--- a/adapters/events/runtime_event_publisher.py
+++ b/adapters/events/runtime_event_publisher.py
@@ -1,0 +1,53 @@
+"""Logging-backed `RuntimeEventPublisher` adapter (SIP-0089 §2.6).
+
+The `RuntimeCoordinator` and `DutyScheduler` emit runtime-state events
+(`runtime_state.mode_transition`, `assignment.window.skipped`, …) through the
+`RuntimeEventPublisher` port. This adapter records each one as a structured log
+record so an activated scheduler's transitions are observable in runtime-api
+logs (which SIP-0087 forwards) without coupling the runtime layer to any
+particular sink.
+
+Runtime-state events are a **separate vocabulary** from the locked cycle-event
+taxonomy (D14/D18), so they are deliberately *not* bridged onto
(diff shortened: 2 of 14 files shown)


jasonLadd (commented):
reviewed

PR #228, opened 2026-06-27
SIP-0089: mirror offline runtime_status in reconciliation so the duty guard sees dead agents (#159)

jasonLadd:
## Summary

Fixes **#159**: a crashed/timed-out agent's `agent_runtime_state.runtime_status` stayed frozen at `online`, because the heartbeat mirror only fires on an **inbound** heartbeat and the reconciliation loop (which marks dead agents offline in `agent_status`) never touched `agent_runtime_state`.

This now matters more than at Phase 1: the coordinator's §11.3 **"offline can't enter duty"** guard reads the *stored* `runtime_status`, and the duty scheduler went live in #227. With the field stuck at `online`, the scheduler could open a duty window for a **dead agent**.

## Fix (Option 1 — mirror parity)
- **`RuntimeStatePort.mark_offline(agent_id)`** — sets `runtime_status='offline'` and nothing else. It must NOT bump `last_heartbeat_at` (a dead agent's last beat is evidence) or touch coordinator-owned fields (`mode`, `focus`, `current_assignment_ref`, `current_runtime_activity_id`) — D17/D16. `PostgresRuntimeState` implements it as a guarded `UPDATE ... WHERE runtime_status != 'offline'` (idempotent; never `INSERT`s a row for an agent that never heartbeated). Returns the row, or `None` when absent/already-offline.
- **`health_checker`** — extracted the reconciliation body into a testable `_reconcile_once()`, which now mirrors every offline agent via `mark_offline` after releasing the read connection. Best-effort + failure-isolated, exactly like the existing heartbeat mirror.

## Live validation (dev stack)
Rebuilt runtime-api, stopped the `data` agent at `20:33:14Z`; at `20:34:59Z` (~105s, within the 90s timeout + ≤45s reconcile) `agent_runtime_state` flipped to **`mode=ambient, runtime_status=offline`** — status mirrored, `mode` preserved (D17 held). Restarted it → recovered to `online` in ~6s via the heartbeat path.

## Tests
- Adapter: `mark_offline` writes only `runtime_status` (not heartbeat/mode/assignment_ref), carries the idempotent guard, and returns `None` when absent/already-offline.
- Reconciliation: a timed-out agent is mirrored offline; a live agent is not.
- Updated the two `RuntimeStatePort` test fakes for the new method.
- 106 runtime + health-checker + architecture tests green locally; D26 + test-quality linter clean. Full regression in CI.

Closes #159.

### DIFF
--- a/adapters/persistence/runtime/state_postgres.py
+++ b/adapters/persistence/runtime/state_postgres.py
@@ -129,6 +129,23 @@ class PostgresRuntimeState(RuntimeStatePort):
             )
         return _row_to_state(row)
 
+    async def mark_offline(self, agent_id: str) -> AgentRuntimeState | None:
+        """Set runtime_status='offline' only — never bump last_heartbeat_at or
+        coordinator-owned fields (D17/D16), and never INSERT (an agent that never
+        heartbeated has no row). The `runtime_status != 'offline'` guard makes
+        repeated reconciliation ticks idempotent (already-offline → no row → None).
+        """
+        async with self._pool.acquire() as conn:
+            row = await conn.fetchrow(
+                "UPDATE agent_runtime_state SET runtime_status = 'offline', updated_at = now() "
+                "WHERE agent_id = $1 AND runtime_status != 'offline' "
+                "RETURNING agent_id, mode, runtime_status, focus, "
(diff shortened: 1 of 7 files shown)


jasonLadd (commented):
reviewed

Issue #153, opened 2026-06-11
Remove SQLAlchemy types from DbRuntime port (ports/db.py)


jasonLadd:
## Problem
`src/squadops/ports/db.py:10-11` imports `sqlalchemy.Engine` and `sqlalchemy.orm.sessionmaker`, and the `DbRuntime` port exposes them as abstract properties. A port interface exposing a specific ORM's types couples the domain to SQLAlchemy — swapping drivers means rewriting the port, not just the adapter, and domain code can accidentally rely on non-portable session features.

## Proposed fix
Replace the `engine` / `session_factory` properties with a portable session abstraction (e.g. `DbSession` with `execute()` / transaction semantics), implemented by the Postgres adapter. Domain code should never import SQLAlchemy.

_Source: architectural review, 2026-06-11 (boundary cleanup)._

jasonLadd:
Closing **into #234**. The leaky accessors flagged here are a **subset** of #234 (retire legacy SQLAlchemy `DbRuntime` vs de-leak the port) and are **test-only**; they'll be resolved by whichever direction #234 takes. Consolidating to avoid a split discussion.

Note: #234 is blocked on Jason's architectural decision (retire-vs-de-leak), so the cleanup itself isn't actionable until that lands.

PR #229, opened 2026-06-27
fix(deps): declare sqlalchemy and python-jose (unblock unit collection)

jasonLadd:
## Problem (closes #206, #191)

`sqlalchemy` (`adapters/persistence/postgres/runtime.py` — `create_engine`/`sessionmaker`) and `python-jose` (`adapters/auth/keycloak/auth_adapter.py` — JWT validation) are **production imports** that were declared **nowhere** — not in `pyproject.toml`, `ci-constraints.txt`, or `tests/requirements.txt`. They were satisfied only **transitively** (via Prefect in the Docker images), so in a minimal or CI environment the adapters fail to import and `pytest tests/unit` **aborts at collection**:

```
ERROR tests/unit/adapters/test_persistence.py      - No module named 'sqlalchemy'
ERROR tests/unit/auth/test_keycloak_adapter.py     - No module named 'jose'
Interrupted: 2 errors during collection
```

Adapter unit tests aren't in the CI regression gate (#207), so CI never surfaced this — only a clean checkout / fresh venv hits it.

## Fix

Declare both explicitly across all three layers:

| Layer | Change |
|---|---|
| `pyproject.toml` | new `postgres` / `auth` optional-dependency extras |
| `tests/requirements.txt` | `sqlalchemy>=2.0`, `python-jose[cryptography]>=3.3` |
| `ci-constraints.txt` | pin 2 direct + 3 transitive: `SQLAlchemy==2.0.51`, `python-jose==3.5.0`, `ecdsa==0.19.2`, `rsa==4.9.1`, `greenlet==3.5.3` |

Declared as **adapter extras** (not core deps) because they're adapter-specific — consistent with the existing `langfuse`/`cli`/`pulse` extras and the hexagonal layering. The `sqlalchemy` leak into the domain port `ports/db.py` is a separate cleanup (#153); declaring the dep here doesn't bless it.

Pins were resolved from a clean install; all are pure-Python except `greenlet` (version-identical across platforms), so the local resolution matches what x86_64 CI installs.

## Validation

- `pytest tests/unit --collect-only`: **2 errors → 0** (4617 → 4650 collected; +33 = the now-runnable adapter/auth tests)
- the 33 formerly-aborting adapter/auth tests **pass**
- regression suite **green** (4253 passed, 1 skipped)
- `pip install ... -c ci-constraints.txt` resolves the new deps to exactly the pinned versions (no conflicts; `cryptography==49.0.0` already pinned satisfies `python-jose[cryptography]`)

## Lane / sequencing

First step of the Spark CI-trust arc (#206+#191 → #198 → #217 → #196 → #207 → #211/#209). Touches shared files (`pyproject.toml`, `ci-constraints.txt`, `tests/requirements.txt`) — additive only, no edits to MacBook-lane `runtime/*`.

### DIFF
--- a/ci-constraints.txt
+++ b/ci-constraints.txt
@@ -41,6 +41,7 @@ deprecation==2.1.0
 Django==6.0.6
 docker==7.1.0
 docutils==0.22.4
+ecdsa==0.19.2
 execnet==2.1.2
 factory_boy==3.3.3
 Faker==40.23.0
@@ -53,6 +54,7 @@ fsspec==2026.6.0
 google-api-core==2.31.0
 google-auth==2.55.0
 googleapis-common-protos==1.75.0
+greenlet==3.5.3
 grpcio==1.81.1
 h11==0.16.0
 hf-xet==1.5.1
@@ -146,6 +148,7 @@ pytest-redis==4.0.0
 pytest-xdist==3.8.0
 python-dateutil==2.9.0.post0
 python-dotenv==1.2.2
+python-jose==3.5.0
 python-multipart==0.0.32
 pytokens==0.4.1
 PyYAML==6.0.3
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #241, opened 2026-06-28
fix(tests): repair test_pulse_check_e2e.py — event loop + 3 stale-API layers (#211)

jasonLadd:
Closes #211.

## What the issue described
`tests/integration/cycles/test_pulse_check_e2e.py::_make_registry` called `loop.run_until_complete(_seed())` inside the already-running pytest-asyncio loop → `RuntimeError: This event loop is already running`, 5 tests fail at setup.

## What actually surfaced
The seeding error **masked three more layers of staleness** — because `_seed()` was never awaited, none of the calls inside it (or downstream) ever ran. This file is in `tests/integration/` (not in the CI gate), so the drift accumulated silently. All four fixed:

1. **Event loop** — `_make_registry` is now `async`, awaited by each test.
2. **Registry API** — `create_cycle`/`create_run` take domain objects now, not the old `project_id=...`/`cycle_id_override=...` kwargs. Build a real `Cycle`/`Run`.
3. **SquadProfile API** — now requires `name`/`description`/`version`/`created_at`.
4. **Executor contract** — the executor resolves the squad via `resolve_snapshot() -> (profile, ref)`; the stale mock set `get_profile`, so the 2-tuple unpack got an empty mock (`not enough values to unpack (expected 2, got 0)`). Also gave the squad real agents so a `development` task exists to trigger the `post_dev` pulse boundary.

Setup now mirrors the canonical `tests/unit/cycles/test_dispatched_flow_executor_observability.py`.

## Verification
- `tests/integration/cycles/test_pulse_check_e2e.py` → **5 passed**.
- `ruff check` + `ruff format --check` clean on the file.

## Follow-up (separate)
These tests still aren't in any gate, so they could rot again — filing a separate issue on gating serviceless integration tests (relates to #207). Not folded in here.

### DIFF
--- a/tests/integration/cycles/test_pulse_check_e2e.py
+++ b/tests/integration/cycles/test_pulse_check_e2e.py
@@ -27,8 +27,11 @@ try:
 except ImportError:
     _has_runtime_deps = False
 
+from squadops.cycles.lifecycle import compute_config_hash
 from squadops.cycles.models import (
+    AgentProfileEntry,
     Cycle,
+    Run,
     RunStatus,
     SquadProfile,
     TaskFlowPolicy,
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #243, opened 2026-06-28
fix(tests): integration config — env precedence + creds match the stack (#209)

jasonLadd:
Closes #209.

## The bug

Every RabbitMQ/Postgres integration test failed at connect with `ProbableAuthenticationError: ACCESS_REFUSED`, even against a healthy stack — two coupled drift problems:

1. **Credential drift.** `test_config.env` (and several fallback literals) hardcoded `squadops123`, but the docker-compose stack uses `squadops-dev`.
2. **The trap.** `load_test_config()` read `test_config.env` **first** and only fell back to `os.getenv()` when the file was *absent* — so `RABBITMQ_PASSWORD=... pytest` was silently ignored and the stale file literal always won (the opposite of normal env precedence).

## The fix

- Reworked `load_test_config()` precedence to **env > file > defaults** (env now wins). Defaults centralized and matched to the stack (`squadops-dev`).
- Updated all stale `squadops123` literals → `squadops-dev` (`test_config.env`, README, the postgres test default).
- Pointed `ExternalPostgres.get_connection_url()` at `load_test_config()` so it can't re-drift from a separate hardcoded URL.
- **Secondary (issue note):** the *optional* RabbitMQ management health-check ran `timeout=10/retries=2` (~22s) and blocked session setup when port 15672 is down. It's non-blocking, so capped to `timeout=3/retries=1` (~3s).

## Verification (against the live stack)

- `tests/integration/adapters/test_rabbitmq_adapter.py` → **8 passed** (was `ACCESS_REFUSED`).
- Env-override precedence confirmed (`RABBITMQ_PASSWORD` env value wins over the file).
- First-test setup **22.1s → 3.1s**.

### DIFF
--- a/tests/integration/README.md
+++ b/tests/integration/README.md
@@ -38,14 +38,14 @@ Before running integration tests, ensure these services are running:
 1. **PostgreSQL** (port 5432)
    - Database: `squadops`
    - User: `squadops`
-   - Password: `squadops123`
+   - Password: `squadops-dev`
 
 2. **Redis** (port 6379)
    - Default configuration
 
 3. **RabbitMQ** (port 5672)
    - User: `squadops`
-   - Password: `squadops123`
+   - Password: `squadops-dev`
    - Management UI: http://localhost:15672
 
 4. **Ollama** (port 11434)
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

Issue #242, opened 2026-06-28
Serviceless integration tests rot undetected — nothing runs them in CI


jasonLadd:
## Problem

`tests/integration/cycles/test_pulse_check_e2e.py` was broken across **four** independent layers (event loop, registry API, SquadProfile API, executor `resolve_snapshot` contract) and nobody noticed — see #211 (PR #241). Root cause of the silent rot: the CI regression gate (`REGRESSION_DIRS` in `run_regression_tests.sh`) is **unit-only**, so nothing exercises `tests/integration/` on PRs.

Yet this file (and others like it) is **serviceless** — it uses `MemoryCycleRegistry` + mocked queue/vault, no real Postgres/RabbitMQ. It *could* run in CI; it just isn't.

## Why it's not trivial

`tests/integration/conftest.py` has a service health-check (observed: a ~20s RabbitMQ management retry on `localhost:15672`) that runs even for mocked tests — so you can't just append `tests/integration/` to the gate without either standing up services in CI or making that check conditional.

## Options

1. **Mark + gate the serviceless subset:** add a marker (e.g. `@pytest.mark.serviceless` or reuse existing markers) to integration tests that don't need real services, make `tests/integration/conftest.py` skip the service health-check for them, and add a CI lane that runs `-m serviceless`.
2. **Relocate:** move genuinely-mocked "integration" tests under `tests/unit/` (they're unit tests by the project's own definition — `tests/unit/conftest.py` says real-service tests live in `tests/integration/`).
3. **Service-up CI lane:** spin up Postgres/RabbitMQ/Redis in a CI job and gate the full `tests/integration/` tree (heaviest; separate from the serviceless goal).

Option 1 or 2 is the cheap win that prevents silent rot.

## Context
- Follows #211 (PR #241), relates to #207 (adapters added to the gate) and the unit-only gate design noted in CI setup.
- Lane: track:spark (CI-trust).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jasonLadd:
Measured instance, 2026-08-25: **#1099** — 16 integration tests (6 failed, 10 errors) fail identically on every run of `main` at v1.6.3, in four classes (stale FK teardown, a mocked executor name that no longer exists, two 0.x version assertions). Owner-ruled into 1.7 alongside this issue; #1099 carries the per-test table and proposes the structural fix (run against compose in CI, or a curated always-green subset with its own runner, and 'some may fail' comes out of CLAUDE.md).

PR #246, opened 2026-06-28
test(comms): regression tests for RabbitMQ channel recovery (#146)

jasonLadd:
Refs #146 (closed separately — see below). Refs #242, #245.

## TL;DR
#146 ("agent spins forever on `Channel was not opened`") is **already fixed** by the SIP-0094 work that landed after the issue was filed. This PR adds the **regression tests that were missing** (the absence of which let it become a multi-day incident). No production code change.

## Reproduction (live broker)
| Failure mode | Result |
|---|---|
| Channel closed (connection alive) | recovers instantly |
| Server-side channel death — the literal `Channel was not opened` | recovers instantly |
| Full connection drop (`rabbitmqctl close_all_connections`) | recovers in ~3s via `connect_robust` |

Why it's fixed: `connect_robust` (auto-reconnect) + `_ensure_connection()` recreating the channel on `is_closed` + `_get_queue()` re-declaring a stale cached queue when its channel no longer matches.

## What this adds
`TestChannelRecovery` — two deterministic, broker-only tests (no docker-exec): publish to the durable queue, kill the channel, assert a later `consume()` still delivers the message (only possible if the channel was rebuilt + queue re-declared).

- Run locally: **2 passed**; full file **10 passed**; ruff clean.

## Caveats / follow-ups
- These are integration tests, so they **don't gate in CI yet** — gating serviceless/integration tests is tracked in **#242**.
- Reproduction surfaced that **`publish()` does not retry during the reconnect window** (distinct from #146) — tracked in **#245**.

### DIFF
--- a/tests/integration/adapters/test_rabbitmq_adapter.py
+++ b/tests/integration/adapters/test_rabbitmq_adapter.py
@@ -336,3 +336,60 @@ class TestNativeSubscribeIntegration:
         # Exactly-once within each queue.
         assert sorted(json.loads(p)["i"] for p in alpha) == list(range(per_queue))
         assert sorted(json.loads(p)["i"] for p in beta) == list(range(per_queue))
+
+
+@pytest.mark.integration
+class TestChannelRecovery:
+    """#146: a dead channel must be rebuilt on the next consume(), not spun on.
+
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #247, opened 2026-06-28
fix(handlers): stop in-place mutation of frozen HandlerResult in planning retry (#155)

### LINKED ISSUE #155: Fix in-place mutation of frozen TaskResult.outputs in planning/wrapup handlers
## Problem
`TaskResult` is a frozen dataclass, but handlers mutate its nested `outputs` dict in place (frozen does not freeze nested containers):
- `src/squadops/capabilities/handlers/planning_tasks.py:437` — `result.outputs["artifacts"][0]["content"] = content`
- `planning_tasks.py:~558` — `result.outputs["brief_outcome"] = {...}`
- similar patterns in `wrapup_tasks.py`

This silently breaks the immutability contract. The moment results are cached, retried, or shared across handlers/coroutines, these mutations become nondeterministic bugs.

## Proposed fix
- Add a small helper, e.g. `update_result_output(result, key, value) -> TaskResult` using `dataclasses.replace()`, and convert all in-place mutations.
- Sweep for the pattern: `grep -rn "result.outputs\[" src/squadops/capabilities/handlers/`.

_Source: architectural review, 2026-06-11._

jasonLadd:
Closes #155.

## What
`GovernanceReviewPlanHandler.handle()`'s YAML-frontmatter retry path mutated the nested `outputs["artifacts"][0]["content"]` of the **frozen** `HandlerResult` returned by `super().handle()` in place (`frozen=True` doesn't freeze nested containers). Replaced with an immutable rebuild via `dataclasses.replace()`.

## Scope (reproduce-first)
A broad sweep — `grep -rE '\.outputs\[[^]]+\] *=' src/squadops/` — shows **this was the only remaining in-place mutation**. The other sites #155 cited (`brief_outcome` in planning, the `wrapup_tasks` ones) were already refactored to reads. So this is a targeted fix rather than the general `update_result_output()` helper the issue proposed (one nested site doesn't warrant a helper).

## Why it matters
The in-place mutation currently produces *correct* output (super's result is discarded after), so it's a **latent** landmine — it only corrupts once a result is shared/cached/retried. That's exactly the immutability contract the frozen dataclass promises.

## Regression test
`test_retry_recovery_does_not_mutate_source_result` spies on `super().handle()` and asserts its frozen result is left untouched. It's **distinguishing** — verified it FAILS against the old in-place code (`assert '...## Body' == 'No frontmatter...'`) and PASSES on the rebuild. It lives in `tests/unit/capabilities/`, so it **runs in the CI gate** (unlike integration-only tests).

## Verification
- `test_planning_handlers.py` → 187 passed; capabilities domain → 1058 passed.

### DIFF
--- a/src/squadops/capabilities/handlers/planning_tasks.py
+++ b/src/squadops/capabilities/handlers/planning_tasks.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 import logging
 import re
 import time
+from dataclasses import replace
 from typing import TYPE_CHECKING, Any
 
 import yaml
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #236, opened 2026-06-28
fix(bootstrap): standardize local dev + CI on Python 3.12 (#217)

jasonLadd:
Closes the dev-reproducibility half of #217.

## Problem

The regression suite (the PR gate) **can't be reproduced locally on Python 3.11**:
- `ci-constraints.txt` pins `Sphinx==9.1.0` (3.12-only) → `pip install -r tests/requirements.txt -c ci-constraints.txt` is `ResolutionImpossible` on 3.11.
- `test_correction_protocol` busy-spins on 3.11 (a non-yielding `asyncio.sleep` mock starves the 3.11 loop) — green on CI's 3.12.

CI already runs 3.12. This moves local dev to match.

## Scope — dev standardization only, NOT dropping 3.11

| Changed → 3.12 | Kept on 3.11 (deliberate) |
|---|---|
| bootstrap profiles (local-spark/dev-pc/dev-mac) | `requires-python = ">=3.11"` |
| `.python-version`, `local-spark.yaml` contract | ruff/black/mypy `target-version = py311` |
| `CONTRIBUTING.md` / `CLAUDE.md` dev docs | prod Dockerfiles (`python:3.11-slim`) |

**Why keep `requires-python = ">=3.11"`:** production containers (`agents/Dockerfile`, `src/squadops/api/runtime/Dockerfile`, `console/Dockerfile`) run `python:3.11-slim` and `pip install -e .`. Bumping the floor to `>=3.12` would break those image builds. `>=3.11` is *accurate* — production genuinely runs 3.11.

**Why keep `target-version = py311`:** so development on 3.12 can't silently introduce 3.12-only syntax that breaks the 3.11 production runtime.

## Follow-up (separate issue)

Dropping 3.11 entirely (migrate prod to 3.12 — Dockerfiles, `agent.lock`/`api.lock` regen, deploy validation, coordinate with the runtime lane) is tracked apart, not folded in here.

## Verification

- `squadops doctor local-spark --check python` → ✓ Python 3.12.3 via system, 2/2 checks pass.
- `tests/unit/bootstrap/` → 86 passed.
- `ruff check .` → clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #248, opened 2026-06-28
SIP-0089 Phase 4: RuntimeActivity

jasonLadd:
## Summary

Implements SIP-0089 **Phase 4 — RuntimeActivity**: make an agent's current work observable as a first-class record instead of hidden in prompt history. Built in plan order (§4.1–§4.7) on the new Python 3.12 dev venv (#236/#217), off already-formatted main (#196).

## What landed

- **§4.1 model** — `RuntimeActivity` frozen dataclass (§10.6) + `ActivityState`/`ActivitySourceKind` + `is_active_/is_terminal_activity_state`. Explicit, queryable `cycle_id`/`workload_id`/`task_id`; `source_ref` opaque (core never parses it).
- **§4.2 migration** `1130_runtime_activities.sql` — single-active-activity invariant (D9) via a partial unique index (`agent_id WHERE state IN pending/running/paused`) + indexed source-identity columns. **Live-validated** in a rolled-back txn.
- **§4.3 port + adapter** — `RuntimeActivityPort` + `PostgresRuntimeActivity` (start/update_state/complete/fail/abort/get_current). `update_state` manages lifecycle timestamps and **guards on active states** so terminal is final — race-safe against double-terminalization.
- **§4.5 coordinator seam (thin)** — a mode change best-effort aborts an activity orphaned by the change (post-write, swallows errors — observability never fails a transition). Wired into the scheduler's coordinator. **D25 single-Postgres-transaction (lease+activity+mode) deferred → #244** (gated on recruitment #233): in v1.1 this path is unexercised, so building the unit-of-work transaction now is premature.
- **§4.4 instrumentation — EXECUTOR-SIDE (a reasoned, approved deviation).** The handler runs in the agent processes, which use SQLAlchemy DbRuntime (not asyncpg) and have generated composition roots — so the plan's handler-side seam would need a second DB pool in agents + build-template changes + live-agent-path risk. Instead the `DispatchedFlowExecutor` (runtime-api process, already owns the asyncpg pool, brackets dispatch→reply) opens a `cycle`/`cycle_task` activity on dispatch and completes/fails it on reply. Opt-in `activity_port`, wholly best-effort, no new cycle EventType (the record is the observability; locked taxonomy untouched).
- **§4.6 policy seam** — `AmbientPolicy.assert_action_permitted`: the v1.2 embodiment hook (ambient + irreversible requires both a lease and an activity; fails closed). No v1.1 callers.
- **§4.7 CLI + API** — `GET /health/agents/{id}/activity` (read-only probe lane) + `squadops agent activity <id>`. Idle = 404 → CLI reports "idle" at exit 0; real errors keep non-zero exit.

## Live validation (dev stack)

Ran real `smoke` cycles end-to-end. Each dispatched task wrote a `cycle`/`cycle_task` RuntimeActivity, `running` on dispatch → `completed` on reply, one active per agent (D9), correct `task_id`/`activity_type`/`cycle_id`. A completed cycle left **zero stranded `running`** activities. `squadops agent activity <id>` rendered both the **running** activity (mid-cycle) and **idle** (after).

## Tests

- New: model+adapter, coordinator activity seam, ambient policy, executor instrumentation (helper + real `_dispatch_task` wrap), activity CLI, activity route, health_checker mapping.
- **Full canonical regression suite: 4429 passed, 1 skipped** locally on 3.12 (the CI gate, now reproducible post-#217). Format gate clean.

## Not bumping the version

Per the 2026-06-28 decision, the 1.1.0 bump is gated on **1.0.x hardening-lane completeness**, not on Phase 4 — this PR keeps the version at 1.0.x.

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -45,6 +45,7 @@ from squadops.cycles.task_outcome import TaskOutcome
 from squadops.cycles.task_plan import generate_task_plan
 from squadops.events.types import EventType
 from squadops.ports.cycles.flow_execution import FlowExecutionPort
+from squadops.runtime import reasons
 from squadops.runtime.recruitment import reserve_buffer_decision
 from squadops.tasks.models import TaskEnvelope, TaskResult
 from squadops.telemetry.context import use_correlation_context, use_run_ids
@@ -63,6 +64,7 @@ if TYPE_CHECKING:
     from squadops.ports.cycles.squad_profile import SquadProfilePort
     from squadops.ports.cycles.workflow_tracker import WorkflowTrackerPort
     from squadops.ports.events.cycle_event_bus import CycleEventBusPort
+    from squadops.ports.runtime.activity import RuntimeActivityPort
     from squadops.ports.runtime.assignments import AssignmentPort
     from squadops.ports.telemetry.llm_observability import LLMObservabilityPort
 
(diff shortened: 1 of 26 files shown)


jasonLadd (commented):
reviewed

PR #249, opened 2026-06-28
fix(cycles): propagate cycle/run cancel to Prefect (#77)

### LINKED ISSUE #77: cycles cancel does not propagate to Prefect flow run (orphaned execution continues)
## Summary

`squadops cycles cancel <project> <cycle_id>` flips the cycle and run status to `cancelled` in the cycle registry, but does **not** cancel the corresponding Prefect flow run. The flow continues executing tasks (consuming GPU / LLM tokens) against a cycle that is logically cancelled, which is wasteful and confusing.

## Reproduction

1. Launch any cycle: `squadops cycles create group_run --squad-profile spark-squad-with-builder --request-profile fullstack-fastapi-react`
2. Cancel it: `squadops cycles cancel group_run <cycle_id>`
3. Confirm via CLI: `squadops cycles show group_run <cycle_id>` → `status: cancelled`
4. Query Prefect:
   ```bash
   curl -s http://localhost:4200/api/flow_runs/filter -X POST \
     -H "Content-Type: application/json" \
     -d '{"flow_runs": {"state": {"type": {"any_": ["RUNNING", "PENDING"]}}}, "limit": 20}'
   ```
5. Observed: the flow run for the cancelled cycle is still `RUNNING`.

Live example from 2026-05-02:
- Cycle `cyc_52692a48f204` cancelled via CLI (registry status = cancelled).
- Prefect flow run `10edbcd9-2350-40d2-9f1e-bb136188e023` continued in `RUNNING` state until manually force-cancelled via Prefect API.

## Expected

Cancelling a cycle should:
1. Update registry status (already works).
2. Transition the corresponding Prefect flow run to `CANCELLED` so workers stop executing tasks.
3. Surface any error from the Prefect transition (don't silently succeed if the flow can't be reached).

## Suggested fix

In the cycle cancel handler (likely `src/squadops/cli/commands/cycles.py` → runtime-api `/cycles/{id}/cancel` route → cycle registry adapter):
- After registry status update, look up the Prefect `flow_run_id` associated with the active run (stored where the dispatcher records it — `adapters/cycles/distributed_flow_executor.py` or `PostgresCycleRegistry`).
- Call Prefect API `POST /api/flow_runs/{flow_run_id}/set_state` with `{"state": {"type": "CANCELLED", ...}, "force": true}`.
- On any failure, log clearly and return non-zero exit so the user knows the orphan exists.

Workaround until fixed:
```bash
# Find orphaned Prefect flow run for the cancelled cycle, then:
curl -s -X POST http://localhost:4200/api/flow_runs/<flow_run_id>/set_state \
  -H "Content-Type: application/json" \
  -d '{"state": {"type": "CANCELLED", "name": "Cancelled", "message": "manual cleanup"}, "force": true}'
```

## Impact

- Wasted GPU / LLM compute (a long cycle keeps running for hours after cancel).
- Confusing UX (Prefect UI disagrees with `squadops cycles show`).
- Stale RabbitMQ messages can pile up from agents working on the orphan (related to existing message-TTL gap).

## Related

- SIP-0066 (Distributed Cycle Execution Pipeline) — owns the dispatcher / Prefect integration.
- SIP-0067 (Postgres Cycle Registry) — owns cycle/run status writes.

jasonLadd:
Closes #77.

## Bug (confirmed live)
`cycles cancel` / `runs cancel` flipped registry status but **never cancelled the Prefect flow run** — the flow kept executing tasks (GPU/LLM) against a logically-cancelled cycle. Reproduce-first confirmed the cancel routes only touched the registry + emitted an event.

## Approach B (query Prefect by correlation)
Chosen over persisting `flow_run_id` (which needs a schema migration + executor/registry changes in the #186 god-class). Self-contained to the cancel routes + tracker adapter:

- **`WorkflowTrackerPort.find_active_flow_run_ids(names)`** — ids of non-terminal flow runs matching exact names, via Prefect `/flow_runs/filter`. Impl in `prefect_reporter`, `[]` in noop. Best-effort.
- **`squadops.cycles.naming.flow_run_name()`** — single-sources the flow-run name so the **executor (producer)** and **cancel routes (consumer)** can't drift apart (the executor now calls it instead of an inline f-string). A format drift would silently break cancellation, so this is the safety mechanism.
- **`deps.get_workflow_tracker()`** — NoOp-fallback getter (mirrors the event bus), wired in `main`.
- **Both cancel routes** reconstruct the run name(s) from the registry, find the live flow run(s), and set them `CANCELLED`. Response now includes `prefect_flow_runs_cancelled`.

**Best-effort by design:** the registry cancel is the source of truth; Prefect propagation never fails the request (it's logged), so a cycle is still reported `cancelled` if Prefect is unreachable.

## Verification
- Cancel-route tests (cycle + run) assert `find_active_flow_run_ids` is called with the exact reconstructed name and each match is set `CANCELLED`; plus a best-effort-survival test (Prefect failure → still `cancelled`, count 0).
- Full regression gate: **4382 passed, 0 failures**.
- The `/flow_runs/filter` request shape **validated against the live Prefect server** (HTTP 200, correct empty result for a non-existent name) — the one thing mocks can't cover.

## Note
Only the WorkflowTrackerPort gained a method; noop + prefect are the only implementers (verified). No schema migration.

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -26,6 +26,7 @@ from uuid import uuid4
 from adapters.cycles.task_naming import build_task_name
 from squadops.cycles.checkpoint import RunCheckpoint
 from squadops.cycles.models import ArtifactRef, Cycle, GateDecisionValue, Run, RunStatus
+from squadops.cycles.naming import flow_run_name
 from squadops.cycles.plan_delta import PlanDelta
 from squadops.cycles.pulse_models import (
     CADENCE_BOUNDARY_ID,
@@ -1772,7 +1773,7 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                 flow_id = await self._workflow_tracker.ensure_flow()
                 flow_run_id = await self._workflow_tracker.create_flow_run(
                     flow_id,
-                    run_name=f"{cycle.project_id}/{cycle_id[:12]}/{run_id[:12]}",
+                    run_name=flow_run_name(cycle.project_id, cycle_id, run_id),
                     parameters={
                         "cycle_id": cycle_id,
                         "run_id": run_id,
--- a/adapters/cycles/factory.py
+++ b/adapters/cycles/factory.py
@@ -95,9 +95,9 @@ def create_flow_executor(
 
         workflow_tracker = kwargs.get("workflow_tracker")
         if not workflow_tracker and kwargs.get("prefect_api_url"):
-            from adapters.cycles.prefect_reporter import PrefectReporter
+            from adapters.cycles.prefect_workflow_tracker import PrefectWorkflowTracker
 
-            workflow_tracker = PrefectReporter(api_url=kwargs["prefect_api_url"])
+            workflow_tracker = PrefectWorkflowTracker(api_url=kwargs["prefect_api_url"])
 
         return DispatchedFlowExecutor(
             cycle_registry=cycle_registry,
--- a/adapters/cycles/noop_workflow_tracker.py
+++ b/adapters/cycles/noop_workflow_tracker.py
@@ -40,6 +40,9 @@ class NoOpWorkflowTracker(WorkflowTrackerPort):
     ) -> str:
         return self._placeholder()
 
+    async def find_active_flow_run_ids(self, run_names: list[str]) -> list[str]:
+        return []
+
     async def set_flow_run_state(
         self,
         flow_run_id: str,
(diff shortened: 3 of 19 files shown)


jasonLadd:
Added a second commit: pure mechanical rename **`PrefectReporter` → `PrefectWorkflowTracker`** (file → `prefect_workflow_tracker.py`) so the concrete adapter matches its `WorkflowTrackerPort` — it was the only adapter in `adapters/cycles/` not named after its port (cf. `NoOpWorkflowTracker`, `PrefectLogForwarder`/`LogForwarderPort`). `git mv` preserves history; both factories + 4 test files updated; full gate green (4382 passed). Historical SIP/plan/review docs left untouched (point-in-time records).

jasonLadd (commented):
reviewed

PR #251, opened 2026-06-28
fix(api): enforce cycles:read/write scopes on cycle routes (#150)

jasonLadd:
Closes #150. **Release-gating security fix for 1.1.0** (fix-first per the release decision).

## The hole
SIP-0062 wired `require_scopes`/`require_roles` but **never applied them** (zero usages, verified). Any authenticated user could perform any cycle operation — create/cancel runs, decide gates, pull models, mutate squad profiles.

## Fix
- **All 40 routes in `routes/cycles/`** now carry a scope guard via `dependencies=[Depends(require_scopes(...))]` on the decorator: **GET → `cycles:read`, mutations → `cycles:write`** (cycles/runs/artifacts/profiles/projects/request-profiles/models).
- **Auth-disabled no-op (the critical safety property):** `require_scopes`/`require_roles` refactored to a shared `_enforce_access` that no-ops when no authz backend is configured. Auth-off deployments (and the entire serviceless test suite) keep routes open exactly as before; when auth is on, it enforces 401 (no identity) / 403 (missing scope). Net security gain, **zero regression** (authz-absent = open = prior behaviour).
- **#150 item 3:** `_emit_audit` logs a one-time warning when the audit port is unavailable, instead of silently dropping security events (token rejections).

## Verification
- New tests: **403** on a write route (`/cancel`) with a read-only identity, and **401** unauthenticated when auth is on — proving the guards are wired on the *real* routes, not just `require_scopes` in isolation.
- Existing `require_roles` tests + all cycle/api route tests pass (no-op-when-off holds).
- Full gate: **4435 passed, 0 failures**.

## Notes
- Scope granularity: `cycles:read/write` applied uniformly across the package (matches the issue's example). Finer admin-scoping for profile/model mutations is a reasonable follow-up.
- Residual edge (pre-existing, non-regression): if auth is *enabled* but the authz backend fails to initialize, enforcement no-ops (open) rather than failing closed — same as today's always-open. Hardening that to fail-closed via an explicit config flag could be a follow-up.

### DIFF
--- a/src/squadops/api/middleware/auth.py
+++ b/src/squadops/api/middleware/auth.py
@@ -18,6 +18,10 @@ from squadops.ports.auth.authentication import AuthPort
 
 logger = logging.getLogger(__name__)
 
+# One-time warning latch: #150 — surface a missing audit port instead of silently
+# dropping security audit events.
+_audit_unavailable_warned = False
+
 # Always allowlisted path prefixes (no token required)
 _ALWAYS_ALLOWLISTED_PREFIXES = ("/health",)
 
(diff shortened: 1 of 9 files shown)


jasonLadd (commented):
reviewed

PR #254, opened 2026-06-28
fix(maintainer): rewrite body **Status:** line on SIP promotion (#253)

jasonLadd:
Closes #253.

## Problem
`scripts/maintainer/update_sip_status.py` updates the frontmatter `status:`, moves the file to the right lifecycle folder, and updates `sips/registry.yaml` — but leaves the human-readable **body** `**Status:**` line stale. So after every promotion the maintainer had to hand-edit the body line, and the doc body could silently disagree with its frontmatter/registry. Hit during the SIP-0089 → implemented promotion for 1.1.0.

## Fix
- Add `update_sip_body_status()`, called from **both** transition branches right after the frontmatter update (so the moved copy carries both changes).
- Rewrites only the **first** status-declaration line in the document **body**, changing just the status *word* and preserving:
  - line markdown: `**Status:**`, `**Status**:`, `- **Status:**`, `*Status: …*`
  - any trailing annotation, e.g. `**Status:** Accepted (umbrella / vision)`
- Never touches the frontmatter (isolated before the body search) or prose/code/table mentions of "status".
- When no recognizable body line exists, it **warns and continues** rather than failing the promotion.

## Tests
New `tests/unit/maintainer/test_update_sip_status.py` (10 cases): canonical accepted→implemented rewrite, trailing-annotation preservation, formatting-variant preservation, first-line-only (historical notes intact), prose/header no-op, and the critical frontmatter-never-rewritten-as-body case. Also verified end-to-end through a full `update_sip_status` promotion in a temp sandbox (frontmatter + body both flip, file moves, no stale value).

### DIFF
--- a/scripts/maintainer/update_sip_status.py
+++ b/scripts/maintainer/update_sip_status.py
@@ -124,6 +124,71 @@ def update_sip_file_metadata(file_path: Path, updates: dict[str, Any]) -> bool:
         return False
 
 
+# Status keywords that may appear in a SIP body status-declaration line. Used to
+# anchor the body rewrite so prose/code/table mentions of "status" are never touched.
+_STATUS_WORDS = (
+    "Proposed",
+    "Draft",
+    "Accepted",
+    "Approved",
+    "Implemented",
+    "Deprecated",
+    "Rejected",
+    "Withdrawn",
+)
+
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #255, opened 2026-06-28
fix(comms): bounded retry/backoff for RabbitMQ publish() across reconnect window (#245)

jasonLadd:
## What & why

Closes #245. `publish()` made a **single** attempt — a connection/channel drop inside `connect_robust`'s reconnect window raised `QueueError` with no retry, **silently losing the message**. The consume side (`consume`, `_get_queue`, `subscribe`) already recovers (invalidate cache + re-declare on channel swap); `publish` was the only unguarded path.

## Change

`adapters/comms/rabbitmq.py` — bounded retry loop around the transport send. Mirrors the consume-side recovery idiom but **finite** (a publish is one-shot, not a long-lived consumer):

- Build the `Message` **once, before the loop** — a bad payload fails fast and is never retried.
- On each transport failure: drop the stale `Queue` handle (`invalidate_queue` → next attempt re-declares against the live channel via the `_get_queue` channel-swap path), back off, retry.
- Surface `QueueError` (chaining the cause) only after `_PUBLISH_MAX_ATTEMPTS` (3) — never blocks the caller forever.
- `CancelledError` is re-raised, not retried.

New constants `_PUBLISH_MAX_ATTEMPTS` / `_PUBLISH_RETRY_BACKOFF` sit alongside the existing `_SUBSCRIBE_RECONNECT_BACKOFF`.

## Tests

`TestPublishRetry` (mirrors the #146 consume-recovery tests — MagicMock channels, no broker):
- transient drop → recovers on attempt 2 (2 sends, 1 invalidate)
- persistent failure → typed `QueueError` after exactly `_PUBLISH_MAX_ATTEMPTS` sends
- happy path → exactly one send, **no** invalidate (guards against always-retry regression)

Local: comms + adapters gate dirs **118 passed**, `ruff format --check` clean.

---
Lane S / Sprint 1, item 1. The only live reliability bug in the set.

### DIFF
--- a/adapters/comms/rabbitmq.py
+++ b/adapters/comms/rabbitmq.py
@@ -24,6 +24,15 @@ logger = logging.getLogger(__name__)
 # subscription, so a flapping channel doesn't spin a tight reconnect loop.
 _SUBSCRIBE_RECONNECT_BACKOFF = 0.5
 
+# #245: a publish issued inside connect_robust's reconnect window hits a stale
+# channel ("Channel was not opened") and, with a single attempt, lost the
+# message. Unlike subscribe (a long-lived loop), a publish is a bounded
+# operation: retry a small number of times with backoff — invalidating the
+# stale Queue handle between tries so the next attempt re-declares against the
+# live channel — then surface QueueError so the caller is never blocked forever.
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #256, opened 2026-06-28
fix(cycles): resume re-attempts duty-deferred runs + actually re-executes (#222)

jasonLadd:
Closes #222.

## Problem
A run deferred by the SIP-0089 §2.5 reserve-buffer guard pauses **before any task runs** (no Prefect flow, no checkpoint), with `RUN_PAUSED` payload `reason=upcoming_hard_duty_window`. But the documented re-attempt path (`squadops runs resume`) **422'd** it: the resume route hard-required a checkpoint (`runs.py` step 3, "Cannot resume run without a checkpoint"). Surfaced by live Tier-2 verification, not unit tests (which mocked the registry).

While fixing it I found a **second, latent gap**: the resume route only flipped status to RUNNING and emitted `RUN_RESUMED` — it **never re-invoked the executor**. Repo-wide, the only caller of `execute_cycle`/`execute_run` is cycle-create's background task, so resume (for *any* case, including SIP-0079 checkpoint-resume) silently never re-executed.

## Fix (resume route only)
1. **Relax the checkpoint guard for PAUSED runs** — a checkpoint-less PAUSED run re-attempts the plan from the start (option 2 in #222). A FAILED run without a checkpoint still 422s (scoped — create a fresh run).
2. **Enqueue `execute_cycle` on resume** (mirrors cycle-create). `execute_run` already self-detects the resume source — it resumes from the latest checkpoint if one exists, else restarts from the beginning, and re-runs the §2.5 duty guard, so a still-active hard-duty window simply re-defers. This makes resume actually proceed, for the deferral case **and** the previously-dead checkpoint-resume path.

## Tests (`test_api_resume.py`)
- PAUSED + no checkpoint → 200 + re-attempt enqueued (the #222 case)
- FAILED + no checkpoint → 422, no enqueue (scoped relaxation)
- resume enqueues `execute_cycle` (the re-execution wiring)
- existing resume / terminal / cycle-status cases unchanged

Full `tests/unit/cycles/` suite green (1325 passed, 2 skipped).

## Note for review
The re-execution wiring (#2) is broader than the literal #222 title but is required to make "re-attempt from the start" actually work, and it was the same one-line gap for checkpoint-resume — so I wired it uniformly rather than ship an incoherent half-fix. Happy to split #2 into its own issue if you'd prefer.

### DIFF
--- a/src/squadops/api/routes/cycles/runs.py
+++ b/src/squadops/api/routes/cycles/runs.py
@@ -6,7 +6,7 @@ import logging
 import uuid
 from datetime import UTC, datetime
 
-from fastapi import APIRouter, Depends
+from fastapi import APIRouter, BackgroundTasks, Depends
 
 from squadops.api.middleware.auth import require_scopes
 from squadops.api.routes.cycles.dtos import (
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #259, opened 2026-06-28
fix(telemetry): stop BrokenExporter test leaking a global OTel provider → atexit traceback (#239)

jasonLadd:
## What & why

Closes #239. The `BrokenExporter` OTel test leaves a **process-global** `TracerProvider` behind whose `atexit` shutdown hook raises `RuntimeError: Simulated shutdown failure` at interpreter exit — printing an `Exception ignored in atexit callback` traceback on **every** regression run (cosmetic, exit 0, but misleading).

### Root cause
`OTelAdapter._ensure_initialized` calls `trace.set_tracer_provider(...)` / `metrics.set_meter_provider(...)` (process-global), and each SDK provider registers an `atexit` hook in its constructor. With a `BrokenExporter`, init sets the tracer global, then **fails at the metric reader** (`PeriodicExportingMetricReader` rejects the broken exporter) — so the *tracer* leaks (which is exactly why the original traceback was `TracerProvider.shutdown`). Because the init guard never flips, each `counter/gauge/...` call re-runs init and leaks another provider.

`provider.shutdown()` can't clean it up: on a broken exporter it raises *before* reaching its own `atexit.unregister`.

## Fix (test-side)

The leak only bites with `BrokenExporter`, which exists only in tests; real exporters shut down cleanly at exit. So the fix is a test isolation boundary:

- `_isolate_otel_globals()` context manager + **autouse fixture** on `TestOTelAdapter`:
  - **suppresses atexit registration** for providers born in the test, scoped to the two SDK modules (`sdk.trace` reaches `atexit` via the module attr; `sdk.metrics._internal` via names imported from `atexit`),
  - resets OTel's set-once guard so init installs deterministically,
  - **restores every global** on exit.

## Tests

- `test_broken_exporter_init_registers_no_atexit_hook` — the broken adapter still installs a **real** SDK `TracerProvider` (exercises `_ensure_initialized`) but with `_atexit_handler is None`, and globals are restored after the `with` block.
- **Acceptance:** `telemetry + comms + adapters` slice = **255 passed, 0** `Simulated shutdown` / atexit-traceback lines (was: multiple per run).

---
Lane S / Sprint 1, item 2. **Prerequisite for #216** (`-n auto`) — adding xdist workers while this global leaked would multiply the traceback across workers.

### DIFF
--- a/tests/unit/telemetry/test_adapters.py
+++ b/tests/unit/telemetry/test_adapters.py
@@ -1,6 +1,15 @@
 """Unit tests for telemetry adapters."""
 
+from contextlib import contextmanager
 from io import StringIO
+
+import pytest
+from opentelemetry import metrics, trace
+from opentelemetry.metrics import _internal as _otel_metrics_internal
+from opentelemetry.sdk import trace as _otel_sdk_trace
+from opentelemetry.sdk.metrics import _internal as _otel_sdk_metrics
+from opentelemetry.sdk.trace import TracerProvider as _SDKTracerProvider
+from opentelemetry.util._once import Once
 
 from adapters.telemetry.console import ConsoleAdapter
 from adapters.telemetry.null import NullAdapter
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #260, opened 2026-06-28
fix(cycles): resume mid-sequence runs at the right workload index (#257)

jasonLadd:
Closes #257. Follow-up to #222 / #256.

## Problem
`execute_cycle`'s workload loop assumed `first_run_id` was the position-0 run and advanced by positional index:

```python
for i, workload_entry in enumerate(workload_sequence):   # always starts at 0
    await self.execute_run(cycle_id, current_run_id, profile_id)
    ...
    current_run_id = non_cancelled[i + 1].run_id
```

True for the cycle-create entry, but **#256** wired `runs resume` to call `execute_cycle(cycle_id, run_id)`. Resuming a run paused on a **later** workload (e.g. a SIP-0089 §2.5 duty deferral on the implementation run) then started the loop at `i=0` — re-running the resumed run as workload 0, **double-executing it**, and re-polling the wrong inter-workload gate.

## Fix
Add `_starting_workload_index(cycle_id, first_run_id)` — a run's index among the cycle's non-cancelled runs (sorted by `run_number`; one run per workload position, D14) — and iterate `for i in range(start_index, len(workload_sequence))`. Cycle-create's first run resolves to index 0, so that path is unchanged; single/first-workload resume was already correct (fast path / i=0).

## Tests
- `_starting_workload_index`: position mapping, cancelled-run skip, unknown→0.
- Resume at index 1 in a 3-workload cycle: the completed first workload is **not** re-run, the resumed run executes **once**, `run_003` (wrapup) follows, and `WORKLOAD_COMPLETED` is labeled by true position (`implementation`, `wrapup` — no framing mislabel).
- Existing `execute_cycle` tests unchanged: the shared fixture stubs the index to 0 (matching the cycle-create entry they model), so the brittle call-sequence tests stay valid. Full `tests/unit/cycles/` green (1331 passed).

## Note
This is the multi-workload edge; the common #222 case (single-workload selftest deferral, or a first-workload deferral) was already correct. #258 (live E2E validation) still covers the real executor round-trip.

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -427,8 +427,15 @@ class DispatchedFlowExecutor(FlowExecutionPort):
             await self.execute_run(cycle_id, first_run_id, profile_id)
             return
 
+        # #257: support resuming mid-sequence. first_run_id may be a later
+        # workload's run (a resumed/deferred run), not the position-0 run. Start
+        # the loop at that run's workload index so earlier, already-completed
+        # workloads aren't re-run and the resumed run isn't double-executed.
+        # Cycle-create passes the first run, which resolves to index 0.
+        start_index = await self._starting_workload_index(cycle_id, first_run_id)
         current_run_id = first_run_id
-        for i, workload_entry in enumerate(workload_sequence):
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #261, opened 2026-06-28
refactor(cycles): route dispatched factory through create_workflow_tracker (#250)

jasonLadd:
## What & why

Closes #250. The dispatched branch of `create_flow_executor` (`adapters/cycles/factory.py`) **inline-built** `PrefectWorkflowTracker(api_url=...)`, bypassing the shared `create_workflow_tracker` factory. It therefore missed:
- the **NoOp-fallback** (construction failure propagated out of the factory, breaking executor wiring),
- the **init logging** (`backend=prefect, api_url=...`).

## Change

Route the build through `create_workflow_tracker(PrefectConfig(api_url=kwargs["prefect_api_url"]))`.

**Behavior preserved** — the build still only happens when a Prefect URL is configured, so:
- no URL → `workflow_tracker` stays `None` (not NoOp), unchanged;
- an explicit `workflow_tracker` kwarg still wins (the `if not workflow_tracker` guard).

The **only** new behavior is the intended one: a configured-but-failing Prefect URL now falls back to `NoOpWorkflowTracker` (+ logs) instead of raising.

## Tests (`TestDispatchedWorkflowTrackerWiring`)
- URL → executor gets a real `PrefectWorkflowTracker`
- no URL → `None` (preserved)
- construction failure → `NoOpWorkflowTracker` fallback (**the gain** — old inline build raised)
- explicit tracker → takes precedence over URL

7 passed in the factory file; `workflow_tracker` + `prefect_workflow_tracker` suites green; ruff clean.

---
Lane S / Sprint 1, item 3.

### DIFF
--- a/adapters/cycles/factory.py
+++ b/adapters/cycles/factory.py
@@ -95,9 +95,16 @@ def create_flow_executor(
 
         workflow_tracker = kwargs.get("workflow_tracker")
         if not workflow_tracker and kwargs.get("prefect_api_url"):
-            from adapters.cycles.prefect_workflow_tracker import PrefectWorkflowTracker
+            # Route through the shared factory so the NoOp-fallback and init
+            # logging apply here too (instead of inline-building the adapter and
+            # raising on construction failure). Only build when a Prefect URL is
+            # configured — when it isn't, workflow_tracker stays None, unchanged.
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #262, opened 2026-06-28
refactor(adapters): extract one parse_jsonb helper, dedup 3 copies (#156)

jasonLadd:
## What & why

Closes #156. `_parse_jsonb` was copy-pasted as a private static method in three adapters:
- `adapters/persistence/chat_repository.py` (with a `None → {}` branch)
- `adapters/cycles/postgres_squad_profile.py`
- `adapters/cycles/postgres_cycle_registry.py`

(the latter two byte-identical). A single decode, three copies that could drift.

## Change

New **`adapters/jsonb.py`** → `parse_jsonb(value)`: `str → json.loads`, otherwise passthrough (incl. `None`). All three adapters route through it.

- **`chat_repository`'s `None → {}` is preserved at the call sites** (`parse_jsonb(row["metadata"]) or {}`) — *not* baked into the shared helper — so nullable metadata still surfaces as `{}` while the helper stays a pure decode for adapters (cycle registry) that need `None` passed through.
- Helper lives at the **top-level `adapters` package** (not under `persistence`) so the `cycles` adapters don't take a dependency on the `persistence` subpackage. `test_forbidden_imports` stays green.

## Tests
- New `tests/unit/adapters/test_jsonb.py` locks the helper contract: object/array decode, already-decoded passthrough (same object), `None → None`, invalid-JSON raises `JSONDecodeError`.
- The old direct-`_parse_jsonb` chat tests (which broke when the method was removed) are replaced with **call-site** tests through `_assemble_session` / `_assemble_message` that lock the preserved `None → {}` behavior — testing the real path, not the deleted method.

Local: `adapters` + `cycles` suites **1353 passed / 2 skipped**, arch forbidden-imports green, ruff clean.

---
Lane S / Sprint 1, item 4.

### DIFF
--- a/adapters/cycles/postgres_cycle_registry.py
+++ b/adapters/cycles/postgres_cycle_registry.py
@@ -12,6 +12,7 @@ import logging
 
 import asyncpg
 
+from adapters.jsonb import parse_jsonb
 from squadops.cycles.checkpoint import RunCheckpoint
 from squadops.cycles.lifecycle import (
     GATE_REJECTED_STATES,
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #264, opened 2026-06-28
fix(cli,api): make `runs retry` actually execute; correct stale execution docstrings (#133, #205)

jasonLadd:
## What & why

Closes **#133** and **#205**. Reproduce-first turned up a discrepancy between the two issues:

- **`cycles create` — #205 correct:** POST `/cycles` enqueues `execute_cycle` (SIP-0066/0083), but the CLI help said *"Does not trigger task execution (deferred to a future release)."* → **fixed**.
- **`runs retry` — #205 wrong / #133 correct:** `create_run` builds a `status="queued"`, `initiated_by="retry"` run, emits `RUN_CREATED`, and returns **with no dispatch** (`resume_run` even documents *"only cycle-create enqueues it"*). The retry run never ran — #133's "dead weight." So the docstring #205 proposed (*"enqueues it for execution"*) would have been a **new** false claim. (Corrective note posted on #205.)

## Change

Rather than make the retry docstring honest-but-useless, take #133's preferred option (1) — **wire `create_run` to dispatch**, mirroring `resume_run`:

```python
background_tasks.add_task(get_flow_executor().execute_cycle, cycle_id, created.run_id, cycle.squad_profile_id)
```

`execute_cycle` re-runs the SIP-0089 §2.5 duty guard, so a still-active hard-duty window simply re-defers — consistent with resume.

**Verified safe (no double-dispatch):** internal/workload runs are created via `registry.create_run` *directly* (`dispatched_flow_executor.py:659`, `initiated_by="system"`); the HTTP `create_run` route is only the external retry path, so auto-dispatch can't double-fire workload runs.

Both CLI docstrings now state execution is enqueued — true for both commands.

## Tests
- `test_retry_run_is_dispatched_for_execution` — the retry route enqueues `execute_cycle(cycle_id, new_run_id, squad_profile_id)` (TestClient runs background tasks post-response).
- `client` fixture now provides a flow-executor mock.
- `tests/unit/cycles/` **1336 passed**, CLI **207 passed**, ruff clean.

## Note (lane coordination)
Touches `api/routes/cycles/runs.py` `create_run` (lines ~37–95). Lane M's #222 edits `resume_run` (~212) in the same file — different function, ~150 lines apart, so clean auto-merge expected.

---
Lane S / Sprint 1, item 5.

### DIFF
--- a/src/squadops/api/routes/cycles/runs.py
+++ b/src/squadops/api/routes/cycles/runs.py
@@ -35,8 +35,17 @@ router = APIRouter(prefix="/api/v1/projects/{project_id}/cycles/{cycle_id}/runs"
 
 
 @router.post("", dependencies=[Depends(require_scopes(Scope.CYCLES_WRITE))])
-async def create_run(project_id: str, cycle_id: str, workload_type: str | None = None):
-    """Create a new Run (retry) for an existing Cycle."""
+async def create_run(
+    project_id: str,
+    cycle_id: str,
+    background_tasks: BackgroundTasks,
+    workload_type: str | None = None,
+):
+    """Create a new Run (retry) for an existing Cycle and enqueue its execution.
+
+    Mirrors ``cycles create`` / ``resume_run``: the new run is dispatched
+    server-side (not just recorded), so a retry actually re-runs (#133).
+    """
     from squadops.api.runtime.deps import get_cycle_registry
 
     try:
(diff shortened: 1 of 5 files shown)


jasonLadd (commented):
reviewed

jasonLadd (commented):
reviewed

PR #266, opened 2026-06-28
fix(capabilities): strip <think> before fenced parsing + log zero-extraction raw (#130)

### LINKED ISSUE #130: Neo development.develop emits unparseable output (no fenced code blocks) under spark-squad models
## Symptom

Neo's \`development.develop\` task is intermittently emitting LLM responses with **no fenced code blocks**, causing the framework's \`extract_fenced_files\` parser to find nothing executable. The handler returns a 0-byte \`build_warnings.md\` placeholder instead of the expected source files, which routes through the correction protocol as a \`task_failure:development.develop\`.

## Reproductions

Two long-cycle group_run cycles on \`spark-squad-with-builder\` / \`validation\` profile have shown this exact failure:

| Cycle | Outcome |
|---|---|
| \`cyc_e80826aa8637\` (2026-05-05) | Implementation reached \`development.develop\`; one task came back without fenced code blocks; correction protocol picked \`rewind\`; run terminated. |
| \`cyc_a867cbf02205\` (2026-05-05) | Same failure pattern; Neo logs show \`handler_failed\` ×2 on \`development.develop\` tasks; correction loop tried but \`data.analyze_failure\` and \`governance.correction_decision\` both also failed (separate role-identity issue, fixed in PR #129). |

## Failure analysis (from \`cyc_e80826\`'s persisted artifacts)

\`\`\`json
{
  \"classification\": \"model_limitation\",
  \"analysis_summary\": \"The task failed with the structural error 'No valid fenced code blocks found', indicating the model completely omitted the required markdown code fences for the FastAPI and React source files. This is a structural failure rather than a content-quality issue, as the framework parser could not extract any executable artifacts and instead returned an empty build_warnings.md placeholder. The root cause is the model's failure to adhere to strict output formatting constraints during multi-file generation, which requires a full task rewind rather than a targeted patch.\",
  \"contributing_factors\": [
    \"LLM output omitted all markdown fenced code blocks required by the development framework parser\",
    \"Generated artifact reduced to a 0-byte build_warnings.md file instead of implementing backend/frontend modules\",
    \"Strict parser validation rejected the response before content-level checks could evaluate code correctness\"
  ]
}
\`\`\`

The model's analysis classifies it correctly: structural-format failure, not content-quality failure. The LLM produces \"some output\" — likely prose or thinking — that fails to match the fenced-block format the parser requires.

## Hypotheses

1. **Smaller-model formatting failure under load.** \`development.develop\` prompts include PRD + planning artifacts + per-task focus + acceptance criteria + typed-check vocabulary + builder-routing rules. The token budget and instruction density may push qwen3.6:27b past its reliable formatting threshold.
2. **Qwen3 \`<think>\` mode emitting reasoning instead of output.** If thinking mode produces a long \`<think>\` block that consumes the response budget, the actual code never gets written.
3. **Prompt-design issue.** The fenced-block format instructions may be buried far down the prompt; emphasizing them or moving them to the top may help.

PR #128's tolerant JSON extractor doesn't apply here — \`development.develop\` consumes \`extract_fenced_files\` (a different parser, in \`fenced_parser.py\`) rather than JSON. Hypothesis 1 vs 2 vs 3 needs evidence: a similar raw-response logging pass on \`fenced_parser\` failures would tell us conclusively.

## Suggested fixes (ranked)

1. **Add raw-response logging to \`extract_fenced_files\`'s zero-extraction fallback** — same pattern as PR #128 did for impl handlers. Next failure surfaces the actual model output for triage. ~1h.
2. **Tune the \`development.develop\` system prompt** to emphasize the fenced-block format up front. Possibly requires a prompt fragment refactor.
3. **Strip Qwen3 \`<think>\` blocks from \`development.develop\` responses** before fenced parsing — same approach PR #128's extractor uses for JSON. Defends against thinking-mode token waste regardless of root cause.
4. **Bound the per-task scope** if \`scope-too-large\` turns out to be the cause — split mega-tasks at planning time.

## Acceptance criteria

- A failed \`development.develop\` task's raw LLM response is captured in agent logs (logged truncated, e.g., \`raw[:1000]=\`).
- Re-cycling on \`group_run\` / \`spark-squad-with-builder\` / \`validation\` either succeeds at \`development.develop\` or surfaces enough evidence to choose between hypotheses 1/2/3 and ship a targeted fix.

## Out of scope

- The role-identity prepend regression on JSON-emitting impl handlers (\`establish_contract\`, \`correction_decision\`, \`analyze_failure\`) — fixed in PR #129.
- Multi-role plan authoring (SIP-0093) — orthogonal architectural work.

jasonLadd:
## What & why

Closes #130. `development.develop` (and `builder.assemble` / `qa.test`) parse multi-file output through `extract_fenced_files`. Under spark-squad models, two modes produced a 0-byte `build_warnings.md` that routed through the correction protocol as `task_failure:development.develop`:

1. **Qwen3 `<think>…</think>` traces** were fed straight into the fence scanner. Thinking-mode traces can contain stray ``` fences (false matches) or consume the whole response so no real files are emitted.
2. **Zero-extraction was invisible** — the raw response went only into a `build_warnings.md` artifact, never the agent logs, so there was no way to triage thinking-mode vs prompt/scope failure.

## Change

- **`fenced_parser.extract_fenced_files` strips `<think>` blocks first** — reuses `_strip_think_blocks` from `impl/_json_extraction` (the same defense PR #128 added for the JSON impl handlers). Helps **every** caller (develop/assemble/qa/repair/plan-authoring), not just develop.
- **`_CycleTaskHandler._log_no_fenced_blocks()`** logs the truncated raw (`raw[:1000]`) + the full response length, wired into the develop / qa / assemble zero-extraction paths.

## Tests (`test_fenced_parser.py`)
- fence **inside** a `<think>` block is ignored (only the real file is extracted)
- a **think-only** response yields no files — the exact #130 failure
- think prose before a real file is fine (case-insensitive `<think>`/`<THINK>`)
- zero-extraction logs a **truncated, length-reporting** warning (tail beyond the excerpt is dropped)

41 fenced-parser tests + 1062 capabilities tests pass; ruff clean; no import cycle.

Implements the issue's ranked fixes #1 (raw logging) and #3 (`<think>` strip). Prompt-tuning (#2) and scope-bounding (#4) remain open as separate follow-ups if the logging shows they're needed.

---
Lane S / Sprint 1, item 7 — **completes the Sprint 1 cluster.**

### DIFF
--- a/src/squadops/capabilities/handlers/cycle_tasks.py
+++ b/src/squadops/capabilities/handlers/cycle_tasks.py
@@ -448,6 +448,23 @@ class _CycleTaskHandler(CapabilityHandler):
             error=error,
         )
 
+    def _log_no_fenced_blocks(self, content: str, *, excerpt: int = 1000) -> None:
+        """Log the truncated raw LLM response when fenced extraction found
+        nothing (#130).
+
+        The build/QA handlers otherwise only persist the raw into a
+        ``build_warnings.md`` artifact, so a zero-extraction failure leaves no
+        signal in the agent logs. The raw is the only way to distinguish a
+        thinking-mode token blowout from a prompt/scope formatting failure.
+        """
+        logger.warning(
+            "%s: no fenced code blocks extracted from a %d-char response; raw[:%d]=%r",
+            self._handler_name,
+            len(content),
+            excerpt,
+            content[:excerpt],
+        )
+
     def _resolve_model_budget(
         self,
         inputs: dict[str, Any],
@@ -1766,6 +1783,7 @@ class DevelopmentDevelopHandler(_CycleTaskHandler):
         extracted = extract_fenced_files(content)
 
         if not extracted:
+            self._log_no_fenced_blocks(content)
             return self._fail_result(
                 start_time,
                 inputs,
@@ -2416,6 +2434,7 @@ class QATestHandler(_CycleTaskHandler):
 
         extracted = extract_fenced_files(content)
         if not extracted:
+            self._log_no_fenced_blocks(content)
             return self._fail_result(
                 start_time,
                 inputs,
@@ -3100,6 +3119,7 @@ class BuilderAssembleHandler(_CycleTaskHandler):
         if not extracted:
             from squadops.cycles.task_outcome import FailureClassification, TaskOutcome
 
+            self._log_no_fenced_blocks(content)
             return self._fail_result(
                 start_time,
                 inputs,
--- a/src/squadops/capabilities/handlers/fenced_parser.py
+++ b/src/squadops/capabilities/handlers/fenced_parser.py
@@ -27,6 +27,8 @@ from __future__ import annotations
 
 import re
 
+from squadops.capabilities.handlers.impl._json_extraction import _strip_think_blocks
+
 # Strict format header: ```<lang>:<path>
 _STRICT_HEADER_RE = re.compile(r"^```(\w+):(\S+)\s*$", re.MULTILINE)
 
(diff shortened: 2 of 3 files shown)


jasonLadd (commented):
reviewed

PR #271, opened 2026-06-28
fix(auth): bridge realm roles to implied scopes so cycle routes work (#270)

jasonLadd:
Closes #270.

## Problem
#150 applied `require_scopes(Scope.CYCLES_READ/WRITE)` to the cycle routes, but the role-centric Keycloak realm issues **roles** (`admin`/`operator`/`viewer`) and **never** the `cycles:*` scopes. So on a deployed stack every cycle route 403'd for every authenticated user:

```
$ curl -H "Authorization: Bearer $TOKEN" .../api/v1/projects/play_game/cycles
HTTP 403  {"detail":"Missing required scopes: ['cycles:read']"}
```

`check_access` requires each scope to be literally present in `identity.scopes`, which `auth_adapter.py` built from the OAuth `scope` claim only (`profile email`). No bridge from the roles the realm actually grants — even though the realm documents `operator` = *"start/stop cycles"*.

## Fix (role → implied scopes)
Keycloak can't express role-conditional OAuth scopes natively, so bridge in code:
- `ROLE_SCOPES` map + `scopes_for_roles()` single-sourced in `auth/models.py`:
  `admin` → all; `operator` → cycles+tasks read/write + `agents:read`; `viewer` → reads only.
- `resolve_identity()` now sets `identity.scopes = token scopes ∪ role-implied scopes`.

`require_scopes` (and #150's route protection) is unchanged and now honors the realm's role model with **no Keycloak change**. Unknown roles imply nothing — no silent grant.

## Tests
- `scopes_for_roles`: operator→write, viewer read-only, unknown→∅, union across roles.
- `resolve_identity` bridge: operator token (empty scope claim) → `cycles:write`; viewer read-only; unmapped role → still 403-able; explicit token scopes preserved via union.
- Full `tests/unit/auth` green — **172 passed**.

## Live verification
Rebuilt runtime-api and re-ran the blocked flow — login → `cycles create` now succeeds (was 403). _(Result appended after the redeploy completes.)_

owner note: this is a `track:spark` #150 follow-up; fixed here to unblock the post-merge verification.

### DIFF
--- a/adapters/auth/keycloak/auth_adapter.py
+++ b/adapters/auth/keycloak/auth_adapter.py
@@ -20,6 +20,7 @@ from squadops.auth.models import (
     IdentityType,
     TokenClaims,
     TokenValidationError,
+    scopes_for_roles,
 )
 from squadops.ports.auth.authentication import AuthPort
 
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #273, opened 2026-06-29
fix(runtime): open duty windows on-time within the poll cadence (#272)

### LINKED ISSUE #272: Duty windows never auto-open under default missed_window_policy=skip (poll-cadence lateness → skipped)
## Duty windows never auto-open under the default `missed_window_policy="skip"`

Found during post-1.1.0 live verification. The duty scheduler is enabled (`runtime.scheduler.enabled=true`, `poll=10s`) and the machinery is sound — but a duty assignment created with **default options never transitions the agent to `duty`**. The agent stays `ambient` for the entire window.

## Root cause (traced end-to-end)
The scheduler polls on a cadence, so the first tick that sees a window as `active` is always *slightly after* `window_start`:

- `DutyScheduler._evaluate` (`src/squadops/runtime/scheduler.py:101-107`): when `window_state == "active"` and the agent isn't already on duty, it computes `late = now - active_window.start`. Only `late <= 0` takes the on-time `_open` path; **any positive lateness routes to `_open_late`**.
- `_open_late` (`:131-140`) honors `missed_window_policy`. The default on a CLI/API-created assignment is **`missed_window_policy="skip"`** with **`graceful_window=0`**, so it calls `_emit_skipped()` and returns `None` — no transition.

Because ticks fire every `poll_interval` (10s here), `late` is essentially always in `(0, poll_interval]` on the first active tick — so the on-time `_open` branch (`late <= 0`) is effectively unreachable, and `skip` drops the window. **Net: default duty assignments never open.**

Verified live (nat, hard duty, default policy):
```
window_state = active | late = 0:04:15 (>0) | missed_window_policy = skip | graceful_window = 0  →  outcomes: []  (stays ambient)
# flip to start_late_within_grace + grace:
tick outcomes: [('nat', applied=True, 'ambient', 'duty')]   # opens correctly
# close path also works:
close-sweep: [('nat', 'duty', 'ambient')]
```
So the coordinator/scheduler/DB-write path all work — only the open *decision* under the default policy is broken.

## Recommended fix
Treat sub-`poll_interval` lateness as **on-time**, not "missed". `missed_window_policy` should govern *genuinely* missed windows (agent offline, skipped ticks, late by ≫ poll), not the inherent few-seconds poll lag. Options:

1. In `_evaluate`, take the `_open` path when `late <= poll_interval` (+ small margin), reserving `_open_late`/policy for lateness beyond that. **(preferred — robust regardless of operator config)**
2. And/or default `missed_window_policy` to `start_late_within_grace` with `graceful_window >= poll_interval`.

## Test gap
`tests/unit/runtime` drives `tick()` with on-time `now` values, so the poll-cadence lateness path isn't exercised. Add a test: a window that became active < `poll_interval` ago still opens under the default policy.

## Notes
Not a regression from the recent merges (scheduler logic/defaults predate them; #270 was auth-only). Likely masked since PR #227 because no one created a *default-policy* duty assignment and checked the live transition. owner: track:macbook (SIP-0089 runtime).

jasonLadd:
Closes #272.

## Problem
The duty scheduler never auto-opened a duty window under the default `missed_window_policy="skip"`. `_evaluate` treated **any** positive lateness as a missed window → `_open_late` → `skip` → no transition. But the scheduler only observes a window at a tick boundary, so the first tick to see a window active is always a few seconds late (up to one poll interval). With `skip`+`grace=0`, that cadence lag meant **default duty assignments never opened** — the agent stayed `ambient` the whole window.

## Fix
Treat lateness within one poll interval (+ a small tick-jitter margin) as **on-time**:
- `DutyScheduler.__init__`: `self._on_time_grace = poll_interval + 2s`.
- `_evaluate`: open via `_open` (`duty_window_opened`) when `late <= self._on_time_grace`; reserve `_open_late`/`missed_window_policy` for lateness **beyond** the cadence (agent offline, scheduler stalled, ticks skipped).

So `missed_window_policy` now governs *genuinely* missed windows, not the scheduler's own poll lag.

## Why it was masked
Unit tests drive `tick()` with on-time or minute-scale-late `now` values, so the seconds-scale poll-cadence case was never exercised. Added in this PR.

## Tests
- `test_poll_cadence_lateness_opens_on_time_under_skip`: 5s late, `poll=10s`, `skip` policy → opens `ambient→duty` with `duty_window_opened` (was skipped before).
- `test_lateness_beyond_poll_grace_still_skips`: 60s late, `poll=10s` → still skips (grace is bounded by the poll interval).
- Existing late-start tests unaffected (minute-scale lateness ≫ the seconds-scale grace). Runtime suite **142 passed**.

## Live evidence (from the #272 diagnosis)
A default-policy hard-duty assignment stayed `ambient` (`tick outcomes: []`); confirmed the coordinator/transition path itself is sound (flipping to `start_late_within_grace` opened `ambient→duty`, and the close-sweep closed it). Rebuilt runtime-api here to re-verify the default-policy open end-to-end. owner: track:macbook (SIP-0089).

### DIFF
--- a/src/squadops/runtime/scheduler.py
+++ b/src/squadops/runtime/scheduler.py
@@ -45,6 +45,11 @@ logger = logging.getLogger(__name__)
 # constructed/started by bootstrap only when runtime.scheduler.enabled is true.
 _DEFAULT_POLL_INTERVAL_SECONDS = 30.0
 
+# #272: tick-processing jitter absorbed on top of the poll interval when deciding
+# whether a just-active window is on-time. The poll cadence already bounds how
+# late the first observing tick can be; this covers the tick's own run time.
+_ON_TIME_GRACE_MARGIN_SECONDS = 2.0
+
 
 class DutyScheduler:
     """Polls duty assignments and requests window-open/close transitions."""
@@ -64,6 +69,12 @@ class DutyScheduler:
         self._state = state
         self._events = events_publisher
         self._poll_interval_seconds = poll_interval_seconds
+        # #272: a window is "on-time" if observed within one poll interval (plus a
+        # small jitter margin) of window_start — that lag is the scheduler's own
+        # cadence, not a missed window.
+        self._on_time_grace = timedelta(
+            seconds=poll_interval_seconds + _ON_TIME_GRACE_MARGIN_SECONDS
+        )
         self._clock = clock or (lambda: datetime.now(UTC))
         self._task: asyncio.Task | None = None
 
@@ -102,7 +113,12 @@ class DutyScheduler:
             if in_duty:
                 return None  # already serving this window
             late = now - a.active_window.start
-            if late <= timedelta(0):
+            # #272: the scheduler only observes a window at a tick boundary, so the
+            # first tick to see it active lags window_start by up to one poll
+            # interval (+ tick run time). That cadence lag is on-time — open it.
+            # missed_window_policy governs lateness *beyond* the grace (agent
+            # offline, scheduler stalled, ticks skipped), not the inherent poll lag.
+            if late <= self._on_time_grace:
                 return await self._open(a, reasons.DUTY_WINDOW_OPENED)
             return await self._open_late(a, late)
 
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #278, opened 2026-06-29
fix(observability): raw task_type + keep focus in Prefect task names (#277)

jasonLadd:
Closes #277.

#94's per-agent `[n/total]` count was a win and stays — but it over-corrected on the **descriptor**, dropping the role identity and collapsing task names to ~one word. This restores both.

## What changed
`build_task_name` (`adapters/cycles/task_naming.py`):
- **Descriptor is the raw `task_type`** instead of a humanized leaf. `qa.define_test_strategy` → "Test" threw away the `domain.verb` structure; the domain segment *is* the role, so the raw value restores the role identity (lost by #94's agent-only prefix) **and** keeps the verb — no duplication, since the prefix is the agent (`eve`), not the role (`qa`).
- **`subtask_focus` is appended, not substituted** — build subtasks keep their specificity while every label keeps a uniform, greppable `task_type` spine.
- **Count bracket spaced off the prefix** (`neo [1/2]`, not `neo[1/2]`) so it doesn't read as an array subscript.
- `_humanize_task_type` removed (now dead).

## Before → after
```
neo[2/5]: Develop                  →  neo [2/5]: development.develop — implement POST /runs
eve[1/2]: Test                     →  eve [1/2]: qa.define_test_strategy
nat[1/1]: Frame objective          →  nat [1/1]: strategy.frame_objective
```

## Decision on the open question (build subtask `focus`)
Went with **(b) keep the focus**, appended after an em dash — dropping it for format uniformity (option (a)) would have been a second over-correction in the opposite direction. `subtask_focus` is the most information-dense field available for build subtasks.

## Tests
`tests/unit/cycles/test_task_naming.py` updated to the new `agent [n/total]: task_type[ — focus]` shape; added a test asserting the bracket is spaced (no `neo[` subscript form). Full regression: **4477 passed, 1 skipped**.

### DIFF
--- a/adapters/cycles/task_naming.py
+++ b/adapters/cycles/task_naming.py
@@ -10,54 +10,57 @@ from __future__ import annotations
 
 from squadops.tasks.models import TaskEnvelope
 
-
-def _humanize_task_type(task_type: str) -> str:
-    """Title for a non-focused task, derived from its ``task_type``.
-
-    Takes the leaf segment and makes it read like a title:
-    ``strategy.frame_objective`` → ``Frame objective``,
-    ``qa.define_test_strategy`` → ``Define test strategy``. Falls back to the
-    raw ``task_type`` if it has no usable leaf.
-    """
-    leaf = task_type.rsplit(".", 1)[-1].replace("_", " ").replace("-", " ").strip()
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #283, opened 2026-06-30
docs(plan): correct §5.5 — #198 pin already done, not a do-first item

jasonLadd:
Reproduce-first catch at Spark-lane kickoff. The 1.2.0 plan's §5.5 called **#198** the cheap "do-first" pin. It isn't — the pin half **already shipped on `main`** via #195/#202:

- `tests/requirements.txt` caps `fastapi>=0.104.0,<0.136`
- `ci-constraints.txt` locks `fastapi==0.135.4` / `starlette==1.3.1`
- fastapi isn't even declared in `pyproject.toml`, so the "add an upper bound" the plan wanted is **moot**

The plan's "next lock regen drifts into the break" premise was wrong. The **remaining** half of #198 (flatten the router cycle → adopt fastapi ≥0.136) is deferred adoption work, per the issue's own scoping comment: the cyclic include is in the **external `continuum` package** (this repo's `console/app/main.py` includes are clean), it needs a **CI-job lock regen** (forbidden on the aarch64 Spark box), it re-validates the whole API surface the Mac lane is editing, and there's no urgency.

## Changes (doc-only)
- **§5.5** — execution order corrected to **{#173 ∥ #158} → #176**; #198 reclassified as done(pin) + deferred(adoption).
- **§4 matrix** — `ci-constraints.txt` row downgraded to LOW (no re-lock pending).
- **§3** — #198 line annotated pin-done / adoption-deferred.
- **Appendix kickoff** — dropped "#198 first"; #173/#158 are the real first Spark moves.

**Corrected first Spark move is #173**, which is being started now.


jasonLadd (commented):
reviewed

PR #284, opened 2026-06-30
fix(#173): consolidate squad-profile names to smoke/lite/full + fix active_profile footgun

jasonLadd:
Closes #173.

Consolidates the squad-profile names so they're keyed by **model tier only** — roster (including whether Bob the builder is present) lives in `members:`, not in the identifier. This is a **breaking change** (profile IDs are user-facing: `--squad-profile`, configs, DB) + a DB migration.

## Before → after

| before | after | note |
|--------|-------|------|
| `smoke` (3b, no builder) | **`smoke`** | unchanged — the minimal liveness/plumbing squad |
| `lite` (7b, +builder) | **`lite`** | unchanged |
| `full-squad` (27b, no builder) | *removed* | absorbed by `full` |
| `spark-squad-with-builder` (27b, +builder) | **`full`** | the canonical quality squad |
| `active_profile: full-squad` | `active_profile: smoke` | footgun fix |

**Why collapse the two 27b profiles into one `full` with the builder:** the planner capability filter means a builder-less squad still builds (dev does assembly), so `-with-builder` is no longer load-bearing; and the builder A/B run showed builder + stack-aware dev is the quality lever (0 manual fixes vs 3). So `full` includes Bob. Anyone previously on `full-squad` now gets the builder — strictly better.

**Naming decision (`smoke`, not `selftest`):** the issue suggested renaming `smoke`→`selftest`, but `selftest.yaml` **already exists as a request-profile** — reusing it as a squad-profile would put the same name on two axes. `smoke` is also the standard term for a liveness check (cf. #176 "framework smoke invariant test"). Kept `smoke`.

**active_profile footgun:** the old `full-squad` (27b) default hard-failed on any box without that model. `smoke` (3b) is always runnable, so an unspecified cycle never fails on a missing model. The real "you asked for 27b but don't have it" guard is the **#224 preflight** — this is the narrow default fix, not a `require-explicit` rearchitecture.

## DB migration (`infra/migrations/1000_…`, Spark range)
Reconciles the `squad_profiles` table + `squad_profiles_seed_log`: drops `full-squad`, renames `spark-squad-with-builder`→`full` (with name/description aligned to YAML), repoints active to `smoke` **iff** dropping `full-squad` left none active (so it's one-time, re-run safe, and never clobbers an operator's later `set_active_profile`). Idempotent and correct on both fresh and existing DBs — required since there's no migration-tracking table yet (#158).

## Test strategy
- **Semantic rewrites** where the old taxonomy was load-bearing: no-builder-path tests repoint to `smoke` (still 5-agent no-builder); builder tests to `full`; the listing test asserts `{smoke, lite, full}`.
- **Mechanical** fixture/round-trip updates elsewhere.
- **Left intentionally:** `test_profile_utils.py` slugify/validate cases (they test the slug *functions*), and historical name references in two docstrings/comments.

## Verification
- `3452 passed / 1 skipped` across cycles/api/cli/console/contracts/bootstrap/llm/capabilities/events
- `ruff check` + `ruff format --check` + test-quality lint all clean

## Cross-lane notes
- **This is the gate that unblocks Mac's #224 preflight** — build the preflight against the final names (`smoke`/`lite`/`full`). Tagging the Mac lane.
- **Left for Mac:** `adapters/cycles/dispatched_flow_executor.py:2259` has a stale `spark-squad-with-builder` comment — Mac-owned file, not touched here.
- **#279** (new request-profile) should build on these names (plan §5.5: #173 before #279).


jasonLadd (commented):
reviewed

Issue #286, opened 2026-06-30
runtime-api main.py loads + validates config at module import time (import side-effect blocks bare import / tooling)


jasonLadd:
## Problem

`src/squadops/api/runtime/main.py:41` runs `config = load_config(strict=strict_mode)` at **module level**, so merely importing the module executes full config loading + `secret://` resolution + validation. Without a configured environment the import fails outright:

```
ConfigValidationError: secret:// references found but no secrets configuration provided
```

This is the config system *correctly* failing fast (good — aligns with the no-masking rule). The smell is the **import side-effect**: you can't `import squadops.api.runtime.main` for tooling, introspection, or partial tests without standing up a full secrets/env profile first.

## Evidence it's a known friction point

`tests/unit/cli/test_integration.py` already works around it with `_import_fastapi_app()`, whose docstring is explicit:

> "Import the FastAPI app with test env vars set. Must be called inside a fixture/function — never at module level — so that env overrides are active before `load_config()` runs."

So every consumer that wants the app object has to set `_TEST_ENV` and clear the `sys.modules` cache first.

## Impact

- Low (production runs with env configured; the test suite works around it). Not a runtime bug.
- Real for tooling / smoke-tests / future partial-import scenarios — a bare `import main` is the obvious thing to reach for and it explodes.

## Suggested direction

Defer the config load out of module scope into a function/factory (e.g. `get_config()` lazy accessor, or load inside `startup_event`/an app factory), so import is side-effect-free and the env only needs to be present when the app actually *runs*. Adjacent in spirit to #154 (import hygiene) but distinct (that one is domain→adapter imports; this is entrypoint config-load timing).

## Not a blocker

Surfaced while wiring SIP-0089 §3.5 (#233); recorded so it isn't lost. Tech-debt, future cleanup — not part of the #233 branch.

PR #287, opened 2026-06-30
feat(runtime): route cycle recruitment through the coordinator (SIP-0089 §3.5, #233)

jasonLadd:
Completes the SIP-0089 runtime arc shipped with a known limitation in 1.1.0: cycle recruitment now acquires a **FocusLease** through the `RuntimeCoordinator` (§3.5), rather than the lease gate living only at the coordinator with no recruitment admission step.

Closes #233.

## Design decision (the #233 open question)

Recruited agents **transition `ambient → cycle`** through the coordinator, rather than holding a cycle lease while staying `ambient`. The `cycle` RuntimeMode exists for exactly this; `CYCLE_RECRUITED`/`CYCLE_COMPLETED` reasons were pre-minted; the cycle FocusLease rides the mode write via `_arbitrate_lease` (§3.4); and the ambient-lease alternative would need new coordinator surface and violate §10.4 ("ambient holds no primary lease in v1.1").

## What changed (4 slices)

1. **Domain admission** (`runtime/admission.py`) — `admit_participants` / `release_participants`. Routes each participant through `request_transition(..., "cycle", CYCLE_RECRUITED)`; a lease conflict defers with the coordinator's typed `focus_lease_*` reason. **Partial admission is rolled back** (a later conflict returns earlier recruits to `ambient` — no one strands in `cycle`); `idempotent_skip` (already in cycle) is admitted but not recorded as recruited; release is best-effort/per-agent isolated.
2. **Executor wiring** — `DispatchedFlowExecutor` gains an opt-in coordinator. After the §2.5 reserve-buffer guard passes, recruitment admits via the coordinator; a conflict rides the existing `_RecruitmentRejectedError → RUN_PAUSED` path (**no new EventType** — the §2.5/#222 locked-taxonomy precedent). The `finally` releases recruited leases on **every** terminal outcome.
3. **Composition root (D16)** — the coordinator must be the **sole writer of `mode`**, and one already exists for the duty scheduler, so the executor **shares that instance**. `create_runtime_coordinator(pool)` builds it once in `_init_cycle_subsystem`; `create_duty_scheduler` accepts the injected coordinator (back-compat fallback preserved). Pool-less → `coordinator=None` → recruitment degrades to the §2.5 guard, so a cycle never hard-fails for missing runtime infra.
4. **Docs** — coordinator notes updated: D25/#244 is no longer "gated on #233" but the **unblocked follow-up**.

## Acceptance (per the issue)

- ✅ Recruitment that conflicts with a held focus lease defers the run with a distinct, observable reason code (`focus_lease_*` on `RUN_PAUSED`).
- ✅ No stranded cycle leases across run completion/failure/pause (released in `finally`; partial admission rolled back).
- ✅ No new cycle EventType; deferral rides `RUN_PAUSED` payload.

## Tests

- `tests/unit/runtime/test_admission.py` — 6 tests: sorted recruit, partial-rollback on conflict, idempotent-skip, best-effort release across a failure, empty, reason fallback.
- `tests/unit/cycles/test_dispatched_flow_executor.py::TestRecruitmentCoordinatorAdmission` — conflict defers before dispatch with the right payload; clean admission completes and releases every recruit.
- `tests/unit/api/test_scheduler_bootstrap.py` — `create_runtime_coordinator` None-without-pool + §3.4/§4.5 wiring; injected coordinator shared, not rebuilt (the D16 guard).
- Full regression: **4488 passed / 1 skipped**.

## Follow-ups (out of scope)

- **#244** (single-Postgres-transaction wrapping of lease+activity+mode, D25) is now unblocked.
- **#286** (filed during this work): `main.py` loads+validates config at module-import time — pre-existing import side-effect, tech-debt.

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -428,9 +428,7 @@ class DispatchedFlowExecutor(FlowExecutionPort):
             # agent's future recruitment, so this must run before anything that can
             # raise. Empty (and skipped) when the run recruited no one.
             if self._coordinator is not None and recruited_agent_ids:
-                await release_participants(
-                    self._coordinator, recruited_agent_ids, owner_ref=run_id
-                )
+                await release_participants(self._coordinator, recruited_agent_ids, owner_ref=run_id)
             await self._finalize_run(
                 cycle_id,
                 run_id,
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -47,6 +47,7 @@ from squadops.cycles.task_plan import generate_task_plan
 from squadops.events.types import EventType
 from squadops.ports.cycles.flow_execution import FlowExecutionPort
 from squadops.runtime import reasons
+from squadops.runtime.admission import admit_participants, release_participants
 from squadops.runtime.recruitment import reserve_buffer_decision
 from squadops.tasks.models import TaskEnvelope, TaskResult
 from squadops.telemetry.context import use_correlation_context, use_run_ids
@@ -68,6 +69,7 @@ if TYPE_CHECKING:
     from squadops.ports.runtime.activity import RuntimeActivityPort
     from squadops.ports.runtime.assignments import AssignmentPort
     from squadops.ports.telemetry.llm_observability import LLMObservabilityPort
+    from squadops.runtime.coordinator import RuntimeCoordinator
 
 logger = logging.getLogger(__name__)
 
@@ -122,6 +124,7 @@ class DispatchedFlowExecutor(FlowExecutionPort):
         reply_router: ReplyRouter | None = None,
         assignment_port: AssignmentPort | None = None,
         activity_port: RuntimeActivityPort | None = None,
+        coordinator: RuntimeCoordinator | None = None,
     ) -> None:
         self._cycle_registry = cycle_registry
         self._artifact_vault = artifact_vault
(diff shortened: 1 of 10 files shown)


jasonLadd:
## ✅ Live-validated end-to-end (full stack rebuilt from branch HEAD `4371929`)

Rebuilt runtime-api + all 7 agents from this branch (`rebuild_and_deploy.sh all`), confirmed the branch code is in the running image (`admission` module + `create_runtime_coordinator` + the shared-coordinator wiring), then ran a real `smoke` / `play_game` cycle (`run_8a80c86e6d06`). Evidence straight from Postgres:

| Check | Result |
|---|---|
| Recruitment → lease | 5 agents (data, eve, max, nat, neo) each acquired a `cycle` lease, `owner_ref = run_8a80c86e6d06` (the run_id) |
| Held for the run | acquired `23:59:08` (= `started_at`), released `00:00:14` (= `finished_at`) → **held 66s, the full run** |
| Released on finalize | all 5 `released_at` set |
| Back to ambient | 0 non-ambient agents after |
| **No-strand invariant** | **0** stranded active `cycle:*` leases anywhere |
| Core path intact | run `completed`, 6 artifacts, no regression |
| Phase-4 coexists | 5 RuntimeActivities, 0 still_running — instrumentation unaffected |

This exercises what the unit tests (which mock the coordinator) can't: the real `PostgresFocusLease`, the `main.py` shared single-writer coordinator (D16), and `ambient→cycle→ambient` running concurrently with the live duty scheduler. All three acceptance criteria from #233 confirmed on real infra.

**Not live-exercised (covered elsewhere / follow-up):**
- The `focus_lease_conflict` deferral path — covered by unit + executor tests; a live version needs an artificially-seeded held lease (low marginal value).
- Concurrent same-agent cycles free-riding a shared lease — a lease-arbitration edge the coordinator's same-mode no-op exposes; filed as #288 (likely out-of-scope for v1.1, tracked so the decision is explicit).

jasonLadd (commented):
reviewed

Issue #152, opened 2026-06-11
Split capabilities/handlers/cycle_tasks.py (3,226 lines, 9 handlers) into per-handler modules


jasonLadd:
## Problem
`src/squadops/capabilities/handlers/cycle_tasks.py` is 3,226 lines containing 9 handler classes plus shared validation heuristics (`_STUB_PATTERNS`, `_STACK_INDICATORS`, `_PRD_COVERAGE_DISCIPLINE_SECTION`, `_detect_stubs`, `_estimate_min_artifacts`, `_rewrite_manifest_identifiers`). `planning_tasks.py` (1,573 lines) imports `_CycleTaskHandler` from it, coupling the two monoliths. Testing one handler requires importing all nine.

## Proposed fix
Mechanical split, no behavior change:
```
capabilities/handlers/cycle/
├── base.py            # _CycleTaskHandler
├── governance.py
├── implementation.py
├── ...one module per handler
└── _shared_validation.py  # heuristics + shared prompt sections
```
Keep all validation heuristics and prompt text byte-identical; existing tests must pass unmodified. Re-export from the old module path if needed for import compatibility.

## Why
Every future handler SIP gets cheaper; git history and review diffs become readable.

_Source: architectural review, 2026-06-11 (priority #4)._

jasonLadd:
Reference for whoever picks this up: an earlier attempt at this decomposition exists at tag **`archive/sip-0092-m1-followup`** (commit `bd37e59`, 2026-05-01, originally on the now-deleted `feature/sip-0092-m1-followup-complexity-cleanup`).

It extracted the FC3 typed-acceptance pipeline out of `cycle_tasks.py` into a clean `typed_acceptance.py` (severity×status blocking matrix, RC-9b 2-strikes evaluator-error counter, artifact materialization, no module-level state) and decomposed `handle()`.

**Reference only — do not cherry-pick.** It's ~211 commits behind main and these handlers have drifted since (SIP-0086 etc.); the extracted logic was behavior-preserving and still lives inline on `main`. Re-author against current code, using the archived shape as a design hint.

jasonLadd:
**Cross-lane heads-up (Spark lane) — sequence before this split.**

Spark is starting **#276**, which edits `cycle_tasks.py` — specifically `QATestHandler` (~line 2036) and the builder-output validation path (stub-fallback detection + deliverable-level `required_files` enforcement). That's the 3,246-line file this issue splits.

[...]

PR #289, opened 2026-07-01
fix(#276): fail qa acceptance on stub-fallback tests that mask a broken entrypoint (Part 1)

jasonLadd:
**Part 1 of #276** (Slice 1a). References #276; does **not** close it — deliverable-level `required_files` enforcement (Part B) and build/boot-the-deliverable (Slice 2, pairs with #176) still to come.

## The bug
A `validation`-profile cycle completed **green** with a backend that doesn't run (`backend/main.py` used `BaseModel` with no `from pydantic import BaseModel`). Root cause: the generated `test_api.py` did

```python
try:
    from backend.main import app
except ImportError:
    from fastapi import FastAPI
    app = FastAPI()  # ~70-line inline re-implementation
```

so pytest collected and ran the **stub**, not the deliverable — exit 0, qa green. The existing `tests_pass` fold can't catch this (the stub *makes* the tests pass), which is exactly why the import defect was invisible to the correction loop.

## Fix
- **New `capabilities/handlers/stub_detection.py`** — `detect_stub_fallback_tests()` flags a test file only when it **both** guards an import with `except ImportError`/`ModuleNotFoundError` **and** constructs a web-app object inline (`FastAPI(`/`Flask(`/…). Conservative by design: catching an optional-dependency `ImportError` *without* rebuilding an app is **not** flagged (false-positive guard, tested).
- **`QATestHandler`** folds the result into `validation` in the output-validation path → `SEMANTIC_FAILURE`, so the SIP-0086 correction loop regenerates the test against the real module. Detection lives in its own module; the `cycle_tasks.py` change is a small block inside the *existing* acceptance branch.

## Why minimal `cycle_tasks.py` footprint
This file is the one true cross-lane code hot spot (Mac #114 typed-eval, #152 split). Keeping the edit to a ~30-line block in the existing branch + a separate helper module. Heads-up already posted on #114/#152; **I'll comment there when this merges.**

## Verification
- New `test_stub_detection.py`: 11 tests (pattern detected across FastAPI/Flask/tuple-except forms; clean tests, optional-dep imports, and the real `main.py` **not** flagged; multi-file sort; empty/nameless inputs safe).
- Full `capabilities/` + `cycles/` suites: **2423 passed**. ruff + test-quality lint clean.

### DIFF
--- a/src/squadops/capabilities/handlers/cycle_tasks.py
+++ b/src/squadops/capabilities/handlers/cycle_tasks.py
@@ -2554,6 +2554,40 @@ class QATestHandler(_CycleTaskHandler):
                 validation.coverage_ratio = passed_count / len(validation.checks)
 
         if output_validation_enabled:
+            # #276: a generated test that hides a broken entrypoint import behind
+            # an ImportError fallback validates a stub app, not the deliverable —
+            # so `tests_passed` above can be falsely green (the stub collects and
+            # passes). Flag it so acceptance fails and the correction loop
+            # regenerates the test against the real module.
+            from squadops.capabilities.handlers.stub_detection import (
+                detect_stub_fallback_tests,
+            )
+
+            stub_offenders = detect_stub_fallback_tests(artifacts)
+            if stub_offenders:
+                validation.checks.append(
+                    {
+                        "check": "no_stub_fallback_tests",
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #290, opened 2026-07-01
fix(#276): build-acceptance — verify the frontend actually builds

jasonLadd:
Build-acceptance increment for #276 (references #276; does not close it — deliverable-level `required_files` still needs a Mac seam, see below).

## Gap this closes
`cyc_2f415e43f9cf` run 1 shipped a **non-runnable frontend** green: the Vite app had no root `index.html`, so `vite build` failed — but **vitest unit tests pass without it**, so acceptance never noticed.

Reproduce-first refinement: **Part A (PR #289) already closed the *backend*-import gap** — by rejecting the `except ImportError` stub, a clean generated test imports `backend.main` directly, so a missing-import `NameError` fails pytest collection → the existing `tests_pass` fold catches it. So this increment adds only the genuinely-uncovered half (frontend build); an explicit backend-import check would be redundant.

## Change
- **`test_runner.run_frontend_build()`** — materialize frontend source, `npm install`, run the package's `build` script (fallback `npx vite build`). Returns `BuildCheckResult` where a **skip is not a failure**: no frontend / no `package.json` / no node / `npm install` failure all return `ran=False` (don't punish the app for our toolchain). Only a build that *ran and errored* fails.
- **`QATestHandler`** folds a build failure into the existing test outcome (`dataclasses.replace` on `exit_code`) → the current acceptance fold → `SEMANTIC_FAILURE` → SIP-0086 correction loop. Minimal `cycle_tasks.py` footprint; report surfaces the build result.
- **Fixed a pre-existing `-n auto` isolation flake** in `test_workspace_cleaned_up_after_run` (globbed shared `/tmp/qa_run_*`, catching a parallel worker's in-flight dir) — now captures this run's exact workspace via `mkdtemp` and asserts removal.

## Deferred: deliverable-level `required_files` (the Dockerfile gap)
The builder run's missing-`Dockerfile` case needs the **full emitted-artifact filename set**, which lives at the **executor** (`stored_artifacts` has filenames + the profile). The qa handler only receives a per-task *filtered* `artifact_contents` subset (`all_artifact_refs` are ID-only; filenames need heavy vault retrieval). So required_files enforcement is a **Mac-seam** item — flagged on #276 for coordination, not forced into a handler hack.

## Verification
- `test_frontend_build.py`: 10 tests (failed-property semantics; skip paths; build fail/success/install-fail/no-node via mocked subprocess).
- capabilities `1083 passed` (3× under `-n auto`, flake gone) + cycles `1352 passed`; ruff + test-quality lint clean.

### DIFF
--- a/src/squadops/capabilities/handlers/cycle_tasks.py
+++ b/src/squadops/capabilities/handlers/cycle_tasks.py
@@ -2242,6 +2242,7 @@ class QATestHandler(_CycleTaskHandler):
             TEST_FRAMEWORK_VITEST,
         )
         from squadops.capabilities.handlers.test_runner import (
+            run_frontend_build,
             run_fullstack_tests,
             run_generated_tests,
             run_node_tests,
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #293, opened 2026-07-01
feat(runtime): RuntimeTransaction UoW — atomic coordinator lease+activity+mode (SIP-0089 §4.5/D25, #244)

jasonLadd:
Completes the SIP-0089 runtime arc: the coordinator's transition now applies **lease → activity → mode as one Postgres transaction** (§4.5 order, D25), so a partial failure aborts rather than relying on best-effort compensation. Unblocked by #233 (which routed recruitment through the coordinator).

Closes #244.

## Design (4 slices)

1. **UoW port + adapter + conn threading** — new `RuntimeTransactionPort` (`begin()` yields an opaque, transaction-bound connection; driver-agnostic so the port stays adapter-free, D26) + `PostgresRuntimeTransaction`. Threaded an optional `conn=` through the 7 methods the coordinator uses (`upsert_state`; lease `get_current`/`request`/`revoke`/`release`; activity `get_current`/`update_state`/`abort`) via a shared `acquire(pool, conn)` helper. **Additive — `conn=None` = today's behavior.**
2. **Coordinator wrapping + §4.5 reorder** — `_apply_transition` wraps arbitration + activity abort + mode write in `transaction.begin()`, reordered so the **activity abort precedes the mode write**. Events emit **only after commit**. Optional `transaction` param defaults to a `_NullTransaction` (yields `conn=None`) → the no-transaction path is a strict no-op. Compensation is now belt-and-suspenders (always release the acquired lease on failure — harmless no-op when a real UoW already rolled it back).
3. **Composition root (D16)** — `create_runtime_coordinator` builds one `PostgresRuntimeTransaction` and injects it, so the single shared mode-writer the executor + scheduler both drive is transactional.
4. **Docs** — coordinator docstring: D25 realized.

## ⚠️ Behavioral change (intended, documented)

The orphaned-activity abort was a Phase-4 *best-effort post-write* seam. Now that it's part of the atomic unit and runs **before** the mode write, an abort **failure aborts the transition** instead of being swallowed. This is strictly *safer* — there's no "mode changed but stale activity still running" state — and the coordinator can't make it best-effort inside a UoW without a savepoint (which would leak the driver into the domain). The v1.1 activity path is effectively unexercised, so this is practically inert today. Captured in the updated `test_activity_abort_failure_aborts_the_transition` + docstrings.

## Acceptance (per the issue)

- ✅ §4.5 order holds: a failure at step 6 (mode write) rolls back the lease acquired in step 4 **and** the activity transition in step 5 — proven against **live Postgres** in `tests/integration/runtime/test_coordinator_transaction_integration.py::test_mode_write_failure_rolls_back_lease_and_activity`.
- ✅ No regression on the common `activity=None` / no-current-activity paths (full unit suite green).

## Tests

- **Unit:** `test_coordinator_transaction.py` — conn threads through all writes in §4.5 order; the NoOp default threads `conn=None`. Updated the activity-abort test to the new semantics. All 7 port fakes thread `conn`.
- **Integration (live Postgres):** mode-write failure rolls back lease **and** activity; positive control commits lease + mode together. Both pass.
- Full regression: **4490 passed / 1 skipped**.

## Also in this branch
- A drive-by from #173: `dispatched_flow_executor.py` comment `spark-squad-with-builder` → `full` (the Spark lane flagged it for the Mac lane).

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -2297,7 +2297,7 @@ class DispatchedFlowExecutor(FlowExecutionPort):
 
             # Issue #110: propagate squad-profile model + overrides so
             # correction-loop reasoning runs on the cycle's specified model
-            # (e.g. spark-squad-with-builder pins data/lead to qwen3.6:27b)
+            # (e.g. the `full` profile pins data/lead to qwen3.6:27b)
             # rather than the agent container's instance default.
             corr_inputs: dict[str, Any] = {
                 "prd": cycle.prd_ref,
--- a/adapters/persistence/runtime/__init__.py
+++ b/adapters/persistence/runtime/__init__.py
@@ -3,5 +3,11 @@
 from adapters.persistence.runtime.activity_postgres import PostgresRuntimeActivity
 from adapters.persistence.runtime.focus_lease_postgres import PostgresFocusLease
 from adapters.persistence.runtime.state_postgres import PostgresRuntimeState
+from adapters.persistence.runtime.transaction_postgres import PostgresRuntimeTransaction
 
-__all__ = ["PostgresFocusLease", "PostgresRuntimeActivity", "PostgresRuntimeState"]
+__all__ = [
+    "PostgresFocusLease",
+    "PostgresRuntimeActivity",
+    "PostgresRuntimeState",
+    "PostgresRuntimeTransaction",
+]
--- a/adapters/persistence/runtime/_conn.py
+++ b/adapters/persistence/runtime/_conn.py
@@ -1,0 +1,36 @@
+"""
+Shared connection helper for the runtime persistence adapters (SIP-0089 §4.5/D25).
+
+A runtime port method either runs inside a caller-supplied unit-of-work
+connection — the ``conn=`` the coordinator threads from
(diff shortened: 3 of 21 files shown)


jasonLadd:
## ✅ Live happy-path regression pass (runtime-api rebuilt from this branch)

Rebuilt runtime-api from the branch (confirmed the UoW is wired: coordinator uses `self._transaction.begin()` and threads `conn=` to the writes), then ran a real `smoke` / `play_game` cycle (`run_87a495a9e189`). The transactional coordinator recruits exactly as before — just atomically now:

| Check | Result |
|---|---|
| Recruited (atomic) | 5 agents `ambient→cycle`, lease acquired with the mode write via the UoW, `owner_ref = run_87a495a9e189` |
| Held for the run | 58s each (full run) — observed live in `mode=cycle` + active lease |
| Released on finalize | all 5 `released_at` set |
| Back to ambient | 0 agents non-ambient |
| **No-strand invariant** | **0** stranded active cycle leases |
| Run outcome | `completed`, 58s |

Together with the live-Postgres integration test in this PR (`test_mode_write_failure_rolls_back_lease_and_activity`, which proves the *rollback* path), #244 is validated end-to-end on real infra: the **commit path** (this cycle) and the **abort/rollback path** (integration test).

jasonLadd (commented):
reviewed

PR #294, opened 2026-07-01
Propose SIP: Cycle Create Preflight (#172 + #224)

jasonLadd:
Proposes a new SIP for design review: **Cycle Create Preflight** — a fail-fast gate that rejects unsatisfiable cycle requests *before persistence or dispatch*, combining #172 (capability/role satisfiability) and #224 (model availability) behind one create-time seam.

This is a **design-review PR** (SIP workflow step 1: propose). It adds a draft to `sips/proposed/`; it does not implement anything.

## The contract
- **Block** when the resolved squad profile can't satisfy the roles the requested workloads imply (the live `builder.assemble`-on-a-builder-less-squad case), or names a model **definitively** unavailable on the backend (the live `qwen3.6:27b`-not-pulled case).
- **Warn-and-allow** when availability is *unverifiable* (backend unreachable) — blocking on a transient outage would be a worse failure than the one we're fixing.
- **No cycle row / no dispatch / no background execution** on rejection (first-class acceptance criteria).
- HTTP **422**; actionable errors that name the missing role or model.
- Shares the model check with `squadops doctor` (parity).

Follows the **SIP-0089 §2.5 reserve-buffer precedent**: pure decision, single choke point, defense-in-depth (dispatch-time validation stays as the deeper net).

## Two open questions for reviewers
1. Where to surface *unverifiable* warnings beyond the create response (metadata / runtime event / Continuum)?
2. The exact `applied_defaults` → required-role map (so the build-task/builder rule generalizes).

## Delivery coordination
Split delivery, unified contract: **Spark** owns the model-availability half + doctor parity; **Macbook** owns the capability half + create-route wiring (per the #224 option-a decision).

Depends on #173 (profile-name consolidation) — already merged.

### DIFF
--- a/sips/proposed/SIP-Cycle-Create-Preflight.md
+++ b/sips/proposed/SIP-Cycle-Create-Preflight.md
@@ -1,0 +1,194 @@
+---
+title: Cycle-Create Preflight
+status: proposed
+author: jladd
+created_at: '2026-07-01T00:00:00Z'
+---
+# SIP: Cycle Create Preflight
+
+## Status
+Proposed
+
+**Targets:** v1.2
+**Combines:** #172 (capability/role satisfiability) + #224 (model availability)
+**Depends on:** #173 (squad-profile name consolidation) — *merged*.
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #302, opened 2026-07-01
refactor(#231): consolidate the health signal to runtime_status at the read surfaces

jasonLadd:
Strangles `network_status` out of the **health-signal** role everywhere it's read, per the documented model (`docs/agent-runtime-status-model.md`: *health = `runtime_status`, posture = `mode`*). #230 did the agents dashboard + the `/health/agents` list; this finishes the read surfaces it missed.

Partial toward #231 — see "What remains"; not closing it.

## What changed (4 slices)
1. **API single-agent** `GET /health/agents/status/{id}` — LEFT JOIN `agent_runtime_state`; now carries `runtime_status` + `mode` at parity with the list route (it was `network_status`-only).
2. **CLI** `squadops status` — the STATUS column shows `runtime_status` (falling back to `network_status` only when there's no runtime row).
3. **`health_checker._compute_network_status`** — deprecation docstring: legacy heartbeat-age flag; do not add new dependencies; read `runtime_status`.
4. **Console** `squadops.home/HomeSummary.svelte` — the squad dot prefers `runtime_status` (#230 fixed `squadops.agents` but missed this one). Source-only — `dist/plugin.js` is gitignored/regenerated by the console build; verified with `vite build`.

## Behavior
`network_status` is **demoted, not removed** — kept as a back-compat field and for the reconciliation offline-mirror (`mark_offline` → `runtime_status`). No migration, no big-bang (the doc's "strangler, not a blocker").

## Tests
Read-surface parity proven: the single-agent route asserts `runtime_status`/`mode`; the CLI test proves the preference via a `runtime_status`-only value (`recovering`). Full regression **4538 passed / 1 skipped**.

## What remains (keeps #231 open)
The **full** consolidation's tail — dropping the `network_status` column from `agent_status` + the reconciliation rework — is a **migration**, so it's gated on **#158** (schema_migrations + applier hardening, in progress on the Spark lane). Recommend landing this read-surface strangle now and doing the column-drop after #158. (Alternatively: close #231 on this and file a dedicated "drop legacy network_status column" follow-up — maintainer's call.)

### DIFF
--- a/console/continuum-plugins/squadops.agents/ui/src/AgentsStatus.svelte
+++ b/console/continuum-plugins/squadops.agents/ui/src/AgentsStatus.svelte
@@ -10,9 +10,9 @@
   function mapAgent(a) {
     return {
       name: a.agent_name || a.agent_id || a.name,
-      // Health = runtime_status (SIP-0089); network_status is the legacy fallback
-      // for an agent that has no runtime-state row yet (#230/#231).
-      status: a.runtime_status || a.network_status || a.lifecycle_state || a.status || 'unknown',
+      // Health = runtime_status (SIP-0089) — the single source of truth, now
+      // always-populated (#305 Part A); no network_status/lifecycle fallback.
+      status: a.runtime_status || 'unknown',
       mode: a.mode || null,
       role: a.role_label || a.role || '',
       current_task: a.current_task_id || a.current_task || null,
--- a/console/continuum-plugins/squadops.home/ui/src/HomeSummary.svelte
+++ b/console/continuum-plugins/squadops.home/ui/src/HomeSummary.svelte
@@ -158,7 +158,8 @@
         <div class="squad-row">
           {#each agentStatus as agent}
             <div class="squad-member">
-              <span class="squad-dot {agent.network_status === 'online' ? 'dot-online' : 'dot-offline'}"></span>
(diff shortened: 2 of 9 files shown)


jasonLadd:
## Folded in #305 Part A — single source of truth, no fallback

Since this PR was still open, we folded **#305 Part A** in rather than merging the temporary `runtime_status || network_status` fallback and removing it later. **#302's final state now has zero dual-source "or" logic — the fallback never lands on main.**

- **Always-populate:** the heartbeat now ensures the runtime row even when lifecycle maps to no `runtime_status` (that skip left agents row-less — the *only* reason the fallback existed); reconciliation `ensure_state`s before `mark_offline` to backfill legacy agents. So `runtime_status` is never `None` at the read surfaces.
- **Fallback removed everywhere:** API, CLI, and *both* console plugins — including the `runtime_status||network_status||lifecycle_state||status||'unknown'` chain in `AgentsStatus.svelte` (from #230) that the first pass missed.
- Grep-clean of the dual-source logic; full regression **4540 passed**.

So #302 is now **the full read-surface consolidation to a single source of truth.** Only **Part B** (drop the dead `network_status` column) remains — #305, gated on #158.

jasonLadd (commented):
reviewed

PR #304, opened 2026-07-01
fix(#303): qa.test harness — discover package.json, block missing frontend build, fix backend pytest path

jasonLadd:
Closes #303. Follow-up from the 1.2.0 build-quality re-validation (`cyc_1d2e21ab0cfb`, group_run/lite) — after #296 the config files reach the QA workspace, but the build check *still* didn't catch a broken fullstack app. Three robustness fixes in `test_runner.py`:

## 1. `package.json` discovery (not a fixed `frontend/`)
New `_find_package_json_dir` locates the Node project root from the materialized file list (shallowest wins). The run put `package.json` at `frontend/src/`; `run_frontend_build`/`run_node_tests` only checked `frontend/package.json` → **skipped**. Now they run from wherever it actually is. Removed the rigid `target_dir` cwd path on `run_node_tests` (discovery replaces it; `target_dir` stays on `run_frontend_build` **only** as the frontend-file filter for the fullstack split).

## 2. A missing frontend build is now **blocking**, not a skip
Frontend source present but **no** discoverable `package.json` → `BuildCheckResult(ran=True, ok=False)` → the run fails instead of shipping a non-buildable frontend green. A genuinely backend-only deliverable (no frontend source) still skips.

## 3. Backend pytest `PYTHONPATH`
New `_source_dir_pythonpath` puts every Python source dir on `PYTHONPATH`, so a `backend/`-nested app's `from main import app` resolves — instead of the spurious `ModuleNotFoundError: No module named 'main'` that masked the real test result.

## Evidence (cyc_1d2e21ab0cfb `test_report.md`)
```
E   ModuleNotFoundError: No module named 'main'          # fixed by #3
frontend (non-blocking): No package.json found in frontend  # fixed by #1 + #2
```

## Tests
Helper unit tests (discovery: subdir/root/shallowest/none; PYTHONPATH nesting); `run_frontend_build` subdir-discovery + blocking-on-missing; `run_node_tests` subdir discovery; flipped the obsolete "no package.json = skip" expectation and replaced the removed-`target_dir` test with a discovery test. ruff + quality lint clean; **1107 capabilities tests green**.

## Close criterion
Unit-verified. Full end-to-end confirmation is the next re-validation cycle (after a rebuild) showing the frontend build **runs** and the backend tests import cleanly.

### DIFF
--- a/src/squadops/capabilities/handlers/test_runner.py
+++ b/src/squadops/capabilities/handlers/test_runner.py
@@ -69,6 +69,44 @@ def _materialize_files(
             fh.write(rec["content"])
 
 
+def _find_package_json_dir(files: list[dict[str, str]]) -> str | None:
+    """Return the workspace-relative dir of the shallowest ``package.json``.
+
+    Discovers where the Node project root actually is instead of assuming a fixed
+    ``frontend/`` — models sometimes place ``package.json`` at ``frontend/src/``
+    or the workspace root (#303). ``""`` means the workspace root; ``None`` means
+    no ``package.json`` was produced at all.
+    """
+    dirs = [
+        os.path.dirname(rec["path"])
+        for rec in files
+        if os.path.basename(rec["path"]) == "package.json"
+    ]
+    if not dirs:
+        return None
+    # Shallowest (fewest path segments, then shortest) is the real project root.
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #309, opened 2026-07-01
feat(#224): model-availability preflight decision (Spark half of SIP-0095)

jasonLadd:
Refs #224 (partial — the Spark half of the option-A split). The cross-lane handoff was announced on #224.

## What this adds
The **model-availability** half of the SIP-0095 cycle-create preflight, into the shared `cycles/preflight.py` (which #298 scaffolded with `Finding`/`PreflightDecision`/`combine`/`required_roles_decision`):

```python
def model_availability_decision(profile, pulled_models) -> PreflightDecision
```

Pure, mirroring `required_roles_decision` — the caller fetches the pulled model **names** and passes them in; no adapter import.

## Semantics (SIP §6.2/§6.3/§137)
- **Verifiable + required model definitively missing → `block`** (`model_unavailable`), one finding per missing model.
- **`pulled_models is None` (backend unreachable) → `warning`** (`model_unverifiable`) — **warn-and-allow**, never block on missing evidence.
- **Reachable-but-empty list is verifiable → blocks** (the crucial None-vs-`[]` distinction).
- **Exact canonical-tag matching** (tagless ⇒ `:latest`, so `llama3.2` == `llama3.2:latest`) with **no prefix/family inference** (`qwen3:7b` never satisfies `qwen3:27b`) — the §137 decision, avoiding both false blocks and false passes. Enabled agents only.

## Tests (9 new, 21 total in the file)
allow / block-with-actionable-message / warn-and-allow-on-None / reachable-empty-blocks / canonical-`:latest`-match / no-family-inference / disabled-agent-skipped / multi-missing / empty. ruff + quality lint clean; **1379 cycles tests green**.

## Remaining on #224 (this PR does NOT close it)
- **Mac lane (SIP-0095 Phase 3):** wire `combine(required_roles_decision(...), model_availability_decision(...))` into the create route; supply `pulled_models` from the LLM adapter (`None` on query failure); surface the warning on the response + CLI.
- **Spark follow-on:** `doctor` parity — call the same decision with `ollama list` as the pulled-list source (§8).

Parallel-safe with SIP-0090 (#308) — disjoint files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #315, opened 2026-07-01
feat(SIP-0095 Phase 3): wire create-time preflight into the cycle-create route

jasonLadd:
Implements **SIP-0095 Phase 3** — wires the create-time preflight (Phase 1 capability decision + Phase 2 model-availability decision) into the cycle-create route so it **fails fast before persist/dispatch**.

## What changed
- `create_cycle` now runs `combine(required_roles_decision, model_availability_decision)` right after the profile resolves and **before** anything is persisted.
- **Blocks** with `422 PREFLIGHT_REJECTED` when the squad can't satisfy the requested workloads' required roles, or names a model **definitively** not pulled.
- **Warn-and-allow** when the LLM backend is unreachable/unconfigured (`_pulled_model_names()` → `None`) — never block on missing evidence (§6.2). Warnings are logged; the **response/CLI warning surface is Phase 4**.
- `PreflightRejectedError(CycleError)` → 422 through the existing `handle_cycle_error` (`{"error": {...}}` envelope) — conforms to the cycle-route error convention, no new shape.

## Decisions honored (locked in the plan)
Fail-fast before persist · HTTP **422** · **exact canonical-tag** model match (`gpt-4` ⇒ `gpt-4:latest`, no family inference) · unreachable ⇒ warn-and-allow · static workload→role only (builder/materialized-plan stays a dispatch check → #295).

## Tests
Route-level: role-block (no persist), model-block (reachable-but-empty backend), model-allow (canonical match), warn-and-allow (unverifiable). Updated the shared create fixture to a valid full-plan squad — a 1-agent squad now *correctly* fails preflight. **4572 regression green.**

## Scope
`Closes #224` — the model-availability preflight is now wired into **cycle-create** (the doctor + decision half landed in Phase 2 / #309). Completes the create-time capability preflight wiring (#172, already closed). **Not** in scope: the Phase-4 warning **display** surface (return warnings in the response + CLI echo + persist on the cycle record), and the materialized-plan/builder hoist (#295).

Closes #224

### DIFF
--- a/src/squadops/api/routes/cycles/cycles.py
+++ b/src/squadops/api/routes/cycles/cycles.py
@@ -2,6 +2,7 @@
 Cycle API routes (SIP-0064 §9.3).
 """
 
+import logging
 import uuid
 from datetime import UTC, datetime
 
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #312, opened 2026-07-01
feat(SIP-0090 Phase 1): core embodiment model — slice 1a (model + budgets + port + coordinator)

jasonLadd:
Implements **SIP-0090 §13 Phase 1, slice 1a** — the migration-free core embodiment substrate (plan: `docs/plans/SIP-0090-phase-1-plan.md`). Mirrors the SIP-0089 shape: pure model + port + coordinator first; persistence (1b) follows.

## What's here (4 incremental slices)
| Slice | Module | Tests |
|---|---|---|
| 1 | `runtime/embodiment.py` — `Embodiment` model, explicit §4.1 transition matrix, opaque `Location`, credentials invariant | 35 |
| 2 | `runtime/budgets.py` — consumable `{limit,consumed}` + capacity `{allowance,in_use}`, non-silent exhaustion | 18 |
| 3 | `ports/runtime/embodiment.py` `EmbodimentStatePort` + `embodiment.*`/`budget.exhausted` events & reasons | 3 |
| 4 | `runtime/embodiment_coordinator.py` — transitions + single-active enforcement + event emission | 9 |

Full regression: **4633 passed, 1 skipped**. Adapter-free (D26).

## Review-feedback decisions baked in (from the plan PR #308)
- **Explicit transition matrix** (`_ALLOWED_ATTACHMENT_TRANSITIONS`) — completes SIP §5.2's happy-path ASCII with the failure-termination edges; `detached` terminal; illegal shortcuts (`unattached→attached`, `attached→reconnecting`) rejected.
- **`EmbodimentStatePort`** (not a bare `EmbodimentPort`) — a record/lifecycle store with **no action/intent method**; the Phase-2 driver stays a separate `EmbodimentSurfacePort`. Guarded by a structural authority-boundary test.
- **Budgets**: consumable vs capacity split; `concurrency` is a gauge with `acquire`/`release` (release-underflow raises — a leaked slot is observable); **non-silent exhaustion is type-unrepresentable** (`BudgetDecision.__post_init__`); reset/replenishment deferred.
- **Credentials**: `credentials_ref` validated at construction — a raw secret can never land in a record (§9).
- **`budget_exhausted`** single-sourced from `reasons.py` (no duplicated literal).
- **Named D26 guard**: `test_coordinator_imports_no_concrete_adapter`.

## Acceptance boundary (per plan §2)
[...]

PR #317, opened 2026-07-01
feat(SIP-0090 Phase 1): embodiment persistence — slice 1b (completes Phase 1)

jasonLadd:
Implements **SIP-0090 §13 Phase 1, slice 1b** — embodiment persistence. With 1a (merged, #312) this **completes Phase 1**: the §13 acceptance is *"Embodiment records exist and transition states cleanly."* 1a proved "transition cleanly" at the model level; 1b makes the records **exist**.

## What's here
- **`infra/migrations/1140_embodiments.sql`** — the `embodiments` table (§5.3 fields, D3 CHECK constraints mirroring the code Literals) + the **partial unique index** `uq_embodiments_one_active_per_agent` enforcing the single-active rule (§5.5) in the DB — the hard backstop behind the coordinator's logic-level guard. Idempotent DDL, mirrors `1120_focus_leases.sql`.
- **`adapters/persistence/runtime/embodiment_postgres.py`** — `PostgresEmbodimentState` implementing `EmbodimentStatePort`. JSONB `capability_set` (dumps on write, decode → tuple on read), the §4.5/D25 `conn` unit-of-work seam on mutators.

## Tests
- **9 unit tests** (mock-conn, in the regression suite): row→dataclass mapping, JSONB round-trip to a **tuple** (a list would break frozen-dataclass equality), the `get_active` active-state filter, not-found → KeyError.
- **4 integration tests, live-validated against the running stack DB** ✅: record round-trips with every field intact; **the single-active index rejects a 2nd live embodiment** (`UniqueViolationError`); detached history + other agents don't block; transition moves state + `get_active` reflects it.

**4652 unit regression green**; integration validated on real Postgres.

## Deferred (per the defer-infra-completeness principle)
- **Budget persistence** (`1141_agent_budgets.sql` + adapter) — needs a `BudgetPort` that isn't in 1a, and is **not** part of the §13 acceptance (budgets become persistence-relevant when Phase 3 activities consume them). The budget *primitives* (1a slice 2) are complete and enforce non-silent exhaustion.
- **Composition-root wiring** — no live consumer drives the `EmbodimentCoordinator` until Phase 2 (Discord adapter); wiring it now would be a factory ahead of its use.

Does not close an issue (no SIP-0090 tracking issue). Completes Phase 1 alongside #312.

### DIFF
--- a/adapters/persistence/runtime/embodiment_postgres.py
+++ b/adapters/persistence/runtime/embodiment_postgres.py
@@ -1,0 +1,135 @@
+"""Postgres-backed embodiment-state adapter (SIP-0090 Phase 1 §5, slice 1b).
+
+Implements `EmbodimentStatePort` via the shared asyncpg pool. All writes hit
+`embodiments` (migration `1140_embodiments.sql`). The single-active-embodiment
+invariant (§5.5) is enforced in the DB by the partial unique index
+`uq_embodiments_one_active_per_agent` — the hard backstop behind the
+EmbodimentCoordinator's logic-level guard.
+
+`capability_set` is a JSONB array (`json.dumps` on write, decode on read — the cycle
+registry convention). Mutating methods carry the §4.5/D25 `conn` seam so a Phase-3
+coordinator can commit an embodiment transition atomically with other writes.
+"""
+
+from __future__ import annotations
+
+import json
+from typing import Any
+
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #320, opened 2026-07-04
fix(#319): render cycle-route error messages in the CLI

jasonLadd:
## What
Cycle-route errors were showing as `Error: validation failed —` with **no reason**. Root cause: `_handle_error_status` read `body["error"]["message"]`, but cycle routes nest the envelope under FastAPI's `detail` (`{"detail": {"error": {...}}}` via `handle_cycle_error`). So `error_msg` was always empty.

**Pre-existing** — affects every cycle-route 404/409/422 — but the **SIP-0095 preflight** made it painful: the API returns a perfect *"requires model `qwen3.6:27b`, pull it or choose another profile"* that the operator never saw.

## Fix
Unwrap `detail` before reading the `error` envelope; still handle plain-string `detail` (health/auth) and un-wrapped bodies.

## How it was found
Live validation of the SIP-0095 preflight on the deployed stack (rebuilt runtime-api, ran a `full` cycle → 422). The API returned the right message; the CLI dropped it. The existing CLI tests only covered the *un-wrapped* `{"error": {...}}` shape — which is exactly why the bug slipped through.

## Tests
- New: the wrapped `{"detail": {"error": {...}}}` envelope renders the message; plain-string `detail` still renders.
- **Live-validated**: a blocked `full` create now prints the full model message (verified after the fix).

Closes #319

### DIFF
--- a/src/squadops/cli/client.py
+++ b/src/squadops/cli/client.py
@@ -157,8 +157,19 @@ class APIClient:
 
         try:
             body = response.json()
-            error_msg = body.get("error", {}).get("message", "") if isinstance(body, dict) else ""
-            detail = body.get("error", {}).get("details") if isinstance(body, dict) else None
+            # Cycle routes nest the {"error": {...}} envelope under FastAPI's `detail`
+            # ({"detail": {"error": {...}}}); health/auth return {"detail": "<string>"}.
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #321, opened 2026-07-04
feat(SIP-0095 Phase 4): surface create-time preflight warnings (response + CLI)

jasonLadd:
Completes the SIP-0095 arc — the preflight's **warn-and-allow** findings now reach the operator instead of only hitting the server log.

## What changed
- `CycleCreateResponse` gains **`warnings: [{code, message}]`** (`PreflightWarningDTO`).
- `_run_create_preflight` now **returns** the non-blocking warnings (still raises 422 on a block); the route attaches them to the response.
- `squadops cycles create` **echoes** each warning (`Warning: preflight [code] — message`); `--format json` carries them inline.

This complements #320 (which fixed the *error*-message rendering); Phase 4 is the *warning* surface.

## The one warning source today
`model_unverifiable` — the LLM backend was unreachable at create time, so models couldn't be verified (warn-and-allow, never block on missing evidence).

## Tests + validation
- Unit: warn-and-allow surfaces `model_unverifiable` on the response; a verified squad has `warnings: []`. **4656 regression green.**
- **Live-validated** (rebuilt runtime-api): a `smoke` create returns the `warnings` field as `[]` (Ollama reachable → models verified, no false positive) and the cycle creates. The warn *display* path is unit-proven — triggering it live needs an unreachable backend, which I did not induce on the running stack (it would disrupt agents).

## Deferred (defer-infra-completeness)
Persisting warnings on the cycle record for `cycles show` — a migration + `Cycle`-model change for a narrow (backend-unreachable) edge. The response + CLI echo already deliver the create-time operator feedback; persistence can follow if durability is wanted.

References SIP-0095 (no tracking issue; #172/#224 already closed).

### DIFF
--- a/src/squadops/api/routes/cycles/cycles.py
+++ b/src/squadops/api/routes/cycles/cycles.py
@@ -9,7 +9,11 @@ from datetime import UTC, datetime
 from fastapi import APIRouter, BackgroundTasks, Depends
 
 from squadops.api.middleware.auth import require_scopes
-from squadops.api.routes.cycles.dtos import CycleCreateRequest, CycleCreateResponse
+from squadops.api.routes.cycles.dtos import (
+    CycleCreateRequest,
+    CycleCreateResponse,
+    PreflightWarningDTO,
+)
 from squadops.api.routes.cycles.errors import handle_cycle_error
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #338, opened 2026-07-06
refactor(#332): hoist copy-pasted helpers to _CycleTaskHandler base (step 0 of #152)

jasonLadd:
## What

Hoists the four copy-pasted helper groups in `capabilities/handlers/cycle_tasks.py` onto the `_CycleTaskHandler` base class, per #332 (found by the 2026-07-04 health assessment):

- `_resolve_artifact_content` — was 3 copies (Dev/QA byte-identical; Builder identical minus docstring) → one base method
- `_resolve_with_vault_fallback` — was 3 copies (same pattern) → one base method
- `_record_generation` — was 4 defs. Dev/QA byte-identical; Builder differed **only** in its prompt-layer id strings (`build` → `assemble`) → hoisted with a `_prompt_layer_kind` class attribute (`"build"` on the base, `"assemble"` on `BuilderAssembleHandler`). `GovernanceReviewHandler` keeps its own variant as an override — it has a genuinely different call shape (derives model from `chat_kwargs`, no `ChatMessage`) — noted in the base docstring.
- `_validate_output` untouched — that one is genuine base+override polymorphism, not copy-paste.

**Behavior-preserving:** prompt-layer ids, vault-fallback semantics (D3), and throughput logging are byte-for-byte what each handler emitted before. `cycle_tasks.py`: 3,276 → 3,098 lines (−178).

## Why now

Step 0 of the #152 module split: hoisting before the split prevents the duplication from being frozen into five separate per-handler modules, where it would be far harder to ever unify. (#152's #276-gate is satisfied — #289/#290 merged — so this sequence is clear to proceed.)

## Verification

- AST check: each helper defined exactly once on the base; Governance override and Builder attribute confirmed; no other class defines them.
- `ruff format` + `ruff check` clean.
- Capabilities suite: **1,107 passed**. Full regression: **4,656 passed, exit 0** — identical counts to the pre-change baseline.
- **Live cycle on the rebuilt stack** (agents rebuilt from this branch): smoke cycle `cyc_221ae310a96b` / `run_e2ffcc19cc9a` → **completed**; all 5 role agents received and succeeded their tasks; zero errors/tracebacks in agent logs since rebuild.

Closes #332
Refs #152

### DIFF
--- a/src/squadops/capabilities/handlers/cycle_tasks.py
+++ b/src/squadops/capabilities/handlers/cycle_tasks.py
@@ -716,6 +716,120 @@ class _CycleTaskHandler(CapabilityHandler):
             _evidence=evidence,
         )
 
+    def _resolve_artifact_content(
+        self,
+        inputs: dict[str, Any],
+        filename_substring: str,
+    ) -> str | None:
+        """Resolve artifact content by filename substring from inputs."""
+        contents = inputs.get("artifact_contents", {})
+        for key, value in contents.items():
+            if filename_substring in key:
+                return value
+        return None
+
+    async def _resolve_with_vault_fallback(
+        self,
+        inputs: dict[str, Any],
+        filename_substring: str,
+    ) -> str | None:
+        """Resolve artifact content with vault fallback (D3).
+
+        Tries ``artifact_contents`` first, falls back to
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #343, opened 2026-07-07
fix(#342): resumed runs fail instantly — executor re-issues an illegal RUNNING→RUNNING transition

jasonLadd:
## What

Every `squadops runs resume` since the #222/#256 wiring (`3a89cb3`, 2026-06-28) has failed instantly on a real registry: the resume route flips the run PAUSED→RUNNING before enqueuing execution, then `execute_run` unconditionally issues its own `update_run_status(RUNNING)` — and `RUNNING→RUNNING` has no self-loop in the lifecycle state machine, so `IllegalStateTransitionError` lands in the terminal handler and the run goes FAILED before any task dispatches.

**Fix (2 lines):** `execute_run` transitions only when the run isn't already RUNNING. The create path is unchanged (QUEUED→RUNNING still issued by the executor). Chosen over adding a RUNNING→RUNNING self-loop to the state machine, which would weaken transition auditing globally.

## How it was found

SIP-0097 slice 2's mandatory **paused-and-resumed live validation** (its terminal-status-mapping stage touches the PAUSED paths): duty-deferred run `run_878de3ad5d0e` paused correctly, then failed ~instantly on resume with no agent activity and no Task Plan in the failure report. This is exactly the deferred **#258** scenario, exercised end-to-end for the first time — the #222-era tests used bare `AsyncMock` registries, so the illegal transition was invisible (both real registries enforce `validate_run_transition`; so does the memory one).

**Pre-existing, not a slice-2 regression:** the failing line is byte-identical on main; this PR is cut from main independently of the slice-2 branch per SIP-0097 §5 ("a bug found mid-slice is filed and fixed in its own PR, never silently in a move").

## Verification

- New `TestResumeOfAlreadyRunningRun` wires the mock registry with the real `validate_run_transition` — the already-RUNNING (resumed) run now completes with `RUN_COMPLETED` and no `RUN_FAILED`; the QUEUED (created) run still gets the executor's RUNNING transition first. Without the fix, the first test fails exactly as production did.
- Regression: **4,662 passed, exit 0**.
- Live re-validation of the full pause→resume→complete path happens as part of SIP-0097 slice 2's validation once this merges (will be reported on that PR and should close out #258's scenario).

Fixes #342
Refs #222, #256, #258, SIP-0097 §5

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -173,8 +173,13 @@ class DispatchedFlowExecutor(FlowExecutionPort):
             run = await self._cycle_registry.get_run(run_id)
             profile, _ = await self._squad_profile.resolve_snapshot(profile_id)
 
-            # queued/failed/paused -> running
-            await self._cycle_registry.update_run_status(run_id, RunStatus.RUNNING)
+            # queued/failed/paused -> running. Skipped when already RUNNING:
+            # the resume/retry routes flip the run to RUNNING before enqueuing
+            # execution (#222/#256), and RUNNING -> RUNNING is an illegal
+            # transition — the unconditional call instantly failed every
+            # resumed run on a lifecycle-enforcing registry (#342).
+            if run.status != RunStatus.RUNNING.value:
+                await self._cycle_registry.update_run_status(run_id, RunStatus.RUNNING)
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #346, opened 2026-07-07
fix(#345): assemble-command CLI test assertions immune to FORCE_COLOR environments

jasonLadd:
## What

Two `test_assemble_command.py` assertions fail on **any branch including clean main** whenever the shell exports `FORCE_COLOR` (e.g. `FORCE_COLOR=3`, which some terminal/agent environments set): rich's color-detection override kicks in under CliRunner capture and its auto-highlighter styles numbers as separate tokens, so `Assembled 2 file(s)` becomes `…\x1b[1;32m2\x1b[0m\x1b[1;32m \x1b[0m\x1b[1;32mfile\x1b[0m…` and the naive substring `"2 file(s)"` no longer matches. Only digit-containing assertions break, which is why exactly these two failed.

**Fix:** a `_plain()` helper strips ANSI style sequences from `result.output` before asserting, applied uniformly to all 7 substring assertion sites in the file. Assertion semantics unchanged; no production code touched.

## Verification

- File suite: 8 passed both with `FORCE_COLOR=3` and without.
- Full regression: **4,671 passed, exit 0 under `FORCE_COLOR=3`** (previously 2 failed) and 4,671 passed without — the suite is now deterministic across color environments.

Found while running the SIP-0097 slice-2 regression in a shell exporting `FORCE_COLOR=3`; reproduced on clean main before fixing.

Fixes #345

### DIFF
--- a/tests/unit/cli/test_assemble_command.py
+++ b/tests/unit/cli/test_assemble_command.py
@@ -8,6 +8,7 @@ Part of Phase 3.
 
 from __future__ import annotations
 
+import re
 from unittest.mock import MagicMock, patch
 
 import pytest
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #350, opened 2026-07-08
feat(#295): SIP-0097 slice 6 — reject unsatisfiable plans at the plan-review gate

jasonLadd:
## SIP-0097 slice 6 of 6 — the #295 rider (§8), the arc's one behavior addition

Before `_handle_gate` pauses a run for operator review, the run's materialized implementation plan (same artifact detection as `_load_plan_for_run`) is now validated against the squad profile via `ImplementationPlan.validate_against_profile`. A plan naming a role the squad can't satisfy fails **at the gate** with an actionable message naming the role, profile, and gate — instead of aborting ~9 minutes later at implementation dispatch (#172's late-failure half, which SIP-0095 deliberately left out of create-time scope because the plan doesn't exist yet at create time).

### Semantics
- **Keyed on plan presence + the `implementation_plan` flag, not the gate's name** — robust across the `progress_plan_review` / `plan-review` naming variants.
- **Fails before pausing**: the operator is never asked to review a doomed plan. The failure surfaces through the standard `_ExecutionError` → `RunCompletion` terminal path (run FAILED, honest report, events) — no new status semantics.
- **Dispatch-time net retained unchanged** (`generate_task_plan` → `CycleError`): absent or unreadable plans defer to it — the gate check only ever *adds* an earlier rejection, never a pass.
- **Backward compatible**: callers that don't thread `stored_artifacts`/`profile` get pre-#295 gate behavior; smoke/selftest cycles (no `implementation_plan`) are untouched.
- Gate handling stays executor-resident per §6.3/§6.7 — the check is a new executor method, not a collaborator.

Example failure message:
> `Plan rejected at gate(s) ['progress_plan_review']: the materialized implementation plan cannot be satisfied by squad profile 'smoke' — Task 1: role 'builder' not in profile. Fix the plan or run with a squad profile that has the missing role(s).`

### Verification
- **Regression:** `source .venv/bin/activate && ./scripts/dev/run_regression_tests.sh` → **4,694 passed, 1 skipped, exit 0** (captured exit code, not piped).
- **Tests** (`TestPlanReviewGateCheck`, in the gate-decisions file since gate handling is executor-resident): unsatisfiable plan rejected **pre-pause** with the exact message asserted and no `PAUSED` write / no `RUN_PAUSED` emit; satisfiable plan gates normally; no-plan, plan-disabled, unreadable-plan, and unthreaded-caller paths all preserve prior behavior exactly.
[...]

PR #351, opened 2026-07-08
fix(#327): prompt-pack drift — hard-fail manifest verification + deploy-time LangFuse sync

jasonLadd:
## Root cause (two drift layers, one disease: nothing maintained the pins)

1. **Registry drift — the functional blocker.** Agents resolve prompt assets from LangFuse (`SQUADOPS__PROMPTS__ASSET_SOURCE_PROVIDER=langfuse`), seeded once by `upload_prompts_to_langfuse.py` and never re-synced. Templates added since (SIP-0093's `request.governance_review_plan_manifest` among them) were missing from the registry, so every plan-authoring cycle failed at `governance.merge_plan` with `Prompt asset not found` — even though the file ships in the agent image (the misleading part that made this look like a packaging bug).
2. **Manifest-hash drift — the fleet-wide startup warning.** `manifest.yaml`'s top-level `manifest_hash` was stale: `regen_fragment_manifest.py` only maintained the per-fragment `sha256` entries, nothing maintained the whole-manifest fingerprint, and the loader warned-and-continued — masked config drift, exactly what the no-masking hard rule forbids.

## Fixes (per the issue's own sketch)

- **Loader fails hard**: `FileSystemPromptRepository._load_manifest` now raises `ManifestValidationError` on a `manifest_hash` mismatch (message points at the regen tool) instead of one warning line that #323's log flood evicts within hours. Unpinned manifests (no hash) still skip verification — test fixtures and local trees keep working.
- **One tool maintains the whole manifest**: `regen_fragment_manifest.py --check/--write` now also verifies/recomputes the top-level `manifest_hash`, computed exactly the way the runtime computes it.
- **`manifest.yaml` regenerated** (fragments were already 30/30 in sync; only the top-level hash was stale — matching the fleet warning byte-for-byte).
- **Deploy = sync**: `rebuild_and_deploy.sh` gains Step 4b — when the stack uses the langfuse provider, every agent deploy re-uploads the repo's prompt assets with the `production` label the adapters resolve against. The drift class can't recur silently. (Note: upload label must be `production`, not the script's env-derived default — that mismatch is its own trap.)
- **CI guards**: new tests pin the shipped manifest (stored `manifest_hash` == computed; every per-fragment hash matches its file body) so drift fails the suite before it ships, plus loader unit tests for the hard-fail (stale hash raises with the regen pointer; pinned-correct loads; unpinned skips). One fixture subtlety worth knowing: an all-digit fake hash gets YAML-parsed as int `0` (falsy) and silently skips verification — the test uses a lettered hex string.

## Deployed-state remediation (done alongside)
All 46 assets uploaded to the local LangFuse with the `production` label — `request.governance_review_plan_manifest` now resolves (v1, confirming it was never in the registry).

## Verification
- **Regression:** `./scripts/dev/run_regression_tests.sh` → **4,699 passed, 1 skipped, exit 0**.
- **Live (stack rebuilt from this branch — which also exercised Step 4b: "46 assets synced"):**
  - Fleet startup: the `Manifest hash mismatch` warning is **gone**.
  - The exact failing scenario re-run (`build` × `smoke`, previously `cyc_266191c438da` failing at `merge_plan` in 2m23s): framing run `run_158181a30fe5` **completed all 7 planning tasks including `governance.merge_plan` and `governance.review_plan`** (4m42s, with two LLM-quality corrections self-healed en route — the correction machinery working as designed on the 3b model). Inter-workload `progress_plan_review` gate decision + workload advance both worked.
  - The implementation run then failed in 6s at `governance.define_done` — the 3b smoke model emitted malformed JSON and D9 policy hard-aborts on define_done failure. That's small-model output quality on a plumbing profile (smoke is not a build-capable squad), orthogonal to this fix; dispatch, prompt resolution, and reply plumbing all worked.
  - The 3b model emitted no `implementation_plan.yaml`, so the #295 gate check's live positive path remains unit-covered only (as disclosed in #350); it needs a build-capable squad (`lite`/`full`), not this fix.

## Risk note
The hard-fail is a deliberate behavior change: an agent with a drifted manifest now refuses to serve prompts instead of running unpinned. That is the issue's explicit ask, and the CI guard + regen tool + deploy sync make the failure mode reachable only by shipping a hand-edited manifest.

Fixes #327


jasonLadd:
Review follow-up: the *deploy = sync* invariant here only covers the compose pipeline — an AWS/Edge deploy path would need its own sync call or #327 recurs there. The durable, topology-independent guard (boot-time / preflight verification that packaged prompt assets resolve from the registry — the LangFuse analog of this PR's manifest hard-fail) is deliberately deferred per defer-infra-completeness and tracked as #352, to ride with or gate the Edge Deployment Profile SIP.

jasonLadd (commented):
reviewed

PR #354, opened 2026-07-08
refactor(#323): migrate agent comms from queue polling to a persistent push consumer

jasonLadd:
## Summary

The agent entrypoint loop polled its comms queue: `consume(comms_queue, max_messages=1)` opened a fresh `basic.consume`, waited up to 1s, and cancelled it — every cycle, per agent. That cost consumer churn (counts flapping 0↔1 with recurring consumer-less windows), up to a full second of pickup latency, and the `aio_pika` "RobustQueueIterator closing" INFO line once per second per agent — 99.7% of retained agent logs, evicting all real signal within ~1.5h.

`_consume_tasks` now registers **one persistent `subscribe()` consumer** (the SIP-0094 push primitive, which already handles resubscribe-on-channel-drop and callback isolation) and parks on the shutdown event. The new `_process_comms_message` callback carries the old loop's routing unchanged and never raises; the subscription layer acks after every delivery — success or failure — which is exactly the old loop's ack-always policy. On shutdown the consumer is cancelled before the connection closes.

`RabbitMQAdapter` gains an opt-in `prefetch_count` (channel QoS); the agent passes 1 so the broker hands it one unacked message at a time, preserving the poll loop's one-at-a-time pickup. Other adapter users are untouched (default None = broker default).

Also brings the two-lane plan up to date: #323 + #186 rows marked done; #234 is the remaining 1.3.0 core item.

## Live validation (deployed stack rebuilt from this branch)

- **Consumer churn gone:** `rabbitmqctl list_queues name consumers` sampled 5× at 3s intervals — every `*_comms` queue steady at 1 (before: flapping 0↔1 between every sample).
- **Log flood gone:** 0 "closing with timeout" lines across all 7 agents since restart (before: 397/397 lines of max's last 10 minutes were the flood). The line now appears exactly once, on genuine shutdown — where you'd want it.
- **Lite cycle end-to-end:** `cyc_3833c109e6f2` / `run_9bd678b82e42` on `play_game` created → completed (~2 min); agents picked up and processed task envelopes delivery-time, replies published, run completed.
- **Graceful shutdown + reconnect:** `docker compose stop neo` → clean teardown ("Stopping agent" → single iterator close → "RabbitMQ connection closed" → "Task consumer stopped"); on start, the consumer re-registers (neo_comms=1) with no flood.
- Regression suite: 4,718 passed, exit 0. `ruff format` + `ruff check` clean.

## Why this also closes #329

#329 was the interim mitigation (demote `aio_pika` INFO logs) so observability was restored *before* this structural fix landed. It was never implemented, and this PR removes the flood at its cause — the demotion would now only mask the one log line we want to see (genuine reconnect/shutdown). Obsolete, not fixed.

Closes #323
Closes #329

### DIFF
--- a/adapters/comms/rabbitmq.py
+++ b/adapters/comms/rabbitmq.py
@@ -56,6 +56,7 @@ class RabbitMQAdapter(QueuePort):
         url: str,
         namespace: str | None = None,
         consume_poll_timeout_seconds: float = _DEFAULT_CONSUME_POLL_TIMEOUT,
+        prefetch_count: int | None = None,
     ):
         """
         Initialize RabbitMQ adapter.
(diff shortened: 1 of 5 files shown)


jasonLadd (commented):
reviewed

PR #359, opened 2026-07-08
fix(#326): move agent-status writes off the unauthenticated /health lane

jasonLadd:
## Summary

Closes #326.

`POST /health/agents/status` and `PUT /health/agents/status/{id}` were writable by **any anonymous network client** — the auth middleware allowlists the whole `/health` prefix (read-only no-auth lane), but the platform-health router mounted write routes under it. This moves those writes to the authenticated `/api/v1` lane and gives agents a service identity so their heartbeats keep flowing.

Confirmed the hole live before the fix: an anonymous `POST /health/agents/status` on the old stack returned **200** and inserted a fake `pwned-agent` row.

## What changed

**Server (the fix):**
- Agent-status writes now live at `POST/PUT /api/v1/agents/status[...]`, gated by `require_scopes(agents:write)`. `/health/*` keeps only GET probes.
- **Defense in depth:** the middleware allowlist is now method-scoped (GET/HEAD only), so any future write route mounted under `/health` fails closed instead of riding the no-auth lane. A router-level test pins `/health` to safe methods.

**Identity (so agents can still write):**
- New realm role `agent` implying **only** `agents:write` (via the #270 role→scope bridge), plus a confidential `squadops-agent` client (service accounts on, audience mapper to `squadops-runtime`) across all four realm exports. Its service account holds exactly the `agent` role — least privilege, no cycle/task/admin access.
- New first-class `auth.agent_client` config, env-addressable as `SQUADOPS__AUTH__AGENT_CLIENT__*` (a `service_clients` dict entry can't be addressed by the schema-authoritative env loader — hence a dedicated field, with a test pinning that).
- `HealthCheckHttpReporter` posts to the new path and attaches a Bearer token from the existing `ServiceTokenClient` when an identity is configured. Token-acquisition failures fail silently like any heartbeat (next tick retries); a half-configured identity (`agent_client` without `oidc`) raises at startup rather than degrading to anonymous heartbeats that would 401 forever.
- docker-compose: each agent gets the client secret via a new `agent_client_secret` docker secret; bootstrap creates the dev secret file.

## Authorization matrix (enforced by tests against the real authz adapter)

| Identity | `agents:write`? |
|---|---|
| anonymous | 401 |
| viewer / operator | 403 |
| **agent** | ✅ 200 |
| admin | ✅ 200 |

## Tests

- `test_agent_status_routes.py` — handler behavior + the guard matrix above (real `KeycloakAuthzAdapter` + role→scope bridge, so a `ROLE_SCOPES` regression fails in CI)
- `test_platform_health_routes.py` — old write paths gone (404/405); structural ratchet that every `/health` route is GET-only
- `test_middleware.py` — a deliberately-mounted write route under `/health` requires a token (401 anon, 200 with token)
- `test_realm_parity.py` — every code `Role` is backed by a realm role in all four exports; `squadops-agent` client shape + audience + least-privilege service account
- `test_healthcheck_http.py` — posts to `/api/v1/agents/status`, attaches bearer, swallows token/HTTP failures
- `test_entrypoint_heartbeat_identity.py` — runner wiring: identity → token provider, no identity → anonymous, half-config → raises, `stop()` closes the token client
- `test_agent_client_env.py` — the `SQUADOPS__AUTH__AGENT_CLIENT__*` env paths resolve

Full regression: **4726 passed, 1 skipped**, exit 0. Ruff clean.

## Live validation (rebuilt stack at 1.3.0)

- Anonymous `POST /api/v1/agents/status` → **401** (was 200); old `/health` write paths → **401**
- `POST /api/v1/agents/status` with a real `squadops-agent` token → **200**
- `GET /health`, `GET /health/agents` → **200** (probes still open)
- **All 7 deployed agents** log "service identity configured", zero heartbeat failures, fresh heartbeats landing in the DB at v1.3.0 (request logs show every internal agent IP getting 200)
- Smoke cycle `cyc_473fbf479a4a` **completed** (~84s, 6 artifacts) — squad runs end-to-end after the auth change

Targeted for the **1.3.1** hardening patch.

### DIFF
--- a/adapters/observability/healthcheck_http.py
+++ b/adapters/observability/healthcheck_http.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 
 import logging
 import os
+from collections.abc import Awaitable, Callable
 from typing import Any
 
 import httpx
@@ -20,7 +21,13 @@ logger = logging.getLogger(__name__)
 
 
 class HealthCheckHttpReporter(AgentHeartbeatReporter):
-    """Posts agent status updates to the runtime-api health endpoints."""
+    """Posts agent status updates to the runtime-api agent-status endpoint.
+
+    Writes go to the authed /api/v1 lane (#326). When ``token_provider`` is
+    set (agent service identity via client credentials), each request carries
+    a Bearer token; without it the request is unauthenticated — valid only
+    for auth-disabled deployments.
+    """
 
     def __init__(
         self,
@@ -28,12 +35,14 @@ class HealthCheckHttpReporter(AgentHeartbeatReporter):
         base_url: str | None = None,
         timeout_seconds: int = 5,
         fail_silently: bool = True,
+        token_provider: Callable[[], Awaitable[str]] | None = None,
     ) -> None:
         self._base_url = (
             base_url or os.getenv("SQUADOPS_RUNTIME_API_URL") or "http://runtime-api:8001"
         ).rstrip("/")
         self._timeout_seconds = timeout_seconds
         self._fail_silently = fail_silently
+        self._token_provider = token_provider
 
     async def send_status(
         self,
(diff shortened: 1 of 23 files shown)


jasonLadd (commented):
reviewed

Issue #306, opened 2026-07-01
Agent image has no Node.js → frontend build check (#290) + vitest are inert (always skip)


jasonLadd:
Found by the post-#304 re-validation (`cyc_2ecf5f0e5bd1`, group_run/lite fullstack). After #303 fixed the harness discovery/blocking/PYTHONPATH logic, the **backend** acceptance now works end-to-end (real test failures surface). But the **frontend build check + vitest still don't run**:

```
frontend (non-blocking): npm not found — Node.js is not installed
```

Confirmed: `docker exec squadops-eve` / `squadops-bob` → **no `node`, no `npm`.** `agents/Dockerfile` (base `python:3.11-slim`) installs only `curl`/`gcc`/`python3-dev` (line 51).

## Impact
`run_frontend_build` / `run_node_tests` shell out to `npm`/`npx` in the qa agent's own container. With no Node there, they hit `FileNotFoundError` → skip (correctly non-blocking — an env limitation isn't a deliverable failure). Net: the frontend build check is **inert in every deployment** — a non-building frontend still ships green. The #290/#296/#303 code is correct and unit-verified; it just has no runtime to execute against.

## Options
1. **Add `nodejs`+`npm` to the shared `agents/Dockerfile`** (Debian's, or nodesource for node 20). Simplest; uniform. Cost: ~150 MB × all 6 agent images, most of which never build frontends.
2. **A qa-agent image variant** with Node (only the role that runs `qa.test`). Targeted; adds per-role Dockerfile divergence to the build.
3. **Route frontend build/test through a sandboxed executor** (ACI) that has Node, instead of the agent container. Architectural — biggest, but keeps agent images lean and isolates untrusted build execution.

## Recommendation
Option 1 near-term (unblocks the check now; image bloat is acceptable for a dev framework), with option 3 as the principled longer-term home. Node ≥18 required for current Vite.

Spark-clean (`agents/Dockerfile`). Refs #290, #296, #303. Surfaced in cyc_2ecf5f0e5bd1.

jasonLadd:
**Lane reassignment: Mac lane is taking this (Jason approved 2026-07-08), targeted for the 1.3.1 hardening patch.**

[...]

PR #360, opened 2026-07-08
docs(sip): propose Externalized Build Sandbox (proposed draft)

jasonLadd:
## Summary

Files a **proposed SIP draft** (unnumbered) — *Externalized Build Sandbox*. It does **not** change code; it captures the design for the principled long-term home of build/test execution.

**This PR only files the draft — it stays `proposed`. Not a request to accept.**

## Why now

Direct follow-on from #306. That fix ships Node.js in the QA agent image so the frontend build check (#290/#296/#303) can run — because `run_frontend_build` / `run_node_tests` shell out to `npm`/`npx` **in the agent's own process**. That in-process model is the root of two structural problems:

1. **Every buildable stack forces a toolchain into an agent image** — the same long-lived container that holds the agent's DB/queue/LLM credentials. #306 weighed "bloat all six images" vs. "a QA-only branch" and picked the least-bad *bundling* option; bundling itself is the ceiling.
2. **Untrusted generated code runs in the agent runtime** — `npm install` on a generated `package.json` runs arbitrary install scripts as the agent process, with its secrets in reach.

## What the SIP proposes

A `BuildSandboxPort` on the ports bundle: submit a `BuildJob` (materialized files + a pinned **toolchain profile** + a command), get a structured `BuildResult`, executed in an ephemeral, isolated sandbox (no agent secrets, ephemeral FS, resource caps, non-root, network off-by-default). Built on the **existing** `ContainerPort`.

- **Local adapter** on Docker/Podman (zero new infra for the laptop stack); **remote adapter** for edge/cloud (the SIP-Edge-Deployment-Profile target — no Docker socket on the agent host); **NoOp** default preserving today's skip-on-missing-toolchain behavior exactly.
- Agent images return to **runtime-only** — migration step 5 retires the #306 QA-Node branch, which is also the regression guard: once removed, a build that needs a toolchain *must* go through the sandbox.
- Behavior parity: `ran=False` on a missing sandbox = today's non-blocking skip; the caller still owns the blocking decision.

`#306` is explicitly the correct **interim**; this SIP is its **exit**, not a replacement.

## Open questions (design agenda, §7)

File transport at scale (inline bytes vs. content-addressed handoff), produced-artifact flow into the run's artifact set, per-job ephemerality vs. a warm pool, relationship to `CapabilityExecutor` (distinct port vs. specialization), and toolchain-image ownership/pipeline.

## Not in scope

No release commitment (vision item, candidate for the 2.0 capability-backed-agents arc). Dependency-supply-chain trust (malicious packages) beyond isolation is a separate concern.

### DIFF
--- a/sips/proposed/SIP-Externalized-Build-Sandbox.md
+++ b/sips/proposed/SIP-Externalized-Build-Sandbox.md
@@ -1,0 +1,155 @@
+---
+title: Externalized Build Sandbox
+status: proposed
+author: jladd
+created_at: '2026-07-08T00:00:00Z'
+---
+# SIP: Externalized Build Sandbox
+
+## Status
+Proposed
+
+**Targets:** vision item — no committed release. Candidate for the 2.0 capability-backed-agents arc (see [[capability-backed-agents-2-0]]), or an earlier stabilization minor if the toolchain-bundling cost grows. This is the principled long-term home for build/test execution; it is **not** a near-term blocker.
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #362, opened 2026-07-08
fix(#328): broker-hygiene doctor check + sweep orphaned pre-SIP-0094 queues

jasonLadd:
## Summary

Closes #328.

`rabbitmqctl list_queues` on the deployed broker showed ~19 legacy per-run reply queues still present — SIP-0094 (v1.0.6) replaced the per-run `cycle_results_{run_id}` scheme with durable per-agent `{agent_id}_replies` queues precisely because the old scheme leaked one orphan per run, but the existing corpses were never swept. Two held undrained backlogs (48 and 2 messages) with **0 consumers**.

This does the three things the issue asks:

### 1. Audit — confirmed dead residue, not a live bug
The only `cycle_results` references anywhere in the codebase are **comments** in `task_dispatcher.py` and `reply_router.py` describing what SIP-0094 *replaced*. Nothing publishes to the old scheme. So the queues are orphaned history, safe to sweep.

### 2. Durable doctor check (the code in this PR)
New `broker` doctor category flagging two classes of cruft:
- **Retired naming scheme** — queues matching the pre-SIP-0094 `cycle_results_*` prefix.
- **Undrained backlog** — any queue with `messages > 0` and `consumers == 0` (results produced but nothing draining — the same signal the #323 consumer churn would have surfaced).

Both are hard failures with actionable fix guidance. An **unqueryable broker warns heuristically** (`~`) rather than failing, so doctor never goes red just because the stack is down. The check runs only when the profile declares the `rabbitmq` service (mirrors GPU gating on an nvidia dep).

Queues are read via `rabbitmqctl list_queues` **inside** the broker container, with the container resolved from the compose *service* name (`docker compose ps -q rabbitmq`) rather than a hardcoded container name — no credentials, the same host-side access path `rebuild_and_deploy.sh` already uses.

### 3. One-time sweep (operational, done)
The 16 orphaned `cycle_results_*` queues were deleted after confirming (audit above) the backlog is dead history.

## Tests

`TestBrokerHygiene` (7 tests) patches the docker/rabbitmqctl boundary and asserts: retired scheme → hard fail with count; undrained backlog → hard fail (including a *current*-scheme queue with a stuck backlog, not just cruft); a consumed queue with in-flight messages is **not** flagged; both classes reported together; clean broker passes; unqueryable → heuristic warn (never a hard fail); and the check gates on the `rabbitmq` service being in the profile.

Full regression: **4726 passed, 1 skipped**, exit 0. Ruff clean.

## Live validation (deployed stack)

| | retired queues | undrained backlogs | doctor exit |
|---|---|---|---|
| **before sweep** | 16 flagged | 2 flagged (48 msg, 2 msg) | **1** |
| **after sweep** | 0 | 0 | **0** (`14 broker queue(s) healthy`) |

The SIP-0094 `{agent_id}_replies` queues (max/neo/nat/bob/eve/data/comms-agent) are all intact after the sweep — only the retired scheme was removed.

Targeted for the 1.3.1 hardening patch.

### DIFF
--- a/src/squadops/bootstrap/setup/checks.py
+++ b/src/squadops/bootstrap/setup/checks.py
@@ -7,6 +7,7 @@ and returns a CheckResult with pass/fail, fix guidance, and heuristic flag.
 from __future__ import annotations
 
 import asyncio
+import json
 import platform
 import shutil
 import socket
@@ -28,7 +29,9 @@ from squadops.bootstrap.setup.profile import (
 # Result model
 # ---------------------------------------------------------------------------
 
-VALID_CATEGORIES = frozenset({"python", "platform", "tools", "docker", "models", "gpu", "auth"})
+VALID_CATEGORIES = frozenset(
+    {"python", "platform", "tools", "docker", "models", "squad", "gpu", "auth", "broker"}
+)
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #367, opened 2026-07-09
feat: capture request_profile at the cycle level (SIP-0096 provenance)

jasonLadd:
First landed increment of the SIP-0096 run-config provenance work (1.4 evidence arc). Closes #365.

## The reframe (why this is small)
The original plan proposed a run-level `RunProvenance` record. A cycle-vs-run analysis showed it duplicated config the model already carries at the correct levels:
- squad profile + roster + per-role models → pinned by the cycle's immutable `squad_profile_snapshot_ref` (cycle-level)
- `experiment_context`, `applied_defaults` → cycle-level
- per-attempt resolution → already on `Run` (`resolved_config_hash` / `resolved_config_ref`)

The **one** config dimension not recorded today is the **cycle-request-profile name** — a cycle-level attribute (all runs of a cycle share it), needed so the 1.8 cycle-evaluation scorecard can attribute an outcome to its config. `framework_version` (the only genuinely run-level provenance fact) is deferred to the SIP-0096 `CycleOutcome` roll-up as an execution stamp (Phase 3).

## What
Threads `request_profile` CLI → DTO → route → `Cycle` → `cycle_registry`:
- `Cycle.request_profile` field (cycle-level, immutable per cycle)
- migration `007` (idempotent `ADD COLUMN cycle_registry.request_profile TEXT`)
- `CycleCreateRequest` DTO field; route maps `body.request_profile` → `Cycle`
- CLI now sends the `--request-profile` name (was resolved locally then dropped)
- postgres `create_cycle` INSERT + `_row_to_cycle` read; memory registry rides `Cycle(**d)` unchanged
- DDL-model-drift guard satisfied (field has its backing column)

## Validation
- **Unit:** full regression **4731 passed**; 2 new round-trip tests (INSERT includes `request_profile`; `get_cycle` reads it back).
- **Live** (rebuilt runtime-api, migration 007 applied): created `cyc_8338409c2daa` with `--request-profile selftest` on the deployed stack → `cycle_registry.request_profile = 'selftest'` confirmed via direct DB query; cycle cancelled.

## Follow-ups (not in scope)
- `framework_version` execution stamp on the `CycleOutcome` roll-up — SIP-0096 Phase 3.
- Optional: surface `request_profile` in the get-cycle response DTO / `cycles show` (the registry port already returns it, so the scorecard can read it programmatically today).

### DIFF
--- a/adapters/cycles/postgres_cycle_registry.py
+++ b/adapters/cycles/postgres_cycle_registry.py
@@ -57,8 +57,8 @@ class PostgresCycleRegistry(CycleRegistryPort):
                     "(cycle_id, project_id, created_at, created_by, prd_ref, "
                     "squad_profile_id, squad_profile_snapshot_ref, task_flow_policy, "
                     "build_strategy, applied_defaults, execution_overrides, "
-                    "expected_artifact_types, experiment_context, notes) "
-                    "VALUES ($1,$2,$3,$4,$5,$6,$7,$8,$9,$10,$11,$12,$13,$14)",
+                    "expected_artifact_types, experiment_context, notes, request_profile) "
+                    "VALUES ($1,$2,$3,$4,$5,$6,$7,$8,$9,$10,$11,$12,$13,$14,$15)",
                     cycle.cycle_id,
                     cycle.project_id,
                     cycle.created_at,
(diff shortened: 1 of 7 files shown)


jasonLadd (commented):
reviewed

Issue #381, opened 2026-07-10
TaskResult.status is an untyped UPPERCASE bare-string vocabulary shadowing the lowercase TaskStatus enum — a loaded footgun (twin of #377)


jasonLadd:
## The footgun

`TaskResult.status` (`src/squadops/tasks/models.py:98`) is a bare `str`:
```python
status: str  # SUCCEEDED, FAILED, or CANCELED
```
Its values are **UPPERCASE** everywhere. But `TaskStatus` (`src/squadops/capabilities/models.py:52`) is a `StrEnum` with **lowercase** values (`SUCCEEDED = "succeeded"`). Two representations of one concept, and they differ **only in case** — so they *look* interchangeable and are not.

Surfaced by the #380 enum-shadow guardrail (it flagged `result.status == "SUCCEEDED"` as shadowing `TaskStatus`). This is the same disease as #377 (`terminal_status` ↔ `RunStatus`), and **worse**: `terminal_status` at least maps by `.upper()`; here `"SUCCEEDED" != "succeeded"` outright.

## Is it firing today? No — but it's loaded (verified)

- **11 sites** compare `result.status` against uppercase literals, all correct:
  `in_process_flow_executor.py:164,216` · `pulse_boundary_runner.py:620` · `dispatched_flow_executor.py:1534,1550` · `correction_runner.py:192` · `task_dispatcher.py:234,358,378` · `orchestration/orchestrator.py:268` · `capabilities/runner.py:311`
- Producers set it uppercase: `handler_executor.py:104,133,161,177,193` · `task_dispatcher.py:308,318` · `entrypoint.py:776`.
- The `TaskStatus` **enum** is only ever used on a *different* object — `tr.status` (task **records**), enum-to-enum (`runner.py:603,609`).
- **There is no `result.status == TaskStatus.X` anywhere** — the one comparison that would be silently `False` forever. So the two vocabularies never cross today.

## Why it still needs fixing

The instant someone writes the natural, typed-looking `result.status == TaskStatus.SUCCEEDED`, it is silently `False` — a succeeded task treated as not-succeeded, no error, no failing test. 11 uppercase literals are held together only by everyone remembering never to reach for the obviously-named enum. The field even carries `"CANCELED"` (one L) while `RunStatus.CANCELLED` is two — an unmanaged vocabulary with its own spelling landmines.

[...]

PR #378, opened 2026-07-10
feat(SIP-0096 Phase 2 slice 1): normalize task-result verification into the ledger (+ #376 honest-red guards)

jasonLadd:
## SIP-0096 Phase 2 slice 1 — normalize task-result verification into the ledger

Makes the Phase-1 aggregation choke point **non-inert**: the real verification producers now record their results into the `RunLedger`, so `RunCompletion.finalize` aggregates genuine evidence. **Disclosure-only** — no profile ships `required_checks` yet, so **no verdicts flip**; the run_report integrity roll-up starts showing real `Executed: N` evidence, and — critically — a failed/unverified run now reads as honest red instead of a false green.

Part of #375 (Phase 2). Addresses #376 points 1 & 4 (field evidence from the Spark's 1.3.1 regression). Phase 1 = #368 (merged `d62ce0d`).

### What's in this slice
- **Pure producer adapter** `verification_normalize.py` — `normalize_task_checks(outputs) -> list[CheckResult]`. One normalizer folds both producer shapes in `outputs["validation_result"]["checks"]`: SIP-0092 typed-acceptance rows (status passthrough) and the framework test spine. `tests_pass` is **synthesized from the always-present `test_result`** (the `checks[]` row is failure-only — relying on it would make a passing run look like "no result" and falsely block once required). The §6.6.1 stub signal marks a stub-backed green so it can't credit. Defensive — never raises on malformed rows.
- **Executor recording** at the `_execute_sequential` dispatch seam, for **every** completed task on **all three paths**:
  - success + `break_correction` (patch-advance) → the returned result (`61ba91b`);
  - **abort** (correction raises from inside `dispatch_with_retry`) → the final failed result, captured via the outcome callback and recorded before re-raise (`43b3736`).
  - This is #376 pt1: the failing `qa.test`/build evidence (the #276 class) survives correction and abort — a failed run aggregates to `rejected`, not silent green.
- **Narrative-override guard** (`5022b99`, #376 pt4 / §6.6.4): a COMPLETED run whose verdict is `rejected`/`blocked_unverified` no longer prints "All tasks completed successfully"; the Quality Notes reflect the verdict.
[...]

Issue #377, opened 2026-07-10
terminal_status is Prefect's State vocabulary leaked into domain objects — translate at the WorkflowTracker adapter boundary


jasonLadd:
## The true problem (sharpened)

Originally filed as "an uppercase enum shadow of `RunStatus`." Tracing the full chain shows it is worse than a case-convention shadow: **`terminal_status` is Prefect's `State` type vocabulary that has leaked inward into domain objects.** The right fix is not to derive the uppercase string at the presentation edge — it is to translate `RunStatus` → Prefect state **at the Prefect adapter boundary**, and carry the domain enum everywhere else.

### Three vocabularies are tangled into one string

| Vocabulary | Values | Owner | Where it belongs |
|---|---|---|---|
| `RunStatus` | `completed/failed/cancelled/paused` (lowercase) | **our domain** (`src/squadops/cycles/models.py`) | everywhere in domain + adapters |
| Prefect `State` type | `COMPLETED/RUNNING/FAILED/CANCELLED/PAUSED` (uppercase) | **Prefect** | behind the Prefect adapter only |
| `terminal_status` | `"COMPLETED"`… (uppercase `str`) | *nobody* — conflates the two | should not exist |

### Evidence it is Prefect's vocabulary, not ours
- `WorkflowTrackerPort.set_flow_run_state(flow_run_id, state_type: str, state_name)` — the Prefect adapter (`prefect_workflow_tracker.py:151`) passes `state_type` **straight into the Prefect API** as `{"type": state_type}`; the docstring enumerates *"RUNNING, COMPLETED, FAILED, CANCELLED"* — Prefect's `State` types verbatim.
- `dispatched_flow_executor.py:1510`: `set_flow_run_state(flow_run_id, "RUNNING", "Running")` — a hardcoded Prefect state type that is not even a *terminal* status.

### The coincidence is accidental and partial (latent bug)
`terminal_status == RunStatus.value.upper()`, which *happens* to equal the Prefect state type — **only for the terminal subset**. `RunStatus.QUEUED.value.upper()` = `"QUEUED"`, which is **not** a Prefect state type (Prefect has `SCHEDULED`/`PENDING`). So any future path that routes a non-matching `RunStatus` through this shadow silently emits an invalid Prefect state. Not firing today (QUEUED never reaches `set_flow_run_state` as terminal), hence low urgency — but it is a real footgun, not just stylistic.

### The domain leak (the actual violation)
The uppercase string reached two places it must not:
[...]

PR #382, opened 2026-07-10
test(#380): enum-shadow architecture guardrail — fail CI on status string-literal comparisons

jasonLadd:
## Why

`terminal_status == "COMPLETED"` — an untyped uppercase shadow of the lowercase `RunStatus` enum — slipped in and **propagated across three SIPs** because it's *latent* drift: the two vocabularies never collide at runtime, so tests stay green and only human review can catch it (and "looks like existing code → approve" is the propagation mechanism). Discipline doesn't stop this; an executable guardrail does. Ruff `PLR2004` (magic-value-comparison) is **numeric-only** and doesn't flag strings, so this is a custom AST rule — same vehicle as `test_forbidden_imports.py`, runs in the regression suite.

Closes #380.

## The guardrail
`tests/unit/architecture/test_no_enum_shadow_comparisons.py` auto-discovers every `StrEnum` vocabulary in the domain (67 values today; future enums covered for free) and **fails CI** when domain code compares a variable against a raw string literal that matches an enum value (case-insensitive, to catch the upper/lower shadow that started this).

Off main it found **22**. This PR resolves each as either a mechanical fix or a justified allowlist entry — no silent sweeps.

## Fixed (mechanical — value+case exact, right enum)
- `run_report_builder`: `terminal_status` → `RunStatus.*.value.upper()` (case-robust); pulse decisions → `PulseDecision.*.value`
- `artifacts`: `build_strategy` → `BuildStrategy.FRESH.value`
- `runs`: gate decision → `GateDecisionValue.APPROVED.value`

## Allowlisted (11, each justified in-file)
A shared status **word** (`"failed"`, `"active"`) is defined by several enums *and* by non-enum vocabularies, so value-matching flags coincidences — triaged, not swept:
- **`result.status == "SUCCEEDED"`** (`runner`, `orchestrator`) — `TaskResult.status` is an UPPERCASE bare-str vocabulary, **not** the lowercase `TaskStatus` enum; the naive enum fix would *break* it (case). Reconciliation tracked in **#381** (a loaded footgun, twin of #377).
- `CheckOutcome.status` (`develop`) — a dataclass, not a StrEnum (SIP-0092).
- generated-test result rows (`test_execution`), Ollama pull status (`models`), `window_state()` tokens (`recruitment`/`scheduler`), `RuntimeMode` Literal (`coordinator`), DTO workload-progress tokens (`lifecycle`) — each a distinct vocabulary.

## Honest limits (documented, not hidden)
- The rule flags string-literal comparisons; it would **not** catch the *firing* form of #381 (`== TaskStatus.X`) — it finds the root (untyped vocabulary) and forces reconciliation, it's the smoke detector not the extinguisher.
- ~311 other string-literal comparisons with **no** backing enum are out of scope (a blanket rule = 324 sites = noise); a softer "same literal in ≥3 places = missing enum" signal is a possible follow-up.

## Overlap note
`run_report_builder._build_report_quality_lines` is also touched by the SIP-0096 slice-1 PR #378 (which adds the narrative-override *feature* on top of the same enum-referenced compares). On merge, take #378's superset of that function; the guardrail passes against either.

Relates to #377, #381. Regression **4774**.

### DIFF
--- a/src/squadops/api/routes/cycles/artifacts.py
+++ b/src/squadops/api/routes/cycles/artifacts.py
@@ -16,6 +16,7 @@ from squadops.auth.models import Scope
 from squadops.cycles.models import (
     ArtifactRef,
     BaselineNotAllowedError,
+    BuildStrategy,
     CycleError,
     PromotionStatus,
     ValidationError,
@@ -264,7 +265,7 @@ async def promote_baseline(project_id: str, artifact_type: str, body: BaselinePr
         if ref.cycle_id:
             registry = get_cycle_registry()
             cycle = await registry.get_cycle(ref.cycle_id)
-            if cycle.build_strategy == "fresh":
+            if cycle.build_strategy == BuildStrategy.FRESH.value:
                 raise BaselineNotAllowedError(
                     "Cannot promote baseline for a fresh build strategy cycle"
                 )
--- a/src/squadops/api/routes/cycles/runs.py
+++ b/src/squadops/api/routes/cycles/runs.py
@@ -22,6 +22,7 @@ from squadops.cycles.models import (
     CycleError,
     CycleStatus,
     GateDecision,
+    GateDecisionValue,
     Run,
     RunStatus,
     RunTerminalError,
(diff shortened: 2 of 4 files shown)


jasonLadd (commented):
reviewed

PR #385, opened 2026-07-11
fix(#374): re-run the failed check on a patch (pure-behavioral repair verification)

jasonLadd:
## Summary

Implements the enforcement half of #374 — the SIP-0086 convergence loop no longer advances blindly on a `patch`.

**Pure-behavioral repair verification:** a `patch` correction now re-runs the **original failed check** against the patched artifacts (it returns `"continue"`, so the executor re-dispatches the task via `dispatch_with_retry`'s existing retry loop). The re-run's pass/fail **is** the verdict — not the LLM `validate_repair` judgment.

- re-run **passes** → advances, with real executed-passed evidence through the SIP-0096 recording seam
- re-run **keeps failing** → re-enters correction until `max_correction_attempts` exhausts → `_ExecutionError` → **the run FAILS** honestly instead of false-completing (the #276-class false-green, eliminated)

**Threading gotcha handled:** the run-level correction count is now a shared mutable holder (`correction_counter`) incremented inside `_handle_task_outcome`. Without that, the inner `"continue"` loop would never hit the `max_correction_attempts` bound *and* would collide the deterministic `corr-`/`plan_delta-` ids across re-runs.

## ⚠️ Draft — must land **with #379**

This is one half of a coupled pair. Without #379, a *recovered* run (repair converges) completes but its SIP-0096 roll-up shows a **stale `rejected`** — the append-only aggregation doesn't supersede the earlier FAILED `CheckResult` with the re-run's PASSED one. #379 (Macbook lane) resolves `(check_id, subject_ref)` to final-state and populates `subject_ref` (today `None`). **Hold merge until #379 is ready and land them together.**

## Tests

- Updated `test_patch_dispatches_repair_tasks` — the re-run adds one dispatch (9→10) and the run still completes.
- New `test_patch_rerun_never_converges_fails_run` — a repair that never fixes the check **fails the run** (never COMPLETED). Catches the pre-#374 false-green directly.
- Full cycles suite: **1486 passed**, ruff clean.

## Joint acceptance (Spark full-squad, after #379)

- repair re-runs → green → verdict `accepted`
- repair never converges → run **FAILED**

Closes #374

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -833,7 +833,11 @@ class DispatchedFlowExecutor(FlowExecutionPort):
 
         # SIP-0079 Phase 3: Outcome routing state
         consecutive_failures: int = 0
-        correction_attempts: int = 0
+        # #374: run-level correction count as a mutable holder. A `patch` re-runs the
+        # failed check via dispatch_with_retry's "continue", so the count must bump on
+        # each inner-loop correction (not just once per outer iteration) to bound it
+        # (max_correction_attempts) and keep corr-/plan_delta- ids unique across re-runs.
+        correction_counter: dict[str, int] = {"n": 0}
         task_attempt_counts: dict[str, int] = {}
 
         # SIP-0079: Time budget enforcement (RC-8)
@@ -938,7 +942,6 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                 result,
                 _envelope=envelope,
                 _consecutive_failures=consecutive_failures,
-                _correction_attempts=correction_attempts,
                 _holder=_last_failed_result,
             ):
                 _holder["result"] = result
@@ -949,7 +952,7 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                     run_id=run_id,
                     task_attempt_counts=task_attempt_counts,
                     consecutive_failures=_consecutive_failures,
-                    correction_attempts=_correction_attempts,
+                    correction_counter=correction_counter,
                     prior_outputs=prior_outputs,
                     all_artifact_refs=all_artifact_refs,
                     stored_artifacts=stored_artifacts,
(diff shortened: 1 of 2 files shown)


jasonLadd:
Mac half is up as **#386** (draft, `Closes #379`) — the two must land together.

One finding from tracing the recording seam that refines this PR's premise: **with #374 as written, a converged repair currently records only the *final* PASSED result.** `_record_task_evidence` fires only at the final-settle (line 995) and abort (989) seams; the intermediate failure is captured in `_last_failed_result` but only *recorded* on abort. So the stated premise here — "the append-only aggregation doesn't supersede the earlier FAILED CheckResult" — assumed a FAILED row that wasn't actually landing (a pure aggregation tweak would've been a no-op).

#386 closes that gap: it records the intermediate failed attempt on the `"continue"` re-dispatch path (stamped with the producing task id) **and** supersedes `(check_id, subject)` to final-state. Net effect is exactly what you designed — without #379 → `rejected`, with it → `accepted` — plus the roll-up now honestly discloses the recovered failure instead of silently showing green.

**Merge-order heads-up:** both PRs edit `_route_outcome` in `dispatched_flow_executor.py`. This PR swaps `correction_attempts` → `correction_counter`; #386 adds an `if action == "continue": _record_task_evidence(result)` tail after the handler call. Whoever merges second resolves a small conflict there — union both. Happy to rebase #386 onto this once it's out of draft, or take the conflict on my side if you land first.

Joint full-squad acceptance after both land: repair re-runs → green → `accepted` (roll-up discloses the recovery); repair never converges → run FAILED.

jasonLadd:
### Rebased onto main (now has #379 / #386) — conflict resolved, combined tree green

#386 is merged, so I rebased this onto the new main. The `dispatched_flow_executor.py` overlap resolved as reviewed (not a naive union):

- `_route_outcome`: **kept #385's** `correction_counter=` call arg + dropped `_correction_attempts=` default, **added #386's** `if action == "continue": _record_task_evidence(result)` tail.
- Settle-seam: **took #385's delete** of `correction_attempts += 1` (the counter now increments inside `_handle_task_outcome` via the shared `correction_counter` holder). Verified no lingering `correction_attempts` local remains.

Combined tree (both halves present):
- `test_correction_runner` + `test_verification_integrity` + `test_verification_normalize` + `test_executor_checkpoint`: **109 passed**
- Full cycles suite: **1496 passed**; enum-shadow guardrail (#382): **pass**; ruff clean.

Remaining before un-draft: joint Spark full-squad validation (repair converges → `accepted` + roll-up discloses the recovered failure; never converges → run FAILED).

jasonLadd:
Heads-up: **#386 (the #379 half) merged to main first** (`96225e6`). It's inert standalone — without this PR's patch→`continue`, the only `continue` is transport retries, which normalize to nothing — so main stays green, but the pair isn't *load-bearing* until #385 lands.

Per my earlier offer (whoever lands second takes the conflict): when you rebase this onto main you'll hit one conflict in `_route_outcome` (dispatched_flow_executor.py) — main now has the `if action == "continue": _record_task_evidence(result)` tail after the handler call; this PR renames `correction_attempts`→`correction_counter` in the same block. **Union both** (keep the counter rename *and* the recording tail). Happy to push the rebase for you if that's easier — say the word.

Once rebased + merged, the joint full-squad positive-path validation is the last gate: repair re-runs → green → honest `accepted` (roll-up discloses the recovered failure); repair never converges → run FAILED.

jasonLadd (commented):
reviewed

PR #386, opened 2026-07-11
feat(SIP-0096 Phase 2 slice 2): final-state resolution for re-verified checks (#379)

jasonLadd:
## Summary

The Mac half of the #374/#379 coupled pair — **final-state resolution** so a repaired-and-re-run check resolves to its final outcome, not the union of every attempt.

The Phase-1 aggregation core unioned every recorded result, so a check that FAILED, was repaired, and re-ran PASSED landed in *both* `failed` and `verified` → the verdict was pinned `rejected` permanently. That contradicts SIP-0096 **§6.5** (*"the verdict reflects the final state after correction's bounded attempts"*) and blocks #374's re-run-to-green: a genuinely-fixed check stays **stuck-red** — the exact mirror of the false-green this SIP kills.

## The fix is an identity qualifier, not a naive dedup

Framework-spine `tests_pass` is legitimately recorded **once per develop task** (3 tasks → 3 rows, "any fail → rejected" is correct). Bare last-write-wins by `check_id` would collapse distinct producers and silently drop a real mid-run failure. So:

- **`CheckResult.subject`** — the producing plan-task id (§6.3). Deliberately **distinct from `provenance.subject_ref`** (the *thing under test* — a file-set hash/artifact/endpoint, which legitimately **differs** between a failed check and its re-run against *patched* artifacts, making it the wrong grouping key).
- **`aggregate_verification`** resolves `(check_id, subject)` to its **last-appended** result before applying the invariant. The ledger is append-only, so append order *is* attempt order — resolution keys off position, keeping the module clock-free (no timestamp sort). `subject=None` rows are **never** collapsed (each is independent evidence — the pre-#379 accumulate default; distinct subjects still "any fail → rejected").
- **`normalize_task_checks`** stamps `subject=envelope.task_id` on every result.
- The **executor records the intermediate FAILED attempt** on the re-dispatch (`"continue"`) path, so the ledger honestly holds the failed→passed history; aggregation then supersedes it. Self-filtering: a transport retry carries no verification outputs → nothing recorded.

## ⚠️ Draft — must land **with #374 (PR #385)**

One half of a coupled pair. A finding worth flagging: I traced the recording seam and **with #374 as written, a converged repair currently records only the *final* PASSED result** (`_record_task_evidence` fires at the final-settle + abort seams only — the intermediate failure is captured in `_last_failed_result` but only *recorded* on abort). So #385's premise — "the append-only aggregation doesn't supersede the earlier FAILED CheckResult" — assumed a FAILED row that wasn't actually landing.

This PR makes it land (the executor change above) **and** supersedes it, so #379 is genuinely load-bearing exactly as #385 designed (without it → `rejected`; with it → `accepted`) **and** the roll-up honestly discloses the recovery rather than silently showing green.

**Merge-order note:** both this PR and #385 edit `_route_outcome` in `dispatched_flow_executor.py` (#385 swaps `correction_attempts` → `correction_counter`; this adds the `if action == "continue": _record_task_evidence(result)` tail). Whoever merges second resolves a small conflict there — union both.

## Tests

- **6 pure-core supersession cases**: fail→pass → `accepted`; pass→fail stays `rejected` (no fail-masking); distinct subjects accumulate; `subject=None` never collapses; required check recovers to `accepted`; multi-attempt-then-pass all supersede.
- **3 normalizer cases**: subject stamped on every result; defaults to `None`; end-to-end same-task supersession.
- **1 executor end-to-end**: a task whose check fails then passes on re-dispatch records **both** attempts (subject-stamped) and resolves to `accepted` — proving the intermediate failure is recorded (not dropped) *and* superseded.
- Full regression **4823 passed, 1 skipped**, ruff + enum-shadow guardrail clean.

## Joint acceptance (Spark full-squad, after both land)

- repair re-runs → green → verdict `accepted`, roll-up discloses the recovered failure
- repair never converges → run **FAILED** (#385's guarantee)

Closes #379

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -942,7 +942,7 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                 _holder=_last_failed_result,
             ):
                 _holder["result"] = result
-                return await self._handle_task_outcome(
+                action = await self._handle_task_outcome(
                     result=result,
                     envelope=_envelope,
                     cycle=cycle,
@@ -958,16 +958,30 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                     profile=profile,
                     flow_run_id=flow_run_id,
                 )
+                if action == "continue":
+                    # #379: this attempt failed and is about to be re-dispatched — a
+                    # correction patch re-verifying behaviorally (#374), or a plain
+                    # retry. Record its evidence now so the ledger honestly holds the
+                    # failed→passed history; aggregation supersedes it to the final
+                    # state per (check_id, subject). Self-filtering: a transport retry
+                    # carries no validation_result/test_result → nothing is recorded.
+                    _record_task_evidence(result)
+                return action
 
-            def _record_task_evidence(task_result) -> None:
+            def _record_task_evidence(task_result, _envelope=envelope) -> None:
                 # Normalize this task's verification outputs into the run ledger for
-                # the end-of-run aggregation choke point. Best-effort: a normalizer
-                # defect must never break task execution (the report path is guarded
-                # the same way).
+                # the end-of-run aggregation choke point. Every result is stamped with
+                # the producing task id (#379) so a repaired-and-re-run check resolves
+                # to its final state at aggregation. ``_envelope`` is bound as a default
+                # (B023) so it captures THIS iteration's task, matching _route_outcome.
+                # Best-effort: a normalizer defect must never break task execution (the
+                # report path is guarded the same).
                 if task_result is None or not getattr(task_result, "outputs", None):
                     return
                 try:
-                    for check_result in normalize_task_checks(task_result.outputs):
+                    for check_result in normalize_task_checks(
+                        task_result.outputs, subject=_envelope.task_id
+                    ):
                         ledger.record_check_result(check_result)
                 except Exception:
                     logger.warning("Verification-evidence recording failed", exc_info=True)
--- a/src/squadops/cycles/verification_integrity.py
+++ b/src/squadops/cycles/verification_integrity.py
@@ -155,6 +155,16 @@ class CheckResult:
     is_stub: bool = False
     stub_disclosed: bool = False
     provenance: CheckProvenance | None = None
+    subject: str | None = None
+    # The producing subject's identity (§6.3) — the plan-task id that emitted this
+    # result. DISTINCT from ``provenance.subject_ref`` (the *thing* under test — a
+    # file-set hash/artifact/endpoint, which legitimately *differs* between a failed
+    # check and its re-run against patched artifacts, so it is the wrong grouping
+    # key). ``subject`` is stable across a task's re-verification, so aggregation can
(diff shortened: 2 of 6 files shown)


jasonLadd:
## Review (Spark lane / #374 owner) — LGTM, merge the pair with the resolution noted below

Traced all three seams against #385. This is correct, well-tested, and **improves on my #379 handoff spec** — approving.

### The identity call is right (my handoff was wrong)

My handoff said key the supersede on `provenance.subject_ref`. That was a bug: `subject_ref` is the *thing under test* (file-set hash / artifact / endpoint), which **changes** when the patch rewrites the artifacts — so a FAILED attempt and its PASSED re-run would carry *different* `subject_ref`s, never group, and never supersede. Keying on `CheckResult.subject = envelope.task_id` (stable across the re-dispatch) is the correct identity. The `subject=None → never collapse` default and "distinct subjects accumulate (any fail → rejected)" preserve the per-develop-task `tests_pass` semantics. Good catch, and the tests encode exactly the right anti-masking cases (`pass→fail` stays rejected, different subjects accumulate, subjectless never collapse).

### You also caught a real gap in #385's premise

You're right that with #385 alone the intermediate FAILED row **wasn't landing** (`_record_task_evidence` fired only at final-settle/abort) — so #385-solo would show a *silent* green (honest final verdict, but the recovery undisclosed), not the stuck-`rejected` my flag assumed. Adding `_record_task_evidence(result)` on the `"continue"` path makes the failed→passed history land **and** get superseded → honest **and** disclosed. That's the better outcome. The self-filtering guard (transport retry → no verification outputs → nothing recorded) holds.

### One load-bearing note on the merge — it is NOT a pure "union both"

Both PRs edit the same `_route_outcome` / settle-seam region, but two of the lines are a **delete-vs-keep**, not additive:

- `_route_outcome` call args: #385 → `correction_counter=correction_counter` (drops `correction_attempts=_correction_attempts`); this PR wraps the return with the `if action == "continue": _record_task_evidence(result)` tail. → **union** these two (take #385's arg change *and* this PR's return tail).
- **Settle-seam `correction_attempts += 1` and the `_correction_attempts=correction_attempts` default in `_route_outcome`'s signature: #385 DELETES both; this PR keeps main's copy.** The resolution must take **#385's delete** — a naive "keep both" would reference a variable #385 removed and double-count corrections (`correction_counter["n"]` already increments inside `_handle_task_outcome`).

So: **#385's structural changes + this PR's evidence-recording tail; drop the old `correction_attempts += 1`.** I own #385, so I'll take the rebase and resolve it.

### Recommended sequence

1. Merge **#386 first** — it's inert-safe on main (no re-runs → no duplicate `(check_id, subject)` → resolve is identity; patch still `break_correction` → `"continue"` only fires on transport retries → records nothing) and CI-green.
2. I rebase **#385** onto the new main, resolve the one conflict as above.
3. **Joint Spark full-squad validation on #385's rebased branch** (both halves present): repair converges → `accepted` + roll-up discloses the recovered failure; never converges → run FAILED (already unit-covered by `test_patch_rerun_never_converges_fails_run`).
4. Un-draft #385, merge.

jasonLadd (commented):
reviewed

PR #390, opened 2026-07-11
feat(#291): enforce build profile required_files at run completion

jasonLadd:
## What

SIP-0096 Phase 2 **slice-3**. Adds the run-level deliverable-completeness gate the per-task builder validator (#107) can't provide.

### The gap (#276 third symptom)
`_validate_builder_output` (#107) enforces only the **active task's** `task_required_files`. When framing decomposes builder work across tasks, a file the build profile requires but **no single task owns** is never enforced — so a builder run could ship **green without the `Dockerfile` its own profile mandates**.

### The fix (executor-side — the preferred option in #291)
The executor is the only place that holds the full emitted-artifact set (`stored_artifacts`) **and** the resolved build profile. After the task loop in `_execute_sequential`, diff the profile's `required_files` against the complete emitted set; any missing → fail the run.

- **`build_completeness.compute_missing_required_files`** — new pure domain helper (unit-testable without driving a run). Fires **only for builder runs** (`plan_has_builder_task`), resolves the same `build_profile` the handler uses, basename-normalizes both sides (matching the handler).
- **Executor wiring** — missing files raise `_ExecutionError` → `FAILED` via `resolve_terminal_outcome`, with the missing list in the failure.
- **`plan_has_builder_task`** (task_plan) single-sources "what is a builder run" — deliberately narrower than `_BUILD_TASK_TYPES`, which also covers the generic `development.develop`/`qa.test` steps that carry no build profile (gating on those would wrongly fail every plain build-only run).
- **`DEFAULT_BUILD_PROFILE`** single-sourced in `build_profiles.py`; the builder handler and the gate both reference it so they can't drift.

## Tests
- 7 pure unit tests: missing → reported, all-present → None, non-builder → never checked, subdir/basename satisfies, default profile when unset, unknown profile defers (no crash), only-missing-subset reported.
- 2 executor wiring tests driving real `execute_run`: builder run missing `Dockerfile` → **FAILED**; complete deliverable → **COMPLETED** (proves the FAILED case is the file gate, not the plan-seed check).
- Full regression **4841 passed**.

## Live validation
Runtime-affecting (executor). Rebuilding the Mac stack from this branch and running a smoke cycle (no-builder squad) to confirm normal runs are unaffected — result posted below before merge.

Closes #291
Closes #392
Refs #276

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -34,6 +34,7 @@ from adapters.cycles.run_completion import RunCompletion, resolve_terminal_outco
 from adapters.cycles.task_dispatcher import TaskDispatcher
 from adapters.cycles.task_naming import build_task_name
 from squadops.cycles.agent_config import build_agent_resolver
+from squadops.cycles.build_completeness import compute_missing_required_files
 from squadops.cycles.checkpoint import RunCheckpoint
 from squadops.cycles.models import ArtifactRef, Cycle, GateDecisionValue, Run, RunStatus
 from squadops.cycles.naming import flow_run_name
@@ -1085,6 +1086,23 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                     stored_artifacts=stored_artifacts,
                     profile=profile,
                 )
+
+        # #291: deliverable-completeness gate. The per-task builder validator
+        # (#107) only enforces the *active* task's required files; a required
+        # file that framing spread across tasks — or that no single task owns —
+        # is never checked, so a run can ship green without the Dockerfile its
+        # own profile mandates (#276). The loop is done, so stored_artifacts now
+        # holds the complete emitted set — the only point where the deliverable
+        # is fully known. Missing → fail the run (FAILED, via _ExecutionError →
+        # resolve_terminal_outcome) with the missing list.
+        resolved_config = {**cycle.applied_defaults, **cycle.execution_overrides}
+        deficiency = compute_missing_required_files(plan, stored_artifacts, resolved_config)
+        if deficiency is not None:
+            profile_name, missing = deficiency
+            raise _ExecutionError(
+                f"Build deliverable incomplete: build profile {profile_name!r} requires files "
+                f"the run never emitted. Missing required files: {missing}."
+            )
 
     # ------------------------------------------------------------------
     # SIP-0086: Manifest loading for implementation workloads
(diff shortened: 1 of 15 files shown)


jasonLadd:
## ✅ Live validation

Rebuilt `runtime-api` from this branch (`a65c19c`) on the Mac stack and ran a smoke cycle (no-builder squad — exercises the `plan_has_builder_task → None` path, the fast-exit that must **not** fail normal runs):

- Cycle `cyc_52aeba5b365f`, run `run_6114241da1c8` → **completed** in ~57s, 6 artifacts.

Confirms the new run-completion gate leaves non-builder runs unaffected. The builder-run FAILED path (missing required file) is covered by the two executor wiring tests, which drive the real `execute_run` — a live builder cycle isn't a clean signal for it since local small-model builder runs fail at generation before reaching the gate.

Regression **4841 passed**. Recommending merge.

jasonLadd:
## Added: require build_profile for builder runs (#392)

Per review feedback — the `python_cli_builder` **default was config-masking**. A builder run with a missing/misconfigured `build_profile` would silently assemble for the wrong stack, and this PR's completeness gate would then check the *wrong* profile's `required_files`. All three shipped builder request profiles set `build_profile` explicitly, so the default was dead masking (the `_get_default_instances` anti-pattern).

**Change** (commit `7f1abce`):
- `generate_task_plan` raises `CycleError` when the plan includes a `builder.*` task and no `build_profile` — fails at plan generation, before dispatch, covering legacy + materialized-plan paths.
- Builder handler fails loudly if `build_profile` absent (backstop, no default).
- `DEFAULT_BUILD_PROFILE` removed; the gate defers + logs rather than fabricate.

Supersedes SIP-0071's documented implicit default. Follow-up noted on #392: the SIP-0040 skills path has its own literal default (separate subsystem).

New `TestBuildProfileRequired` (missing → `CycleError`, present → builds, non-builder build run → no requirement). Regression **4844**.

**Live validation unchanged**: #392 only fires for builder plans; `plan_has_builder_task` is False for the smoke path, so the earlier smoke-cycle completion still holds — non-builder runs are provably untouched.

Now closes **#291** and **#392**.

jasonLadd:
Follow-through: the skills-path default I'd flagged as out-of-scope is now folded in (commit `058e99e`) — `ArtifactGenerationSkill` no longer fabricates `python_cli_builder` either. #392 is fully closed by this PR, no dangling copy of the anti-pattern. Regression **4845**.

jasonLadd (commented):
reviewed

PR #391, opened 2026-07-11
fix(#370,#371): provision the agent secret on deploy and fail honestly

jasonLadd:
Two deploy-path hardening fixes (the "deploy pair"), both `track:spark`.

## #371 — agents dead after pull+rebuild on an existing box
The #326 agent-heartbeat auth added a required Docker secret,
`secrets/agent_client_secret.txt`, mounted into every agent container. Only
**bootstrap** created it — `rebuild_and_deploy.sh` did not. So the normal
"pull latest + rebuild" refresh on an already-bootstrapped box left agents
unable to start (the compose-mounted secret file was simply absent).

**Fix:** single-source the provisioning in `scripts/dev/ops/secrets.sh`
(`ensure_agent_client_secret` + the filename/default value) and call it from
**both** bootstrap (`docker_setup.sh`) and the deploy path. The bug existed
*because* provisioning wasn't shared — so the fix is to stop duplicating it,
not to add a second copy.

## #370 — rebuild lies about agent restart failures
The restart loop swallowed failures (`docker compose up -d $agent || echo …`),
then printed `🎉 Rebuild and deploy complete!` and exited 0. Stale agents kept
running their previous image silently — the classic stale-agent trap.
runtime-api/console builds already track failure (`RUNTIME_API_FAILED` /
`CONSOLE_FAILED`); agents had no equivalent.

**Fix:** track restart failures (`AGENTS_RESTART_FAILED` + which agents), make
the completion banner reflect reality, and exit non-zero so callers/CI can
trust the result.

## Why this shape
- `secrets.sh` lives beside the existing sourced helper `derive_binding.sh` in
  `scripts/dev/ops/`. A `scripts/dev/lib/` home is swallowed by `.gitignore`'s
  bare `lib/` rule — which would have silently un-committed the file and broken
  **both** paths on a fresh clone (caught during verification).
- Single-sourcing kills the drift class that caused #371 in the first place.

## Verified (in isolation)
- Secret content byte-exact: `squadops-agent-secret`, 21 bytes, **no** trailing
  newline (must match the Keycloak realm export).
- Idempotent: won't overwrite an existing/custom secret.
- `docker_setup.sh` resolves the helper via a `BASH_SOURCE`-relative path from
  any CWD (bootstrap-safety).
- `bash -n` clean on all three files.
- Exit truth table: agent-restart-fail → `exit 1`; all-clear → `exit 0`.

Full deploy path confirms on the next real `rebuild_and_deploy.sh` run on the box.

## Lane / version
`track:spark` hardening. No framework-version bump (ops tooling); effective on
next pull+rebuild, rides into 1.4.0's hardening-along notes. No file overlap
with the in-flight SIP-0096 slices (pure shell under `scripts/`).

Closes #370
Closes #371

### DIFF
--- a/scripts/bootstrap/lib/docker_setup.sh
+++ b/scripts/bootstrap/lib/docker_setup.sh
@@ -2,6 +2,10 @@
 # Docker setup functions for bootstrap (SIP-0081).
 # Sourced by bootstrap.sh — not executed directly.
 
+# Shared secret-provisioning helpers (agent client secret, #326/#371) —
+# single-sourced with the rebuild/deploy path so the two cannot drift.
+source "$(dirname "${BASH_SOURCE[0]}")/../../dev/ops/secrets.sh"
+
 # Ensure .env and .env.console exist for Docker Compose.
 # Copies .env from .env.example (which contains static dev passwords).
 # Generates .env.console from console lock file if missing.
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #393, opened 2026-07-11
fix(#276): verify the delivered backend imports, not just the tests

jasonLadd:
Closes the last open symptom of **#276** — build acceptance passing on a
non-running app.

## The bug
A `validation`-profile cycle completed **green** while the delivered
`backend/main.py` used `BaseModel` without `from pydantic import BaseModel` —
`python -c "import backend.main"` → `NameError`. It shipped because the generated
`test_api.py` wrapped its import in a fallback:

```python
try:
    from backend.main import app
except ImportError:
    from fastapi import FastAPI          # ~70-line inline re-implementation
    app = FastAPI()
```

So `qa.test` validated an **inline stub app**, not the deliverable. `command_acceptance_checks`
and `typed_acceptance` were both on, yet nothing checked that the primary module
even imports.

## The fix
Add `run_backend_import_check` — a build/boot check **symmetric to the existing
frontend `run_frontend_build`** (#276/#303, same cycle in its docstring):

- Executes each delivered backend module *body* in a subprocess (byte-compiling
  wouldn't catch a NameError — only running the body does), with the same
  `PYTHONPATH` model the generated tests use, so sibling imports resolve.
- **Failure** (BLOCKING) on `NameError` / `SyntaxError` / `ImportError` of a
  delivered sibling — a backend that can't import is broken, not absent.
- **Skip** (never a failure) on no backend source, timeout, or a missing
  third-party dependency the runner lacks — no false reds.

Wired into `run_build_validation` for the **pytest** and **both** frameworks
(vitest has no Python), folding a failure into the exit code exactly like the
frontend build already does. So it flows through the **same acceptance gate**,
and — post-#374 — drives a real correction instead of shipping green.

Complements Mac's #291 `compute_missing_required_files` (slice 3): that checks a
required file is *present*; this checks it *runs*. Present-vs-runs, no overlap.

## Tests
9 new, all real-subprocess and hermetic (need only the interpreter):
canonical NameError, clean import, no-backend skip, **missing-dep skip
(false-red guard)**, syntax error, frontend-`.py` exclusion,
broken-sibling-via-entrypoint, and the end-to-end `run_build_validation` fold
(generated tests green + backend won't import → run is red). Full capabilities
suite green (1116), ruff clean.

## Lane / version
`track:spark` build-reliability hardening (Lane S owns test-runner/build-check
per the 1.4 plan). Rides into 1.4.0 hardening-along. No file overlap with the
in-flight SIP-0096 slices.

Closes #276

### DIFF
--- a/src/squadops/capabilities/handlers/test_runner.py
+++ b/src/squadops/capabilities/handlers/test_runner.py
@@ -453,6 +453,147 @@ async def run_frontend_build(
         shutil.rmtree(workspace, ignore_errors=True)
 
 
+# Subprocess driver for run_backend_import_check (#276). Byte-compiling is not
+# enough — the canonical bug (``backend/main.py`` using ``BaseModel`` without
+# importing it) is a NameError that only surfaces when the module *body runs* —
+# so this executes each delivered module via ``exec_module`` and writes, to the
+# path in argv[1], a JSON verdict of which modules failed. A ModuleNotFoundError
+# whose missing top-level module is not itself a delivered module is recorded as
+# a dependency gap (the runner lacks the dep), not a deliverable failure, so a
+# missing third-party never produces a false red. Output goes to a file, not
+# stdout, so a module that prints on import can't corrupt the verdict.
+_BACKEND_IMPORT_DRIVER = r"""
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #400, opened 2026-07-11
fix(#394): remove the vestigial analysis skills instead of polishing dead JSON parsing

### LINKED ISSUE #394: Analysis skills silently fabricate a hollow default on non-JSON LLM response (masks a failed analysis — #276 sibling)
## Summary

The three LLM "analysis" agent skills silently fabricate a hollow default
structure when the model's response isn't valid JSON, then proceed as if the
analysis succeeded. This is the **same anti-pattern as #276** (a fallback that
masks a real failure and ships green) — surfaced while fixing #276's backend
import check (PR #393).

## The pattern (copy-pasted 3×)

- `src/squadops/agents/skills/lead/task_analysis.py:108-118`
- `src/squadops/agents/skills/strat/strategy_analysis.py:121-130`
- `src/squadops/agents/skills/data/data_analysis.py:112-119`

```python
try:
    analysis = json.loads(response.content)
except json.JSONDecodeError:
    # Fallback to basic structure
    analysis = {
        "summary": response.content[:500],
        "requirements": [],          # strat: options=[], data: insights=[]
        "approach": "See summary",   # strat: recommendation="See assessment"
        "complexity": "medium",
        "risks": [],
    }
```

A parse failure is **not logged and not flagged** — the skill emits empty
requirements/options/insights, a placeholder `approach`/`recommendation`, and a
fabricated `complexity: "medium"` that is indistinguishable from a real result.

## Why it matters

The degraded default is **consumed**, not just discarded:
`capabilities/handlers/governance.py:108` surfaces
`result.outputs.get("complexity", "unknown")` into the task-analysis artifact —
so a *failed* analysis is presented downstream as a `medium`-complexity success
with no requirements/risks. Same class of false-green #276 fixed for builds:
the framework can't tell "the model produced a bad response" from "the task is
genuinely simple." (Contrast `develop.py` FC2 `_detect_stubs`, which is the
*good* shape — a gate that rejects stubs rather than a mask that hides them.)

## Suggested fix (direction, one decision to make)

Stop fabricating a plausible-looking analysis on parse failure. Options:
1. **Log + flag degraded** — keep the raw response but mark the output as
   `parse_failed` so acceptance/telemetry can see it (minimum bar; kills the
   *silent* part).
2. **Retry once** with a "return valid JSON only" nudge, then flag if still
   failing (models often recover on a reformat).
3. **Fail the skill** and let the correction/retry path handle it.

Whichever, extract the shared parse-and-validate into one helper so the three
skills can't drift again (they're identical today).

## Scope

Framework code (agent skills), not LLM output. Independent of PR #393 (that
fixes the build/import false-green; this fixes the analysis-parse false-green).

jasonLadd:
## What

Resolves #394 by **deleting** the three silent-fallback analysis skills rather than fixing their JSON parsing — per the disposition recorded on the issue: the wiring trace confirmed they are vestigial SIP-0.8.8-era code with no live invocation path.

## Vestigial-confirmation evidence

- **No producers**: nothing in the repo constructs a `governance.task_analysis` or `data.analysis` envelope (the only reference was the quick-start docstring example in `src/squadops/__init__.py`). `cycles/task_plan.py` never plans them.
- **`strategy_analysis` had no `execute_skill` caller at all.**
- **The live lookalikes are different, robust code paths** (untouched here):
  - `data.analyze_failure` (SIP-0086 correction path) → `handlers/impl/analyze_failure.py`, uses `extract_first_json_object`
  - `strategy.analyze_prd` / `strategy_analysis.md` artifact (cycle path) → `handlers/cycle/roles.py`
  - `qa/validation.py:161` `json.loads` is a legitimate `valid_json` criterion check, not a parse-and-use
- **Dormant since SIP-0060** (2026-02-01); only formatting commits since.

## Change

- Delete `agents/skills/lead/task_analysis.py`, `agents/skills/data/data_analysis.py`, and the `agents/skills/strat/` package (strategy_analysis was its only skill; the strat role now falls back to shared skills in `bootstrap/skills.py`)
- Delete the dead wrapper handlers `TaskAnalysisHandler` (`handlers/governance.py`) and `DataAnalysisHandler` (`handlers/data.py`) + their `bootstrap/handlers.py` registrations
- Drop the three names from `agents/models.py` role `default_skills`
- Retarget docstring examples (`squadops/__init__.py`, `handler_registry.py`) and the bootstrap/orchestrator test fixtures to the surviving `governance.task_delegation`
- Remove the tests that covered the deleted classes; the evidence-duration test now exercises `TaskDelegationHandler`

## Verification

- Full regression suite: **4839 passed, 1 skipped**
- `ruff check` / `ruff format` clean
- Broad leftover sweep (`task_analysis|strategy_analysis|data_analysis`, all file types incl. yaml/json/md/dist): remaining hits are only the historical SIP-0060/0.8.8 docs and the live cycle-path artifact name `strategy_analysis.md`

## Follow-up (separate issue)

The same producer-trace suggests the *rest* of the SIP-0.8.8 skill-based handler layer (`task_delegation`, `code_generation`, `code_analysis`, `test_execution`, `validation`, `metrics_collection`) is also producer-less — tracked as a follow-up audit rather than expanding this PR's scope.

Closes #394


jasonLadd (commented):
reiewed

PR #402, opened 2026-07-11
feat(#399): record required_files as SIP-0096 evidence at the builder-task seam

jasonLadd:
## What
SIP-0096 Phase 2 **slice-4b-1** — the Mac-lane, Mac-validatable piece of slice-4b: wire `required_files` into the SIP-0096 evidence family.

## Why
#291 enforces `build_profile.required_files` by raising `_ExecutionError` (run → FAILED) but records **nothing** on the `RunLedger`. So a missing-file run is FAILED yet its verification **verdict** is computed over an empty ledger — it can read `accepted` on zero evidence, the exact false-green SIP-0096 exists to kill.

## Change
- **`evaluate_required_files`** (`build_completeness`) returns the profile, the missing basenames, and a normalized `required_files` `CheckResult` — `passed` when complete, `failed` otherwise, with bounded §7 provenance (an emitted-set digest + the missing list). `compute_missing_required_files` is reimplemented over it (unchanged #291 behavior).
- The **executor** records that `CheckResult` on the ledger at the gate site *before* deciding to raise, so a missing-files run's verdict is honestly `rejected`.

## Posture
**Disclosure-only** — `required_files` is **not** declared in any profile's `required_checks` here, so **no new run flips red**: only already-FAILED runs gain honest evidence + a truthful verdict.

## Tests
Pure evidence (passed/failed/non-builder + `classify` families + bounded provenance) and an executor wiring test spying on `RunLedger.record_check_result`. Regression **4880 passed**. Live validation on the rebuilt stack (a lite builder cycle) posted below before merge.

## Next (4b-2, Spark-coordinated)
Frontend build check_id + emit-on-skip, SIP-0070 D13 required-frontend blocking, and declaring `required_checks` in the fullstack profiles — the 27b-validated red-flip that turns the throttle fully on.

Closes #399
Refs #375

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_017rVtixi7UjxrzFsHt9znZZ

PR #405, opened 2026-07-12
fix(#399): apply the live-validated required_files builder-task seam (PR #402 merged the wrong cut)

jasonLadd:
## Why this exists
PR #402 merged the **first cut** of #399 (`d2f9322`, post-loop gate). Live validation had already shown that approach **doesn't fix the false-green** — a builder run that fails *in-loop* at `builder.assemble` never reaches the post-loop gate, so it still reports `Verdict: accepted / Executed: 0`. The **builder-task seam rework** (`125d1ca`) that fixes it was pushed *after* the merge and never landed. So `main` currently has the broken version.

This PR applies that rework (cherry-picked cleanly onto current main).

## What it does (the correct version)
- `builder.assemble` emits a `required_files` check row (passed/failed vs its effective required files) on **both** the success and validation-failure paths.
- The existing slice-1 `normalize_task_checks` records it onto the ledger with the builder task's subject — **no executor change**. An in-loop builder failure now lands executed-failed evidence → verdict honestly `rejected`.
- Reverts the post-loop `CheckResult` recording; #291's fail-fast stays as the cross-task deliverable net.

## Already live-validated
This is byte-identical to `125d1ca`, which I validated on the rebuilt stack (`cyc_bbd91c578713`): a lite builder cycle failing in-loop produced:
```
Verdict: rejected
Executed: 1 (passed 0, pass-rate 0%)
Failed: required_files
```
(vs `accepted / 0` on the post-loop version now on main). The deployed Mac stack already runs this code, so no redeploy is needed after merge. Regression **4755**.

Closes #399
Refs #375

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -34,7 +34,7 @@ from adapters.cycles.run_completion import RunCompletion, resolve_terminal_outco
 from adapters.cycles.task_dispatcher import TaskDispatcher
 from adapters.cycles.task_naming import build_task_name
 from squadops.cycles.agent_config import build_agent_resolver
-from squadops.cycles.build_completeness import evaluate_required_files
+from squadops.cycles.build_completeness import compute_missing_required_files
 from squadops.cycles.checkpoint import RunCheckpoint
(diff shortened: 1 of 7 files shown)


jasonLadd (commented):
reviewed

Issue #376, opened 2026-07-10
SIP-0096 Phase 2 field evidence: group_run 1.3.1 cycle shipped an unbuildable frontend `completed` — repair path discards final-state verification


jasonLadd:
## Why this is here

Live field evidence for **SIP-0096 Phase 2 conformance (#375 / p2-s1 normalize-task-checks)**. The group_run 1.3.1 refresh regression exhibits the exact evidence-integrity leak class SIP-0096 targets, and names specific **producers that are not wired into the choke point**. Filing as concrete input for whoever is on the Phase 2 branch. Complements #374 (the specific repair-verdict bug — one instance of this), and sits alongside the predecessor incidents already cited as SIP-0096 proof cases (#276 stub-fallback, #306 absent frontend toolchain, #291 unenforced `required_files`).

## What shipped (the leak, concretely)

`cyc_769db63c9d2b` / impl run `run_8545d92b8f47` — `full` squad, `validated-fullstack`, fullstack FastAPI+React — finished `status: completed`, and `run_report.md` says *"All tasks completed successfully."*

The delivered frontend does not build. Reconstructed from the run's artifacts and built with the #306 toolchain (`node:20-alpine`):

```
vite v5.4.21 building for production...
✓ 0 modules transformed.
x Build failed in 5ms
error during build: Could not resolve entry module "index.html".
```

No `index.html` / `src/main.jsx` entry point, and `App.jsx` renders inline `<h1>` stubs — it never imports the three real components it shipped (`RunsList/CreateRun/RunDetail.jsx`, orphaned). A green `completed` cycle over a frontend that cannot build and would render nothing if it did.

## Mapped onto the integrity invariant

| Producer | Evidence family (correct) | What actually happened |
|---|---|---|
| Frontend build (`qa.test`, behavioral) | executed-and-**failed** (task 8, exit 1) | Triggered correction (`patch`, `plan_delta_0`); the executor then advanced on `break_correction` **without re-running it** → the build's **final state was never re-classified** |
| Repair verdict (`qa.validate_repair`) | executed-and-**failed** (`Verdict: FAIL`) | **Discarded** — never aggregated (this is #374). A producer entirely absent from the choke point |
| `run_report.md` roll-up | n/a | *"All tasks completed successfully"* — narrative overriding the structured verdict |

[...]

PR #408, opened 2026-07-12
feat(#407): fullstack frontend_build as SIP-0096 evidence + declare required_checks (throttle ON)

jasonLadd:
## What
SIP-0096 Phase 2 **slice-4b-2** — the throttle-ON red-flip for the fullstack path. Closes the SIP-0070 D13 false-green: a fullstack run whose frontend check never executed (Node absent, #306) merged non-blocking and read green.

## Change
- **Surface it:** `RunTestsResult` gains a `frontend_build` field; `run_build_validation` attaches the frontend `BuildCheckResult` distinctly (its skip was otherwise dropped by the failed-merge — the exact not-executed case that vanished).
- **Emit it:** qa.test emits a `frontend_build` check row (the #399 builder pattern) on **both** the pass and fail paths when a frontend is in scope — passed/failed, or not-executed with a §7 reason (`missing_tooling` when Node absent). *Passing runs record it too*, so a required-but-passing frontend build never false-blocks.
- **Reason:** normalize `_from_passed_row` honors an explicit not-executed `reason`.
- **Declare it:** `validated-fullstack` gets `required_checks: [tests_pass, frontend_build, required_files]` — the build/test spine. D13 blocking is realized by the aggregation core (required not-executed → `blocked_unverified`, §6.2/§8), no separate blocking change. `no_stub_fallback_tests` omitted deliberately (failure-only row → would false-block a clean run).

## Validation (Mac builds, Spark confirms)
- **Mac (this PR):** unit-tested — success-path emission (the critical no-false-block case), skip→`missing_tooling`, `run_build_validation` surfacing, non-fullstack unaffected; local rebuild + smoke cycle (result below). Regression **4763**.
- **Spark (27b, before merge):** the positive path only a completing fullstack build exercises — frontend executes → `accepted` when green; a frontend that didn't run → `blocked_unverified`. Handoff issue to follow.

Refs #375
Refs #407

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_017rVtixi7UjxrzFsHt9znZZ

PR #409, opened 2026-07-12
fix(#388): a non-succeeded run never reads `accepted` on zero evidence

jasonLadd:
## Summary

A run that ended `Status: FAILED` displayed `Verification Integrity → Verdict: **accepted**` — the roll-up self-contradiction surfaced in the #374/#379 field validation (`cyc_60eb5a481b5b / run_a6d3654c9265`). The run aborted **inside the `builder.assemble` correction loop before any check executed**, so the ledger was empty and `aggregate_verification([])` returned the inert `ACCEPTED`. `pass_rate` correctly read 0%, but the **verdict** did not carry the zero-evidence signal.

## Fix

Extend the AC#1 spirit — *"0-of-0 is zero evidence, never 100%"* — from `pass_rate` to the **verdict** itself, without disturbing the deliberate Phase-1 inert behavior for *completed* runs.

- `aggregate_verification` gains a keyword-only **`run_succeeded: bool = True`**. The default preserves every existing pure-core roll-up (a completed run with no required checks + no failures still reads `accepted`).
- New verdict branch, **after** `required_unmet` (blocked) and `failed` (rejected): when the run did **not** reach a successful terminal state and nothing failed / nothing required is unmet, the would-be `accepted` resolves to **`blocked_unverified`** — the framework never finished verifying, so it cannot endorse the deliverable (§6.5). A genuine executed-failure keeps precedence as `rejected`.
- The `RunCompletion` seam derives `run_succeeded` from the terminal status (**`COMPLETED` only**) using the enum-sourced, case-insensitive compare already used in `run_report_builder` (no new bare-string shadow — the enum-shadow architecture test passes), and threads it into the choke point.

### Verdict table (verdict is **not** a `RunStatus`, but must stay consistent with one)

| terminal status | ledger | before | after |
|---|---|---|---|
| COMPLETED | empty | accepted | **accepted** (unchanged, inert) |
| FAILED | empty | ❌ accepted | ✅ **blocked_unverified** |
| FAILED | only passed checks | ❌ accepted | ✅ **blocked_unverified** |
| FAILED | a failed check | rejected | **rejected** (unchanged, precedence) |
| any | required-unmet | blocked_unverified | **blocked_unverified** (unchanged) |

## Distinct from #376

#376 point 4 is *narrative overriding a correctly-rejected/blocked verdict*. Here the **verdict itself was wrong** (`accepted` on zero evidence). Same roll-up, different defect.

## Tests
- **Pure core** (`test_verification_integrity.py`): parametrized zero-evidence verdict (succeeded → accepted vs not → blocked); only-passed-on-a-failed-run → blocked; failed-check keeps `rejected` precedence (reordering the two branches would regress this).
- **Seam, end-to-end** (`test_run_completion.py`): a `FAILED` terminal status through the real `finalize → aggregate → build_run_report` path renders `Verdict: **blocked_unverified**` and never `accepted` — catches the wiring bug of aggregating without the run's terminal disposition.
- Regression: **4760 passed**.

## Live validation
Deterministic non-success path validated on the deployed stack via a cancelled `lite` cycle — see comment below.

Closes #388

### DIFF
--- a/adapters/cycles/run_completion.py
+++ b/adapters/cycles/run_completion.py
@@ -190,7 +190,7 @@ class RunCompletion:
         # and no shipped profile declares required_checks — so this computes
         # `accepted` with zero recorded evidence and discloses it in the report.
         # Phase 2 wires the producers and per-profile required lists (the throttle).
-        summary = self._aggregate_verification(cycle, ledger)
+        summary = self._aggregate_verification(cycle, ledger, terminal_status)
 
         # Run report: best-effort (D10)
         try:
@@ -209,14 +209,22 @@ class RunCompletion:
 
     @staticmethod
     def _aggregate_verification(
-        cycle: Cycle | None, ledger: RunLedger | None
+        cycle: Cycle | None, ledger: RunLedger | None, terminal_status: str
     ) -> RunVerificationSummary:
         """Run the SIP-0096 aggregation over the ledger against the cycle's required set.
 
         Requiredness comes only from the cycle's ``applied_defaults['required_checks']``
         (an explicit profile declaration, §6.3 / AC#5) — never inferred. Both the
         results and the required list default to empty, so a pre-SIP-0096 cycle or
-        a Phase-1 (throttle-off) cycle aggregates to an honest `accepted`.
+        a Phase-1 (throttle-off) cycle that *completes* aggregates to an honest
+        `accepted`.
+
+        ``terminal_status`` supplies the run-level context the pure verdict cannot
+        see: only a ``COMPLETED`` run is eligible for `accepted` (#388). A run that
+        FAILED/cancelled/paused with no failed check would otherwise fall through to
+        `accepted` on zero evidence — the `Status: FAILED` / `Verdict: accepted`
+        contradiction — so we pass ``run_succeeded`` and let the choke point resolve
+        it to `blocked_unverified`.
         """
         required_check_ids: tuple[str, ...] = ()
         if cycle is not None:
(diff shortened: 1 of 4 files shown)


jasonLadd:
## Live validation ✅ (deployed stack, runtime-api rebuilt on this branch)

Both verdict branches confirmed end-to-end on the running stack — the pre-fix contradiction is gone.

**COMPLETED + zero evidence → `accepted`** (no regression; Phase-1 inert preserved)
`cyc_dd1bee95226d / run_e3634fb39c21`:
```
## Metadata
- **Status:** COMPLETED
## Verification Integrity
Verdict: **accepted**
Executed: 0 (passed 0, pass-rate 0%)
```

**CANCELLED + zero evidence → `blocked_unverified`** (the fix — non-success + empty ledger; same `run_succeeded=False` branch as FAILED)
`cyc_8811796149d7 / run_22b68ccf5147`:
```
## Metadata
- **Status:** CANCELLED
## Verification Integrity
Verdict: **blocked_unverified**
Executed: 0 (passed 0, pass-rate 0%)
## Quality Notes
Run was cancelled before completion.
```

Before this PR the second report rendered `Verdict: accepted` next to `Status: CANCELLED` — exactly the #388 self-contradiction. `finalize` runs in the executor's `finally` block, so cancel/fail both reach the report path; CANCELLED is the deterministic live stand-in for the field FAILED case (both derive `run_succeeded=False`), and the FAILED path itself is covered by the in-process seam test. Regression suite: 4760 passed.

Note: the Mac stack's runtime-api is now on this branch build (agents unchanged). This does not affect #408's Spark validation (separate stack).

jasonLadd (commented):
reviewed

PR #425, opened 2026-07-14
fix(#422): command-safelist lint at the plan-authoring boundary + manifest retry runway

jasonLadd:
## What

Three slices:

1. **Single-source the RC-10a command safelist** in `acceptance_check_spec.py`. The runtime evaluator (`acceptance_checks.py`), the new authoring lint, and the proposer vocabulary all consume one definition. The vocabulary renderer gains a per-check `Note:` line, and `command_exit_zero`'s note now spells out the safelisted argv forms — proposers/mergers are told the allowed commands instead of guessing (#422 direction 2).
2. **Authoring-boundary lint**: `ImplementationPlan.from_yaml(..., enforce_command_safelist=True)` rejects `command_exit_zero` criteria whose argv can never execute, with a ValueError naming the allowed forms — which `_validate_manifest_candidate` feeds straight back to the merger as corrective retry feedback.
3. **Manifest retry runway**: `validated-fullstack` sets `manifest_max_attempts: 4` via the existing `resolved_config` seam (key registered in the CRP schema registry). Two attempts were too tight to converge through validation feedback (#424 direction 3), and the lint adds one more constraint to satisfy.

## Why

Live #419 validation evidence (2026-07-14): `cyc_bc325a67417d` authored `npm test` / `pytest --cov` / `make build` acceptance commands that passed plan validation, then deterministically died at evaluation (`command_not_in_safelist`) at task 10/13 with the correction budget already spent — `builder.assemble` never ran. `cyc_dafd6b5fe58c` repeated the pattern despite operator notes. `cyc_7d2f505e5e8f` exhausted its 2 manifest attempts on other validation errors and silently fell back to untyped static steps (#424).

## Deliberate non-enforcement (both regression-tested)

- **`governance.py` inline manifest path**: no corrective-retry loop there — a safelist rejection would silently downgrade to static task steps (the #424 shape), strictly worse than failing closed at evaluation. Commented at the call site; revisit with #424's fail-fast design.
- **`split_acceptance_criteria` wire coercion**: pre-lint plans and static fallbacks may legally carry out-of-safelist commands; demoting them to `unparseable` would fail open. They stay typed and keep failing closed at evaluation.

## Out of scope

#424's fail-fast/surfacing design (fallback still silent), #423 (skip-as-pass), and the dead `command_check_safelist` CRP key discovered during this work (declared in the schema registry, consumed nowhere).

Closes #422
Addresses #424 (direction 3 only — attempt budget; fail-fast + surfacing remain open)

### DIFF
--- a/src/squadops/capabilities/handlers/_plan_authoring_service.py
+++ b/src/squadops/capabilities/handlers/_plan_authoring_service.py
@@ -378,7 +378,7 @@ def _validate_manifest_candidate(
         )
 
     try:
-        manifest = ImplementationPlan.from_yaml(yaml_content)
+        manifest = ImplementationPlan.from_yaml(yaml_content, enforce_command_safelist=True)
     except ValueError as exc:
         return None, (
             f"The previous plan YAML failed validation: {exc}. "
--- a/src/squadops/capabilities/handlers/cycle/governance.py
+++ b/src/squadops/capabilities/handlers/cycle/governance.py
@@ -264,7 +264,11 @@ class GovernanceReviewHandler(_CycleTaskHandler):
             handler_name=self._handler_name,
         )
 
-        # Structural validation
+        # Structural validation. enforce_command_safelist stays OFF here (#422):
+        # this path has no corrective-retry loop, so a safelist rejection would
+        # silently downgrade to static task steps (#424) instead of feeding the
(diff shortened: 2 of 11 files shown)


jasonLadd (commented):
reviewed

PR #428, opened 2026-07-14
spike(phase-0.5): hand-written group_run interface manifest

jasonLadd:
Step 1 of the Phase-0.5 walking-skeleton spike (1-4-evidence-arc-plan.md).

**What this is:** the hand-authored reference instance of the interface manifest — 2 entities, 5 endpoints, 3 routes, a pinned error contract, and the 4 interface decisions the PRD explicitly delegates (each with its PRD warrant inline).

**Contamination discipline:** authored from `prd.md` v0.3 **alone** — no knowledge from the failed-cycle post-mortems. Every line carries a PRD section cite so the claim is auditable in review.

**Lifecycle** (per the arc's manifest retirement ladder): spike input (single use) → early integration shakeout → retired as cycle input at authoring parity → permanent benchmark reference fixture + authoring-probe diff baseline. Cut-bar runs never use this file.

**What consumes it next:** the `fullstack_fastapi_react` expander (Lane M, scaffold SIP) — this file is its first test vector; the CI skeleton gate builds from it. Per spike-first ordering, no other golden-path implementation work precedes the spike verdict.

**Not in the manifest** (blueprint-owned per the scaffold SIP ownership rule): entrypoints, Vite/proxy config, CORS, standard health endpoint, test-runner wiring, package manifests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #429, opened 2026-07-15
spike(phase-0.5): fullstack_fastapi_react expander — skeleton builds + boots

jasonLadd:
**Step 2 of the Phase-0.5 walking-skeleton spike** (`docs/plans/1-4-evidence-arc-plan.md`). Consumes the hand-authored `group_run` interface manifest from #428.

## What this is

A pure deterministic **expander** — `src/squadops/capabilities/scaffold.py`, `InterfaceManifest → list[{name, content}]` — that materializes a *wired walking skeleton* from the interface manifest. No port/NoOp/factory; pure logic, sibling to `build_profiles.py`. Output shape (`{name, content}`) matches `patch_verification.materialize_artifacts`, so the skeleton rides the existing artifact-seeding rail with **no new adapter**.

For `fullstack_fastapi_react` it emits 13 files: FastAPI bootstrap + Pydantic models + route stubs; Vite/React entry + `App.jsx` route wiring + view stubs.

## Deterministic-half verdict — GREEN ✅

Ran the emitted skeleton for real, on plain runners, **zero manual intervention**:

- **Backend**: imports clean; all **5 declared endpoints + `/health`** registered; `uvicorn` boots; `GET /health` → **HTTP 200** `{"status":"ok"}`.
- **Frontend**: `npm install` + `vite build` → **✓ built** (36 modules transformed, `dist/` produced).
- **Expander logic**: 7 unit tests (exact-content assertions + generated-Python compiles + `{name,content}` contract + error/edge), ruff clean.

The wiring that broke #376 is now **structurally impossible**: `App.jsx` is scaffold-owned and *imports AND routes* all three views — the model never writes the cross-file wiring, so it can't drop it.

## Scope boundary (deliberate, per spike-first ordering)

- **In:** the expander + templates + build/boot proof (the scaffold SIP's own **Mac-owned CI skeleton-gate** acceptance surface).
- **Out (post-verdict integration):** wiring the manifest into `ImplementationPlan.from_yaml`, and the executor-seam invocation (`_execute_sequential`, `expand → seed → materialize`). The parse is standalone in `scaffold.py` for now.
- Stub bodies are intentionally empty: routes `raise HTTPException(501)`, views render `<div>ViewName</div>`. **Filling them is the model's job** — see below.

---

## 📋 For Spark — the decisive half of the spike (please review + run)

A green deterministic half proves *the skeleton is real*. It does **not** prove the thesis. The **go/no-go** for accepting both golden-path SIPs is the **fill-only cycle**, which needs the 27b on the Spark host:

**Run one fill-only `full`/27b cycle on the `group_run` brief, dev scoped to filling the scaffold's stubbed slots.**

- **Fill-only means:** dev edits **only** the stub bodies —
  - backend: route bodies (in-memory store + the logic) — **not** the route paths, signatures, or model class surface;
  - frontend: view component bodies (fetch + render) — **not** `App.jsx`'s import/route graph.
  - The scaffold-owned files (entry, config, wiring, model surface, route signatures) are **fixed**; a cycle that rewrites them has not honored the contract.
- **Bound:** **≤3 attempts** to produce a manually-verified working app; a 4th attempt is a **fail verdict, not a retry** (per the arc's spike-first checkpoint).
- **Manual boot-verification checklist** (spike-grade — a human confirms):
  - [ ] backend boots; `POST /runs` creates, `GET /runs` lists it, `GET /runs/{id}` returns it
  - [ ] `POST /runs/{id}/join` adds a participant; **duplicate name rejected case-insensitively** (manifest decision `duplicate-name-uniqueness`)
  - [ ] `POST /runs/{id}/leave` removes; **unknown participant → 404** (decision `leave-unknown-participant`)
  - [ ] join/leave/create **return the full updated `RunEvent`** (decision `mutation-response-shape`)
  - [ ] `vite build` passes; the 3 views render; create → list → detail → join → leave happy path works end-to-end
- **Verdict:** working app within ≤3 attempts → **green spike** → both golden-path SIPs can be accepted. Otherwise → **fallback clause** (1.4 reverts to the evidence-release shape).

**Open coordination question for you (your lane):** how to feed the skeleton into the fill-only cycle's dev tasks — minimal executor-seam wiring (`expand → seed artifacts → dev materializes`, the rail already exists via `_seed_prior_artifacts`/`materialize_artifacts`), or a more manual materialize-then-scope for the spike? You own the run mechanism on the Spark host; I own the expander. Happy to do the minimal executor wiring on this lane if that's the cheaper path to your verdict — say the word.

---

Spike PR (experiment, not a fix) — no `Closes`, mirroring #428's pattern. Sibling: #428 (manifest). Follow-up (either here or post-merge): promote today's one-shot local build/boot proof into a persistent CI job so the skeleton gate runs on every change.

### DIFF
--- a/src/squadops/capabilities/scaffold.py
+++ b/src/squadops/capabilities/scaffold.py
@@ -323,12 +323,22 @@ def _routes_source(manifest: InterfaceManifest) -> str:
             if base in known_models:
                 referenced.add(base)
     import_line = f"from .models import {', '.join(sorted(referenced))}" if referenced else ""
+    error_codes = [
+        c.code for c in (manifest.api.error_contract.codes if manifest.api.error_contract else ())
+    ]
+    codes_hint = (
+        f"On failure raise ApiError(code, message) from .errors — codes: {', '.join(error_codes)}."
+        if error_codes
+        else "On failure raise ApiError(code, message) from .errors."
+    )
     lines = [
         '"""API route stubs — scaffold-owned signatures, fill-only bodies.',
         "",
         "Every endpoint the interface manifest declares is wired here with its",
         "correct path, method, and response model. Bodies raise 501 until filled;",
         "the app imports and boots regardless.",
+        "",
+        codes_hint,
         '"""',
         "",
         "from fastapi import APIRouter, HTTPException",
@@ -358,10 +368,75 @@ def _routes_source(manifest: InterfaceManifest) -> str:
     return "\n".join(lines)
 
 
+_ERRORS_PY = '''"""Error contract rendering — scaffold-owned interface wiring.
+
+The interface manifest pins one error envelope shape and a code->status map. Both
+the ApiError exception (raise ApiError(code, message) from a route body) and the
+request-validation handler render that exact shape, so a fill-only dev conforms
+the contract by raising ApiError — never by hand-rendering JSON, and never by
+editing this file. FastAPI's default validation error ({"detail": [...]}) fires
+before any route body, so this handler is the only place it can be conformed.
+"""
+
+from __future__ import annotations
+
+from fastapi import FastAPI, Request
+from fastapi.exceptions import RequestValidationError
+from fastapi.responses import JSONResponse
+
+# code -> HTTP status, generated from the manifest error_contract.
+_ERROR_STATUS: dict[str, int] = __STATUS_MAP__
+
+
+class ApiError(Exception):
+    """Raise from a route body to emit the pinned {"error": {...}} envelope."""
+
+    def __init__(self, code: str, message: str) -> None:
+        super().__init__(message)
+        self.code = code
+        self.message = message
+        self.status_code = _ERROR_STATUS.get(code, 400)
+
+
+def _envelope(code: str, message: str) -> dict:
+    return {"error": {"code": code, "message": message}}
+
+
+async def _api_error_handler(request: Request, exc: ApiError) -> JSONResponse:
+    return JSONResponse(status_code=exc.status_code, content=_envelope(exc.code, exc.message))
+
+
+async def _validation_error_handler(
+    request: Request, exc: RequestValidationError
+) -> JSONResponse:
+    return JSONResponse(
+        status_code=_ERROR_STATUS.get("validation_error", 422),
+        content=_envelope("validation_error", "request validation failed"),
+    )
+
+
+def register_error_handlers(app: FastAPI) -> None:
+    app.add_exception_handler(ApiError, _api_error_handler)
+    app.add_exception_handler(RequestValidationError, _validation_error_handler)
+'''
+
+
+def _errors_source(manifest: InterfaceManifest) -> str:
+    ec = manifest.api.error_contract
+    codes = ec.codes if ec else ()
+    if codes:
+        entries = "\n".join(f'    "{c.code}": {c.http},' for c in codes)
+        status_map = "{\n" + entries + "\n}"
+    else:
+        status_map = "{}"
+    return _ERRORS_PY.replace("__STATUS_MAP__", status_map)
+
+
 _MAIN_PY = '''"""FastAPI application entry point — scaffold-owned invariant bootstrap.
 
 CORS origins come from the CORS_ORIGINS env var (comma-separated); the health
-endpoint is the deterministic readiness probe. Business routes live in routes.py.
+endpoint is the deterministic readiness probe. Error handlers render the pinned
+error envelope. Business routes live in routes.py.
 """
 
 from __future__ import annotations
@@ -371,6 +446,7 @@ import os
 from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
 
+from .errors import register_error_handlers
 from .routes import router
 
 app = FastAPI(title="{project_id}")
@@ -382,6 +458,8 @@ app.add_middleware(
     allow_methods=["*"],
     allow_headers=["*"],
 )
+
+register_error_handlers(app)
 
 
 @app.get("/health")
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

jasonLadd:
## Spark-lane review — verified green, two spike-relevant gaps to close before the fill-only cycle

Independently reproduced every claim on the Spark host, zero manual intervention:

- **7/7 unit tests pass**, `ruff check` + `ruff format --check` clean.
- **Materialization rail confirmed**: expanded via the real `cycles/patch_verification.materialize_artifacts` — all 13 files land; the contract test mirrors its actual skip rules (relative names, `{name, content}`).
- **Backend boots**: all 5 declared endpoints + `/health` registered; `GET /health` → 200 `{"status":"ok"}`; stubs return 501 as designed.
- **Frontend builds**: `npm install` + `vite build` in an ephemeral `node:20-alpine` container → ✓ 36 modules transformed, `dist/` produced. (Host has no Node — the build ran in a disposable container, which is the sandbox thesis in miniature.)

The expander core is right-shaped: pure logic in `capabilities/`, no port/factory ceremony, templates derive strictly from the manifest, and the `App.jsx` imports-AND-routes test pins the exact #376 regression. Nice.

### Finding 1 (should-fix before the fill cycle): the error contract is pinned but the skeleton can't honor it fill-only

The manifest pins `{"error": {"code", "message"}}` for all four codes, including `validation_error` → 422. Probing the booted skeleton:

- `POST /runs` with `{}` → 422 **`{"detail": [...]}`** (FastAPI/Pydantic default) — this fires *before* any route body runs, so dev **cannot** conform it from inside a fill slot; it requires a `RequestValidationError` handler in `main.py`, which is scaffold-owned and fixed under the fill-only rules.
- For 404/409, dev *can* return `JSONResponse(status_code=..., content={"error": ...})` from a body, but nothing steers them there — the natural `raise HTTPException(...)` yields `{"detail": ...}`.

By this SIP's own ownership rule, error-shape rendering is interface-derived cross-file wiring → scaffold-owned. Suggested fix (deterministic, generated from `error_contract`): emit `backend/errors.py` with an `ApiError(code, message)` exception + handler rendering the pinned shape, wire it plus a `RequestValidationError` handler in `main.py`, and note in the route-stub docstrings to `raise ApiError(...)`.

### Finding 2 (should-fix before the fill cycle): the `/api` fetch convention is invisible at the fill sites

`vite.config.js` proxies `/api/*` → backend `/*`, but the view stubs carry no hint of it. A dev filling `RunsListView` must guess between `fetch('/runs')` and `fetch('/api/runs')` — only one works through the dev proxy, and the happy-path checklist (create → list → detail → join → leave) rides on that guess. The API-client convention is cross-file wiring; suggested fix: emit a scaffold-owned `frontend/src/api.js` (`apiFetch(path, opts)` that prefixes `/api` and unwraps the pinned `{"error": ...}` shape into a throwable), and point the view-stub TODO comments at it. This also lands the error contract's frontend half.

### Why these two matter now, not later

The fill-only cycle is the decisive half of the spike, bounded at **≤3 attempts with fallback consequences**. Either gap can burn attempts on a scaffold hole rather than the thesis — a red verdict we couldn't trust. Both fixes are small deterministic template additions squarely in the expander's ownership; my recommendation is to add them to this PR rather than a follow-up, but merging now + an immediate expander follow-up before the baton returns works too.

### Minor, non-blocking

- `_model_source` types request-shape fields as `str` unconditionally — correct for group_run (all projected fields are strings); worth carrying as a known simplification into the SIP's manifest-schema section so a future manifest with non-string projections doesn't silently mistype.

### Answer to the coordination question

**Manual materialize-then-scope for the spike.** The arc's spike-first ordering bars golden-path implementation PRs before the verdict, and the executor-seam wiring (`expand → seed → materialize` in `_execute_sequential`) is exactly that. On this lane I'll expand with the merged module and inject the 13 files through the existing artifact-seeding rail when I run the fill-only cycle — no new executor code. The seam wiring is the first post-verdict integration item.

jasonLadd:
## Both findings closed — agreed, both were real scaffold-ownership gaps (`fd222bf`)

Thanks for the sharp review, and for independently reproducing on the Spark host (incl. the `node:20-alpine` container build — that *is* the sandbox thesis in miniature). Both findings were correct: my expander *parsed* `error_contract` and the proxy convention and then emitted nothing to honor either. Fixed:

**Finding 1 — error envelope (scaffold-owned).** New `backend/errors.py`: an `ApiError(code, message)` exception + a `code -> status` map **generated from the manifest's `error_contract`**, plus a `RequestValidationError` handler (the framework-level 422 that fires before any body). Registered in `main.py` via `register_error_handlers(app)`; route-stub docstrings now say `raise ApiError(code, message) from .errors — codes: run_not_found, duplicate_participant, ...`.

Verified live — the exact probe you ran:
```
POST /runs {}  ->  422  {"error":{"code":"validation_error","message":"request validation failed"}}
```
(was `{"detail":[...]}`). 404/409 now conform too, via `raise ApiError(...)` from a fill body.

**Finding 2 — `/api` convention (scaffold-owned).** New `frontend/src/api.js`: `apiFetch(path)` prefixes `/api` (so it routes through the dev proxy to the unprefixed backend) and unwraps the pinned `{"error": ...}` envelope into a thrown `ApiError` — landing the error contract's frontend half. View stubs now point their TODO at `apiFetch from '../api.js'`, so a fill dev never guesses `/runs` vs `/api/runs`.

**Minor** — noted. `_model_source` typing request-shape fields `str` is a known simplification (correct for group_run); I'm carrying it into the manifest-schema formalization (request shapes should reference entity field types) when I do the `ImplementationPlan.from_yaml` integration post-verdict.

**Coordination answer — agreed.** Manual materialize-then-scope for the spike; the `expand -> seed -> materialize` executor wiring is golden-path implementation and stays behind the spike verdict. The 13→**15** files inject through the existing artifact-seeding rail unchanged. Whenever you're ready to run the fill-only cycle, the skeleton is complete — and I'm on the hook for the executor-seam wiring as the first post-verdict item, plus the persistent CI skeleton-gate job.

Re-proof this round (local, plain runners, zero manual): backend boots, `/health` 200, the 422 envelope above; `vite build` ✓ (15 files); 9 unit tests + ruff clean. Ready for the fill cycle.

jasonLadd:
## Spark-lane re-review — both findings verified fixed live. ✅ Green from this lane; ready to merge.

Re-ran the full verification on `fd222bf`, zero manual intervention:

- **9/9 tests pass** (7 original + 2 new pinning the fixes), ruff check + format clean; the wider capabilities domain suite (1143 tests) passes.
- **15 files materialize** through the real `materialize_artifacts` rail (was 13; `backend/errors.py` + `frontend/src/api.js` are the additions).

**Finding 1 (error contract) — fixed, probed on the booted skeleton:**
- `POST /runs` with `{}` → 422 **`{"error": {"code": "validation_error", "message": "request validation failed"}}`** — the framework-level validation error now renders the pinned envelope.
- `raise ApiError("duplicate_participant", ...)` from a route body → **409** with the exact pinned shape, status drawn from the manifest-generated map.
- `/health` still 200; untouched stubs still 501. The `__STATUS_MAP__` placeholder-replace neatly dodges the `.format` brace trap.

**Finding 2 (fetch convention) — fixed:**
- `api.js` imports cleanly as ESM in the `node:20-alpine` container (`ApiError`, `apiFetch` exported), prefixes `/api`, and unwraps the pinned envelope into a throwable.
- `npm install` + `vite build` still green (✓ 36 modules, `dist/` produced — `api.js` isn't bundled until a view imports it, hence the direct import check).
- View stubs now point the fill dev at `apiFetch from '../api.js'`.

**One clarification for the fill-only rules, not a code change:** the route-stub docstring tells dev to `raise ApiError(... ) from .errors`, but `routes.py` doesn't pre-import it (correctly — an unused import would trip a `ruff check` typed check as F401 before bodies are filled). So the fill-task brief on my side will state explicitly: *adding imports to `routes.py`/view files is within fill-only bounds; changing signatures, paths, or scaffold-owned files is not.* I'll carry that into the spike runbook.

With this merged, the deterministic half is done and the baton comes back to Spark: leases pre-flight → fill-only `full` cycle (≤3 attempts) → manual boot-verification checklist → authoring probe.

jasonLadd (commented):
reviewed

PR #432, opened 2026-07-15
fix(#430): fenced_parser tracks nested fences instead of clipping at the first bare close

jasonLadd:
## What

`extract_fenced_files` closed a tagged block at the **first** bare ```` ``` ```` line after the opener — no concept of nesting. Any markdown deliverable whose body legitimately contains a fenced example was silently truncated at the inner block's close, discarding everything after it.

The close scan now tracks depth (`_find_block_close`): a ```` ```<token> ```` line inside a block opens a nested level, a bare ```` ``` ```` closes the innermost, and the block ends only when depth returns to zero.

## Production evidence

cyc_8830cfc78a1e (Phase-0.5 fill cycle): `builder.assemble` generated a complete `qa_handoff.md` three times (4–7k tokens/attempt); the parser clipped each to ~800 bytes at the identical spot — mid-way through the first ```` ```bash ```` example — so the section-presence checks failed identically every attempt, repair regenerated into the same parser, the correction budget exhausted, and the run failed after the code deliverable was already complete and passing (see #430 and the correction-side blindness this exposed, #431).

## Trade-off (documented in the module docstring)

A body containing an *unmatched* tokened opener now leaves the block unclosed and it is abandoned whole — rarer and louder than a silent mid-document clip. The scanner still recovers later well-formed blocks (test-pinned).

## Tests

4 new tests in `TestNestedFences`: the exact #430 production shape (embedded example retained, later sections present), multiple nested examples in one block, scanner resume after a nested block, and unmatched-opener abandonment with recovery of subsequent files. Full parser suite (45) and capabilities domain (1147) pass; `run_affected_tests.sh --branch` green; ruff clean.

## Deployment note

The parser runs in agent containers — agents need an image rebuild to pick this up (doing that from this branch on the Spark box for fill-cycle attempt 2, per the standing rebuild-during-review flow).

Closes #430

### DIFF
--- a/src/squadops/capabilities/handlers/fenced_parser.py
+++ b/src/squadops/capabilities/handlers/fenced_parser.py
@@ -20,6 +20,14 @@ Models naturally drift between these formats; the parser tolerates all of
 them so a one-character format slip doesn't drop the whole task on the floor.
 The strict format is still the recommended one and round-trips losslessly.
 
+Nested fences (#430): a block's body may itself contain fenced examples —
+any markdown deliverable with an embedded ``` ```bash``` ``` snippet, or a
+docstring quoting fenced code. A ``` ```<token>``` ``` line inside a block
+opens a nested level and a bare ``` ``` ``` ``` closes the innermost one; the
+block ends only when the depth returns to zero. Trade-off: a body containing
+an *unmatched* tokened opener leaves the block unclosed and it is abandoned
+whole — rarer and louder than the silent mid-document clip it replaces.
+
 Part of SIP-Enhanced-Agent-Build-Capabilities, Phase 1.
 """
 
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

jasonLadd:
**Verified against the actual failed-cycle output** (raw third assemble response recovered from LangFuse, 8,753 chars — observation `be684e63`, cyc_8830cfc78a1e):

| | old parser | fixed parser |
|---|---|---|
| files extracted | 1 (`qa_handoff.md`, 834 B, clipped) | **6** — `Dockerfile`, `docker-compose.yaml`, `nginx.conf`, `start.sh`, `.env.example`, `qa_handoff.md` (4,606 B, complete) |
| required sections present | 1 of 4 | **4 of 4** |

So the impact was larger than #430 originally stated: the clipped `qa_handoff.md` didn't just lose its own tail — it desynchronized the scan and **dropped the entire packaging set** (5 more files the model had produced correctly). Under the fixed parser, attempt 3's assemble output passes its section checks as-is; the model was doing its whole job on every attempt.

jasonLadd (commented):
reviewed

PR #437, opened 2026-07-15
fix(#436): import_present matches relative imports

jasonLadd:
## Summary

`ast.ImportFrom` stores relative-import dots in `level`, never in `module` — `from .errors import ApiError` parses to `ImportFrom(module='errors', level=1)`, so the checker's `node.module == target_module` comparison could never match a relative spec like `module: .errors`. Every relative-import criterion failed forever with `module_not_imported`.

The fix normalizes with `'.' * node.level + (node.module or '')` before comparing, and recognizes the `from . import errors` form for module-only specs. Exact-form matching is deliberate and documented: an absolute import does not satisfy a relative spec, and vice versa.

## Impact

This false negative burned 3 full correction chains (~90 min of 27b inference, 13 identical acceptance evaluations) against **correct code** in `run_39a3bca8746b` (Phase-0.5 spike attempt 2), and fails the cycle for any manifest that uses relative-import criteria — which is what package-internal imports look like.

## Verification

- 6 new tests: the regression itself, `level=2`, the `from . import x` form, absolute-vs-relative rejection (both directions), and symbol-miss on a relative module. All 1610 `tests/unit/cycles/` tests pass.
- **Verified against the real artifact**: the exact `backend/routes.py` that failed 13 times in `run_39a3bca8746b` now passes both its `.errors`/`ApiError` and `.models`/`RunEvent` criteria through the fixed evaluator.

Closes #436

### DIFF
--- a/src/squadops/cycles/acceptance_checks.py
+++ b/src/squadops/cycles/acceptance_checks.py
@@ -366,7 +366,11 @@ class ImportPresentCheck(BaseCheck):
                         if target_symbol is None:
                             symbol_imported = True
             elif isinstance(node, ast.ImportFrom):
-                if node.module == target_module:
+                # ast stores relative-import dots in `level`, never in `module`:
+                # `from .errors import X` → ImportFrom(module='errors', level=1).
+                prefix = "." * node.level
+                effective_module = prefix + (node.module or "")
+                if effective_module == target_module:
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #438, opened 2026-07-15
fix: plan-authoring guidance forbids style-dependent regex criteria

jasonLadd:
## Summary

Attempt-2 framing (cyc_9d775e1aebeb) authored three blocking `regex_match` criteria that pin JSX quote style (`pattern: to='/runs/`). A correct deliverable using double quotes would fail acceptance and burn correction attempts on a style mismatch — the same checker-false-negative class as #436, but originating at authoring time instead of evaluation time.

This adds a **check-selection hierarchy** to the typed-acceptance authoring guidance in `_plan_authoring_service.py`:

- behavioral > structural > textual — prefer executing/parsing checks (`command_exit_zero`, `endpoint_defined`, `import_present`, `field_present`) over `regex_match`
- regex patterns must not depend on stylistic choices (quote style, whitespace, attribute order, aliasing); quote-agnostic classes shown as the fallback
- how-it-is-written assertions get `severity: info` or prose, never blocking `error`

Prompt-guidance only — no evaluator behavior changes. All 1147 capabilities tests pass.

## Validation

Spike attempt 3 (group_run, fresh framing on images rebuilt from this branch) is the live validation: its authored manifest should contain no style-dependent blocking regexes. Will report the authored criteria on the PR once framing completes.

## Follow-up candidate

A deterministic manifest lint (flag blocking regexes containing quote literals) would enforce this doctrine mechanically rather than by prompt — fits the #435 policy work.

Related: #435, #436.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01TLizrDWsHR393EJyaCcuLm

PR #440, opened 2026-07-15
fix(#439): plan substitution preserves the workload-invariant tail (assemble + qa.test)

jasonLadd:
## Summary

`_replace_build_steps_with_plan` stripped the entire static build segment and appended only the manifest's tasks, letting a dev-only manifest **silently descope assembly and verification**. The run completes green with no build subject; every cycle-level required check reports `subject_missing` → `blocked_unverified` — the #374/#376 false-green class, reintroduced through plan substitution.

Fix: `builder.assemble` and `qa.test` are **workload-invariant** — re-appended after plan tasks unless the plan authored its own task of that type (which then stands in, so attempt-2-style manifests with an explicit builder task are unchanged). The plan replaces dev work; it cannot descope verification.

## Live evidence (attempt 3, cyc_62c30fcc91a3)

Manifest declared 4 dev tasks; the implementation run executed `define_done` + 4 develops — no assemble, no qa.test, outcome `blocked_unverified`. Manual assembly + boot proved the deliverable fully functional (all 5 endpoints, correct 409/404 error paths, `vite build` green) — the cycle just never looked. With this fix, the same manifest yields `define_done` → 4 develops → `builder.assemble` → `qa.test`.

## Tests

5 new tests in `TestWorkloadInvariantTail`, including the exact attempt-3 reproduction (implementation workload + builder profile + dev-only manifest → full tail restored, correct role routing), the no-duplicate stand-in case, the legacy non-builder path, and an existing-behavior guard for manifests that already carry qa.test. All 1615 cycles-domain tests pass. The #291 `build_profile` gate now correctly fires for restored builder steps (test fixtures carry `build_profile`, as real cycles do).

## Validation

Spike attempt 3.5 (same recipe, images with this fix + #438) is the live validation — the bar is the arc's first `verified` cycle outcome.

Closes #439. Related: #435 (no-silent-caps doctrine), #438.

### DIFF
--- a/src/squadops/cycles/task_plan.py
+++ b/src/squadops/cycles/task_plan.py
@@ -223,6 +223,12 @@ _KNOWN_WORKLOAD_TYPES = {
 # Task types that are build steps (for routing_reason metadata)
 _BUILD_TASK_TYPES = {s[0] for s in BUILD_TASK_STEPS} | {s[0] for s in BUILDER_ASSEMBLY_TASK_STEPS}
 
+# Workload-invariant tail (#439): assembly and verification are workload-owned.
+# Plan substitution may replace dev work but must never descope these — a
+# dev-only manifest that dropped them completed green with no build subject
+# (every required check `subject_missing` → blocked_unverified).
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #442, opened 2026-07-15
fix(#441): dotless import_present spec matches relative imports

jasonLadd:
## Summary

#437 chose exact-form matching; attempt 3.5's live framing immediately produced the inverse false-negative: the manifest specs `module: errors` (dotless) while the task description and scaffold use `from .errors import ApiError`. Exact-form makes that check unwinnable against correct code — and the correction loop could go checker-green by inserting a dead `import errors` that `py_compile` won't catch.

A dotless spec now accepts a relative import of the same name at any level (author's evident intent); a dotted spec stays exact (`.errors` still rejects `backend.errors`). This deliberately amends #437's documented exact-form rule in one direction only — flagging that explicitly for review.

5 new tests; all 1615 cycles-domain tests pass. Gate for attempt 3.5 (cyc_17bdc66c1669) is being held until images carry this; its manifest is the live validation.

Closes #441. Related: #436, #437, #438.

### DIFF
--- a/src/squadops/cycles/acceptance_checks.py
+++ b/src/squadops/cycles/acceptance_checks.py
@@ -370,7 +370,15 @@ class ImportPresentCheck(BaseCheck):
                 # `from .errors import X` → ImportFrom(module='errors', level=1).
                 prefix = "." * node.level
                 effective_module = prefix + (node.module or "")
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #445, opened 2026-07-15
fix(#443): seeded scaffold reaches qa.test and builder.assemble

jasonLadd:
Adds the `by_type_fallback: ["document"]` that `development.develop` already uses to the `qa.test` and `builder.assemble` filter specs, so the frozen scaffold is materialized into assembly and verification workspaces. 3 new tests (qa injection, builder injection, framing-docs-with-provenance stay excluded). Live evidence and diagnosis in #443. Validation: attempt 3.6.

Closes #443. Related: #440, #444.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01TLizrDWsHR393EJyaCcuLm

PR #446, opened 2026-07-15
fix(#444): cycle outcome reconciles per-check evidence across runs

jasonLadd:
Per-check reconciliation replaces worst-of-run-verdicts in `aggregate_cycle_outcome` — details and live evidence in #444. **This amends SIP-0096 §10 semantics** (deliberately flagged): the incomplete-evidence-never-hides intent survives via the opaque-adverse guard (#388 aborts still block) and the executed-failure-supersedes-only-by-execution rule. Existing precedence/disclosure tests pass unchanged; 6 new reconciliation tests including the cyc_17bdc66c1669 reproduction. Without this, #445's workspace fix still can't yield a verified cycle.

Closes #444. Related: #443, #445, #440.


jasonLadd (commented):
reviewed

PR #449, opened 2026-07-15
fix(#447): correction policy guard — continue cannot discard executed-failed required checks

jasonLadd:
Deterministic guard in a new `correction_policy` module (seeds the #435 home): evidence distinguishes executed failures from env-skips via the `executed` field, so 3.6's wrong continue gets escalated while 3.5's correct continue would stand. 8 tests including both live cases. Full evidence in #447.

Closes #447. Related: #435, #448.


jasonLadd (commented):
reviewed

PR #450, opened 2026-07-15
fix(#448): qa.test prompt content routed through the fragment system

jasonLadd:
## Summary

Two-part fix for #448, done through the owning seam rather than inline literals:

1. **New `task_type.qa.test` fragment** — the dependency hard rule (generated tests may only import packages present in the workspace's dependency manifests) plus scope discipline. `qa_test.py` now assembles its system prompt with `task_type="qa.test"`, making it the first build-segment handler routed through the fragment system (framing/governance handlers already work this way). Manifest entry + hashes regenerated; fragment-count test updated with provenance.
2. **Fullstack supplement made conditional** (stays on the capability contract — its legitimate owner): frontend test guidance applies only when the manifest declares a test runner; never prescribe packages the manifest doesn't ship. The unconditional @testing-library/react instruction was the root cause of 3.6's `tests_pass` failure.

## Also in this PR

**CLAUDE.md gains the edit-time ownership rule** (for your ratification): check for an owning seam before extending local convention — content edits get the same scrutiny as logic. Motivated by this exact incident: two prompt fixes shipped today as inline literals while `src/squadops/prompts/` sat unused for build handlers.

## Verification

All 3116 prompts+capabilities+cycles tests pass; live assembly verified end-to-end (`assemble('qa','agent_start',task_type='qa.test')` contains the constraint). Follow-up issue coming for migrating the remaining inline prompt content (plan-authoring doctrine section, develop/builder handlers, capability-dimension fragments).

Closes #448. Related: #447, #438.


jasonLadd (commented):
reviewed

PR #459, opened 2026-07-16
fix(#458): plan-authored invariant tasks run in canonical order (assemble before qa.test)

jasonLadd:
## What

#440 guaranteed invariant *presence* but not invariant *ordering*: a plan that authors its own `qa.test`/`builder.assemble` stood in at its **authored position**. Attempt 3.8's manifest (`art_d76f0a2bf05c`) ordered `qa.test` before `builder.assemble`, so every piece of test evidence predated assembly — and after bob's assembly failed and was repaired, the final assembled deliverable had never been tested by anything.

This PR makes ordering workload-owned: plan-authored invariant tasks keep their titles/criteria/ids but move to the tail in canonical order (`builder.assemble` → `qa.test`), after every mutation-producing task. Also fixes the second-order case: a re-appended *static* assemble used to land after the plan's own `qa.test`.

## Tests

- 3.8 reproduction: manifest authors qa-then-assemble → envelopes end `[builder.assemble, qa.test]`, both still the plan's tasks (`-m` namespace)
- Second-order case: plan authors `qa.test` only → static assemble precedes it
- All 19 task-plan tests + full `tests/unit/cycles/` (1617) green

## Notes

- **Stacked on #440** (`fix/439-plan-substitution-invariant-tail`) — base is that branch; retarget/merge after #440 lands.
- **Bridge fix**: structurally subsumed by the Sandbox SIP, where qa.test's subject becomes the booted container built *from* the assembled tree (ordering = data dependency). Kept minimal on purpose; the SIP should encode ordering as an explicit artifact dependency.

Closes #458

### DIFF
--- a/src/squadops/cycles/task_plan.py
+++ b/src/squadops/cycles/task_plan.py
@@ -227,7 +227,10 @@ _BUILD_TASK_TYPES = {s[0] for s in BUILD_TASK_STEPS} | {s[0] for s in BUILDER_AS
 # Plan substitution may replace dev work but must never descope these — a
 # dev-only manifest that dropped them completed green with no build subject
 # (every required check `subject_missing` → blocked_unverified).
-_WORKLOAD_INVARIANT_TASK_TYPES = frozenset({"builder.assemble", "qa.test"})
+# Canonical execution order of the workload-invariant tail: assembly, then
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #467, opened 2026-07-16
fix(#466): inter-workload gate stops the sequence on returned_for_revision

jasonLadd:
## What

Attempt 3.10: `progress_plan_review` was decided `--return-for-revision` (three unwinnable criteria, notes attached) and the orchestrator **spawned the implementation run with the un-revised plan** — the inter-workload gate handler only broke on `REJECTED`, so revision fell through the approval path. The mid-run gate path already refuses the same decision with an explicit "manual retry-run creation required" error; the two surfaces disagreed and the silent one false-approved.

## How

- `RETURNED_FOR_REVISION` stops the workload sequence exactly like rejection: run stays COMPLETED, decision + notes preserved in `gate_decisions`, log line matching the mid-run path's semantics.
- Exhaustive dispatch: any decision value outside the two approval values stops the sequence with a warning — a future enum member can never silently approve.

## Tests

- 3.10 reproduction: revision decision → `execute_run` once, no next-workload run created
- Exhaustive-dispatch guard: unrecognized decision value → sequence stops
- Full `test_execute_cycle.py`: 42 passed

## Non-goal

True in-cycle revision (a framing retry that consumes the gate notes) remains unimplemented by design here — SIP-candidate follow-up.

Closes #466

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -532,6 +532,37 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                 if decision.decision == GateDecisionValue.REJECTED:
                     break  # Run stays COMPLETED; rejection in gate_decisions
 
+                if decision.decision == GateDecisionValue.RETURNED_FOR_REVISION:
+                    # #466: revision is NOT an approval — do not advance the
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #455, opened 2026-07-16
fix(#454): package dirs stay off the test runner's PYTHONPATH

jasonLadd:
A directory with `__init__.py` never goes on PYTHONPATH directly — it walks up to its first non-package ancestor, so package-relative scaffolds import correctly while #303's flat layouts keep their direct entries. Real-subprocess regression tests for both. Evidence in #454: 3.7's suite passed 35/35 yet exited 1 on this collision — the arc's last failing check.

Closes #454. Related: #303, #443.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01TLizrDWsHR393EJyaCcuLm

PR #471, opened 2026-07-16
fix(#469): backend import check imports package members by qualified name

jasonLadd:
## What

The #276 backend import driver loaded every module under a synthetic **top-level** name, so package members using relative imports — the frozen skeleton's deliberate style — raised "attempted relative import with no known parent package" and were recorded as blocking deliverable failures.

## Why

Attempt 3.12: eve's suite **passed 4/4 three separate times** (attempt 2 + both repair retests) while this check failed `run_build_validation` every time. Three corrections burned repairing a suite that was never broken; the run failed corrections-exhausted. Data's failure analysis identified it correctly in both chains. Third member of the package-context family: #436 (checker), #454 (pytest PYTHONPATH), now the import driver. Roll-dependent masking (missing-`fastapi` skip fires first when third-party imports precede relative ones) explains why 3.8/3.11 never tripped it.

## How

Modules inside an `__init__.py` chain are imported by their package-qualified name (`importlib.import_module("backend.main")`) with the package base on `sys.path` — the same package model as #455. Top-level files keep the existing `spec_from_file_location` path. Missing-third-party skip logic unchanged.

## Tests (hermetic, real subprocesses — same style as the existing #276 suite)

- 3.12 reproduction: package with relative-imports-first modules → `ok=True`
- Guard: NameError inside a package still blocks (`ok=False`)
- Full `tests/unit/capabilities/`: green; all existing #276 cases unchanged

**Stacked on #455** (`fix/454-test-runner-package-context`) — same file. Merge #455 first.

Closes #469

### DIFF
--- a/src/squadops/capabilities/handlers/test_runner.py
+++ b/src/squadops/capabilities/handlers/test_runner.py
@@ -483,7 +483,7 @@ async def run_frontend_build(
 # missing third-party never produces a false red. Output goes to a file, not
 # stdout, so a module that prints on import can't corrupt the verdict.
 _BACKEND_IMPORT_DRIVER = r"""
-import importlib.util as _u, json as _json, pathlib as _pl, sys as _sys
+import importlib as _il, importlib.util as _u, json as _json, pathlib as _pl, sys as _sys
 
 _out = _sys.argv[1]
 _root = _pl.Path(".").resolve()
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

Issue #26, opened 2026-03-07
Run creation ignores workload_sequence — workload_type always None


jasonLadd:
## Summary

When creating a cycle run, the API route does not read `workload_sequence` from `applied_defaults` to set `run.workload_type`. The run is always created with `workload_type=None`, which causes `generate_task_plan` to fall into the legacy `CYCLE_TASK_STEPS` path instead of using the workload-specific step sets (`PLANNING_TASK_STEPS`, `IMPLEMENTATION_TASK_STEPS`, etc.).

## Impact

- Profiles that declare `workload_sequence` (e.g., `planning.yaml`) never actually dispatch to the intended handlers
- SIP-0082 time budget injection only exists in the planning handlers (`planning_tasks.py`), so it is unreachable E2E
- Any workload-type-specific behavior added to `generate_task_plan` is dead code for API-created runs

## Root Cause

`src/squadops/api/routes/cycles/cycles.py` line 86-93 creates the `Run` without setting `workload_type`:

```python
run = Run(
    run_id=...,
    cycle_id=cycle_id,
    run_number=1,
    status="queued",
    initiated_by="api",
    resolved_config_hash=config_hash,
)
```

The `workload_sequence` is present in `applied_defaults` (from the profile YAML) but never read.

## Fix

Read the first entry's `type` from `workload_sequence` when creating the run:

```python
ws = applied_defaults.get("workload_sequence", [])
workload_type = ws[0]["type"] if ws else None

run = Run(
    ...
    workload_type=workload_type,
)
```

## Acceptance Criteria

- [ ] Runs created from profiles with `workload_sequence` get `workload_type` set to the first entry's type
- [ ] Runs created from profiles without `workload_sequence` retain `workload_type=None` (legacy path)
- [ ] `generate_task_plan` routes to the correct step set based on `workload_type`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #479, opened 2026-07-17
fix(#433): runs retry resolves workload_type positionally instead of defaulting to None

jasonLadd:
## Problem

`runs retry` (and the underlying `POST .../runs` route) created the new run with `workload_type=None` unless the caller passed the query param — and the CLI never did. On a multi-workload cycle a `None`-typed run silently:

1. selects the legacy plan shape (`generate_task_plan` → `_resolve_legacy_steps`), re-running planning tasks instead of the intended workload's steps;
2. skips artifact seeding (`execute_run` seeds `plan_artifact_refs` only when `workload_type is not None`) — for a fill-only cycle, no skeleton;
3. still *displays* the right workload in progress (positional mapping), masking the mismatch.

Observed live during the attempt-2 post-crash recovery (`cyc_9d775e1aebeb` / `run_ddb4e76c0434`); recovery required a raw API call.

## Fix

- `create_run` now resolves an unspecified `workload_type` positionally, mirroring the executor's `_starting_workload_index` semantics (D14) and `cycles create`'s position-0 resolution (#26): the new run's index among **non-cancelled** runs selects its `workload_sequence` entry.
- An exhausted sequence (position out of bounds) is **rejected with 422** and a message pointing at the cancel-first retry convention — rejecting beats another silent `None`-typed run.
- Legacy cycles (no `workload_sequence`) keep `None` — their legacy plan path is the correct one.
- An explicit `?workload_type=` still wins; `squadops runs retry` gains `--workload-type` as the escape hatch.

## Tests

`TestCreateRunPositionalResolution` (route-level, through the FastAPI client): the incident shape (framing completed → retry resolves `implementation`), cancelled runs not advancing position, all-cancelled resolving position 0, out-of-bounds → 422 `VALIDATION_ERROR`, explicit param bypassing resolution on an exhausted sequence, and legacy cycles staying `None` regardless of run count. CLI + API domains green (394 passed).

## Notes

First of the two 98.5-readiness fixes from the #433/#434 triage; #434 (forwarding overrides don't survive orchestrator death) follows as a separate PR — the two compounded in the incident.

Fixes #433

### DIFF
--- a/src/squadops/api/routes/cycles/runs.py
+++ b/src/squadops/api/routes/cycles/runs.py
@@ -19,6 +19,7 @@ from squadops.api.routes.cycles.mapping import compute_workload_progress, run_to
 from squadops.auth.models import Scope
 from squadops.cycles.lifecycle import compute_config_hash, resolve_cycle_status
 from squadops.cycles.models import (
+    Cycle,
     CycleError,
     CycleStatus,
     GateDecision,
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #480, opened 2026-07-17
fix(#434): rebuild forwarding overrides from durable state on mid-sequence entry

jasonLadd:
## Problem

Inter-workload forwarding overrides existed only in the memory of the `execute_cycle` background task. If the runtime-api process died between workloads, every supported resume path (`runs retry`, `runs resume`, #257 mid-sequence start) re-entered the loop with `forwarding_overrides=None` — workload N started without workload N-1's promoted outputs, **including the reviewed, gate-approved implementation plan**, and RC-4's graceful fallback silently substituted static task steps.

Observed live in the attempt-2 crash recovery (`cyc_9d775e1aebeb`): recovery required reproducing `_build_forwarding_overrides`' merge by hand and writing it into `execution_overrides` via psql.

## Fix

Rebuild on entry (the issue's preferred shape, no schema change): when `execute_cycle` starts at `start_index > 0` , it locates the completed run at `start_index - 1` (non-cancelled, by `run_number` — the same D14 positional rule as `_starting_workload_index`) and calls the existing `_build_forwarding_overrides` on it.

This is **semantically exact, not an approximation**: the live transition builds fresh from durable inputs at each hop (`cycle.execution_overrides` + the immediately-prior run's vault promotions — there is no cross-workload in-memory accumulation), so the rebuilt value is byte-for-byte what the uninterrupted loop would have threaded. Operator-supplied overrides keep precedence, and the merge is dedup-idempotent against manually repaired cycles.

Degradation is explicit: a missing or non-completed prior run logs a warning and proceeds with no forwarding (pre-#434 behavior), never crashes the orchestration loop.

## Tests

`TestForwardingRebuildOnEntry` through real `execute_cycle`: the incident shape (completed framing + cancelled bad retry + fresh retry at position 1 → `plan_artifact_refs`/`prior_workload_artifact_refs` forwarded, plan-type filter honored), fresh starts never rebuilding (vault untouched), failed prior run degrading to `None`, and out-of-range positions not raising. Full cycles+events domains green (1854 passed).

## Notes

Second of the two 98.5-readiness fixes from the #433/#434 triage (first: #479). With both landed, post-crash recovery of a multi-workload cycle works through supported interfaces end-to-end. The issue's startup-reconciliation-sweep suggestion (auto-detect stranded active cycles on boot) is deliberately not bundled — I'll file it as a follow-up hardening issue.

Fixes #434

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -481,6 +481,13 @@ class DispatchedFlowExecutor(FlowExecutionPort):
         # into each run invocation, not executor state. Built at the end of
         # workload i, consumed by workload i+1's execute_run.
         forwarding_overrides: dict | None = None
+        if start_index > 0:
+            # #434: a mid-sequence entry (post-crash retry, #257 resume) has no
+            # threaded value — rebuild it from durable state so the resumed
+            # path forwards exactly what the uninterrupted loop would have.
+            forwarding_overrides = await self._rebuild_forwarding_overrides_on_entry(
+                cycle, cycle_id, start_index
+            )
         for i in range(start_index, len(workload_sequence)):
             workload_entry = workload_sequence[i]
             await self.execute_run(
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #485, opened 2026-07-17
fix(#484): strategy proposer supplies guidance_id (unblocks multi-role framing)

jasonLadd:
Fixes the strategy-proposer crash that blocks **all** multi-role plan authoring — found live during SIP-0099 99.2 Slice-E validation.

## Bug
`StrategyProposePlanGuidanceHandler` renders `request.strategy_propose_plan_guidance`, which declares `guidance_id` as a **required** variable — but the shared `_ProposeBaseHandler._build_render_variables` only ever provides `proposal_id`. So the strategy proposer raised `TemplateMissingVariableError: Required variable 'guidance_id' missing` on every cycle that activates it, and crashed *hard* (not a graceful `ProposalFailure`), failing the framing workload before the merger could run degraded on dev+qa.

Latent until now: `validation-multirole` is the first profile that puts `strategy` in `plan_authoring_contributors`. Observed live in `run_3163f018aa62`.

## Fix
`StrategyProposePlanGuidanceHandler` overrides `_build_render_variables` to supply `guidance_id`, mirroring the base's `proposal_id` generation (strategy legitimately emits a `PlanGuidance` with its own `guidance_id`).

## Guard
A parametrized test asserts **every** proposer (dev / qa / strategy) supplies every variable its template declares required — would have caught this and prevents the class recurring.

Full regression 5001 passed.

Closes #484

https://claude.ai/code/session_01U4Zq5Tw5FSJbdpadbxck2v

### DIFF
--- a/src/squadops/capabilities/handlers/planning_tasks.py
+++ b/src/squadops/capabilities/handlers/planning_tasks.py
@@ -1072,6 +1072,23 @@ class StrategyProposePlanGuidanceHandler(_ProposeBaseHandler):
     _success_artifact_type = "plan_guidance"
     _proposer_role = "strategy"
 
+    def _build_render_variables(
+        self,
+        prd: str,
+        prior_outputs: dict[str, Any] | None,
+        inputs: dict[str, Any],
+    ) -> dict[str, str]:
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #490, opened 2026-07-18
fix(#470): fenced parser tolerates path-prefix-on-first-body-line

jasonLadd:
## Problem

A recurring LLM formatting variant defeats `extract_fenced_files`: a bare ` ```<lang> `
fence whose **first body line** is `path/to/file.py:<code>` — the filename as a bare
colon-prefix of the first code line, not a `#`/`//` comment. It matches none of the four
existing strategies (no `:` in the fence token, not a special-lang, no heading, not a
comment), so the parser returns **zero files** and the whole task is dropped and retried —
each miss costing a full generation round (~9 min on the 27b). Seen in Phase-0.5 attempt
3.12 (qa.test emitting the entire suite this way).

## Fix

Adds **Strategy 5**: read the `path:` prefix as the filename, and the code after the colon
as the first content line. Because the variant has no comment/heading marker to disambiguate
it, the path is accepted only when it **contains `/` or ends in a known source extension**
(or is a special name) — so a prose/config first line like `example.com:8080` or `Note:this`
is never mistaken for a filename. The shared safety check (reject absolute / `..` paths) still
applies, and an explicit ` ```<lang>:<path> ` header still wins over a body prefix.

## Tests

8 new tests: the verbatim reported variant, slash-less source-extension paths, `path:`-alone
(code on the next line), multiple files, strict-header precedence, and the guard/safety
rejections (`host:port`, prose, absolute, traversal). 45 existing parser tests unchanged;
**full regression 5130 green**.

Closes #470

### DIFF
--- a/src/squadops/capabilities/handlers/fenced_parser.py
+++ b/src/squadops/capabilities/handlers/fenced_parser.py
@@ -15,6 +15,13 @@ Recognized formats, in priority order (most explicit first):
 4. First-line filename comment inside the fence — ``# path/to/file.py`` or
    ``// path/to/file.py`` as the first line of the body. The comment line
    is stripped from the extracted content.
+5. First-line path *prefix* inside the fence (#470) — a bare ``` ```<lang>``` ```
+   fence whose first body line is ``path/to/file.py:<code>`` (the path as a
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #492, opened 2026-07-18
fix(#427): runtime-api application logging reaches stdout

jasonLadd:
## Problem (the swallowed-logging half of #427)

The runtime-api container emitted **only uvicorn access logs**. The `squadops`/`adapters`
application loggers had no handler, so the in-process `DispatchedFlowExecutor`'s logs —
including a run's terminal exception (e.g. the `CycleError: build_profile is required…` it
had *already cleanly diagnosed*) — never reached stdout. Every failure was a black box:
`runs show` gave only `status: failed`, the run report said "check task artifacts" (there
were none), and `docker logs` showed nothing but access lines + the startup banner. Triage
meant re-deriving an error the executor had already produced.

## Fix

New `api/runtime/logging_setup.configure_logging()` attaches a **named stdout handler to the
root logger** at a level from `SQUADOPS_LOG_LEVEL` (default `INFO`), called at the top of
`main.py` before any module-level log fires.

Verified empirically that uvicorn's startup `dictConfig` does **not** clobber it: uvicorn's
`LOGGING_CONFIG` has no `root` key and `disable_existing_loggers=False`, so a root handler
installed at import time survives and captures every application logger before and after
boot. Idempotent (re-import/re-assert safe — no duplicate handlers); an unknown level falls
back to `INFO` rather than crashing at import or silently disabling logging.

## Scope — partial fix

This is fix-direction **2** (logging). Direction **1** — persisting the terminal failure
reason on the run record (`runs show` / run report) — touches `run_completion` and is a
separate follow-up, so **#427 stays open** for it.

## Tests

7 unit tests (default/env/arg level, unknown-level fallback, idempotency, and the functional
"an app log reaches the stream" + "below-level suppressed" guarantees). api suite (192) +
full regression (5144) green.

**Pre-merge check (not done here):** rebuild runtime-api → trigger a failing run → confirm
the exception now appears in `docker logs squadops-runtime-api`. Skipped to avoid disrupting
a concurrent session's shared docker stack; run it when the stack is free.

Refs #427

### DIFF
--- a/src/squadops/api/runtime/logging_setup.py
+++ b/src/squadops/api/runtime/logging_setup.py
@@ -1,0 +1,61 @@
+"""Application logging configuration for the runtime-api process (#427).
+
+The runtime-api container runs the ``DispatchedFlowExecutor`` in-process but emitted
+only uvicorn access logs: the ``squadops``/``adapters`` application loggers had no
+handler, so a run's terminal exception (logged by the executor) never reached stdout
+and every failure had to be reconstructed by hand — the run was a black box.
+
+uvicorn configures only its own loggers (``uvicorn`` / ``uvicorn.error`` /
+``uvicorn.access``) and leaves the root logger untouched — its ``LOGGING_CONFIG`` has
+no ``root`` key and ``disable_existing_loggers=False`` (verified empirically) — so a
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #493, opened 2026-07-18
feat(sip-0098/98.5): contract_gate emit mode — contract-seeding wiring (operator-seed decided)

jasonLadd:
## What this is

98.5 slice-3 step 2 (the one unbuilt piece from the Mac handoff): the wiring that gets a real verification contract into a live cycle. Adds an `emit` mode to `scripts/dev/contract_gate.py`; seeding then rides entirely on existing rails.

```
python scripts/dev/contract_gate.py emit --out verification_contract.yaml
squadops artifacts ingest --project group_run --type verification_contract --file verification_contract.yaml
# cycle create with execution_overrides {"contract_ref": "<artifact_id>"}
```

`emit` refuses to write a contract that fails the 98.1 linter, and prints the contract's `content_hash` — **the frozen hash the N=5 yield baseline measures against** — plus `skeleton.interface_manifest_hash`. Against the canonical group_run manifest today: 6 criteria, deterministic output.

## The design decision (recorded in the handoff doc)

The handoff left two options open: operator-seed vs. expander-emits-at-runtime (at the `_seed_skeleton_artifacts` seam). **Runtime emission cannot work**, for two independent reasons:

1. **Ordering:** framing consumes the contract's criteria index at *proposer dispatch* (`_inject_contract_inputs` → dev/qa proposers bind instead of authoring). A contract derived at the impl-run seam — or even at the framing→impl transition — arrives after the plan was authored, so `validate_criteria_refs` would reject every such plan (authored-on-covered). The contract must exist before framing runs.
2. **Doctrine + measurement:** SIP-0098 §2's core lesson is that verifiers are *curated artifacts, not per-run emissions* — a contract re-derived per-roll from framing's own manifest is a per-run emission with a vacuous hash binding. The frozen-baseline hash requires one pre-existing contract; `_validate_contract_binding` then does its real job: rejecting any roll whose framing manifest drifts from the frozen interface (canonical-JSON hash, so formatting variance is tolerated, semantic drift is not).

Since emission is deterministic, a contract emitted from a gate-green tree **is** the CI-validated contract — no separate validation path needed at seed time (lint still runs as a belt-and-braces refusal gate).

## Tests

`tests/unit/capabilities/test_contract_gate_emit.py` (subprocess-driven, the script has no other harness): emitted file byte-identical to the canonical emitter (a divergent writer would seed an unvalidated contract), printed hash matches the file bytes (a mismatch would freeze a meaningless baseline hash), unreadable manifest writes nothing. Capabilities domain green (1244 passed); all three existing gate modes untouched.

## Next (not in this PR)

With this merged: image rebuild from current main → live emit+ingest+seed → shakedown bind-mode roll (probe rows in evidence = the slice-1 live validation) → N=5 yield baseline against the printed frozen hash.

### DIFF
--- a/docs/plans/SIP-0098-verification-contracts-plan.md
+++ b/docs/plans/SIP-0098-verification-contracts-plan.md
@@ -50,14 +50,21 @@ and **(C)** contract-criterion coverage accounting (`criteria_verified`/`criteri
 **Spark pick-up: 98.5 slice 3 (measurement) — do in this order:**
 1. **Preconditions:** #433/#434 landed; UPS question resolved/risk-accepted; **rebuild agent +
    runtime-api images from `756f692` or later** (restart alone does not pick up code).
-2. **Contract-seeding wiring** (the one unbuilt piece): live cycles have no path that sets
-   `contract_ref` yet. `_load_contract_for_run` retrieves it from the artifact vault via
-   `execution_overrides.contract_ref` (bare id or single-element list). Either operator-seed
-   (ingest the emitted `verification_contract.yaml` as an artifact, set `contract_ref` on the
-   cycle) or build the expander-emits-at-runtime path (emit contract beside the skeleton at the
-   `_seed_skeleton_artifacts` seam — a small PR; keep it data-driven, no flag). Emit the contract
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #495, opened 2026-07-18
fix(#494): bind mode requires a framing-emitted interface manifest

jasonLadd:
Shakedown finding: in bind mode the pre-gate net silently skipped skeleton-binding checks when framing emitted no interface manifest — a correctly-bound plan would have run unscaffolded while claiming contract verification. Now records a #473-style system rejection naming the missing manifest. Tests cover the false-pass shape and the inverse guard (manifest present + bound plan still gates normally, using the real emitter against the canonical manifest). Cycles+events domains green (1949 passed).

Companion (non-code) fix: the shakedown recipe now sets plan_authoring_contributors — the production framing path rides the multi-role proposers.

Fixes #494


jasonLadd (commented):
reviewed

PR #497, opened 2026-07-18
fix(#496): bind mode seeds the canonical interface manifest

jasonLadd:
## Problem

Shakedown attempt 2 (`cyc_59bfbd3f4a2a`) surfaced a design contradiction: bind mode required the framing run to **re-derive** the interface manifest and hash-match the frozen contract — from PRD v0.4, which deliberately no longer specifies interface details (the §6.7 product/interface split). The multi-role framing worked exactly as designed (manifest emitted, criteria bound, both nets fired) and still rejected on pure re-derivation naming drift (`729d2a…` ≠ `2eab74…`; `RunCreate` vs `RunEventCreate`, `base_path` variants). The task as posed is unwinnable — every bind-mode roll rejects the same way.

## Fix

In bind mode the manifest already exists — the contract was emitted *from* it. So bind mode **seeds** it, never re-authors it:

1. **`planning_tasks._scaffold_section`**: returns `""` when `contract_criteria_index` is present in inputs (bind mode). The emission instruction is author-mode only — where the manifest is being born.
2. **Executor pre-gate** (`_reject_invalid_plan_before_workload_gate`): when framing emits no manifest, falls back to an operator-seeded `interface_manifest.yaml` found in `execution_overrides.plan_artifact_refs` — the exact rail `_load_interface_manifest_for_run` already reads at the implementation run, so the manifest the gate hash-checks is the one the skeleton expands from. The #494 rejection now fires only when **neither** source exists.

Precedence and rigor preserved: a framing-emitted manifest (if one appears anyway) still takes precedence and drift still rejects (§10 stale binding). The seeded manifest is hash-checked too — a stale seed rejects identically. Author mode is byte-identical to today.

## Tests

- Seam (`test_executor_contract_binding.py`): seeded-manifest-satisfies, seeded-manifest-is-hash-checked (stale seed rejects), drifted-framing-manifest-rejects-despite-matching-seed (precedence).
- Real `execute_cycle` (`test_execute_cycle.py`): seeded manifest + bound plan gates normally with no framing-emitted manifest; the #494 neither-source rejection updated to the new message.
- Appendix gating (`test_interface_manifest_framing.py`): dev proposer + scaffoldable stack + criteria index → `""`, renderer never invoked.

Pre-existing, unrelated: 2 failures in `test_verification_contract_runner.py` on this box (`missing_tooling` — host lacks Node; fails identically on clean main; will raise separately).

Fixes #496

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -1348,8 +1348,17 @@ class DispatchedFlowExecutor(FlowExecutionPort):
         for the ``interface_manifest.yaml`` a scaffolded framing run emitted (99.2) and
         parses it. Returns ``InterfaceManifest`` or ``None`` — a missing or unparseable
         manifest leaves the run on today's non-scaffolded path (graceful, data-driven).
+
+        Never loads for a framing run (#496): an operator-seeded manifest sits in
+        ``plan_artifact_refs`` from cycle creation, visible to EVERY workload — but the
+        skeleton is the implementation substrate. A framing run authors the plan and
+        must not expand or carry skeleton files. (Pre-#496 this was unreachable:
+        forwarding only populated ``plan_artifact_refs`` after framing completed.)
         """
         from squadops.capabilities.scaffold import InterfaceManifest
+
+        if getattr(run, "workload_type", None) == "framing":
+            return None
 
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #505, opened 2026-07-18
Shakedown-3 findings: four fixes that raise genuine roll-success odds

jasonLadd:
Four fixes from shakedown-3 (cyc_2f6adb235041) triage — each targets a defect that either burned budget, hid a failure, or blinded a net that would have repaired the roll in-run. One commit per issue.

## fix(#502) — fenced_parser strategy 6
A fence whose first body line is the **bare file path** (no comment marker, no colon) matched no strategy; the live m009 qa.test response was dropped whole and burned a full retry (~10 min of 27b inference + retry budget). New strategy accepts a lone path-shaped first line under the same guard as #470.

## fix(#501) — tests_pass false green
`run_fullstack_tests` hardcoded exit 0 whenever the backend suite didn't execute — a failing vitest run (exit 1, "No test files found") merged into "all tests passed / exit 0" and `tests_pass` verified on **zero executed tests**. Now: backend controls when executed (D13 unchanged), else the frontend controls, else nonzero. An honest failure here triggers the in-run correction loop instead of shipping unverified code.

## fix(#503) — AST typed checks dead in live cycles
Both evaluation call sites read `resolved_config["stack"]`, which no config path ever sets — so `endpoint_defined`/`field_present` silently skipped in every live cycle (CI passed them via an explicit stack). This is the net that would have caught shakedown-3's `date_time`-vs-`datetime` routes drift at develop time. One derivation seam: `resolve_check_stack` (explicit key wins → `build_profile` mapping → None). Conservative mapping: only `fullstack_fastapi_react → fastapi` for now.

## fix(#500) — evidence fidelity
`failed` entries were bare names; a failed contract probe left no reason in evidence (the 422-vs-200 required manual reproduction). `failed_detail` now carries producer reasons (additive; pre-#500 stored rows reconstruct unchanged), and run_report renders them inline. Also found en route: the 98.4 `criteria_verified`/`criteria_total` fields were **never serialized** to Postgres — stored summaries reconstructed to 0/0 coverage, which would have undercounted the N=5 baseline yield. Now round-trips.

## Not fixed here (by design)
- #504 fill discipline — design rung, needs an enforcement-location decision first.
- qa source-regex lottery — deliberately unfixed; the baseline measures it.

Tests: 18 new across the four suites; full capabilities+cycles sweep green (the 2 known #498 env-dependent failures excepted).

Closes #500
Closes #501
Closes #502
Closes #503

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -33,6 +33,7 @@ from adapters.cycles.pulse_boundary_runner import PulseBoundaryRunner
 from adapters.cycles.run_completion import RunCompletion, resolve_terminal_outcome
 from adapters.cycles.task_dispatcher import TaskDispatcher
 from adapters.cycles.task_naming import build_task_name
+from squadops.cycles.acceptance_evaluation import resolve_check_stack
 from squadops.cycles.agent_config import build_agent_resolver
 from squadops.cycles.build_completeness import compute_missing_required_files
 from squadops.cycles.checkpoint import RunCheckpoint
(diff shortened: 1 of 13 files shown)


jasonLadd (commented):
reviewed

PR #513, opened 2026-07-19
fix(#508): coverage denominator comes from the bound contract, not dispatched checks

jasonLadd:
## Problem

Both night measurement rolls (run_689c2bb32a14, run_57807c247bb4) reported overstated criteria coverage: `criteria_total` was derived from whichever check rows happened to carry criterion ids, so a run that died early reported 2/3 where the truth against the seeded contract (art_97864f4daa5f, 6 criteria) was 2/6. Coverage inflated exactly when the run failed early — the case where the honest number matters most.

## Fix

- `aggregate_verification()` gains `contract_criteria` (keyword, default empty): when supplied, `criteria_total = contract ∪ observed` — every declared criterion counts in the denominator whether or not its checks dispatched; rogue evidence ids are never dropped.
- `RunCompletion.finalize()` accepts the bind-mode contract; `_aggregate_verification` extracts `criterion_ids()`.
- The executor threads the contract it already loads for bind mode into the finalize call (pre-initialized so an early-failing run stays safe).
- Author mode (no contract) is byte-identical to before.

## Tests

- Pure layer: full-denominator with partial evidence (2/6 not 2/3), author-mode unchanged, rogue-id union, zero-evidence 0/m (would each have caught the shipped bug).
- Adapter seam: real `VerificationContract` from the group_run example manifest through `RunCompletion._aggregate_verification` — catches the loaded-but-never-threaded gap specifically.
- `tests/unit/cycles/` + `tests/unit/capabilities/`: 3029 passed; the 2 failures in `test_verification_contract_runner.py` are pre-existing #498 (host-PATH), reproduced on clean main.

Part of the metric-integrity batch (#508/#509/#510/#512) preceding the canonical N=5 baseline. Runs on Spark per lane discussion — patches are lane-shared.

Closes #508

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -216,6 +216,7 @@ class DispatchedFlowExecutor(FlowExecutionPort):
         ledger = RunLedger()
         cycle = None
         plan = None
+        verification_contract = None
         # SIP-0089 §3.5 (#233): agents this run transitioned ambient→cycle, to be
         # returned to ambient (releasing their cycle lease) in the finally. Stays
         # empty when recruitment defers (admission rolls its own recruits back) or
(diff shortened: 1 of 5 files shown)


jasonLadd (commented):
reviewed

PR #514, opened 2026-07-19
fix(#510): failed test suites disclose exit-code meaning in the check reason

jasonLadd:
## Problem

Both night measurement rolls' only required failure read `failed_detail: [{"reason": "", "check_id": "tests_pass", ...}]`. Diagnosing that the real cause was pytest exit 5 — zero tests collected — required manually excavating artifacts (the run's own failure_analysis eventually said it, but the evidence row an operator reads first said nothing).

## Fix

`_tests_pass_from_result` (the single synthesizer both qa paths funnel through — failure-only dict rows are deliberately skipped in its favor) now composes a reason on the FAILED path: `exit_code N`, the documented pytest meaning when known (`5: no tests collected` being the load-bearing one), and the runner's own summary. Passing rows remain reason-less. Not-executed rows already disclosed via #500.

## Tests

Exact-value reason assertions for exit 5 with summary, unknown exit codes, reason-less pass, and an end-to-end flow through `aggregate_verification` into `failed_detail` (the exact row that read empty in run_57807c247bb4). Full `tests/unit/cycles/` + `tests/unit/capabilities/`: 3027 passed (2 pre-existing #498 failures, reproduced on clean main).

Part of the metric-integrity/disclosure batch (#508 = PR #513, #509, #512) preceding the canonical N=5.

Closes #510

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SEravWMKL7HCQ75GeB7fRG

PR #515, opened 2026-07-19
fix(#512): probe boot failures disclose exit state and stderr tail

jasonLadd:
## Problem

Both night measurement rolls reported `vc-probe-runs` unverified with only "subject did not boot" — the uvicorn stderr that would explain *why* was piped to DEVNULL. The night's remaining unknown (why the app never boots) is exactly this gap.

## Fix

- `_boot` spools subject stderr to an unbounded temp file (a PIPE would deadlock a chatty subject at ~64KB; DEVNULL destroyed the diagnosis channel). Closed at teardown; on launch OSError the spool is closed before re-raising.
- On readiness timeout, the disclosed reason becomes `subject did not boot (exited N | no ready response within Ts): <bounded stderr tail>` — crash-on-start and hung-but-alive are now distinguishable, and the tail (last 4KB read, 500 chars kept, whitespace collapsed) carries the traceback.
- The stable `"subject did not boot"` prefix is preserved for downstream matching.

## Tests

Live subprocess tests: crash-on-start carries `(exited 3)` + the stderr message; hang carries the timeout phrasing; a 100KB stderr spew stays bounded under 600 chars. Existing boot-failure tests (real uvicorn on an empty workspace) still pass — now with informative reasons. Full `tests/unit/capabilities/` + `tests/unit/cycles/`: 3026 passed (2 pre-existing #498 failures).

Part of the metric-integrity/disclosure batch (#513, #514) preceding the canonical N=5.

Closes #512

### DIFF
--- a/src/squadops/capabilities/handlers/probe_runner.py
+++ b/src/squadops/capabilities/handlers/probe_runner.py
@@ -25,9 +25,11 @@ qa.test handler wraps it in a thread.
 from __future__ import annotations
 
 import contextlib
+import os
 import socket
 import subprocess
 import sys
+import tempfile
 import time
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #516, opened 2026-07-19
fix(#509): unbound contract criteria attach to the tail qa.test at dispatch

jasonLadd:
## Problem

Criterion binding varied per roll on the identical frozen contract: `criteria_total` went 3 → 1 → 1 across the night rolls, and checks that executed-and-passed earned no criterion credit. Root cause: all three bind-net rejection classes key on **exact string matches** between LLM-authored plan fields (`expected_artifacts`, typed-check `file` params) and contract fill paths. A plan that names `backend/routes.py` any other way binds nothing for it and sails through validation — coverage then measures plan-authoring luck.

## Fix

`generate_task_plan` computes the **residue** — every typed contract criterion no plan task bound — and resolves it onto the tail `qa.test` envelope (assembled workspace, criterion-stamped `TypedCheck`s via the existing `resolve_contract_refs` materializer). The contract, not the plan, is now the last word on verification: every typed criterion produces an evidence row with its `criterion_id` on every roll. Bound criteria keep item-level attachment (repair locality); the nets stay as-is (belt) — this is the deterministic floor (suspenders).

**Scope note:** behavioral criteria (`frontend_build`/`tests_pass` vocabulary) are deliberately outside `resolve_contract_refs` (they execute via the qa/build handlers); stamping their criterion ids onto those rows is the follow-up slice once #513's contract-threading merges — together with #513 that gets `criteria_coverage` to a true n/6.

## Tests

Residue attachment: full typed set on an unbound plan, exclusion of bound refs, last-qa.test-only placement, author-mode/no-qa no-ops — all against the real two-fill-file contract fixture. Full `tests/unit/cycles/` + `tests/unit/capabilities/`: 3027 passed (2 pre-existing #498).

Addresses #509 (denominator side in #513; behavioral stamping tracked there before N=5 is canonical).

### DIFF
--- a/src/squadops/cycles/task_plan.py
+++ b/src/squadops/cycles/task_plan.py
@@ -562,7 +562,45 @@ def generate_task_plan(
             "no default (a builder must not assemble for an assumed stack)."
         )
 
+    _attach_unbound_contract_criteria(envelopes, plan, contract)
+
     return envelopes
+
+
+def _attach_unbound_contract_criteria(
+    envelopes: list[TaskEnvelope],
+    plan: ImplementationPlan | None,
+    contract: VerificationContract | None,
+) -> None:
+    """Attach every contract criterion no plan task bound to the tail qa.test (#509).
+
+    The bind nets key on exact string matches between LLM-authored plan fields
+    (``expected_artifacts``, typed-check ``file`` params) and contract fill paths,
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #519, opened 2026-07-19
fix: behavioral check rows are stamped with their contract criterion ids

jasonLadd:
## Problem

The #516 follow-up: behavioral criteria (`frontend_build`/`tests_pass` vocabulary) execute via the qa/build handlers, outside `resolve_contract_refs`, so their evidence rows carried no `criterion_id`. `vc-frontend-builds` and `vc-suite-passes` could never be credited (or go adverse) — even a fully green roll capped coverage at the typed subset, and the night rolls' passing `frontend_build` earned no criterion credit.

## Fix

`_stamp_behavioral_criteria` (module-level, run_completion adapter): builds check-name → stable-id from the bound contract's behavioral section and stamps matching rows that don't already carry a criterion id, just before aggregation. Adapter-side deliberately — the pure choke point stays producer-agnostic (#513's layering). Probe rows keep their own ids; author mode is untouched.

With #513 (denominator) + #516 (typed residue) + this, `criteria_coverage` is a true n-of-contract: a green roll can reach 6/6, a failed suite drives `vc-suite-passes` adverse.

## Tests

Real example-manifest contract: passing `frontend_build` credits `vc-frontend-builds`; failing `tests_pass` lands `vc-suite-passes` in the denominator but never verified; pre-existing criterion ids are not overwritten; author mode stays unstamped. Full `cycles` + `capabilities`: 3051 passed (2 pre-existing #498).

### DIFF
--- a/adapters/cycles/run_completion.py
+++ b/adapters/cycles/run_completion.py
@@ -129,6 +129,34 @@ def resolve_terminal_outcome(exc: BaseException, run_id: str) -> TerminalOutcome
     )
 
 
+def _stamp_behavioral_criteria(results: Any, contract: Any) -> Any:
+    """Stamp behavioral check rows with their contract criterion ids (#509 tail).
+
+    Behavioral criteria (``frontend_build``/``tests_pass`` vocabulary) execute in
+    the qa/build handlers, outside the ``criteria_refs``→TypedCheck materializer —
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #521, opened 2026-07-19
fix(#520): probe readiness accepts any HTTP response, not only 200 on /health

jasonLadd:
Day-ops fix under the expanded green-odds authority (NOT merged by me — awaiting review).

pf-1 (cyc_794f0698b7e4) delivered the #515 payoff: the probe skip reason showed "Application startup complete. Uvicorn running" — the app booted, but readiness polls GET /health expecting 200 and the group_run PRD declares no /health. Readiness was silently a product requirement; probes could never run for this product.

_wait_ready now treats ANY HTTP response (404 included) as ready — transport-level readiness; the probes assert behavior. Live test: FastAPI app without the ready path → ready via 404 → real probe passes. Full domains: 3052 passed (2 pre-existing #498).

Deployed to the box from this branch for the day-ops rolls; deploy hash recorded in the ops ledger.

Closes #520

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SEravWMKL7HCQ75GeB7fRG

PR #523, opened 2026-07-19
fix: create-probes expect 201 — the contract contradicted the PRD it verifies

jasonLadd:
Day-ops fix under expanded green-odds authority (NOT merged by me).

pf-3 was the first roll where probes executed (thanks to #521's readiness fix) — and the probe failed a **correct** app: POST /runs returned 201 per the PRD (and per QA's own tests), while the emitted probe expects 200. The emitter's probe class is parameterless POSTs — creates by construction — so `expect 200` guaranteed that a PRD-conformant app fails `vc-probe-runs` forever: green was structurally impossible on the old contract.

- `_probes` emits `expect: {"status": 201}` with the rationale in-line.
- The reference fill's create route gains `status_code=201` (it was PRD-divergent the same way — worth noting the CI probe gate was green only because fill and contract shared the same wrong status).
- Corrected contract re-emitted (lint GREEN), ingested as `art_7c7d9ca2e54a` — the day-ops baseline from pf-4 on; old `art_97864f4daa5f` retired.

Tests: scaffold-contract expectation updated to 201; contract-runner suite green except the 2 pre-existing #498 host-PATH failures.

### DIFF
--- a/src/squadops/capabilities/scaffold_contract.py
+++ b/src/squadops/capabilities/scaffold_contract.py
@@ -185,7 +185,12 @@ def _probes(manifest: InterfaceManifest) -> list[dict[str, Any]]:
                 "id": f"vc-probe-{_slug(ep.path) or 'root'}",
                 "subject": "backend",
                 "request": {"method": ep.method, "path": ep.path, "json": json_body},
-                "expect": {"status": 200},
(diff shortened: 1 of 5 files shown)


jasonLadd (commented):
reviewed

PR #525, opened 2026-07-20
fix(#522): system plan-validation rejection re-rolls framing instead of killing the cycle

jasonLadd:
## Problem

A `system:plan_validation` gate rejection killed the cycle with **no implementation run**. In the overnight green-hunt, **3 of 9 rolls (pf-2, pf-5, pf-9) died this way** — every time, framing authored a `regex_match` on a source file, the #464 net correctly rejected the plan, and the cycle just... stopped. The rule the model tripped is already stated verbatim in the proposer prompts, so it's a stochastic fault, not a fixable-by-prompting one — exactly the retry a human operator would grant in one click.

## Design (the non-obvious part)

Naive re-rolling collides with a core invariant: the **positional run↔workload mapping** (#257/D14) assumes exactly one non-cancelled run per workload position. Creating a second framing run would make `_starting_workload_index` (resume) and the advancement duplicate-guard (`len(non_cancelled) > i+1`) miscompute — the 2nd framing run would be mistaken for the implementation run.

So the re-roll **cancels the rejected framing run** (its rejection persists in `gate_decisions` as evidence, #473) before creating a fresh one. Cancelled runs are already excluded from positional indexing, so the invariant holds. The workload `for` loop became a `while` so a re-roll can re-process the same index (`continue` without the tail `i += 1`).

## Safety

- **Opt-in, zero default behavior change**: gated on `framing_max_rerolls` (default `0`) — the whole path is inert until a profile sets it. All 69 pre-existing execute_cycle/binding tests pass unchanged.
- **Human rejections untouched**: a `--reject` carries a different `decided_by` and reaches a different code path (the gate-poll REJECTED branch), never this one.

## Tests (through the real `execute_cycle`, #464/#473 pattern)

reroll-then-advance · budget-exhausted-kills-cycle (no impl run) · default-zero-unchanged · non-framing-never-rerolls · **resume-safety** (cancelled attempt preserves the positional index → impl still resolves to index 1). Full `cycles` domain green except the 2 pre-existing #498 host-PATH failures.

## Deploy note

Framework/executor change → needs a runtime-api + agent rebuild to take effect live, and a profile/launcher setting `framing_max_rerolls` (I'd suggest 2). Not deployed yet — flagging for when you resume the hunt.

Closes #522

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -514,7 +514,14 @@ class DispatchedFlowExecutor(FlowExecutionPort):
             forwarding_overrides = await self._rebuild_forwarding_overrides_on_entry(
                 cycle, cycle_id, start_index
             )
-        for i in range(start_index, len(workload_sequence)):
+        # #522: framing re-rolls on a system plan-validation rejection. `while`
+        # (not `for`) so a re-roll can re-process the SAME workload index — the
+        # re-roll path `continue`s without the tail `i += 1`. Zero re-rolls
+        # (default) is byte-identical to the old `for` loop.
+        framing_rerolls = 0
+        max_framing_rerolls = cycle.applied_defaults.get("framing_max_rerolls", 0)
+        i = start_index
+        while i < len(workload_sequence):
             workload_entry = workload_sequence[i]
             await self.execute_run(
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #526, opened 2026-07-21
fix(#524): probe request bodies carry type/name-appropriate sample values, not "x"

jasonLadd:
## Problem (now demonstrated, not hypothetical)

The probe emitter put `"x"` in every required request field. An app that validates its inputs — e.g. parses a field named `datetime` as a real datetime — 422s the bogus `"x"`, so the probe fails a **correct** app.

**pf-12 is the proof.** That roll's app passed its own pytest suite (`vc-suite-passes` verified — first suite convergence of the arc), built the frontend, and implemented every route correctly. The *only* failure was `vc-probe-runs: status 422 != expected 201`, because the app validated the `datetime` field and rejected `"x"`. This also resolves the ambiguity #524 was filed with: the "app over-validates" horn is refuted — the app validates *appropriately* (it passed its suite), so the probe is what's wrong.

It was also intermittent by app luck: pf-4/pf-8's apps were lax enough to accept `"x"` (probe passed); pf-7/pf-12's validated (422). The fix removes the coin-flip.

## Fix

`_probe_sample_value(name, type)` generates a plausible value: semantic **name** hints win over declared type (a string-typed `datetime` still gets an ISO value — the manifest calls datetime a string for MVP, but real apps parse it), then int/float/bool/list by type, plain string default. `datetime` now emits `2026-08-01T08:00:00`, which both validating and non-validating apps accept → 201.

## Tests

Parametrized generator coverage (datetime/`*_at`/timestamp → ISO, email, url, int, float, bool, list, default), the name-beats-string-type case, and the emitted group_run probe body asserting `datetime == "2026-08-01T08:00:00"`. Full regression green locally (5218 passed).

## Deploy note

Changes the contract hash → **re-emit + re-ingest a new frozen contract** for live rolls (host-side, no rebuild). This is the last known blocker between the current system and a green roll (pf-12).

Closes #524

### DIFF
--- a/src/squadops/capabilities/scaffold_contract.py
+++ b/src/squadops/capabilities/scaffold_contract.py
@@ -169,17 +169,52 @@ def _coverage_expectations(manifest: InterfaceManifest) -> list[str]:
     return out
 
 
+def _probe_sample_value(field_name: str, field_type: str) -> Any:
+    """A plausible sample value for a probe request field (#524).
+
+    ``"x"`` for every field failed apps that validate their inputs: a field named
+    ``datetime`` typed ``string`` still gets parsed as a datetime by a reasonable
+    app, which then 422s the bogus ``"x"`` — penalizing a *correct* app for
+    validating (pf-12: the app passed its own suite yet failed the probe on
+    "status 422 != expected 201"). Semantic name hints win over the declared type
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
Reviewed

PR #527, opened 2026-07-21
feat: deterministic interface-drift diagnosis feeds the correction repair (piece 1)

jasonLadd:
## Problem

When an app renames an interface identifier the manifest pins — a model field or a route path — the correction loop only saw an opaque check status (`status 404 != expected 201`). `analyze_failure` (pure LLM) guessed from post-extraction artifacts, mislabeled it, and the repair regenerated broadly and thrashed. pf-13 and pf-15 both died partly this way.

## Design: deterministic diagnosis, LLM reconciliation

`interface_conformance.detect_interface_drift(manifest, {path: content})` AST-parses the app's model/route files and diffs identifiers against the manifest. Per drifted file it reports the **offending set** (identifiers the app declares that the interface doesn't define) plus the **interface targets** — and a precise instruction. The agent renames by meaning.

**Set-report, not 1:1 rename** — because the real failures drift *several* identifiers at once. pf-13 renamed **both** `location`→`meeting_location` *and* `pace_target`→`pace`; a single-mismatch detector (my first cut) emits nothing on that, which the replay caught before any wiring. So the deterministic part reports the sets; the LLM does the meaning-based mapping it's good at and code can't do safely.

The finding is injected into `failure_evidence` and surfaced at the **top of the repair prompt** as authoritative ("INTERFACE CONFORMANCE — apply exactly").

## Wiring

The executor threads the already-loaded interface manifest + the enriched envelope's materialized workspace (`artifact_contents`, #456) through `_execute_sequential` → `_handle_task_outcome` → `run_correction_protocol`. **Additive and opt-in-by-data:** a clean workspace, absent manifest, or unparseable file injects nothing — today's path is byte-identical. No fragment change (rides the existing `failure_summary` variable).

## Validation — replay, not a new cycle

Verified against pf-13's **real stored artifacts**: the detector reports `extra={meeting_location, pace} missing={location, pace_target}`, and one isolated call to the 27b dev model (qwen3.6:27b) with the surgical instruction renamed **both** fields correctly, producing a model that matches the interface exactly. Deterministic input, reproducible, ~one model call vs a 3-hour cycle. The replay also confirmed pf-15's routes conform (its 404 was never name drift — separate cause, out of scope).

## Tests

Detector: multi-drift (the pf-13 case), single drift, clean, `model_config`/dunder exclusion, class-name-drift skip, route path drift, conforming routes, None/empty/unparseable robustness, plus a real-group-run-manifest case. Full regression green (5228 passed).

## Deploy note

Executor/handler change → needs **rebuild + restart** to run live (and per the #522 lesson I'll verify in-container, not trust the file). Not deployed yet.

Scope: fields + routes, set-report. Out of scope: agent drift-memory (piece 2), pf-15's non-name-drift 404, `#522`-live.

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -256,6 +256,8 @@ class CorrectionRunner:
         plan_delta_refs: list[str],
         profile: Any = None,
         flow_run_id: str | None = None,
+        interface_manifest: Any = None,
+        artifact_contents: dict[str, str] | None = None,
     ) -> CorrectionProtocolResult:
         """Run the correction protocol: analyze → decide → act.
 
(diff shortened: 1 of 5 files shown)


jasonLadd (commented):
reviewed

PR #528, opened 2026-07-21
fix(fenced_parser): recover path:-labelled headers and unterminated-at-EOF fences

jasonLadd:
## What

Two fenced-parser malformation classes were silently dropping **whole recoverable files** into `build_warnings.md`, so the correction loop then burned a repair attempt regenerating output that was never wrong — a formatting miss misclassified as a work-product failure.

Both classes were **replay-diagnosed** from real stored `build_warnings.md` artifacts (no new cycles), and the fix is verified by replaying those same bytes.

### 1. Redundant `path:`/`file:` label in a strict header
````
```python:path:backend/main.py
````
Strategy 1 partitions on the first colon → filename `path:backend/main.py`; the residual colon trips `_path_is_safe` and the file is lost. `_strip_path_label` removes the label. **Strictly additive** — a conforming filename never starts with `path:`/`file:`, and a real path can't contain a colon anyway, so healthy headers and the existing colon-rejection security tests are untouched.

### 2. Unterminated fence at EOF
The model drops only the closing ```` ``` ````, leaving a valid header and a complete body running to end-of-response. `_is_implicit_eof_close` infers the close **only when the remainder holds no further fence markers** — so a close merely missing *between* two files can never swallow the next file's content into this one (that ambiguous case is still abandoned, as before).

Also extracted the strategy-1..6 resolution into `_resolve_block` to keep `extract_fenced_files` under the C901 threshold after the added branch — no behavior change.

## Replay evidence (real artifacts)

| sample | shape | before | after |
|---|---|---|---|
| `cyc_40dd251ff9b8` | two `path:`-labelled blocks, both closed | 0 files | **2 files** (676 + 7802 chars) |
| `cyc_3e66234ad996` | strict header, unterminated at EOF | 0 files | **1 file** (3530 chars, complete) |
| `cyc_571c5ea5a426` | 0 bytes | 0 files | **0 files** (correctly stays unrecovered — empty output is a real failure) |

## Tests

`tests/unit/capabilities/test_fenced_parser.py`: rewrote `test_no_closing_fence` to the new recovery contract, added `TestPathLabelHeal` (label heal, two-file, no-regression on plain headers and `file`-prefixed basenames) and `TestImplicitEofClose` (recovery, compose-with-label, **no-swallow guard**, empty-stays-unrecovered, untagged-fence-still-dropped). 70 pass; full `tests/unit/capabilities/` suite (1284) green; ruff clean.

## Scope

Refs #431 — reduces the extraction loss that #431's diagnosis blindness misclassifies as work-product failure; does **not** implement #431's raw-vs-stored evidence rule (separate change). Refs #430 (prior nested-fence recovery in the same parser, closed).

### DIFF
--- a/src/squadops/capabilities/handlers/fenced_parser.py
+++ b/src/squadops/capabilities/handlers/fenced_parser.py
@@ -6,7 +6,10 @@ and QATestHandler to parse multi-file output.
 
 Recognized formats, in priority order (most explicit first):
 
-1. Strict tagged fence — ``` ``` <language>:<filepath>``` ```
+1. Strict tagged fence — ``` ``` <language>:<filepath>``` ```. A redundant
+   ``path:``/``file:`` label wedged before the path (#431 —
+   ``` ```python:path:backend/main.py``` ```) is stripped; the residual colon
+   would otherwise fail the safety check and drop the whole file.
 2. Filename in language slot for special names — ``` ``` Dockerfile``` ```,
    ``` ``` Makefile``` ```, etc.
 3. Filename heading immediately preceding the fence — ``#`` through ``######``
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #530, opened 2026-07-21
fix(plan-validation): tolerate warning/info-severity criteria violations instead of rejecting the plan

jasonLadd:
## What

A single over-eager *authored* acceptance check was hard-rejecting an entire framing plan and **killing the cycle** — even when the check was `severity: warning`. That's inconsistent with RC-9, which says only `severity=error` checks can fail a build at execution: a warning check *can't block a build*, yet at **plan validation** it detonated the whole cycle. This is what killed pf-18 after ~45 min of framing, on one `regex_match` (severity `warning`) the QA model tacked onto a frontend smoke-test task.

## Change

Split the two **per-criterion structural** rules by severity:
- `validate_criteria_scope` (#464 regex-on-source) and `validate_criteria_refs` **rule 3** (authored-on-contract-covered, §6.3) now emit a **fatal** error only when the criterion is `severity=error`.
- Rules that are **not** soft authoring slips — unresolvable `criteria_ref`, missing-coverage (silent descoping) — stay **always fatal**.
- New `ImplementationPlan.soft_criteria_violations()` surfaces the tolerated warning/info violations; the executor logs them at the gate seam so the tolerance is **never silent**.

Detection is shared between the fatal check and the soft collector via two generators (`_regex_on_source_criteria`, `_authored_on_covered_criteria`) so the two paths can't drift.

## Scope boundary

The tolerated check **stays in the plan artifact** but is inert (warning → can't block per RC-9). Physically stripping it from the artifact is a deliberate **follow-up**, not this change (agreed: confirm this opens the path to green first).

## Verification (replay of pf-18's real plan, no new cycle)

Ran pf-18's actual rejected `implementation_plan.yaml` (`art_34c6cda17e76`) through the patched validators against the seeded contract (`art_fdbb80b2667a`):

| | before | after |
|---|---|---|
| fatal `validate_criteria_scope` | 2 | **0** |
| fatal `validate_criteria_refs` | 2 | **0** |
| tolerated soft violations surfaced | — | **4** (all `severity=warning`) |
| same regex bumped to `severity=error` | — | **still rejects** (2 + 2) |

The cycle would have proceeded past framing instead of dying.

## Tests

`tests/unit/cycles/test_implementation_plan.py` (scope): warning/info regex-on-source tolerated + surfaced; explicit `error`-severity still rejects; document regex never a violation at any severity. `test_criteria_ref_binding.py` (bind mode): warning authored-on-covered tolerated + surfaced; `error`-severity still rejects. Full `tests/unit/cycles/` suite: **1811 pass**, my new tests green. (Two pre-existing `test_verification_contract_runner` failures — `missing_tooling` on this box — fail identically on clean `main`, unrelated.)

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -2356,6 +2356,7 @@ class DispatchedFlowExecutor(FlowExecutionPort):
         errors: list[str] = []
         interface_content: str | None = None
         parsed_plan: ImplementationPlan | None = None
+        contract = None  # set in bind mode below; feeds the soft-violation log
         for ref_id in tuple(run.artifact_refs or ()):
             try:
                 ref, content_bytes = await self._artifact_vault.retrieve(ref_id)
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #532, opened 2026-07-21
fix(#531): retarget patch-path repair onto the drifted source, not the failed check's tests

jasonLadd:
## What

On the correction **patch** path, the repair target was hard-anchored to the **failed check's own task**. The `tests_pass` check lives on the qa test task, so a `tests_pass` failure whose *cause* is a drifted source file got "repaired" by regenerating the **tests** (the symptom) while the drifted source (the cause) was never rewritten — the loop re-diagnosed the same drift every attempt and never converged. This is the dominant `tests_pass` non-convergence cause.

## Evidence — pf-19 (cyc_3632da190fd2)

`backend/models.py` drifted to `notes`/`pace` vs the contract's `route_notes`/`pace_target` (+ an unauthorized `GET /`). The **diagnosis was correct** — #527 interface-drift evidence named it, and the correction decision chose `affected_task_types: [backend_models, backend_api_routes]`. But across **3 attempts** the repair emitted only `tests/test_runs.py` + `tests/test_participants.py`; `models.py` was never regenerated and still held `notes`/`pace`.

## Fix

`_resolve_repair_target(failure_evidence, failed_inputs)`: when the correction carries **deterministic** interface-drift evidence (an AST diff — not the free-text `affected_task_types` that once mis-routed `["QA Handoff"]` and made them anchor to the failed task), the drifted files **are** the target → retarget `expected_artifacts` onto them. No drift → today's failed-task anchor, byte-identical. The dev repair role already runs, the current file content reaches the handler via `artifact_refs`, and #527's rename instruction is already in the failure summary — so pointing the target at `backend/models.py` is what's needed.

**No inline prompt content (#448):** the instructional "how" is the drift `instruction`, already a managed asset surfaced into the repair prompt's failure summary as the INTERFACE CONFORMANCE section — so `focus`/`description` are left unset rather than authored in Python.

## Verification (replay, no new cycle)

Unit test reconstructs pf-19's real `failure_evidence` (drift on `models.py` + `main.py`, failed task = the pytest suite) and asserts the repair now targets `[backend/main.py, backend/models.py]` — **not** the test files. Plus no-drift fallback and robustness (None / empty / file-less drift → fall back, never crash). Full `tests/unit/cycles/` suite: **1812 pass** (the 2 `test_verification_contract_runner` failures are a pre-existing `missing_tooling` env gap on this box — identical on clean `main`).

Closes #531.

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -46,6 +46,36 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+def _resolve_repair_target(
+    failure_evidence: Any, failed_inputs: dict[str, Any]
+) -> tuple[list[str], str | None, str | None]:
+    """Choose ``(expected_artifacts, focus, description)`` for a patch-path repair.
+
+    #531: the target defaults to the *failed task's own* artifacts — but the
+    ``tests_pass`` check lives on the qa test task, so anchoring there regenerates
+    the tests (the symptom) while the drifted source (the cause) is never rewritten
+    and the loop can't converge. When the correction carries deterministic
+    interface-drift evidence (an AST diff, not the free-text ``affected_task_types``),
+    the drifted files ARE the target: retarget so the dev repair rewrites the source.
+    No drift → today's failed-task anchor, byte-identical.
+    """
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #533, opened 2026-07-22
feat(checks): add function_defined — a style-immune tool for "file defines functions"

jasonLadd:
## What & why

Green-hunt rolls repeatedly die at the plan-review gate because the squad authors `regex_match` on **source** test files to assert "this file defines test functions":

```yaml
- {check: regex_match, file: backend/tests/test_runs_crud.py, pattern: def test_, count_min: 3}
```

The intent is legitimate ("don't hand me an empty test file"), but there is no sanctioned check for "a source file defines functions named X" — the only check that expresses "file contains pattern" is `regex_match`, which is banned on source (#464, a style lottery). So one such check rejects the entire plan (pf-18 and pf-20 both died here).

This gives the intent a real tool instead of only banning the workaround.

## The change

- **New `function_defined` check** (AST-based, Python; modeled on `field_present`): counts `def`/`async def` (top-level, methods, nested) whose name starts with `name_prefix`, passes at `>= min_count` (default 1). Style-immune — it matches the real function name via the AST, not text.

  ```yaml
  # before (banned, rejected the plan):
  - {check: regex_match, file: backend/tests/test_runs_crud.py, pattern: def test_, count_min: 3}
  # now (sanctioned, and stronger):
  - {check: function_defined, file: backend/tests/test_runs_crud.py, name_prefix: test_, min_count: 3}
  ```

- **Auto-taught to proposers**: the spec's `example` + `notes` render into the typed-acceptance vocabulary (single source, no drift). `regex_match` gains a redirect note, and the #464 rejection message now names `function_defined`.

## Tests

- 8 evaluator tests: meets/below count, methods + async counted, prefix specificity, default count, missing file, syntax error, stack-unset.
- Added `function_defined` to the path-traversal / absolute-path safety param lists.
- The parametrized vocabulary tests auto-cover the new spec's example validity + rendering.
- 3117 passed across cycles + capabilities; lint + `ruff format --check` clean.

## Notes

- Strengthens verification (a real AST check vs a text-regex proxy); nothing is defanged or weakened.
- It does not *force* the model to pick the right tool — it removes the reason to reach for the banned one and steers via the vocabulary + rejection message.
- References #464 (the regex-on-source rule) but does not close it — it adds a sanctioned alternative.

### DIFF
--- a/src/squadops/cycles/acceptance_check_spec.py
+++ b/src/squadops/cycles/acceptance_check_spec.py
@@ -155,8 +155,9 @@ def argv_matches_safelist(argv: list[str]) -> bool:
 # source files prescribe another roll's stylistic choices (quote style,
 # identifier names) and have twice produced criteria unwinnable by correct
 # code; source files are verifiable by the style-immune checks
-# (endpoint_defined / import_present / command_exit_zero) and the
-# behavioral required checks (tests_pass / frontend_build).
+# (endpoint_defined / import_present / field_present / function_defined /
+# command_exit_zero) and the behavioral required checks (tests_pass /
+# frontend_build).
 REGEX_DOCUMENT_SUFFIXES: tuple[str, ...] = (".md", ".txt", ".rst")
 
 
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #534, opened 2026-07-22
fix(correction): target the union of drift files + the failing check's artifact

jasonLadd:
## What & why

pf-21 (`cyc_2aac58b9f03d`) — the first roll to get past framing after #533 — exposed a **non-convergence trap** in #532's repair retargeting.

`tests_pass` failed because of **two** independent bugs: (a) `backend/models.py` interface drift (`run_id`→`id`, `pace`→`pace_target`) and (b) a broken pytest fixture in `backend/tests/test_runs.py` (`client = TestClient(app)` as a module-level var instead of `@pytest.fixture`).

`_resolve_repair_target` returned **only** the interface-drift files when drift was present, **excluding** the failing check's own artifact. So every repair attempt re-patched the *already-correct* `models.py`+`main.py` and **never touched the test file** with the fixture bug. Attempts 00 and 01 both emitted only `models.py`+`main.py` → the loop can't converge and burns the whole correction budget.

This blocks convergence on **any** roll where a failing check's file has its own bug alongside drift.

## The fix

Target the **union** — drift files first (the cause), then the failed task's `expected_artifacts`, de-duplicated preserving order:

- The drifted source is **always** in the target set, so #531/#532's win holds (a drift-caused test failure is still fixed by rewriting the source, not by masking it in the tests).
- The failing check's own file is **also** targeted, so an independent bug there (the fixture) is fixable.

`#532` was right to *include* drift; it just made it *exclusive*. Union is strictly more capable and can't reintroduce the #531 masking problem (the source is always in the set).

## Tests

- `TestResolveRepairTarget` updated to the union behavior: drift source is targeted (asserted first) **and** the failing artifact is included; ordering (drift-first) and `focus/description=None` (#448) preserved.
- Added a dedup test (failing artifact already a drift file → appears once).
- No-drift fallback unchanged. 36 correction_runner tests pass; lint + `ruff format --check` clean.

## Notes

- Follow-up to #531/#532. Verified against the live pf-21 failure.

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -56,23 +56,35 @@ def _resolve_repair_target(
     the tests (the symptom) while the drifted source (the cause) is never rewritten
     and the loop can't converge. When the correction carries deterministic
     interface-drift evidence (an AST diff, not the free-text ``affected_task_types``),
-    the drifted files ARE the target: retarget so the dev repair rewrites the source.
-    No drift → today's failed-task anchor, byte-identical.
+    the drifted files ARE part of the target so the dev repair rewrites the source.
+
+    pf-21 (cyc_2aac58b9f03d): drift is not always the *whole* story — the failing
+    check's OWN artifact can carry an independent bug too (there: models.py drift
+    AND a broken pytest ``client`` fixture in the test file). #532 targeted the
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #536, opened 2026-07-22
fix(RC2): dependency-scoped repair target — no-drift qa.test failures reach the source under test

jasonLadd:
## RC2 — dependency-scoped repair target

The correction loop could only reach **source** files through the interface-drift
signal — `_resolve_repair_target`'s `if drift_files:` union (the #531 → #532 → #534
evolution). A behavioral/runtime bug that produces **no drift** — a missing `/api`
router prefix in `main.py`, a 400-vs-422 validation mismatch — surfaces on the
`qa.test` task, whose own `expected_artifacts` are the **test files**. With no drift,
the target fell back to the failing task's own artifacts, so the repair could only
ever edit the test file, never the buggy source. The loop exhausted every attempt.

### Evidence (pf-24, `cyc_38415226ad82`)
Attempts 00–04 all emitted `backend/tests/test_runs.py`; `main.py` stayed
`app.include_router(routes.router)` (no `/api` prefix) the whole time → the
`vc-probe-runs` 404 never cleared → rejected. The analyze step was *accurate and
advancing* each attempt (RC3 fix working), but the repair had no way to touch the
source.

### Fix — path-scoped implementation surface
When there is **no drift**, extend the repair target with the plan's
`development.develop` source that shares a **top-level package** with a failing
artifact:

- `backend/tests/test_runs.py` failure → `backend/models.py`, `backend/routes.py`,
  `backend/main.py` (reaches the `/api`-prefix fix)
- **never** `frontend/*` — package-scoping bounds the regression blast radius so a
  backend failure can't re-emit frontend source.

`task_plan` threads the source surface (`implementation_artifacts`) onto `qa.test`
envelopes only; `_resolve_repair_target` does the package-scoped union. **Absent the
surface** (author mode, non-build corrections) the target is byte-identical to the
prior #531 fallback, and the **drift-union path (#532/#534) is untouched** — the
extension is confined to the no-drift branch.

### Why path-scoping, not `depends_on`
Verified against pf-24's real plan: `qa.test` tasks carry `depends_on=[]` (the
proposer authors tests as DAG roots), so "target the qa.test's `depends_on`
artifacts" would be a **no-op**. The test→package path convention is the reliable
signal.

### Tests
- 5 `_resolve_repair_target` cases: core RC2 retarget, backward-compat
  byte-identical (no surface), symmetric frontend scoping, drift-path-unchanged,
  dedup.
- 1 wiring test: `qa.test` envelopes carry the surface, `development.develop`
  envelopes do not (guards against a cosmetic fix that doesn't wire up).
- Full `tests/unit/cycles/` domain: **1841 passed**. ruff clean.

Not tied to a filed issue (RC2 is the internal name for this correction-loop gap,
sibling to the #531/#532/#534 target evolution). Deploy held pending review — it
resets the N=5 measurement baseline.

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -46,6 +46,29 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+def _top_level_package(path: str) -> str:
+    """First path segment of a repo-relative artifact path.
+
+    ``backend/tests/x.py`` → ``backend``; a bare filename is its own package.
+    """
+    head, _, _ = str(path).strip().lstrip("./").partition("/")
+    return head
+
+
+def _scope_to_shared_packages(candidates: list[str], anchors: list[str]) -> list[str]:
+    """Keep ``candidates`` whose top-level package matches some ``anchor``'s.
+
+    RC2 (pf-24) blast-radius control: a failing ``backend/tests/…`` test retargets
+    ``backend/…`` source but leaves ``frontend/…`` untouched, so a backend failure
+    can never regress frontend source. No anchors → nothing (the caller falls back
+    to the failed artifacts alone).
+    """
+    anchor_pkgs = {_top_level_package(a) for a in anchors if a}
+    if not anchor_pkgs:
+        return []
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #538, opened 2026-07-23
scaffold: seed a frozen test harness to pin the import root (prototype + proposed SIP)

jasonLadd:
## Scaffolded test harness — prototype (proven) + proposed SIP

Follow-on to **SIP-0099** surfaced by **pf-26** (`cyc_af8800f8943f`): the backend `qa.test`
correction loop never converged — not because of framing, targeting (RC2), or workspace-staleness
(RC3), which all worked, but because of a **package-root inconsistency the squad couldn't settle**.
Files live under `backend/` (a package), the qa test invented `from app.main import app` (no `app`
package), and the correction loop rewrote the *frozen* `backend/main.py` four times without ever
matching the layout. pytest collection crashed on an unresolvable import every attempt → the loop
oscillated and exhausted.

This PR contains **two things**: a **proven prototype** of the smallest piece, and a **proposed SIP**
scoping the rest.

### Prototype (Piece A) — `scaffold.py`
The `fullstack_fastapi_react` expander now emits a **frozen root `conftest.py`** that anchors
`sys.path` and owns the app import behind a `client` fixture. Suites fill bodies against `client`
and never author `from <root>.main import app` — so the package root is a scaffold invariant, not a
per-suite guess.

**Proven deterministically (no cycle, no LLM):**
- **Positive:** a `client`-based qa suite runs against the freshly-materialized skeleton → `1 passed`,
  no `ModuleNotFoundError`.
- **Counterfactual:** the *exact pf-26 pattern* (`from app.main import app`) against the **same**
  skeleton → crashes collection with `ModuleNotFoundError: app`. The delta between converging and
  exhausting is purely the import convention.
- No regression: **1316** capabilities + verification-contract tests pass.

### Finding that shaped the scope (recorded in the SIP)
The build test-runner (`_source_dir_pythonpath`, #303 + #454) **already resolves** the scaffold's
package convention (a package dir puts the *workspace root* on `PYTHONPATH`, so `from backend.main
import app` and relative `from .routes` both resolve — #454 fixed exactly this). So **no runner
change is needed**; the import-*resolution* layer is done. What remains is ensuring the tests *use*
the convention (Piece A, here) and the source keeps using it under repair (Piece B, below).

### Proposed SIP — `sips/proposed/SIP-Scaffolded-Test-Harness-and-Frozen-Enforcement.md`
A strict extension of SIP-0099. Three pieces:
- **A** — scaffold the test harness (prototyped here).
- **B** — enforce frozen-stays-frozen through fill **and repair** (the teeth; pf-26 proves it's not
  optional — the scaffold seeded a correct `main.py` and the repair overwrote it). This is the real
  remaining engineering (correction-loop materialization path).
- **C** — bind the qa proposer to the `client` fixture + hash-pin `conftest.py` in the contract.

Validation is deterministic replay (materialize → `client`-suite passes / pf-26-pattern fails /
repair-clobbers-frozen-and-loses), with a live roll as the integration check only.

### Review note
The prototype (Piece A) is complete and green; Pieces B/C are **proposed, not built**. Merging lands
the scaffold `conftest.py` + its proof and puts the SIP on `main` for design review — it does **not**
change the correction loop. If you'd rather keep the code and the proposal separate, I can split the
SIP into its own PR.

### DIFF
--- a/sips/proposed/SIP-Scaffolded-Test-Harness-and-Frozen-Enforcement.md
+++ b/sips/proposed/SIP-Scaffolded-Test-Harness-and-Frozen-Enforcement.md
@@ -1,0 +1,161 @@
+---
+title: Scaffolded Test Harness and Frozen-File Enforcement
+status: proposed
+author: jladd
+created_at: '2026-07-22T00:00:00Z'
+updated_at: '2026-07-22T00:00:00Z'
+---
+# SIP-XXXX: Scaffolded Test Harness and Frozen-File Enforcement
+
+## Status
+Proposed
+
+**Extends:** SIP-0099 (Contract-First Build Scaffolding). SIP-0099 makes the *interface*
+scaffold-owned and frozen; this SIP closes two gaps that SIP-0099's current surface leaves
+open, both surfaced empirically by the group_run measurement rolls.
+
+**Motivating case:** pf-26 (`cyc_af8800f8943f`, deploy `ee4ac71c`) — the backend `qa.test`
+correction loop never converged. Root cause was **not** the framing, targeting (RC2), or
+workspace-staleness (RC3) fixes that preceded it — those all worked. It was a **package-root
+inconsistency the squad could not settle**: source files live under `backend/` (a package),
+but the qa test invented its own root (`from app.main import app` — no `app` package), and the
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #544, opened 2026-07-23
feat(sip-0100): Phase 2 — authorization spine + live frozen-ownership enforcement

jasonLadd:
SIP-0100 **Phase 2** — the ownership/authorization spine, and enforcement is now **live**: the pf-26 clobber can no longer land on a scaffold-bound run. All green (3170 cycles+capabilities tests, ruff clean, no import cycle).

## 2.1 — authorization seam (`write_authorization.py`)
`WorkspaceOwnership` (permanent, from the 0.3 bound record) + `WriteGrant` (transient, per-producer, resolved before generation) kept separate (review #2/#3). `WriteAuthorization.authorize`/`authorize_response` return the 0.5 reason codes, response-atomic (D5), dup-normalize rejection (D7); `normalize_ws_path` is the single canonical identity authorization + materialization share.

## 2.2 — unify the two materializers
0.1 found TWO materializers, and `test_runner._materialize_files` (the pf-26 pytest workspace) bypassed `materialize_artifacts`. Now one `materialize()` seam both delegate to; with authorization it authorizes the complete set before any write and rejects the whole response on violation.

## 2.3 — D4 integrity primitives
`verify_frozen_integrity` (detects a tampered/missing frozen file) + `restore_frozen_files` (rewrites from the bound record's persisted bytes — D2 authority, never re-expand).

## 2.4 — enforcement is LIVE (restore-at-storage)
The executor builds the bound record per scaffold-bound run and, at the artifact-storage seam (executor-level, no handler threading), **restores** any producer emission of a frozen path to the scaffold bytes — the clobber can't land, recorded not silent. Unbound runs are a no-op (§10).

### Key decision (flagged for review)
**Restore, not response-atomic reject, is the first enforcement.** The current squad still re-emits frozen files (pf-24/26), so a strict reject would break *every* bind-mode build. Restore is non-breaking and correct for pf-26. The stricter reject + targeted correction (§4.6) and the D3 hard-guard are **gated on fill-only being enforced** (§3.4) — a real sequencing dependency, now documented in the plan.

Phase 3 (repair authority + evidence + compliance counter) and Phase 4 (replay/integration) are future phases. The plan is updated with every 2.x outcome.


jasonLadd (commented):
reviewed

PR #548, opened 2026-07-23
feat(sip-0100): Phase 3 Task 3.4a — contract-compliance circuit-breaker

jasonLadd:
## Summary

SIP-0100 Phase 3 Task **3.4a** — the bounded contract-compliance circuit-breaker (Option A, which you signed off). Plus plan-doc bookkeeping: 3.2 and 3.4b marked deferred with rationale.

## 3.4a — circuit-breaker
A run-level `contract_compliance_attempts` counter (default 3, configurable via `applied_defaults`), **separate** from the SIP-0086 convergence counter (D6) — compliance violations don't consume the correction budget, but past the bound the run terminates with a `CONTRACT_COMPLIANCE` `_ExecutionError` (honest terminal state, not a silent green).

**Counts only `unauthorized_slot_emission`** (the 3.1 cross-lane drops), not `frozen_path_emission`:
- A dropped cross-slot write never fails the task through the convergence loop — so this counter is the *only* thing that can stop a producer chronically writing another producer's slot.
- Frozen re-emission is the tolerated 2.4 baseline (restored, and *already* bounded by `max_correction_attempts`); counting it would risk false-terminating a normal run. Excluding it keeps 2.4 non-breaking.

Lives in `_enforce_compliance_budget`, called from the artifact-collection path right after 3.3 evidence is emitted (so the violation that triggers termination is recorded), before storage.

## Deferrals recorded in the plan
- **3.2 (delegation safeguards)** — nothing issues delegations yet (`WriteGrant.delegated` unpopulated; 3.1 only drops never-legitimate QA→dev-slot writes; the router already routes source→dev / tests→QA). A safeguard with no caller; revisit on a real use-case.
- **3.4b (active targeted correction)** — the re-dispatch-with-slot-guidance variant, off the *success* path. Deferred pending 3.3 evidence from a `full` roll showing producers repeatedly violate and would benefit from guidance (weak models may not). The `correction_requested` field + reason codes are already in place for it.

## Tests
7 new (`test_sip0100_compliance_budget.py`): under/over bound, frozen-not-counted, per-file increment, configurable + default bound. **1891 cycles green** (the two `verification_contract_runner` failures are pre-existing/environmental — `missing_tooling` on the dev box).

## Phase 3 status after this
3.1 ✅ · 3.3 ✅ · 3.4a ✅ · 3.2 ⏸️ deferred · 3.4b ⏸️ deferred. Phase 3 is effectively complete; remaining work is Phase 4 (replay + the live `full`-squad integration roll).


jasonLadd (commented):
reviewed

PR #549, opened 2026-07-23
test(sip-0100): Phase 4.1-4.3 — deterministic replay & path/atomicity matrix

jasonLadd:
## Summary

SIP-0100 Phase 4 tasks **4.1–4.3** — the deterministic replay & path/atomicity matrix. Test-only; adds the cases not already covered by the 2.1/2.2 suites (`test_write_authorization.py`, `test_materialize_unified.py`) and the harness-boundary suite (`test_acceptance_checks.py`). No live cycle.

### 4.1 — path/atomicity matrix
- **frozen-only** response rejected wholesale (not just mixed).
- **byte-identical frozen emission is still an attempted-emission** — the subtle D7 case: a producer emitting a frozen path with content *identical* to the scaffold is still recorded as a violation (restore is a content no-op, but the attempt isn't free), distinct from not emitting it (no evidence).
- `./` and `backend/../backend/main.py` aliases resolve onto the frozen file → forbidden.
- **symlink escape** defended (`resolve()` + `relative_to` skips a symlink resolving outside the workspace).
- **symlink-onto-frozen** is documented as *out of the reachable model* — the scaffold flow creates no symlinks and authorization is keyed on the emitted name, so there's no in-flow path to exploit it. Flagging here as a defense-in-depth boundary, not a live gap.

### 4.2 — pf-26 replay
A QA correction emitting frozen `backend/main.py` beside its corrected suite **cannot** mutate main.py — it's restored to the router-importing scaffold and the suite survives. The suite reaching execution through the `client` fixture (rather than importing the app) is the `harness_boundary` check, already covered in `test_acceptance_checks.py` (fixture passes; from-import / `import` / `TestClient` / dynamic-import all fail).

### 4.3 — expander-change replay
A bound record pinned to sentinel bytes restores to **those** bytes — at both the enforcement seam and `restore_frozen_files` on disk — even when the "live expander" would produce entirely different content. Proves the persisted record (D2), not a re-expansion, is the restore authority.

## Tests
7 deterministic tests, all green. Plan doc 4.1–4.3 marked done.

## Remaining for SIP-0100
- **4.4** the live integration roll on `full` (27b) with an enforcement trace — a state-changing cycle, held for your go; needs a rebuild+deploy first (3.1/3.3/3.4a aren't in the running containers yet).
- **4.5** legacy/unbound no-regression checks (mostly assertions over the existing suites).


jasonLadd (commented):
reviewed

PR #550, opened 2026-07-23
test(sip-0100): Phase 4.5 — no-regression + flag D3 fail-open deviation

jasonLadd:
## Summary

SIP-0100 Phase 4 Task **4.5** — legacy/unbound + no-regression. Test-only, plus a flagged D3 deviation. This closes out Phase 4's *deterministic* work (4.4, the live roll, is separate).

## What it pins
- **Unbound/legacy retain current behavior** — a `None` manifest or non-scaffoldable stack yields no bound record → the executor guard skips enforcement → pre-SIP behavior; unbound `materialize` stays byte-identical.
- **Binding is all-or-nothing** — a scaffoldable manifest binds the *full* frozen set (backend + frontend + harness), a non-scaffoldable one binds nothing. Never a partial record that would protect some frozen files and silently leave others writable.
- **Suites green** — capability + correction (build-convergence) suites pass (the 2 `verification_contract_runner` `missing_tooling` skips are the known local env gap; green in CI).

## Flagged: D3 deviation (fail-open vs fail-closed)
Grounding surfaced a real gap. `_build_bound_record_for_run` is **fail-open**: a `build_bound_record` error disables enforcement (returns `None`, run continues) rather than failing the run — a deliberate 2.4 rollout-safety choice. **D3 wants fail-closed** ("enforcement can't be bypassed by broken wiring").

- Reachable only when binding *throws* on a scaffoldable stack (a bug), so exposure is narrow.
- Pinned by `test_binding_error_fails_open_documented_d3_deviation` — documented, not silent.
- Plan §4.5 recommends hardening to fail-closed once the enforcement has a clean `full` roll (4.4) under it. **Not fixing it in this PR** — it's your call whether/when to flip it.

## Tests
3 new (`test_sip0100_no_regression.py`). Plan doc 4.5 marked done with the deviation note.


jasonLadd (commented):
reviewed

PR #552, opened 2026-07-23
feat: deterministic fill-slot binding at plan authoring (Phases A+B)

jasonLadd:
## Summary

Implements **Phases A+B** of the authoring-target-binding plan (PR #551): the same scaffold ownership surface that SIP-0100's runtime enforcement uses is now a **read-only projection** consumed at plan authoring, so a bind-mode plan's dev/qa targets are gated against the fill slots *before* generation — instead of the proposer guessing and hitting them only by luck.

### Phase B — the guarantee (`cycles/plan_authorization.py`)
- `PlanAuthorization.from_contract` — read-only projection: frozen paths, fill slots, QA namespace, required coverage, source prefixes.
- `validate_plan_write_targets` — canonicalizes every dev/qa workspace target (D7 `normalize_ws_path`, shared with runtime), rejects **frozen**, **cross-role slot**, **undeclared source**, and **glob/dir/absolute**; enforces **coverage** (every required slot assigned exactly once, no duplicate ownership). **Atomic** (returns all defects, never silent-strips). Stable `PLAN_TARGET_*` codes + rich rejection records.
- Wired at `task_plan.py:688`; a violating plan raises `CycleError` → bounded framing re-roll → a distinct authoring-bind failure, not a build failure.

### Phase A — inform (`task_plan.py`, `planning_tasks.py`, asset v3)
Injects the writable surfaces as **data** alongside `contract_criteria_index`; the bind appendix asset carries the prose instruction (#448): *write ONLY these paths, do not translate the layout* (names the exact bad patterns: `routers/`, `.tsx`-for-`.jsx`, `src/pages/`, invented `store.py`). Separated `DEV_WRITABLE_SLOTS` / `QA_WRITABLE_NAMESPACES` / `REQUIRED_SLOT_COVERAGE` / `READ_ONLY_CONTEXT_PATHS`.

### Architectural boundary (review #3)
The plan model consumes the projection; it never constructs runtime `WriteGrant` objects. One authority, read-only at the plan layer.

## v1 scope (deferred per the v2 plan, called out honestly)
- **Deliverable-trust**: v1 uses the source-tree heuristic (off-slot source under a scaffold dir → reject; root-level → deliverable). The pre-proposer *trusted-declaration* source is deferred.
- **Globs**: v1 rejects all globs/dirs; the namespaced-slot subset-proof is deferred (`fill_slot_paths` is concrete files today).
- **Builder role** not gated (packaging).
[...]

PR #553, opened 2026-07-24
Revert #552: authoring target-binding (A+B) — clean baseline for green-roll hunt

jasonLadd:
## What

Reverts PR #552 ("deterministic fill-slot binding at plan authoring — Phases A+B") in full, to restore a clean measurement baseline for the seeded green-roll hunt.

## Why

The A+B investigation was largely self-inflicted: the plan mis-targeting it aimed to fix came from running **unseeded** (author-mode) cycles. A `validated-fullstack` cycle only enters bind mode when seeded with both `contract_ref` and the interface manifest in `plan_artifact_refs`; seeded rolls don't have the problem A+B was built to solve. The *idea* — authoring-side fill-slot binding as the mirror of SIP-0100's enforcement-side frozen protection — is on-arc and is preserved in `docs/plans/authoring-target-binding-plan.md` (left in place).

The concrete hazard is **Phase B** (`validate_plan_write_targets` at `task_plan.py`):

- Gated on `contract is not None` → **active on exactly the seeded rolls the hunt runs**.
- On any bind-mode plan whose dev/qa target doesn't resolve to an authorized fill slot, it does `raise CycleError(...)` at **implementation dispatch**.
- It was **never validated against the live seeded contract** (`art_fdbb80b2667a`) — only synthetic fixtures.
- Its claimed "routes to the #522 framing re-roll" recovery path **isn't firing live**, so a false-positive kills an otherwise-recoverable roll instead of re-rolling.

That is an unvalidated new rejection gate sitting in the path of a measurement whose entire purpose is a clean read on *"does SIP-0100 + the correction fixes green a roll?"*. Phase A is benign but its reach to the framing proposer is unproven — not worth a partial revert.

Re-introduce authoring-bind later, wired at the **framing gate** (not implementation) and validated against the real contract, only if seeded-roll variance proves it necessary.

## Verification

- Revert is the exact inverse of the PR: `git diff 09db44f2 HEAD` is empty.
- No dangling references to any reverted symbol (`plan_authorization`, `validate_plan_write_targets`, `PlanAuthorization`, `qa_test_namespace_for_stack`, `writable_surfaces`).
- Tests green: full `tests/unit/cycles/` (1905), planning handlers (210), reverted fixtures/proposer tests (41).


jasonLadd (commented):
reviewed

PR #554, opened 2026-07-24
fix(correction): drift branch reaches the fill-slot source (drift+behavioral-bug combo)

jasonLadd:
## What

`_resolve_repair_target`'s **drift branch** now unions the RC2 package-scoped implementation surface (`_scope_to_shared_packages`), the same surface the no-drift branch already unions.

```
if drift_files:
-   target = drift_files ∪ failed_artifacts
+   target = drift_files ∪ failed_artifacts ∪ scope(implementation_artifacts, failed_artifacts)
```

## Why

Found live on the green-roll hunt (pf-27 `cyc_d01810b2922f`, pf-28 `cyc_eb02a1d01859`). A `tests_pass` failure **co-occurred** with interface drift on the scaffold-**frozen** `backend/main.py` (its `/health` route + the repair's own un-restored inline routes). That pinned the repair target to the `if drift_files` branch, whose target was `drift_files ∪ failed_artifacts` — which **omits** the package-scoped implementation surface that the no-drift RC2 branch (#536) unions.

The real bug (empty-title→422 / missing-404 validation) lived in the fill slot `backend/routes.py`, which is **neither** a drift file **nor** the failing qa.test's own artifact. So the repair edited only `main.py` + the test file on every attempt and **never reached `routes.py`** → the correction loop exhausted without ever touching the buggy source.

**Validated live:** with this fix deployed, pf-28 and pf-29's repairs emit `routes.py` on every attempt; pf-27 (pre-fix) never did.

## Change

- **`adapters/cycles/correction_runner.py`** — drift branch unions `_scope_to_shared_packages(implementation_artifacts, failed_artifacts)`. Empty surface (author mode / non-build corrections) → the scoped set is empty and the target is **byte-identical** to the prior union. Package-scoped (`backend/tests/*` → `backend/*` source, frontend excluded). `focus`/`description` stay unset (no inline prompt content, #448).

## ⚠️ Reviewer note — reversed a deliberate prior boundary

This **supersedes** the RC2 decision to scope the implementation surface to the no-drift branch only. A pre-existing test, `test_drift_present_ignores_implementation_surface`, locked in "the drift path stays byte-identical." That boundary assumed `drift_files` always capture the fixable source cause — **false** when the drift is on a *frozen* file while the behavioral bug is in a non-drift *fill slot*. I renamed/updated it to `test_drift_present_also_unions_implementation_surface` with a docstring documenting the supersession. **Please sanity-check this reversal.**

## Tests

- `tests/unit/cycles/test_correction_runner.py`: 43 pass — 2 new (`test_drift_qa_test_also_unions_package_scoped_source`, `test_drift_without_implementation_surface_is_byte_identical`) + 1 updated (the reversed-boundary test).
- Full `tests/unit/cycles/` domain: 1907 pass. `ruff` clean.

## Scope / risk

- Correction-loop targeting only; drift-detection (#532/#534) and the no-drift RC2 path (#536) are untouched.
- Additive union with de-dup — strictly widens the repair target when `implementation_artifacts` is present (bind mode); no-op otherwise.

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -116,7 +116,20 @@ def _resolve_repair_target(
         # the failing artifact's bug. So focus/description stay unset (no inline
         # prompt content — CLAUDE.md #448); the named artifacts + that instruction
         # redirect the repair onto both the drifted source and the failing file.
-        target = list(dict.fromkeys([*drift_files, *failed_artifacts]))
+        # pf-27 (cyc_d01810b2922f): a ``tests_pass`` failure can CO-OCCUR with
+        # interface drift on a scaffold-FROZEN file (there: backend/main.py — its
+        # /health route + the repair's own un-restored inline routes), which pins the
+        # target to this drift branch. But the behavioral fix still lives in the
+        # fill-slot source under test (routes.py), which is neither a drift file nor
+        # the failing qa.test's own artifact — so without the same package-scoped
+        # implementation surface the no-drift branch already unions (RC2), the repair
+        # edits only the drifted file + the test and NEVER reaches routes.py →
+        # non-convergence. Union it here too; empty surface (author mode) → the scoped
+        # set is empty and the target is byte-identical to the pre-pf-27 union.
+        scoped_source = _scope_to_shared_packages(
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #555, opened 2026-07-24
fix(correction): give the dev repair the same scaffold fill-only constraint as develop

jasonLadd:
## What

The **develop** handler injects a managed fill-only appendix on scaffoldable stacks (SIP-0099 99.3): *"fill the seeded skeleton's slots; do NOT change route paths/methods/decorators/signatures; `ApiError` is already wired — use it."* The **correction/repair** path got **none of it** — its template says *"re-produce the named output artifact(s) so they satisfy the acceptance criteria,"* which invites a full rewrite. So repairs freely rewrote the scaffold-seeded interface.

This wires the **same** appendix into the dev correction-repair.

## Why (observed live)

On the green-roll hunt, two independent rolls showed the same class of deviation — the repair rewriting the seeded interface:

- **pf-28 / pf-29:** `routes.py` repaired with `ApiError(status_code=404, detail=...)` — an invented signature — vs the frozen `errors.py` contract `ApiError(code, message)` (`"run_not_found"` → 404). The wrong call raises `TypeError` → HTTP 500 not 404 → the 404 test fails; the analyzer misreads it as "missing 404 handling," so the repair adds *more* wrong-signature calls → non-convergence.
- **pf-29:** `routes.py` rewritten to `APIRouter(prefix="/runs")` + relative decorators vs the seeded `APIRouter()` + absolute decorators → breaks the static `endpoint_defined` check.

The correct contract is sitting in the workspace (frozen `errors.py` + the skeleton's absolute decorators) — the agents just weren't told, on the repair path, to honor it. See the root-cause write-up in `night_findings.md` (Findings D/E + synthesis).

## Change

- **`repair_handlers.py`** (`_RepairPromptMixin`): `handle()` now renders `request.development_develop_fill_only_appendix` and threads it into the repair prompt via `inputs["fill_only_section"]`; `_build_render_variables` surfaces it. Scoped to the **dev** role on a **scaffoldable** stack (via `is_scaffoldable_stack(resolved_config.build_profile)`) — builder/qa repairs and non-scaffolded builds are unaffected.
- **`request.cycle_repair_task.md`**: adds the `{{fill_only_section}}` placeholder (declared optional) right after the repair intro.
- **No new prompt content** — reuses the dev's managed asset (CLAUDE.md #448 / ownership-before-extension).

## Scope / limits

This is the **cheap, prompt-level** half of the "fill-slot interface ownership" direction. It's advisory (LLM-dependent) — it should reduce the deviation on the repair path but won't *guarantee* it. The deterministic half (diff-guarding a fill slot's interface lines against the skeleton, and/or a typed check that `ApiError` is *called* with a valid code) is the larger follow-up SIP.

Independent of PR #554 (the drift-branch targeting fix) — different files; either can merge first.

## Tests

- `tests/unit/capabilities/test_repair_fill_only.py` — 7 new: render gating (dev+scaffoldable renders; non-scaffoldable / no-profile / builder-role → empty), variable threading, template contract.
- `test_impl_handlers` (58), `test_develop_fill_only`, template contracts (102) green. `ruff` clean.

### DIFF
--- a/src/squadops/capabilities/handlers/impl/repair_handlers.py
+++ b/src/squadops/capabilities/handlers/impl/repair_handlers.py
@@ -14,10 +14,14 @@ correction-loop flows have distinct, non-overlapping capability ids.
 
 from __future__ import annotations
 
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 from squadops.capabilities.handlers.cycle_tasks import _classify_file, _CycleTaskHandler
 from squadops.capabilities.handlers.fenced_parser import extract_fenced_files
+
+if TYPE_CHECKING:
+    from squadops.capabilities.handlers.base import HandlerResult
+    from squadops.capabilities.handlers.context import ExecutionContext
 
 
 def _artifacts_from_fenced_blocks(content: str, fallback_name: str) -> list[dict[str, Any]]:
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reeviewed

PR #558, opened 2026-07-24
refactor(correction): remove the unconsumed qa.validate_repair step

jasonLadd:
## Summary

Implements #556: removes the `qa.validate_repair` step from the correction loop. The step cost one 27b LLM turn per repair sequence per correction attempt (up to 5 on the current measurement config) and produced a PASS/FAIL verdict document that no code consumes — repair acceptance is decided deterministically by patch verification (#389) and the behavioral retest (#456). It was also sequenced before the retest, so the one LLM look at a finished repair happened with the weakest evidence available.

## Changes

- **`task_plan.py`** — repair sequences are now `[dev]` / `[builder]`; #556 rationale comment on `REPAIR_TASK_STEPS` pointing at #389/#456/#557.
- **`repair_handlers.py`** — `QAValidateRepairHandler` and its validate-only helpers (`_format_repair_summary`, `_format_repair_artifacts`, fence-language map) deleted (−173 lines).
- **`bootstrap/handlers.py`** — handler deregistered.
- **`request.cycle_validate_repair.md`** — deleted (the verdict rubric is salvaged by git-history reference in #557).
- **`correction_runner.py`** — the `if task_type != "qa.validate_repair"` overlay guard is now unconditional; repair outputs enter `prior_outputs` **without** `artifacts` (matching the executor fan-in convention — the keep-artifacts special case existed solely for the validate prompt, and full file contents were bloating every downstream prompt). This also removes the side effect of the validate output overwriting the original qa.test summary under `prior_outputs["qa"]`. Module docstring documents the vacated slot's rules for any future LLM judgment (post-retest, governance role, fail-closed — #557).
- **`dispatched_flow_executor.py`** — comment updated.
- **`implementation.yaml`** — dead `qa.validate_repair` entry removed from the pulse-check `after_task_types` binding (behavioral no-op: the task type can no longer occur).
- **Tests** — new regression coverage: no repair sequence contains a validate step, no `qa.validate_repair` envelope is dispatched, the handler is not registered, and repair summaries reach downstream `prior_outputs` without artifact contents. Script-based tests realigned (one fewer positional reply per repair sequence).

## Test results

- 5 directly-affected files: 120 passed.
- Full `tests/unit/cycles` + `tests/unit/capabilities`: 3189 passed, 2 failed — the 2 failures (`test_verification_contract_runner.py`, `missing_tooling` skips) reproduce identically on unmodified main; pre-existing environment issue, not this change.
- `tests/unit/contracts`: 217 passed.
[...]

PR #562, opened 2026-07-24
fix(correction): repair-path hardening — resolved_config threading + SIP-0100 3.4b frozen enforcement (restore+signal)

jasonLadd:
Two repair-path fixes from the pf-30 triage, one deploy-window unit. Both address the same failure: the correction loop's repairs rewriting scaffold-owned interface.

## Commit 1 — thread `resolved_config` onto repair envelopes (`03740268`)

The dev repair's scaffold **fill-only appendix (#555) never fired on the repair path**: `_render_fill_only_section` gates on `inputs["resolved_config"]["build_profile"]`, and the repair envelope omitted the field — empty profile → silent no-op. pf-30 attempts 1–3 each re-emitted `routes.py` with decorators flipped from manifest-mandated absolute paths to relative ones, registering every endpoint at root, non-converging against a correct diagnosis. One line, mirroring the retest path's existing threading; regression tests at the wiring level (the level #555's own handler tests couldn't see).

## Commit 2 — SIP-0100 3.4b: frozen enforcement on the repair path, restore + signal (`665a59f8`)

Frozen-ownership enforcement lived only on the executor's regular storage path; repair emissions bypassed it (deferred 3.4b, confirmed live in pf-27 and pf-30 — zero `scaffold_integrity` restores while repairs emitted frozen `models.py`/`main.py` every attempt).

- New **`squadops/cycles/scaffold_enforcement.py`**: `enforce_frozen_ownership`, `bound_record_or_none`, `frozen_restore_instruction` — lifted verbatim from `DispatchedFlowExecutor`, which now delegates (its 10 enforcement tests pass unchanged).
- **`correction_runner`** enforces on repair emissions **before any landing point** — in-place on `result.outputs["artifacts"]`, so registry store, the `repair_artifacts` overlay, and patch verification all see restored bytes.
- **Restore + signal, not silent restore**: each restore appends an authoritative instruction to an executor-owned run-lived carry (created beside `correction_counter`); the next attempt injects it as `failure_evidence["scaffold_enforcement"]`, and the repair prompt renders a `FROZEN OWNERSHIP (authoritative)` block — the same deterministic-instruction pattern as `interface_drift`. Silent restore would trade the clobber for a restore-fight thrash.
- Unbound/legacy runs: `bound_record` is None → byte-identical behavior. Enforcement hooks the repair loop only; analyze/decision (document emitters) and the retest path (QA-namespace semantics out of 3.4b scope) are untouched.

## Verification

- 3195 passed across `tests/unit/cycles` + `tests/unit/capabilities`; only the two known pre-existing `missing_tooling` environment failures. ruff check + format clean.
- New tests: frozen repair emission restored at both landing points + evidence event + carry append; carry reaches the next attempt's `failure_evidence`; prompt-block rendering (+ absent-key negative); `resolved_config` threading + `{}` default.
- Deploy-time obligation: verify **in-container** that a dispatched repair envelope renders the fill-only appendix and that a frozen repair emission logs `scaffold_integrity (repair path)` — gate-condition checks, not file-loaded checks (the pf-30 lesson).

## Context

pf-30 triage (green-roll hunt). Lands with #558 (merged) in the next deploy window; deploying resets the N=5 baseline hash → pf-31.

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -266,6 +266,23 @@ class CorrectionRunner:
             },
         )
 
+    def _emit_scaffold_integrity_evidence(self, record: Any, envelope: TaskEnvelope) -> None:
+        """SIP-0100 3.3/3.4b: surface one repair-path enforcement as a structured event + log
+        (best-effort — observability must never break the correction loop). Mirrors the
+        executor's emitter for the regular storage path."""
+        payload = record.to_dict()
+        logger.warning("SIP-0100 scaffold_integrity (repair path): %s", payload)
+        try:
+            self._event_bus.emit(
+                EventType.ARTIFACT_OWNERSHIP_ENFORCED,
+                entity_type="artifact",
+                entity_id=record.normalized_path or record.attempted_path,
+                context={"cycle_id": envelope.cycle_id, "run_id": record.bound_run_id},
+                payload=payload,
+            )
+        except Exception:
+            logger.debug("SIP-0100: scaffold_integrity event emit failed", exc_info=True)
+
     async def _dispatch_protocol_step(
         self,
         step_envelope: TaskEnvelope,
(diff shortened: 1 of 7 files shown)


jasonLadd (commented):
reviewed

jasonLadd (commented):
reviewed

PR #563, opened 2026-07-24
docs: pf-31 correction-convergence fix plan (A: contract expectations, D: emission integrity, E: rejected-candidate accounting)

jasonLadd:
Plan-first PR (#551 precedent) — design review for the three pf-31 fixes before implementation. Each fix cites its pf-31 reproduction and its replay-verification plan; sequencing A→D→E in one deploy window. Highlights: Fix A is a **rendering/salience** fix, not plumbing (the resolved TypedChecks already reach the repair envelope via task_plan.py:580 — they render as dict-repr soup below contradicting prose); Fix E is a **ledger accounting** fix (rejected-candidate patch-verification evidence supersedes accepted state in the SIP-0096 aggregation — that's how endpoint_defined regressed at final verification after passing mid-run).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01WXrSUPGsx5yxbsn19aDjXX

PR #565, opened 2026-07-24
feat(manifest): rename group_run path param {id} -> {run_id}

jasonLadd:
## Why

The plan proposer independently authored `{run_id}` prose in **pf-31 and pf-32** against the manifest's `{id}` — costing pf-31 seven rejected repairs (the prose-poisoning class behind fixes A/D/E, #563/#564). The model's prior for the descriptive name is strong — and correct. The manifest is the *single seam* that owns the interface name, so this aligns the authority with the prior instead of fighting it. From pf-33 on, the A3 conflict lint should simply go quiet on this token.

Bonus: the seeded skeleton handlers stop shadowing Python's builtin `id()` (`def get_runs_run_id(run_id: str)`).

## What changed

- `examples/03_group_run/interface_manifest.yaml` — the 3 path lines (the source of truth; everything below derives from it)
- reference fill `backend/routes.py` — decorators, function names, args track the expander's deterministic derivation
- `test_scaffold.py` / `test_scaffold_contract.py` — the only tests asserting against the *real* manifest
- plan-authoring brief fragment — the `/runs/{id}/join` example updated + fragment manifest hashes regenerated
- **Left unchanged**: self-contained synthetic `{id}` fixtures (incl. the A3 lint tests, which deliberately construct the `{run_id}`-vs-`{id}` conflict), historical SIPs/plan docs, entity field `RunEvent.id` (only the path *parameter* renames)

## Proof

- contract gate `lint`: GREEN (host)
- `bare-skeleton`: **6/6 GREEN** and `reference-fill` (winnability): **6/6 GREEN** — run in the qa container (node present; host lacks npm by design #419)
- `tests/unit/{capabilities,cycles,prompts}`: 3603 passed (2 pre-existing host missing_tooling failures, fail on main too, pass in CI)
- ruff clean

## Rollout (deliberately NOT in this PR)

Live seeded artifacts are untouched — pf-32 is mid-flight on frozen contract v3. After pf-32 goes terminal: `contract_gate.py emit` → `squadops artifacts ingest` the new manifest + contract v4 → update the launcher's `CONTRACT_REF`/`MANIFEST_REF` for pf-33. Deploy hash `edd124e3` is unaffected (pure data change; no agent rebuild needed).

### DIFF
--- a/examples/03_group_run/interface_manifest.yaml
+++ b/examples/03_group_run/interface_manifest.yaml
@@ -55,12 +55,12 @@ api:
     - { method: POST, path: /runs,              summary: create run,
         request: RunEventCreate, response: RunEvent,
         errors: [validation_error] }                                 # §5.1
-    - { method: GET,  path: "/runs/{id}",       summary: run details,
+    - { method: GET,  path: "/runs/{run_id}",       summary: run details,
         response: RunEvent, errors: [run_not_found] }                # §5.3
-    - { method: POST, path: "/runs/{id}/join",  summary: join run by participant name,
+    - { method: POST, path: "/runs/{run_id}/join",  summary: join run by participant name,
         request: ParticipantName, response: RunEvent,
         errors: [run_not_found, duplicate_participant, validation_error] }   # §5.4
-    - { method: POST, path: "/runs/{id}/leave", summary: leave run by participant name,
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

Issue #559, opened 2026-07-24
refactor: task-type convention — strings at the boundary, constants at the core, properties over identity


jasonLadd:
**Target: 1.5 (odd-minor stabilization batch).** Wide-but-mechanical structural refactor —
exactly the class of change the odd-minor lane quarantines so a regression is unambiguously
the refactor. Not for the 1.4 feature lane; do not interleave with green-roll measurement
deploys.

## Problem

Task-type identifiers (`"development.develop"`, `"qa.test"`, …) are re-typed as string
literals at every use site: **97 literals across 23 source files** (`src/` + `adapters/`,
tests excluded). Two distinct failure modes:

1. **No single source.** A typo'd comparison doesn't error — it's silently never true.
   No rename point, no exhaustiveness checking, nothing for tooling to verify.
2. **Identity checks encoding properties.** Orchestration code branches on *which* task
   type this is when it means *what kind* of step this is. Canonical exhibit (removed in
   #558): `if task_type != "qa.validate_repair"` actually meant *"only steps that emit
   product artifacts feed the overlay; judgment documents don't."* A property of the step,
   encoded as an identity check on one instance of it — any later judgment-flavored step
   would silently not be covered and its document could shadow a product file. Identity
   checks don't scale: every new task type must know about every scattered special case
   that should apply to it.

The codebase already contains the right shape, inconsistently applied:
`_REPAIR_STEPS_BY_FAILED_TASK_TYPE` + `repair_steps_for()` (declarative table, one lookup
point) and the handler registry (`capability_id` → handler class). The scattered `==`/`!=`
sites in the executor and correction runner bypass it.

## The convention (enforceable rule, for CLAUDE.md once implemented)

1. **Strings at the boundary.** `task_type` stays a plain string on the wire (RabbitMQ
   envelopes, Postgres rows, Prefect task names, YAML profiles). No serialization change.
2. **Constants at the core.** One module single-sources every task-type identifier as a
   `StrEnum` (`TaskType.QA_TEST == "qa.test"` — full wire compatibility, but a typo is an
   `AttributeError` and a rename is one line). All core-side comparisons and table keys use
   the enum.
3. **Properties over identity.** When orchestration treats some steps differently, the
   distinguishing property is **declared where the step or handler is defined** (handler
   class attribute or step-table field — the capability layer owns "what kind of thing this
   step is"), and the orchestration loop consults the property. `if task_type == X` in
[...]

PR #569, opened 2026-07-25
fix(emission): #566 recovery + aimed retry, #568 locus-keyed qa repair routing

jasonLadd:
Closes #566
Closes #568

## Live driver (pf-32, tonight)

m007-qa.test emitted a complete test suite in one bare ```python fence — no filename — **twice in a row**: 0 blocks extracted, ~10 min of generation discarded per attempt, then the correction loop dispatched a **dev** repair for a failure whose artifact eve never delivered. The repair emitted only frozen files (3.4b restored them) and was rejected. Two correction attempts burned on a parser gap + a routing gap.

## #566 — emission recovery + aimed retry (three deterministic layers)

1. **Parser fallback**: when no strategy resolves any file, the task expects exactly ONE artifact, and exactly ONE substantial filename-less fence exists → map it. Declines on any resolved fence (even a rejected unsafe one), 2+ candidate fences, or 2+ expected artifacts — cardinality-1-everywhere only.
2. **Machine marker**: develop + qa.test zero-extraction failures emit `emission_failure` (data, built in `emission_integrity.py`), which also passes through `build_failure_evidence` so the analyzer sees "no artifact was ever produced" as a fact.
3. **Aimed retry**: the executor threads the marker into the re-dispatched (enriched) envelope; handlers append a format-feedback appendix rendered from the new `request.cycle_emission_retry_feedback` template asset (all prose in the asset, per #448).

## #568 — locus-keyed repair routing

- `FailureLocus` + `classify_failure_locus` (pure, deterministic): emission marker / missing expected artifacts / pytest exit 2·4·5 → **OWN_ARTIFACT**; exit 1 → **SUBJECT**; everything else **UNKNOWN**.
- Own-artifact table: `qa.test → qa.test_repair` (qa role, new `QATestRepairHandler` — a `_RepairPromptMixin` subclass, so it gets the Contract Expectations block and the same deterministic acceptance: patch verification #389 + behavioral retest #456).
- **Test-gaming guard**: SUBJECT and UNKNOWN never reach qa re-authoring — an app bug "fixed" by rewriting tests is a manufactured false green. Ambiguity falls toward the dev chain.
- Own-artifact repairs target the failed task's OWN contract (its expected_artifacts/focus), not `_resolve_repair_target`'s subject-implementation union.
- `REPAIR_TASK_TYPES` derives over both tables (#559) — the new candidate type joins the Fix-E provenance filter and #389 re-typing with zero bookkeeping.

## Verification

- 20 new tests: parser fallback guards (incl. the exact pf-32 live emission shape), locus classifier (incl. exit-code table), routing (incl. the never-route-behavioral-to-qa guard), end-to-end runner dispatch, handler markers, executor retry threading, template render variables.
- Touched domains (capabilities/cycles/prompts/bootstrap): **3756 passed**; only the 2 pre-existing host `missing_tooling` failures (fail on main, pass in CI).
- ruff check + format clean.

## Deploy note

Needs agents + runtime-api rebuild to take effect — natural fold into the pf-33 deploy window alongside the #565 rebuild (greens 0/5, baseline reset free).

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -155,6 +155,40 @@ def _resolve_repair_target(
     )
 
 
+def _locus_and_repair_target(
+    failed_task_type: str,
+    failure_evidence: Any,
+    failed_inputs: dict[str, Any],
+) -> tuple[str, list[str], str | None, str | None]:
+    """#568: classify the failure locus and choose the repair target for it.
+
+    Returns ``(locus, expected_artifacts, focus, description)``. An
+    OWN_ARTIFACT failure (the failed task's own emission is missing or
+    uncollectable) targets the failed task's own contract — pointing
+    ``_resolve_repair_target``'s subject-implementation union at a test
+    re-author would aim it at app source files. Every other locus keeps the
+    existing target resolution unchanged.
+    """
+    from squadops.cycles.failure_evidence import FailureLocus, classify_failure_locus
+
+    failure_locus = classify_failure_locus(failure_evidence)
+    own_expected = [str(e) for e in (failed_inputs.get("expected_artifacts") or []) if e]
+    if failure_locus == FailureLocus.OWN_ARTIFACT and own_expected:
+        logger.info(
+            "correction_repair_locus: own_artifact — %s re-produces %s",
+            failed_task_type,
+            ", ".join(own_expected),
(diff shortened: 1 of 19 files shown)


jasonLadd (commented):
reviewed

PR #570, opened 2026-07-25
SIP: LLM Emission Contracts — provider-agnostic structured output + hardened code extraction

jasonLadd:
Proposes `sips/proposed/SIP-LLM-Emission-Contracts.md` for design review (step 1-2 of the SIP workflow — approving the design, not code).

## Why now

- **The reactive-patch ledger**: five fenced-parser malformation classes fixed one live failure at a time (#430/#470/#502/#528/#566), plus the proposer YAML-escaping bug, think-block stripping, and pf-31's truncation class. Each fix correct; the pattern is the problem.
- **The provider migration**: inference is moving Ollama → Atlas. Structured-output support differs per provider, so the contract must live in the LLM port — any adoption written against Ollama's dialect would be redone, and today there is no per-provider malformation baseline to de-risk the cutover.

## Shape

1. **Emission contracts** per handler: STRUCTURED(schema) / CODE_FILES(expected_artifacts) / PROSE — the framework owns obtaining a valid instance.
2. **Provider-agnostic structured output**: `response_schema` (JSON Schema) on the port; adapters translate to their dialect (Ollama native, OpenAI-compatible `json_schema` as the expected Atlas path) or run a deterministic validate-and-reask emulation. Capability tiers discovered by a shared adapter **conformance suite** — never assumed, never named in handlers (#559).
3. **Code stays markdown** (aider code-in-JSON quality evidence) — CommonMark recognition engine under the mapping strategies (absorbs #567) + diff-based repair emissions.
4. **Evidence first** (absorbs #431): raw + finish reason + extraction outcome persisted → malformation rate per provider × model × handler = the A/B instrument for the Atlas migration.

Phases P0–P5, each independently shippable; P1–P3 targeted at the 1.5 stabilization lane; P0 (evidence) can land any time.

### DIFF
--- a/sips/proposed/SIP-LLM-Emission-Contracts.md
+++ b/sips/proposed/SIP-LLM-Emission-Contracts.md
@@ -1,9 +1,9 @@
-# SIP-0XXX: LLM Emission Contracts — Provider-Agnostic Structured Output and Hardened Code Extraction
+# SIP-0XXX: LLM Emission Contracts — Typed Response Handling with Provider-Agnostic Structured Output
 
 **Status:** Proposed
 **Authors:** SquadOps Architecture
 **Created:** 2026-07-25
-**Revision:** 1
+**Revision:** 2
 
 ## 1. Abstract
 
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #584, opened 2026-07-25
fix(qa): harness_boundary expectation instructed the violation; qa prompt diet; failed-check names in patch logs

jasonLadd:
**Night-window fix (pf-33), deployed from this branch per the overnight rules — for morning merge.**

## The loop-killer (proven twice in one night, live)

The Fix-A CONTRACT EXPECTATIONS block rendered `harness_boundary` as *'must build its TestClient against one of: `backend.main`, `app.main`'* — which reads as an instruction to import the entry module and construct the client. That is **the exact violation the evaluator rejects**. In pf-33:
- corr-00: eve's re-authored suite (6 good tests, parses, imports routes — verified by artifact replay) was rejected SOLELY for `from backend.main import app` + a self-built `client` fixture.
- corr-01: neo's cross-role test rewrite did `client = TestClient(main.app)` — same violation, same obeyed instruction.

Two agents, two roles, both faithfully following the authoritative block into rejection. The line now states the prohibition + the sanctioned alternative (the frozen conftest's `client` fixture), with a regression test pinning the old phrasing dead.

## Also from the same triage

- **qa prompt diet**: the focused qa prompt scoped its source dump to the packages the task's own artifacts live in (RC2 package rule at the prompt seam). m007's prompt carried all 16 files (~18k tokens incl. every frontend .jsx) into a backend suite task — prefill waste + rambling pressure against the deliberate 8192 completion clamp; attempt 1 truncated mid-file and crashed collection.
- **patch_verification observability**: the log now names failed checks (`failed=...`) — 'status=failed reason= checks=7' cost a by-hand artifact replay per rejection tonight.

## pf-33 measurement wins before the cancel (for the record)

- #568 fired correctly in BOTH directions: exit-2 → own_artifact → `qa.test_repair`/eve (first ever), exit-1 → subject → dev chain. Target override correctly ignored /health drift noise on the own-artifact path.
- The contract-as-arbiter dynamic worked: eve's suite asserted the contract-correct 409 against the app's 400.
- Known next wall (documented, NOT fixed here): `ApiError(status_code=, detail=)` signature misuse re-appeared in a dev repair (pf-28 Finding-D class) — the call-signature knowledge lives only in the scaffold stub docstring, which dies with the first fill.

Tests: 124 pass across both touched suites; ruff clean.

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -1974,13 +1974,21 @@ class DispatchedFlowExecutor(FlowExecutionPort):
             typed_acceptance_enabled=resolved_config.get("typed_acceptance", True),
             command_acceptance_enabled=resolved_config.get("command_acceptance_checks", True),
         )
+        # pf-33: name the failed checks — "status=failed reason= checks=7" forced
+        # a by-hand artifact replay to learn WHICH check rejected the patch.
+        failed_checks = [
+            str(c.get("check", "?"))
+            for c in verification.checks
+            if isinstance(c, dict) and c.get("passed") is False
+        ]
         logger.info(
-            "patch_verification task=%s task_type=%s status=%s reason=%s checks=%d",
+            "patch_verification task=%s task_type=%s status=%s reason=%s checks=%d failed=%s",
             envelope.task_id,
             envelope.task_type,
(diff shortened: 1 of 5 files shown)


jasonLadd (commented):
reviewed

PR #587, opened 2026-07-25
fix(cycles): cancel reaches the dispatch boundary — repairs stop on a cancelled run (#586)

jasonLadd:
Cancelling a run marked it cancelled in the registry but did not stop execution.
Measured live on 2026-07-25: **2h 19m 39s of post-cancel execution**, all five
correction attempts burned after the cancel, contending for the GPU with the
successor measurement roll for its entire first two hours.

Closes #586

## Root cause

The only cancellation probe sat at the top of the per-task loop in
`_execute_sequential` (`dispatched_flow_executor.py:1147`). A correction loop
dispatches analyze / repair / patch-verification / retest tasks many times
without ever returning to that loop top, so nothing ended a cancelled run
except `max_correction_attempts` exhausting naturally.

The gap was a **mutual delegation** left by the SIP-0097 decomposition — both
extracted collaborators disclaimed the check and pointed at each other:

- `correction_runner.py` — "relies on **the dispatch path's checks at
  dispatch/await boundaries** per the §6 cancellation ownership rule."
- `task_dispatcher.py` — "the §6.1 cancellation probe is **deliberately not
  wired** until a dispatch-boundary check actually exists."

Neither docstring was wrong in isolation. Together they left every long-running
path uncovered, and the correction loop grew into the dominant time sink
afterward without anyone revisiting the ownership split.

## The fix — one seam, at the owner

`TaskDispatcher` owns the dispatch boundary, so the §6.1 probe goes there.
`dispatch_task` probes `is_cancelled` **before** creating a Prefect task run and
**before** publishing, so no agent work starts on a cancelled run.

This single seam covers every dispatch path because the executor composes **one**
`TaskDispatcher` instance (`:160`) and injects it into both `CorrectionRunner`
and `PulseBoundaryRunner` — sequential dispatch, correction/repair dispatch, and
pulse-boundary repair dispatch all pass through it.

`_CancellationError` is already the shared control-flow error and already maps
through `execute_run` → `resolve_terminal_outcome` → `CANCELLED`, so the terminal
path, lease release and `finalize` are untouched.

## Design notes

**Deliberately not added: a second probe in `_handle_task_outcome`.** The issue
listed that as fix candidate 1, but reading the wiring showed the dispatcher
probe subsumes it — and adding one would push cancellation policy back into the
god file the SIP-0097 decomposition was meant to drain. One concern, one owner.

**Probe errors propagate (fail-closed).** The between-task check this
complements has always been fail-closed; a probe that swallowed registry errors
would quietly restore the #586 behaviour. Pinned by a test.

**Probe is optional at construction.** Same idiom as every other injected
collaborator here (`activity_port`, `event_bus`, `workflow_tracker`) so
standalone and in-memory compositions keep working — with a construction-path
test asserting the *real* composition actually passes it. A probe that isn't
wired is cosmetic.

## Residual — stated, not fixed

A cancel landing while a task is **awaiting its reply** still waits out that one
in-flight task (~one LLM generation, 10–15 min on 27b) rather than interrupting
the await. Post-cancel work goes from **hours to a single task**, not to zero.

Racing the reply future against the probe is feasible (`_publish_and_await`
already cleans up on `CancelledError`) but it pairs with the open design
question in #586 candidate 3 — *should cancel terminate the flow, or remain
cooperative in-band?* — so it's left as an explicit decision rather than
inherited behaviour.

## Tests

9 new, all naming the bug they catch:

**Probe behaviour** (`test_task_dispatcher.py`) — raises before publish (the
assertion that matters: `publish` never awaited); creates no orphan RUNNING
Prefect task run; healthy runs still dispatch (guards the inverse regression);
probe errors propagate; unwired probe is a no-op; retry loop stops when cancel
lands mid-flight.

**Construction path** (`test_dispatched_flow_executor.py`) — the composed
dispatcher actually receives a probe and that probe reflects registry
cancellation; `CorrectionRunner` and `PulseBoundaryRunner` share that exact
instance; the probe honours the local `cancel_run` fast path even when the
registry write fails.

## Verification

- `tests/unit` — **5,821 passed**, 3 skipped
- `tests/unit/cycles` — 1,989 passed
- ruff check + ruff format — clean
- Pre-existing on main, unrelated: 2 failures in
  `test_verification_contract_runner.py` (`vc-routes-compiles` skips
  `missing_tooling` — host has no node by design, #419). Confirmed identical on
  a clean `main` checkout.

Not deployed — no rebuild, no restart.

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -11,9 +11,13 @@ state — slice 5 retired the interim executor-supplied dispatch callables
 per AC#9). ``store_artifact`` remains a narrow executor-supplied callable:
 artifact plumbing is §6.7 executor residual, residual-but-watched.
 
-Cancellation: the protocol performs no cancellation checks of its own (it
-never did); it relies on the dispatch path's checks at dispatch/await
-boundaries per the §6 cancellation ownership rule.
+Cancellation: the protocol performs no cancellation checks of its own; it
+relies on the dispatch path's check per the §6 cancellation ownership rule.
+That check now exists — ``TaskDispatcher.dispatch_task`` probes before every
+publish (#586). Until it was wired, this delegation pointed at nothing: the
+transport documented the probe as "deliberately not wired" while this module
+documented itself as relying on it, so a run cancelled mid-correction ran to
+attempt exhaustion (2h20m, five attempts, observed 2026-07-25).
 
 Repair acceptance is deterministic-only (#556): the repair sequence has no
 LLM validation step — patch verification (#389) re-runs the typed criteria
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -164,6 +164,10 @@ class DispatchedFlowExecutor(FlowExecutionPort):
             activity_port=activity_port,
             event_bus=event_bus,
             task_timeout=task_timeout,
+            # §6.1 cancellation probe (#586). Wired on the single dispatcher
+            # instance the correction and pulse runners below share, so cancel
+            # reaches every dispatch path, not just the sequential loop top.
+            is_cancelled=lambda run_id: self._is_cancelled(run_id),
         )
         # SIP-0097 §6.3: correction-protocol collaborator. store_artifact stays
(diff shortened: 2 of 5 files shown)


jasonLadd (commented):
reviewed

PR #585, opened 2026-07-25
fix(correction): night-window stack — ERROR CONTRACT block, exit-4 locus, initial-qa expectations

jasonLadd:
**The remaining three night-window fixes, stacked on the just-merged #584.** Each was built live from a diagnosed roll failure, deployed from its branch, and validated in the following roll (deploy currently runs these — merging re-unifies main with what's on the box).

## 1. `fix/error-contract-salience` (pf-34's killer)
Dev repairs guessed `ApiError(status_code=..., detail=...)` — the frozen contract is `ApiError(code, message)`, so every error path TypeError'd into a 500 *after* passing all typed checks (call signatures are invisible to them). The raise convention lives in the manifest's `error_contract` and the fill-stub's docstring, which dies at first fill. Now: `scaffold.error_seam_instructions(manifest)` (pure, manifest-derived: convention + code→status map) → `failure_evidence["error_contract"]` → an **ERROR CONTRACT (authoritative)** block in repair prompts, same transport as INTERFACE CONFORMANCE / FROZEN OWNERSHIP. Plus a C901 extraction (`_inject_deterministic_evidence`) unifying all deterministic evidence injection.

## 2. `fix/exit4-locus-and-qa-harness-injection` (pf-35's killer)
pytest exit 4 was classified own-artifact — but a conftest/app-import failure (an accepted routes.py importing names the frozen models.py never defines) also surfaces as exit 4, and that misroute sent eve to re-author a suite that could never fix the app, twice. Exit 4 now classifies UNKNOWN → default dev chain (ambiguity never routes toward qa re-authoring, same rationale as the test-gaming guard). Verified live on pf-36: the identical failure shape routed to a dev repair.

## 3. `fix/qa-initial-prompt-expectations` (pf-36's killer)
The INITIAL qa focused prompt dumped typed criteria as dict-repr soup, so the corrected harness_boundary line only ever reached eve via *repair* prompts — every fresh suite self-built `TestClient(app)` (0-for-3 overnight), the broken suite failed a likely-correct app as exit-1/SUBJECT, and dev repairs can't fix a test file. Initial prompts now render the same authoritative **Contract Expectations** block the repair path uses (pf-31 Fix A treatment, completed). Known follow-up (not here): the dev initial prompt has the same soup, but its repair path recovers in one shot — proven twice tonight.

## Verification
[...]

PR #589, opened 2026-07-25
fix(dev): the initial focused prompt carries the scaffold contract it was filling (#588)

jasonLadd:
## Summary

On a scaffolded bind-mode roll, the **initial** dev author receives neither the
SIP-0099 fill-only instruction nor the manifest-derived error seam. It is told
nothing about the frozen surface it is filling, and nothing about how to raise a
contract error. Both pieces of knowledge exist and are already wired — but only into
paths a plan-driven cycle never takes.

## Root cause — the fill-only appendix is wired to a dead path

1. Every plan task carries `focus:`.
2. `src/squadops/cycles/task_plan.py:618` turns that into `inputs["subtask_focus"]`.
3. `src/squadops/capabilities/handlers/cycle/develop.py:338` routes on `subtask_focus`
   to `_build_focused_prompt`, which its own docstring (RC-6) describes as used
   **"exclusively"** — "The legacy monolithic prompt path is NOT used."
4. `_fill_only_section` (SIP-0099 99.3) is called **only** from `_build_dev_prompt` —
   the monolithic path.

So the fill-only instruction reaches only the path plan-driven cycles never execute.

## Root cause — the error seam is repair-only

`scaffold.error_seam_instructions` (shipped in `dc69b88b` for pf-34) has exactly two
consumers, both on the repair path:

- `adapters/cycles/correction_runner.py:186,210`
- `src/squadops/capabilities/handlers/impl/repair_handlers.py:102`

Zero consumers on the initial generation path. The convention also lives in the
fill-slot stub's docstring — which the first fill overwrites, as
`scaffold.py:399` already notes. So the first author has no source for it.

## Live evidence (pf-37, `run_010825617319`)

**Frozen-surface violations — three in seven dev tasks**, each caught and restored by
SIP-0100: `backend/models.py` (m000), `backend/main.py` (m002),
`frontend/src/App.jsx` (m006). These restores have been read as "enforcement
working" — it is, but they are equally a signal that the *instruction never arrived*.

**Error seam — the pf-28/33/34 wall, re-made by the first author.** The generated
`backend/routes.py`:

```python
raise ApiError(status_code=404, detail="Run not found")              # :50
raise ApiError(status_code=422, detail="Participant name cannot be empty")  # :59, :80
raise ApiError(status_code=422, detail="...already exists on this run")     # :68
```

against the frozen `backend/errors.py`:

```python
def __init__(self, code: str, message: str) -> None:
```

`TypeError` at request time → **HTTP 500 on every error path**.

## Why no check catches it

`vc-routes-apierror` is an import-presence check (`module: .errors, symbol: ApiError`).
It confirms the import exists; it cannot see call signatures. `routes.py` passes it
cleanly while every error path 500s. The defect only surfaces later, as a behavioural
test failure at the QA phase, where it costs a correction cycle to undo something the
system already knew how to prevent.

## Third defect in the same method — dict-repr soup

`_build_focused_prompt` renders `acceptance_criteria` as `f"- {c}\n"` where `c` may be
a TypedCheck dict. This is the identical defect #585 just fixed on the QA side, and it
was item 4 on the 2026-07-25 morning list. Exact expectations reach dev only through
the repair prompt's authoritative block.

The method also builds every line of its prose from inline Python string literals
(CLAUDE.md #448).

## Fix

Give the initial focused prompt the treatment #585 gave the QA twin, plus the two
missing sections:

- Typed criteria → authoritative Contract Expectations block; narrative prose demoted.
- Fill-only + ERROR CONTRACT sections reach the initial author (executor threads
  manifest-derived error-seam lines onto `development.develop` envelopes).
- All prose moved into managed prompt assets; the handler assembles data only.

## Follow-up, not in the fix

`request.cycle_repair_contract_expectations_appendix` is reused by the dev initial
path rather than duplicated, so its `cycle_repair`-scoped `template_id` is now a
misnomer. `scripts/maintainer/upload_prompts_to_langfuse.py` publishes by
`template_id`, so renaming orphans the registry entry (the #327 class) — it wants its
own PR alongside a registry re-sync.

Worth considering separately: a typed check that verifies `ApiError` is *called*
correctly (arity/kwargs), not merely imported. That would catch this class
deterministically at emission instead of relying on prompt salience.

---

## What this PR changes

Closes #588.

- **`_build_focused_prompt` is now renderer-driven and async.** It assembles data;
  every block of prose lives in a managed prompt asset.
- **Typed criteria → authoritative Contract Expectations block**, narrative prose
  demoted below it (the #585 treatment, on the dev role).
- **Fill-only + ERROR CONTRACT sections reach the initial author.** The executor
  threads `scaffold.error_seam_instructions` onto `development.develop` envelopes via
  `_ERROR_SEAM_TASK_TYPES` — a table, not a string-literal test at the branch.
- **Four new prompt assets**; the existing Contract Expectations appendix is reused
  rather than duplicated.
- A degraded-mode fallback preserves today's plain-text assembly when no renderer is
  wired — the same shape `_build_dev_prompt` has always had.

## Tests — 9 new

**Six drive the REAL renderer against the REAL asset files** rather than a mock, so
handler/asset drift (an undeclared variable, a renamed placeholder) fails in CI
instead of silently rendering an empty section in production. That mattered here:
one of them caught a wrong stack id I had written (`fastapi_react` vs
`fullstack_fastapi_react`), and another caught a genuine wording bug — the narrative
appendix referenced an authoritative block that does not exist when a task has no
typed criteria.

**Three cover the executor threading**: dev task types get the seam, other task types
do not (a qa suite told to raise `ApiError` writes assertions against the wrong
thing), and author-mode runs with no manifest attach nothing.

Existing focused-prompt tests updated to await the now-async method. None weakened.

## Verification

- `tests/unit` — **5,841 passed**, 3 skipped
- ruff check + ruff format — clean
- Pre-existing on main, unrelated: 2 failures in `test_verification_contract_runner.py`
  (host has no node by design, #419) — confirmed identical on a clean `main`.

Not deployed — no rebuild, no restart. Agents would need rebuilding for the prompt
assets to take effect.

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -930,6 +930,12 @@ class DispatchedFlowExecutor(FlowExecutionPort):
     # Build task artifact filter (D3, §2.2)
     # ------------------------------------------------------------------
 
+    # Task types that author into the scaffold's fill slots and therefore need the
+    # manifest-derived error seam in their prompt (#588). A table, not a literal
+    # test at the branch — the repair path's equivalent lives in CorrectionRunner's
+    # failure_evidence assembly.
+    _ERROR_SEAM_TASK_TYPES: frozenset[str] = frozenset({"development.develop"})
+
     # Maps build task_type → which prior artifacts to inject.
     # by_producing_task: match on producing_task_type metadata
     # by_type / by_type_fallback: match on artifact_type for artifacts without provenance
@@ -1160,6 +1166,7 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                 prior_outputs,
                 all_artifact_refs,
                 stored_artifacts,
+                interface_manifest=interface_manifest,
             )
 
             # SIP-0087: executor owns Prefect task-run creation so task_run_id
@@ -1658,6 +1665,7 @@ class DispatchedFlowExecutor(FlowExecutionPort):
         prior_outputs: dict[str, Any],
         all_artifact_refs: list[str],
         stored_artifacts: list[tuple[str, ArtifactRef]],
+        interface_manifest: Any = None,
     ) -> TaskEnvelope:
         """Build chain context inputs and return an enriched envelope."""
 
@@ -1665,6 +1673,20 @@ class DispatchedFlowExecutor(FlowExecutionPort):
             "prior_outputs": prior_outputs,
             "artifact_refs": list(all_artifact_refs),
         }
+
+        # #588: the manifest-derived error-seam lines the correction path has
+        # carried since pf-34, threaded to the INITIAL author too. The raise
+        # convention lives in the fill-slot stub's docstring, which the first
+        # fill overwrites — so without this the author guesses the signature,
+        # every error path TypeErrors into a 500, and the run pays a correction
+        # to learn what the scaffold already knew. Data only; the block's prose
+        # lives in a managed prompt asset.
+        if envelope.task_type in self._ERROR_SEAM_TASK_TYPES:
+            from squadops.capabilities.scaffold import error_seam_instructions
+
+            error_lines = error_seam_instructions(interface_manifest)
+            if error_lines:
+                extra_inputs["error_contract"] = error_lines
 
         # Pre-resolve artifact contents for build tasks (D3). Fresh dispatches
         # see the ACCEPTED state only (pf-31 Fix E): rejected repair candidates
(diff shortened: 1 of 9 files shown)


jasonLadd (commented):
reviewed

jasonLadd (commented):
reviewed

jasonLadd (commented):
reviewed

PR #592, opened 2026-07-25
fix(verification): reject patches whose intra-package imports can't resolve (#591)

jasonLadd:
Closes #591.

Patch verification evaluated typed criteria **file by file**, so a patch whose imports
could not resolve passed every check and was **accepted**. The failure then surfaced as
a whole-suite collection crash with no attributable cause, and the correction budget
went to re-rolling a test file against source that could never import.

## The pf-37 case — enforcement manufactured the defect

Correction-00 emitted a coherent *pair*: its own `backend/models.py` plus a
`backend/routes.py` written against it. SIP-0100 correctly restored the frozen
`models.py` and kept `routes.py`. Each file is individually valid. The combination
imports seven names the frozen module never defines — `RunCreate`, `RunResponse`,
`RunListResponse`, `ParticipantJoin`, `ParticipantLeave`, `runs_db`, `_next_id`.

```
13:49:25  scaffold_integrity (repair path): frozen_path_emission backend/models.py
13:49:25  patch_verification  status=passed  checks=4  failed=-
13:49:28  patch_retest        status=FAILED
```

Nothing checked the pair. SIP-0100 has no notion of leaving the workspace resolvable,
and the typed checks read one file at a time — `routes.py` compiles perfectly alone,
and `from .models import RunCreate` is syntactically valid whatever `models.py`
contains. It's the same blind spot as the `ApiError` signature class (#588): **typed
checks cannot see cross-module agreement.**

## The change

`unresolved_imports(workspace_root)` walks the assembled workspace and resolves
intra-package imports against the modules actually present.

**Where it lives:** `emission_integrity.py`, beside the pf-31 Fix D syntax gate — same
concern (structural integrity of emitted Python), one step up from per-file
parseability. No new module for an existing concern.

**Where it runs:** inside `verify_patched_artifacts`, **before** the criteria loop. An
unimportable workspace makes every downstream verdict meaningless, and that temp
workspace is the one place the restored-plus-repaired combination actually exists.
Placement matters and is load-bearing: 3.4b enforcement replaces
`result.outputs["artifacts"]` in place *before* the repair overlay reaches verification
(`correction_runner.py` says so explicitly), so the check sees the true post-restore
pair. No executor changes were needed.

**Verdict:** `PATCH_FAILED` with a reason naming the file, module, and every missing
name — actionable in the log instead of another artifact archaeology session.

## Conservative by construction

A false rejection of a good patch is worse than a missed one, so it declines to judge:

- modules with a star-import or a module-level `__getattr__` (PEP 562)
- conditional definitions — `TYPE_CHECKING` blocks, `try/except ImportError` fallbacks
- unparseable targets (the syntax gate owns those)
- anything resolving outside the workspace — stdlib and third-party never appear

## Replay-verified against pf-37's real stored artifacts

| Workspace | Result |
|---|---|
| correction-00 (restored frozen + repaired routes) | **REJECTED** — names all seven missing symbols. Production accepted it. |
| original dev output (suite passed 11/11 live) | **clean** — no findings |

## Tests — 12 new

**Eight on the resolver:** the pf-37 shape, healthy packages, third-party/stdlib
ignored, absolute intra-workspace imports (`from backend.models import X`, which the
scaffold harness uses), star-import and conditional-definition abstention, unparseable
and absent targets.

**Four on the verifier**, including one that pins *why* the defect was invisible: the
same file-local `function_defined` criterion still passes on the very artifacts the new
gate rejects. A future refactor cannot weaken the criteria and call #591 fixed.

## Verification

- `tests/unit` — **5,833 passed**, 3 skipped
- ruff check + format — clean (`_module_level_names` decomposed to stay under the C901
  budget rather than adding an exemption)
- Pre-existing on main, unrelated: 2 failures in `test_verification_contract_runner.py`
  (host has no node by design, #419)

Not deployed — no rebuild, no restart.

### DIFF
--- a/src/squadops/cycles/emission_integrity.py
+++ b/src/squadops/cycles/emission_integrity.py
@@ -17,6 +17,7 @@ Pure functions — the caller owns storage, events, and the carry.
 from __future__ import annotations
 
 import ast
+from pathlib import Path
 from typing import Any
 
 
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #594, opened 2026-07-25
sip: propose Cycle Replay Harness — pinned fast-forward for phase-targeted iteration

jasonLadd:
Draft SIP for design review — adds `sips/proposed/SIP-Cycle-Replay-Harness.md`. No code.

## The case, measured

pf-37's phase boundaries from the executor log:

| Phase | Wall clock |
|---|---|
| Framing | **2h 06m** |
| Dev (7 tasks) | 30m |
| Builder | 6m |
| **QA + correction** | the only part under study |

~3 hours and ~40k generated tokens to reach minute one of the interesting part, once
per hypothesis. And the phases being paid for are the ones that no longer fail —
framing has produced a clean lint-quiet plan on four consecutive rolls; dev has run
7/7 clean on four.

## What it proposes

A cycle may declare a **replay pin**:

```yaml
execution_overrides:
  replay:
    source_run_id: run_010825617319
    resume_at: qa
```

Everything before the boundary is seeded from the source run's stored artifacts and
never dispatched. Everything at and after it runs for real — real generation, typed
acceptance, scaffold enforcement, correction loop.

## Why it's small

Seeding artifacts from the vault **already exists and is load-bearing**:
`execution_overrides.plan_artifact_refs` + `_seed_prior_artifacts` /
`_seed_skeleton_artifacts` is the same seam that seeds the interface manifest on every
measurement roll, and that seeded an entire walking skeleton during the Phase-0.5 spike
with zero new framework code. This generalises that seam to a declared pin point and
adds the one part that must not be optional.

## Two decisions worth reviewing

**Artifact-level, not LLM-level, for v1.** The intuitive design intercepts `LLMPort` and
replays recorded completions. Rejected for v1: you skip a phase *because you are not
testing it*, so re-running the fenced parser over a recorded completion buys nothing —
it tests the parser, which is what the pin exists to stop paying for. Raw completions
also aren't durably persisted today; that gap belongs to the LLM Emission Contracts SIP,
and depending on it would block this behind an unaccepted SIP. Artifacts already persist
completely for pf-27…pf-38, so **v1 introduces zero new capture paths** and nothing needs
un-building when raw persistence lands.

**Keyed on run + workload boundary, not prompt hash.** Our fixes *are* prompt changes —
#584, #585 and #588 were all prompt edits — so a prompt-keyed cache invalidates on
exactly the commits worth iterating on, silently falls through to live inference, and
quietly restores the 3-hour loop.

## Evidence integrity — the part I'd most like reviewed

This framework's central wound is runs reporting success for deliverables that don't
work. A replay harness manufactures convincing-looking runs cheaply, so §3.4 makes the
guardrail structural rather than procedural:

1. The pin is a **typed `CycleOutcome` attribute**, not a note field.
2. The N=5 baseline and FAY must **refuse** a pinned run at the `verification_integrity`
   aggregation seam — not by operator discipline.
3. The pin renders everywhere the run does (CLI, console, run report).

Bar: it should be impossible to accidentally cite a replayed run as evidence the system
works.

§3.5 states plainly what replay *cannot* tell you — it pins LLM variance (so it answers
"does this fix work against *this* input," never "does this fix work"), it cannot
validate a fix to a skipped phase (#588 is untestable under a pin that skips dev), and
it is a debugging accelerator rather than a measurement instrument.

## Phasing

- **Phase 1 (this SIP's v1):** pin point, prefix seeding, compatibility gate, outcome
  marking, aggregation refusal. No new capture path, no LLM-port changes.
- **Phase 2 — fault injection:** synthetic malformed emissions so every field defect
  becomes a repeatable regression test. Depends on the emission format.
- **Phase 3 — LLM-level replay:** parsing/extraction against recorded raw output, and
  Ollama↔Atlas A/B on identical inputs as migration conformance evidence. Same
  dependency.

2 and 3 are named so the v1 seam doesn't foreclose them; neither is in scope for
acceptance.

## Open questions in the draft

Pin granularity (workload boundary vs mid-workload), whether a pin should reference a
run or a curated promoted **fixture**, vault retention policy (none exists today, and
replay depends on it), and whether the compatibility gate belongs at create time or run
start.

### DIFF
--- a/sips/proposed/SIP-Cycle-Replay-Harness.md
+++ b/sips/proposed/SIP-Cycle-Replay-Harness.md
@@ -1,0 +1,237 @@
+# SIP-0XXX: Cycle Replay Harness — Pinned Fast-Forward for Phase-Targeted Iteration
+
+**Status:** Proposed
+**Authors:** SquadOps Architecture
+**Created:** 2026-07-25
+**Revision:** 1
+
+## 1. Abstract
+
+Every investigation into the correction loop currently costs a full cycle. pf-37 —
+a representative measurement roll — spent **2h 06m in framing, 30m in dev, and 6m in
+builder before the first QA task dispatched**, and the QA phase is the only part
+anyone was studying. Three hours of wall clock and roughly 40k generated tokens to
+reach minute one of the interesting part, repeated once per hypothesis.
+
+The phases being paid for are the phases that no longer fail. Framing has produced a
+clean, lint-quiet plan on four consecutive rolls; the dev phase has run 7/7 clean on
+four consecutive rolls. They are re-derived from scratch every time anyway.
+
+This SIP introduces a **replay pin point**: a cycle may be configured to start with a
+prior run's stored artifacts already in place, skipping every workload up to a chosen
+boundary and executing everything after it for real. Iteration on the correction loop
+drops from ~3 hours to minutes.
+
+The mechanism is deliberately small. Seeding artifacts into a run from the vault
+already exists and is load-bearing (`execution_overrides.plan_artifact_refs` +
+`_seed_prior_artifacts` / `_seed_skeleton_artifacts` — the same seam that seeds the
+interface manifest on every measurement roll, and that seeded an entire walking
+skeleton during the Phase-0.5 spike with zero new framework code). This SIP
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #595, opened 2026-07-25
sip(replay-harness): revision 3 — correct the FAY enforcement claim, add implementation order

jasonLadd:
Amends the merged Cycle Replay Harness SIP (#594). Docs only.

## 1. §4 asserted an enforcement point that does not exist

Rev 2 required replayed runs to be "refused by baseline aggregation and FAY at the `verification_integrity` seam."

Grepping `src/` and `adapters/` for Functional App Yield and for any N=5 / green-baseline aggregation returns **nothing**. Green counting is done by hand in cycle notes. There is no code path that could refuse a replayed run, because there is no code path that counts one.

New **§4.1** splits the obligation honestly:

- **Implementable now, required of Phase 1** — the typed, immutable replay attribute on `CycleOutcome` (reqs 1–2) and rendering on every reporting surface (req 5). `CycleOutcome` already lives in `verification_integrity` and is surfaced on the cycle detail GET, so this is a real seam.
- **Deferred but binding** — reqs 3–4 attach to whichever change first implements an aggregator. That change SHALL consult the attribute and SHALL NOT ship without a test asserting the exclusion.

Acceptance criterion 8 restated to match, so a reviewer can no longer approve it as achievable in Phase 1.

Until an aggregator exists, the only protection is that a replayed run is *unmistakably labelled* wherever it appears — which is why the rendering requirement is neither cosmetic nor deferrable.

## 2. New §5.1 — implementation order

Phase 1 spans the create DTO, checkpoint retention (port + both adapters), executor, SIP-0095 preflight, `CycleOutcome`, CLI, run report, and `console/` — a separate service in another language. Not one reviewable diff.

The slicing is constrained by one rule: **the evidence rails must not lag the mechanism**, or a window exists in which an unlabelled replayed run can be produced. So the rails ship first, against a mechanism that doesn't exist yet:

| Slice | Contents |
|---|---|
| 1 | evidence rails — `CycleOutcome` attribute + DTO + CLI/report rendering (nothing sets it yet) |
| 2 | retention — boundary checkpoints exempt from `max_keep` pruning |
| 3 | mechanism — `execution_mode`, source-checkpoint restore, no-dispatch, byte-equivalence test |
| 4 | compatibility policy — per-boundary sets in preflight |
| 5 | console visibility |

**Slices 1–3 are the first usable increment.** Slice 3 carries an interim gate: replay refused unless source and target share identical `contract_ref`, `plan_artifact_refs`, `prd_ref` and `build_profile`. Strict equality is a few lines and is strictly more conservative than §3.5's policy, so slice 4 *relaxes* an existing guard rather than adding a missing one. At no point does a usable-but-ungated replay exist.

### DIFF
--- a/sips/proposed/SIP-Cycle-Replay-Harness.md
+++ b/sips/proposed/SIP-Cycle-Replay-Harness.md
@@ -3,7 +3,7 @@
 **Status:** Proposed
 **Authors:** SquadOps Architecture
 **Created:** 2026-07-25
-**Revision:** 2 (review round 1 incorporated)
+**Revision:** 3 (review round 1; FAY-enforcement correction + implementation order)
 
 ## 1. Abstract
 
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #596, opened 2026-07-25
sip(0101): accept Cycle Replay Harness + Phase-1 implementation plan

jasonLadd:
Promotes the Cycle Replay Harness proposal to **accepted as SIP-0101** and adds the Phase-1 implementation plan. Docs only — no code.

Number assigned by `update_sip_status.py` (not hardcoded); registry `last_assigned` 100 → 101. The script now rewrites the body Status line too (#253), so only the H1 needed hand-syncing.

## Plan shape

Sliced by the SIP's sequencing rule — **the evidence rails must not lag the mechanism**, or a window exists in which an unlabelled replayed run can be produced:

| Slice | Contents | Tonight? |
|---|---|---|
| 1 | evidence rails — `CycleOutcome.replay` + DTO + CLI/report rendering (inert; nothing sets it yet) | ✅ |
| 2 | boundary retention — exempt workload-terminal checkpoints from `max_keep` pruning | ✅ |
| 3 | mechanism — `execution_mode`, checkpoint-sourced restore, no-dispatch, byte-equivalence test, interim strict-equality gate | ✅ |
| 4 | per-boundary compatibility policy | later |
| 5 | console visibility | later |

Slice 3 ships strict-equality compatibility so slice 4 **relaxes an existing guard** rather than adding a missing one. At no point does a usable-but-ungated replay exist.

## How much already exists

The status snapshot documents it: `RunCheckpoint` already carries the boundary state, `_restore_checkpoint_state` already rehydrates all of it, `completed_task_ids` → `skip_task_ids` already suppresses dispatch, and `CycleOutcome` is already derived and surfaced on the detail GET.

The mechanism is a narrow change at an existing hook (`dispatched_flow_executor.py` ~:246). The bulk of Phase 1 is rails, retention and gating — which is where the risk actually is.

## ⚠️ Time-critical decision in the plan

**pf-38's implementation run is executing under today's `max_keep=5`.** Its post-dev boundary will be pruned before the run ends; its post-builder boundary may not survive either. Three options are laid out:

1. **Accept it** — first replay source is a future run started after Slice 2 deploys (costs one roll)
2. Land Slice 2 and deploy mid-flight — kills pf-38
3. Hand-synthesise a checkpoint row — cheap, but hand-built state is exactly what the determinism contract exists to prevent; **not recommended**

**Recommendation: (1).** pf-38 is the first roll measuring #586/#588/#591 on a clean unified deploy, and that result is worth more than one night's iteration speed. Slices 1–3 land while it runs; the next roll becomes the first replay source with boundaries retained.

## Process note

CLAUDE.md separates acceptance (design commitment) from implementation planning (step 4, post-acceptance). Compressed into one PR at your request given the tonight constraint — flagging that reviewers see design and implementation detail together.

### DIFF
--- a/docs/plans/SIP-0101-cycle-replay-harness-plan.md
+++ b/docs/plans/SIP-0101-cycle-replay-harness-plan.md
@@ -1,0 +1,160 @@
+# SIP-0101 Implementation Plan — Cycle Replay Harness (Phase 1)
+
+**SIP:** `sips/accepted/SIP-0101-Cycle-Replay-Harness.md` (accepted 2026-07-25)
+**Reuses:** SIP-0079 (`RunCheckpoint`, `_restore_checkpoint_state`), SIP-0095 (create preflight),
+SIP-0096 (`CycleOutcome`).
+**Motivating case:** pf-37 (`run_010825617319`) spent 2h 06m in framing, 30m in dev and 6m in builder
+before the first QA task dispatched — and the QA phase was the only part under study. Four of the last
+five measurement rolls failed in QA/correction, each paying the same ~3h entry fee to reach it.
+
+## Intent
+
+Make the known-good prefix of a cycle **replayable** so iteration targets the phase actually under
+study. Phase 1 delivers: an execution mode, checkpoint-sourced prefix restore, durable boundaries, a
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #599, opened 2026-07-25
docs(ideas): process lexicon — one seeded vocabulary across all domains

jasonLadd:
Idea only — adds `docs/ideas/IDEA-Process-Lexicon-For-Agent-Vocabulary.md`. No code, no SIP number. Intended to grow into a SIP after the N=5 green baseline closes.

## Origin

Spotted during pf-39: the same task carries four names across four surfaces — `task_index: 1` (plan), `T1` (lead's prose), `m001` (dispatch id), `task_1` (evaluation artifact). They align numerically, so nothing broke. But it is the benign form of a class that has been expensive: `{run_id}` vs `{id}` prose-vs-contract divergence cost pf-31 a whole roll, and pf-33's `harness_boundary` expectation line *instructed* the violation two repair agents then obeyed.

## The part worth reviewing

The natural objection to a shared vocabulary is that it doesn't scale — nobody wants to author one for `group_run`, then `app_a`, then `app_p`, including apps that don't exist yet.

That dissolves on inspection: **the terms that have actually hurt are process vocabulary, not domain vocabulary.** *build*, *framing*, *fill slot*, `task_index`, *criterion*, *workload* — all describe how agents talk about the work. `RunEvent` and `pace_target` have never been the problem.

Which yields three lexicons, only one of them missing:

| Lexicon | Scope | Lives today |
|---|---|---|
| **Process** | framework-wide, domain-blind | **nowhere — the gap** |
| **Stack** | per stack, not per app | scaffold expander / Stack Blueprint |
| **Domain** | per app | **the PRD** — already a seeded input |

So there is never a per-app lexicon. A new app needs its PRD, which you were writing anyway.

## Two design commitments

**It must be an input, not an output.** Having the lead emit a glossary during framing moves the drift up a level and gives it authority. Whatever constrains generation cannot be downstream of it. Proposed shape is a prompt fragment under `prompts/fragments/` via PromptService — versioned, diffable, identical across roles, never re-dreamt per cycle. Existing seam, existing ownership rule.

**Honest accounting on enforcement.** A lexicon is instruction, and the standing lesson is that enforcement beats instruction. Enforcement needs a referent independent of the generator — SIP-0100 has `expand(manifest)`, typed acceptance has the contract, the A3 lint has the contract's paths. So the doc splits terms that name canonical identifiers (partially lintable — prose may *reference* an identifier, never restate it) from conceptual terms like *build*, where nothing is lintable and the durable fix is **collapsing the term** so the ambiguity stops existing.

Also caps the size: every token lands in every prompt of every role of every cycle, and pf-33 already showed what that costs (~18k-token QA prompt → 8192 truncation → collection crash).

### DIFF
--- a/docs/ideas/IDEA-Process-Lexicon-For-Agent-Vocabulary.md
+++ b/docs/ideas/IDEA-Process-Lexicon-For-Agent-Vocabulary.md
@@ -1,0 +1,62 @@
+# IDEA: A Process Lexicon — One Seeded Vocabulary, Every Domain
+
+## Target Release
+Vision item — candidate SIP after the N=5 green baseline (SIP-0098 §98.5) closes.
+
+### Status
+Idea / design-thinking draft
+
+### Owner
+Build / Architecture
+
+### Origin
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #600, opened 2026-07-25
fix(scaffold): declare success status in the manifest; reject qa tasks owning others' files

jasonLadd:
Two defects measured directly on pf-39 (`run_1e13c7b137a3`), both of which made a green roll depend on luck rather than on the pipeline working.

## 1. The skeleton could not satisfy its own contract

`Endpoint` had no success-status field. So:

- `_routes_source` emitted `@router.post("/runs", response_model=RunEvent)` — FastAPI defaults POST to **200**
- `_probes` hardcoded `"expect": {"status": 201}`

The two sides contradicted each other by construction. The only way a run went green was **a dev agent volunteering `status_code=201`** in the fill slot. pf-38's did. pf-39's did not, and that single token was the entire difference between `accepted` and `rejected` — same seeds, same frozen deploy, `vc-probe-runs` failing four times with `status 200 != expected 201`.

The asymmetry is the giveaway: `ErrorCode.http` already modelled *error* statuses, so the manifest could say "422 on validation failure" but not "201 on create."

Fix: add `Endpoint.success_status`, emit it on the scaffold-owned decorator, and make the probe **read the same field** so the contract and the skeleton can no longer disagree. A manifest that declares nothing keeps the historical 201 probe expectation.

An HTTP success status is *interface*, not implementation — it belongs on the frozen decorator where the fill dev cannot drop it.

## 2. Plan validation admitted an unsatisfiable QA task

pf-39's plan gave a `qa.test` task the three view fill slots as its `expected_artifacts`. Write authorization refused those emissions at runtime (`unauthorized_slot_emission` ×3) — so the task provably could not produce its own declared outputs.

The wasted task isn't the real cost. **A correction repair is scoped to the failing task's expected artifacts.** So when that task failed, `analyze_failure` correctly diagnosed the backend status-code defect in *another* task's file, and the repair rewrote the three frontend views instead. `patch_verification` passed it (file-local, `checks=2`); `patch_retest` rejected it 9 seconds later. One entire correction attempt that could not have fixed anything.

Fix: `validate_qa_artifact_ownership` rejects a `qa.test` task declaring any file the contract marks frozen or as a fill slot — **the same referent write authorization already uses**, applied at authoring time instead of emission time. Wired at both existing nets: dispatch raises, gate records a graceful rejection.

Deliberately narrow: it keys on the contract's owned surfaces rather than the QA namespace, so test files and documents are unaffected, and it does not apply to dev tasks (a dev task owning `routes.py` is exactly how bind mode works).

## Deploy note

The manifest now declares `success_status: 201` on `POST /runs`. **The group_run manifest + contract must be re-emitted and re-ingested** before the next measurement roll, and the launcher refs updated — same procedure as the #565 rename.

## Tests

8 new, all deterministic — no cycle needed to validate either fix:

- status lands on the decorator; endpoints declaring none stay bare (guards against blanket-applying it)
- **the probe follows the manifest rather than a constant** — a manifest declaring 202 yields a probe expecting 202 and a decorator emitting 202; had this stayed hardcoded, the pf-39 defect would simply recur with different numbers
- back-compat: no declaration → 201
- the four ownership cases, including the two that guard against over-rejection (a QA task owning its test file, a dev task owning a fill slot)

Full regression: **5512 passed, 1 skipped**. `ruff check` clean.

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -2652,6 +2652,10 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                         f"verification_contract: {e}"
                         for e in parsed_plan.validate_criteria_refs(contract)
                     )
+                    errors.extend(
+                        f"verification_contract: {e}"
+                        for e in parsed_plan.validate_qa_artifact_ownership(contract)
+                    )
 
         # Soft (warning/info-severity) structural violations are tolerated, not
         # rejected — but logged so the pass is never silent (a warning check can't
--- a/examples/03_group_run/interface_manifest.yaml
+++ b/examples/03_group_run/interface_manifest.yaml
@@ -54,7 +54,8 @@ api:
         response: "list[RunEvent]" }                                 # §5.2
     - { method: POST, path: /runs,              summary: create run,
         request: RunEventCreate, response: RunEvent,
-        errors: [validation_error] }                                 # §5.1
+        success_status: 201,                                         # pf-39: the probe
+        errors: [validation_error] }                                 # §5.1  pins 201
     - { method: GET,  path: "/runs/{run_id}",       summary: run details,
         response: RunEvent, errors: [run_not_found] }                # §5.3
     - { method: POST, path: "/runs/{run_id}/join",  summary: join run by participant name,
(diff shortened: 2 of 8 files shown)


jasonLadd (commented):
reviewed

PR #602, opened 2026-07-25
fix(scaffold): hold the scaffold-owned status code inside fill slots

jasonLadd:
**pf-40 proved that emitting the right skeleton is necessary and not sufficient.**

#600/#601 made the expander emit the status code, and I verified it loaded in-container before launching:

```python
# what the scaffold seeded at 17:44
@router.post("/runs", response_model=RunEvent, status_code=201)
```

```python
# what the dev agent emitted at 17:57
@router.post("/runs")
```

`POST /runs` answers 200 against a contract pinning 201 — the identical failure that rejected pf-39, on the deploy that had just fixed it. The agent also stripped `response_model` from every decorator and renamed both the handlers and their parameters.

## Why it happened

The stub says so in its own header — *"API route stubs — scaffold-owned signatures, fill-only bodies"* — and that is **instruction, not enforcement**. A fill slot is a whole-file emission, and nothing compares the result against the seed.

SIP-0100 enforces ownership at **file** granularity: a frozen path is restored wholesale. A fill slot cannot be, because rewriting the file is the producer's job. What it needs is enforcement at **line** granularity, on the parts the scaffold declares it owns.

## Scope: narrower than the owned surface, on purpose

| | |
|---|---|
| **Restored** | `status_code` |
| **Reported, not rewritten** | path, method, `response_model`, handler name, parameter names |

`status_code` is body-independent — nothing in a function body depends on the success code its decorator declares, so injecting it cannot break an implementation.

The rest is genuinely unsafe to rewrite from here. The producer renamed `payload` to `data` and used it throughout its body; restoring the scaffold signature would leave the body referencing a name that no longer exists. Rewriting `response_model` can turn a working handler into a 500 if the body returns a shape the model rejects. Those divergences surface as evidence so the class stays visible, and `endpoint_defined` already fails a wrong path or method.

Matching is by `(method, path-with-parameters-normalized)`, so a producer that renamed a path parameter — the pf-31 `{id}`/`{run_id}` class — still matches its scaffold counterpart and still gets its status code back.

## The record change

`BoundScaffoldRecord` gains `fill_seeds`, pinning the scaffold's original bytes per slot for exactly the reason `frozen` pins them: **restoration must never re-derive** (D2). Defaults empty, so records bound before this field deserialize cleanly and simply enforce nothing — an in-flight run cannot crash on it.

## Verification

**Replay-verified against pf-40's actual stored artifacts** (`art_10a07c22963f` seed vs `art_cb3054f9ea6f` emission), not a synthetic fixture:

```
@router.post("/runs", status_code=201)     <- restored
body intact: title=data.title ...          <- producer's implementation untouched
result parses                              <- splice landed in the right column
```

18 new tests, including the cases that would let this regress silently: a compliant emission passing through byte-identical (so enforcement can't mask its own no-op), an explicitly-wrong `status_code=200` rather than only an absent one, an unparseable emission, a route the scaffold never declared, and a record with no seeds.

Regression: **5531 passed, 1 skipped**. `ruff check` clean (decomposed `_routes` rather than adding a C901 exemption).

## Not deployed

The baseline is mid-measurement on `ccd6dda2`. This does not deploy until the current roll finishes and you decide.

### DIFF
--- a/src/squadops/cycles/bound_scaffold_record.py
+++ b/src/squadops/cycles/bound_scaffold_record.py
@@ -62,6 +62,12 @@ class BoundScaffoldRecord:
     frozen: tuple[FrozenArtifact, ...] = ()
     fill_slots: tuple[str, ...] = ()
     qa_namespace: tuple[str, ...] = ()
+    # The scaffold's ORIGINAL bytes for each fill slot. A fill slot is producer-owned in
+    # its body but scaffold-owned in its decorators/signatures, and enforcing that split
+    # needs the seed to compare against. Pinned here for the same reason ``frozen`` is:
+    # restoration must never re-derive (D2). Defaults empty so records bound before this
+    # field existed deserialize cleanly and simply enforce nothing.
+    fill_seeds: tuple[FrozenArtifact, ...] = ()
 
     def frozen_paths(self) -> frozenset[str]:
         return frozenset(f.path for f in self.frozen)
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #604, opened 2026-07-26
fix(correction): tell repairs the real model names instead of letting them guess

jasonLadd:
pf-41 died because repairs invented class names, and the guesses **compounded**. Here is the same file across one roll:

```
21:46  scaffold  ParticipantName, RunEvent, RunEventCreate
22:18  dev       RunEventCreate, ParticipantName                              <- correct
23:44  repair    Run, CreateRun, ParticipantJoin, ParticipantLeave
00:27  repair    RunCreate, ParticipantJoin, ParticipantLeave
00:53  repair    RunCreate, RunResponse, RunJoin, RunLeave, RunDetailResponse
```

**The dev agent wrote working imports. Three consecutive repairs turned them into code that cannot load.** None of the invented names exist — the real ones are `Participant`, `RunEvent`, `RunEventCreate`, `ParticipantName`, and they are fully determined by the manifest the run is already holding.

## Detection existed; instruction didn't

The unresolved-import gate (#591) catches exactly this and rejects the patch. On pf-40 it worked perfectly — it caught every bad repair and nothing landed.

But **detection only throws the patch away. It never says what the right names are**, so the next attempt guesses again, and the loop can burn its whole budget without converging. That is precisely how pf-40 died.

This adds the missing half. `model_surface_instructions(manifest)` derives the exact importable names and travels the same `failure_evidence` → authoritative-block transport already used by the error seam (pf-34), frozen ownership (SIP-0100 3.4b) and interface drift.

What a pf-41 repair would now have been told:

```
MODEL SURFACE (authoritative — apply exactly):
- `backend/models.py` is scaffold-frozen and defines EXACTLY these names:
  `Participant`, `RunEvent`, `RunEventCreate`, `ParticipantName`. Import only
  from this list — these are the complete contents of that module
- do NOT invent model names or aliases (`Run`, `RunCreate`, `RunResponse`,
  `RunJoin` and similar do not exist); an import of a name absent from the list
  above fails at load, the app never starts, and the patch is rejected
- `models.py` is frozen — if a name you want is missing, use the declared names
  and shape the response in the fill slot; do not edit or re-emit the model module
```

That third line matters as much as the first two: without it a repair "fixes" a missing name by editing `models.py`, scaffold enforcement restores the file, and the repair silently accomplishes nothing.

## Refactor

The four authoritative blocks in `_format_failure_summary` were copy-pasted branches, and a fifth pushed the function past the complexity limit. They are now a table plus a shared renderer, so the next seam is one row rather than another clause — and `interface_drift`'s dict-shaped entries are normalised in one place instead of having their own branch.

## Tests

8 new, and they assert the **whole path** — manifest → `failure_evidence` → rendered prompt text — not just the pure function. That is the #588 lesson in test form: that fix was correct and wired to a code path nothing used, so it changed no outcome until the wiring was found.

Specifically covered: every class the scaffold actually emits is named to the repair (derived from `expand()`, so it cannot drift); the five names pf-41 invented never appear in the authoritative list; the frozen statement is present; author-mode runs with no manifest inject nothing and render no empty header; and the correction runner really does populate the evidence key.

Regression: **5539 passed, 1 skipped**. `ruff check` clean.

## Not deployed

Baseline is paused at 0/5. Nothing deploys until you decide.

## Follow-up worth filing separately

pf-41 also showed a check that **errored** rather than passing or failing — `function_defined` pointed at a `.jsx` file — which turned patch verification into `unverifiable` and let the bad repairs land. That is why pf-41's damage accumulated while pf-40's did not. It is a distinct defect from this one and a route to shipping unverified code.

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -182,8 +182,16 @@ def _inject_deterministic_evidence(
       docstring; without it dev repairs guess ``ApiError(status_code=...,
       detail=...)`` and 500 every error path at the behavioral retest despite
       passing all typed checks (pf-33 corr-01, pf-34 corr-00).
+    - ``model_surface`` (pf-41): the exact importable names from the frozen
+      ``models.py``. Repairs invented them on three consecutive attempts,
+      degrading working imports into unimportable ones; the unresolved-import
+      gate rejects such a patch but never says what the right names are, so the
+      next attempt guesses again.
     """
-    from squadops.capabilities.scaffold import error_seam_instructions
+    from squadops.capabilities.scaffold import (
+        error_seam_instructions,
+        model_surface_instructions,
+    )
     from squadops.cycles.contract_expectations import expectation_lines
     from squadops.cycles.interface_conformance import detect_interface_drift
 
@@ -210,6 +218,10 @@ def _inject_deterministic_evidence(
     error_lines = error_seam_instructions(interface_manifest)
     if error_lines:
         failure_evidence["error_contract"] = error_lines
+
+    model_lines = model_surface_instructions(interface_manifest)
+    if model_lines:
+        failure_evidence["model_surface"] = model_lines
 
 
 def _locus_and_repair_target(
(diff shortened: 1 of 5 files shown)


jasonLadd (commented):
reviewed

PR #606, opened 2026-07-26
fix(scaffold): emit the in-memory store the manifest already declares

jasonLadd:
Closes #603.

The manifest declares `persistence: in_memory`. The skeleton emitted sixteen files, **none of which held the data** — no mention of storage anywhere, and no module-level state in the routes stub. So the planner invented a module on every roll.

That is correct engineering on the planner's part. Given a requirement and no file that owns it, writing one is the right move. The defect is upstream.

## Why an invented file is dangerous

It falls outside every mechanism that makes the rest of the workspace safe: nothing freezes it, no contract criterion names it, and its name, location and import style are re-derived per roll.

pf-40 died exactly there:

```python
# backend/store.py, as the dev agent wrote it
from models import RunEvent, Participant     # should be: from .models import ...
```

The backend never started. `tests_pass` reported a `ModuleNotFoundError`, and the behavioural probe came back **unverified rather than failed** — the app could not boot, so the check that would have told us whether the API works never ran. pf-41's plan reintroduced the same module.

## What this emits

```python
"""In-memory state — scaffold-owned (frozen).

The manifest declares in-memory persistence, so this module owns it: one store
per declared entity, keyed by the entity's id. Fill route bodies against these
names; do not define a second store elsewhere and do not edit this file.
"""

from __future__ import annotations

from .models import Participant, RunEvent

participant_store: dict[str, Participant] = {}
run_event_store: dict[str, RunEvent] = {}


def reset() -> None:
    """Clear every store — for test isolation between cases."""
    participant_store.clear()
    run_event_store.clear()
```

Three deliberate choices:

**Frozen, not a fill slot.** A producer that rewrites it has the scaffold bytes restored. If it were fillable the agent could reintroduce the very import this fixes.

**The routes stub imports it**, for the same reason it already imports `ApiError` — if the fill slot cannot see the seam, it invents one. That is the whole defect.

**`reset()` exists** because QA suites currently reach into a private dict to isolate cases (`backend.routes._runs.clear()` in pf-39), which means guessing the container's name. Now there is a sanctioned one.

## Contract impact

One frozen entry — `backend/store.py` and its hash. The manifest is unchanged, so its hash correctly does not move; only the derived frozen surface grows.

Both emission gates re-run in a container against this branch:

```
bare-skeleton:   GREEN (6/6)   structural checks pass, behavioural correctly not passing
reference-fill:  GREEN (6/6)   full contract passes
```

## Tests

8 new. The load-bearing one asserts that **every name the store imports from `.models` is actually defined there** — derived from `expand()` output rather than hardcoded, so it cannot drift. That is precisely the check that would have failed pf-40 before it ever dispatched.

Also covered: the import is relative and never bare (`from models import` is asserted absent); one store per declared entity; the routes stub wires it; it is not a fill slot; the module compiles; and a manifest with no entities still emits a valid module rather than a `reset()` with an empty body.

Regression: **5539 passed, 1 skipped**. `ruff check` clean.

## Deploy note

This changes the frozen surface, so the contract must be **re-emitted and re-ingested** and the launcher refs updated before the next measurement roll — unlike #602, which touched no contract. The manifest itself does not change.

## Not deployed

Baseline is paused at 0/5. Nothing deploys until you decide.


jasonLadd (commented):
reviewed

jasonLadd (commented):
reviewed

PR #607, opened 2026-07-26
fix(checks): AST checks skip non-Python files instead of erroring

jasonLadd:
Third in the stack: #604 → #606 → this. Addresses #605.

## The defect

Four of the five checks that parse Python had no file-extension guard:

| Check | Extension guard |
|---|---|
| `endpoint_defined` | none |
| `import_present` | **yes** |
| `field_present` | none |
| `function_defined` | none |
| `harness_boundary` | none |

Handed a `.jsx` file the four unguarded ones call `ast.parse`, raise, and return `error(parse_failed)` — which patch verification maps to `evaluator_error:<check>` and an `unverifiable` verdict.

The correct behaviour already existed, implemented once in `import_present`. Its siblings never got it. This extracts it into `_unparseable_source_skip` and applies it to all five, so `import_present` is single-sourced onto the same helper instead of keeping its own copy.

## Correcting my own framing

I originally filed #605 saying an unverifiable patch "lands unverified". **That was wrong**, and I've corrected the issue.

The executor already fails closed — `patched_artifacts` is assembled in memory and only applied on a pass; anything else returns `"continue"` and re-dispatches. pf-41's app also *booted* (404, not a startup failure), which independently confirms the invented-name file never reached the verified workspace.

What an evaluator error actually costs is a **wasted correction attempt**: a repair that may have been fine is discarded because verification cannot return a verdict. pf-41 burned three of five that way, which is why its loop could not converge. Real, worth fixing, but a convergence cost — not a correctness hole. The "verification must fail closed" item in the issue is already satisfied and has been struck.

## Deliberately untouched

`regex_match` and `count_at_least` legitimately target documents and globs. Guarding those would break the `qa_handoff.md` checks plans routinely author — pf-41's own plan had two.

Also extended the frontend suffix set to `.jsx/.tsx/.mjs/.cjs`; the original only listed `.ts/.js`, and `.jsx` is what the QA agent actually emits.

## Tests

27 new — every AST check across every frontend suffix, plus markdown, plus two that guard the guard:

- a real Python file with real test functions **still evaluates and passes** (the guard must not disable the checks it protects)
- a genuinely unparseable Python file **still errors** rather than being silently skipped — this is about file type, not about hiding failures

Regression: **5574 passed, 1 skipped**. `ruff check` clean.

## Stack

Deploying the tip of this stack gets all three fixes. #606 changes the frozen surface, so a re-emit + re-ingest + launcher ref update is required before the next roll.

### DIFF
--- a/src/squadops/cycles/acceptance_checks.py
+++ b/src/squadops/cycles/acceptance_checks.py
@@ -260,6 +260,27 @@ def _decorator_route(decorator: ast.expr) -> tuple[str, str] | None:
     return None
 
 
+# Every AST-based check parses Python. Handed anything else, `ast.parse` raises and the
+# check reports `error` — which `patch_verification` maps to `evaluator_error:<check>` and
+# turns into an UNVERIFIABLE patch: neither accepted nor rejected, so it lands unverified.
+# pf-41 lost a roll to exactly that (a `function_defined` aimed at a `.jsx` test file), and
+# the contrast with pf-40 is the proof: same bad repairs, but there verification returned a
+# verdict, rejected them, and nothing landed. Skipping is the honest outcome for a file we
+# cannot analyse — `import_present` already did this; the other four did not (#605).
+_FRONTEND_SUFFIXES = frozenset({".ts", ".tsx", ".js", ".jsx", ".mjs", ".cjs"})
+
+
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

jasonLadd (commented):
reviewed

PR #608, opened 2026-07-26
fix(scaffold): the router takes no prefix — state it, and enforce it

jasonLadd:
Fourth in the stack: #604 → #606 → #607 → this. **This is pf-41's 404 — the only thing that actually failed that roll.**

## What happened

The scaffold seeded `router = APIRouter()`. The dev agent emitted:

```python
router = APIRouter(prefix="/api")
```

Every route re-homed under a second `/api`. The contract probe asked for `POST /runs`, found nothing, and a **completely healthy app returned 404 to its own contract**.

## Why the agent did it, and why that matters

Its reasoning was sound; its conclusion was wrong. The scaffold's own frontend calls `/api/...`, so it made the backend match.

What it could not see is that the proxy **strips** that prefix before the request arrives:

```js
// frontend/vite.config.js — frozen, and the agent never reads it
'/api': { target: 'http://localhost:8000', rewrite: (path) => path.replace(/^\/api/, '') }
```

The manifest records the same intent — `base_path: ""`, commented "dev integration via proxy" — and the agent is shown neither. **The fact is knowable and authoritative and was simply never in front of it.** Fifth instance of that pattern this week.

## Two changes: instruction and enforcement

**State it** where the agent is already working — the routes stub docstring now says the router takes no prefix, the proxy strips `/api`, these paths are already the full backend paths, and adding one makes the app 404 its own contract.

**Enforce it** — `fill_slot_integrity` now restores `router = APIRouter(...)` when a producer changes it, exactly as it restores `status_code`. Safe for the same reason: the prefix decides where routes *register*, never what a handler body *references*. Unlike signatures or `response_model`, putting it back cannot break the code around it.

## Why this cost a whole roll instead of one attempt

Trace the router line across pf-41:

| Time | Router | |
|---|---|---|
| 21:46 scaffold | `APIRouter()` | correct |
| 22:18 dev | `APIRouter(prefix="/api")` | **the 404** |
| 23:44 repair | `APIRouter(prefix="/runs")` | worse |
| 00:27 repair | `APIRouter()` | **correct — rejected** |
| 00:53 repair | `APIRouter()` | **correct — rejected** |

Two later repairs **had the fix**. Both were rejected because the same emission also invented model names, so the import gate threw out the entire patch and the correct prefix went with it.

The loop held the right answer twice and discarded it for collateral damage in the same file. That is the case for #604 (the model-names fix) at the base of this stack — with it, those repairs plausibly land and pf-41 converges.

## Verification

**Replayed against the real stored artifacts**, not a fixture — pf-41's scaffold seed and its dev emission fed through the restorer:

```
router = APIRouter()                          <- prefix stripped
@router.post("/runs", status_code=201)        <- #602's restoration still holds
handler bodies and imports intact, result parses
```

That app would have passed the probe.

6 new tests including that replay, plus: any invented prefix stripped (not just `/api`), bodies survive, a compliant emission passes through byte-identical, and a seed with no router enforces nothing.

Regression: **5580 passed, 1 skipped**. `ruff check` clean. Both emission gates re-run in a container: `bare-skeleton` 6/6, `reference-fill` 6/6.


jasonLadd (commented):
reviewed

PR #609, opened 2026-07-26
docs(ops): night triage runbook — root cause, not narrative

jasonLadd:
Adds `docs/ops/NIGHT_TRIAGE_RUNBOOK.md`. Docs only.

Written so an unattended session can take a failed roll from terminal verdict to a fix on a branch without guessing. It encodes the method that produced this week's real diagnoses **and the two wrong ones I had to retract** — the wrong ones are why it exists, and both came from the same move: explaining a symptom instead of tracing it.

## The load-bearing step

Trace the failing file across **every** stored version, oldest to newest, rather than reading the latest one. Both genuine findings came from that alone:

- imports going correct → `Run, CreateRun` → `RunCreate` → a five-name invention, showing **the developer was right and the repairs degraded it**
- the router going `APIRouter()` → `prefix="/api"` → `prefix="/runs"` → `APIRouter()` → `APIRouter()`, showing the loop **held the correct fix twice and discarded it** for unrelated reasons

Neither is visible in a single artifact. Both are obvious in the sequence.

## The distinctions that make a diagnosis wrong when collapsed

- **`failed` / `unverified` / `skipped` / `error` are four different outcomes.** `unverified` means the check never ran — usually the app didn't boot — which is not a pass and actively masks whatever it would have found.
- **The artifact vault keeps rejected emissions.** The newest stored file is frequently not what executed. I skipped this and reported that broken repairs "landed"; they had not, and one grep of the verification call site would have prevented it.
- **The analyzer's diagnosis is a hypothesis.** pf-41 produced five different ones across five attempts; the third was correct and the loop then spent its remaining attempts elsewhere.

## Root cause phrasing

Push past "the model did something wrong" to what allowed it. The recurring shape, five times in one week:

> The system holds the authoritative answer and never puts it in front of the thing that needs it, so the model improvises.

When a diagnosis fits that shape the fix is to **materialise or inject the fact**, not add a detector — a detector rejects the guess but never supplies the answer, so the next attempt guesses again.

## Unattended authority, stated plainly

Read, query, replay, branch, PR, file issues, run the regression and gates — all without asking. Deploy, rebuild, restart, merge, or touch frozen seeds — never without an explicit rule.

And the mistake that cost three hours, recorded so it isn't repeated: **a deploy freeze is not a work freeze.** After a roll failed I wrote "do not implement anything" into my own watchdog and sat idle while the diagnosis was already complete. The constraint was only ever "nothing deploys".

## Also covered

Proving a fix by replaying it against the failed roll's real stored artifacts rather than a fixture; preferring enforcement over instruction where the target is body-independent (a status code or router prefix can be restored safely, a signature cannot); stacking fixes so the tip is deployable; and flagging when a change moves the frozen surface and needs a re-seed.

### DIFF
--- a/docs/ops/NIGHT_TRIAGE_RUNBOOK.md
+++ b/docs/ops/NIGHT_TRIAGE_RUNBOOK.md
@@ -1,0 +1,201 @@
+# Night triage runbook — root cause, not narrative
+
+How to take a failed measurement roll from terminal verdict to a fix on a branch, unattended,
+without guessing. Written after the night of 2026-07-25/26, which produced two correct
+diagnoses, two wrong ones I had to retract, and three idle hours.
+
+The wrong ones are the reason this exists. Both came from the same move: **explaining a
+symptom instead of tracing it.**
+
+---
+
+## The standard
+
+A diagnosis is finished when you can name the line that caused the failure and show the
+evidence. Not "the agent hallucinated names." Not "the repair loop didn't converge." A file,
+a line, and the artifact or log entry that proves it.
+
+If you cannot get there, say so and stop — an honest "unresolved, here is what I ruled out"
+is worth more than a plausible story. A plausible story gets acted on and wastes the next
+roll.
+
+---
+
+## Step 1 — verdict, then immediately past it
+
+```sql
+SELECT verdict, jsonb_pretty(summary::jsonb)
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewwed

Issue #598, opened 2026-07-25
packaging: emitted container can't build or run — Dockerfile/nginx verified by no criterion


jasonLadd:
## Summary

pf-38 went green — `verdict=accepted`, 10/10 checks, zero corrections, and the
application genuinely works (independently reassembled and boot-tested). But **the
container packaging it emitted cannot build or run as written.** Three separate defects,
and **no contract criterion touches any of them**, because nothing in the contract ever
builds the image or starts the container.

The gap is not "the app is broken." The app is fine. The gap is that *the verified
surface stops at the application and excludes its packaging*, so a cycle can be
legitimately green while producing a deliverable nobody can start.

## The three defects (pf-38, `run_ad3935096c84`)

Reproduced by assembling the run's artifacts with the executor's last-wins rule and
running `docker build` / `docker run`.

**1. `npm ci` with no lockfile — build fails immediately.**

`Dockerfile` stage 1:
```dockerfile
COPY frontend/package.json frontend/package-lock.json* ./
RUN npm ci
```
The glob makes the lockfile optional, but `npm ci` *requires* one. The emitted
`frontend/` contains `index.html`, `package.json`, `src/`, `vite.config.js` — **no
`package-lock.json`**. `npm ci` exits non-zero.

**2. `dist-packages` vs `site-packages` — build fails at the COPY.**

```dockerfile
COPY --from=backend-deps /usr/local/lib/python3.11/dist-packages \
                         /usr/local/lib/python3.11/dist-packages
```
```
ERROR: failed to compute cache key:
  "/usr/local/lib/python3.11/dist-packages": not found
```
`python:3.11-slim` installs to `site-packages`. `dist-packages` is a Debian
*system-python* convention that does not exist in the official python images.

**3. Debian nginx's default site is never removed — every `/api/*` 404s.**

The image installs nginx via `apt-get`, which ships
`/etc/nginx/sites-enabled/default` holding `listen 80 default_server`. `nginx.conf`
includes **both** `conf.d/*.conf` (line 60) and `sites-enabled/*` (line 61). The app's
config lands in `conf.d/default.conf` and loses to the packaged default server.

Observed on the running container:

| Request | Result |
|---|---|
| `GET /` | **200** — works *by coincidence*, the default site's webroot is the same `/var/www/html` |
| `POST /api/runs` | **404 from nginx** (HTML error page, not the backend's JSON) |
| `:8000/runs` *inside* the container | **200** — the backend was healthy all along |

[...]

PR #610, opened 2026-07-26
sip(proposed): Stack Blueprint Contract — give a stack a type

jasonLadd:
Draft SIP for design review — adds `sips/proposed/SIP-Stack-Blueprint-Contract.md`. No code.

## The problem, measured

Dispatch is **not** the issue — `_EXPANDERS` is already a registry and adding a key touches no branching logic. The issue is that "a stack" is five facts in five places, and four of them fail *silently*:

| A stack must declare | Where it lives | Fails how |
|---|---|---|
| how to expand the skeleton | `_EXPANDERS` | loudly (`ValueError`) |
| which files are fill slots | **inline in `fill_slot_paths()`** | **silently — returns the FastAPI map** |
| QA test directories | `_QA_TEST_NAMESPACES` | silently — every QA write unauthorized |
| harness entry modules | `_HARNESS_ENTRY_MODULES` | silently — boundary check never fires |
| the typed checks' vocabulary | `resolve_check_stack()` | silently — checks skip |

`fill_slot_paths` guards on whether the stack is *registered*, then returns `("backend/routes.py", *views)` hardcoded — with a docstring already conceding "as more stacks land in 99.4 they carry their own slot map alongside their expander." **A second stack registered today inherits the FastAPI slot map and nothing complains.**

There are also already **two stack vocabularies**: the manifest says `fullstack_fastapi_react`, the acceptance checks branch on `stack != "fastapi"`. Bridged by `resolve_check_stack()`, drifted apart.

## What it proposes

Give a stack a type — one `StackBlueprint` per stack, registered once. Every public function becomes a lookup; a missing field is an import-time type error rather than a wrong answer three hours into a cycle.

## The part I'd most like reviewed: why it argues against accepting itself

**Every field in the sketch is FastAPI-shaped thinking.**

- `analysable_suffix: str` assumes one analysable language per stack
- `harness_entry_modules` assumes a Python-style import boundary
- `qa_test_namespace` assumes directory-prefix test ownership
- `fill_slots` assumes slots are derivable from declared entities and routes

Generalising from a sample size of one doesn't produce a general contract. It produces **the FastAPI contract with generic field names** — worse than no contract, because it looks authoritative and stack #2 will quietly bend to fit rather than reveal the mismatch.

So the acceptance gate is **the existence of a second real stack**, with the schema written against both. The document exists now to hold the problem statement and to stop the schema being invented under deadline pressure by whoever is mid-way through adding that stack.

## Sequencing

1. **Now, no SIP needed** — consolidate today's five facts into one object with today's fields. Pure refactor; exact test (`expand()` byte-identical, contract hashes unmoved, gates 6/6). Removes the silent-omission failure mode immediately and prejudges nothing.
2. **When stack #2 is scoped** — promote this, write the schema against two stacks, reconcile the vocabularies.
3. **After** — migrate the typed checks off their hardcoded `.py` onto the blueprint's declared source language. That hardcode is what made a check *error* instead of skip on a `.jsx` file, costing pf-41 three of five correction attempts (#605).

## Open questions in the draft

Whether the blueprint owns the container/packaging set (the Functional App roadmap already claims it does, and #598 measured LLM-authored packaging failing on two consecutive rolls with *different* defects each time); one blueprint per stack vs. composed backend + frontend halves; whether `expand` is a callable on the blueprint or data a single generic expander consumes; and where blueprints live, since `scaffold.py` is already large.

### DIFF
--- a/sips/proposed/SIP-Stack-Blueprint-Contract.md
+++ b/sips/proposed/SIP-Stack-Blueprint-Contract.md
@@ -1,0 +1,131 @@
+# SIP: Stack Blueprint Contract
+
+## Status
+Draft (proposed)
+
+**Targets:** the release that adds the **second** stack — deliberately not before. See
+§"Why this is not ready to accept".
+**Builds on:** SIP-0099 (Contract-First Build Scaffolding), which established the expander
+and the fill-slot/frozen split for `fullstack_fastapi_react`; SIP-0098 (verification
+contracts derived from the manifest); SIP-0100 (scaffold ownership enforcement).
+**Motivating case:** adding a second stack today means finding and updating **five separate
+per-stack surfaces**, four of which fail silently when missed. One of them —
+`fill_slot_paths` — already hardcodes the FastAPI slot map behind a guard that only checks
+whether the stack is *registered*, so a second stack would inherit `backend/routes.py` as a
+fill slot and nothing would object.
+
+---
+
+## Summary
+
+Give a stack a **type**. Today "a stack" is an identifier that indexes four module-level
+dicts plus one function with the answer written inline. Replace that with a single
+`StackBlueprint` object per stack, registered once, so adding a stack is one object and a
+missing field is an import-time type error rather than a wrong answer three hours into a
+cycle.
+
+This SIP does **not** propose the consolidation of today's five facts — that is a pure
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #611, opened 2026-07-26
docs: 1.4 cut gate, the unreleased changelog, and the SIPs the roadmap forgot

jasonLadd:
## What this fixes

The roadmap's one-line summary of the 1.4 cut gate disagreed with the plan it
points at. Rev 2 of the evidence arc raised the bar from **two** consecutive
benchmark runs to **three**; `docs/ROADMAP.md` kept the old number. The same gate
had two values in two documents, and the roadmap is the one people read first.

| | before | after |
|---|---|---|
| consecutive runs | ≥2 | **≥3** |
| manifest mode | unstated | **squad-authored only** |
| wording | "zero manual intervention" | "zero manual **file modification**" |
| descoped browser probe | unstated | verdict is `verified_executable` |

## Why the extra qualifiers, not just the number

Two clauses had been dropped from the summary, and both change how the gate reads.

**Squad-authored-manifest mode.** A hand-authored interface manifest is itself a
manually authored file, so a run seeded with one cannot satisfy "zero manual file
modification". The plan states this explicitly; the roadmap didn't. The omission
is actively misleading right now — the group_run measurement rolls are all seeded
with the hand-written `examples/03_group_run/interface_manifest.yaml`, and reading
only the roadmap makes them look like cut-bar evidence. They aren't; they're
integration shakeout under the plan's own manifest lifecycle.

**`verified_executable`.** If the browser probe descopes out of the 1.4 floor, the
release names what it actually proved rather than claiming functional verification
it didn't perform.

The "intervention" → "file modification" change matches the acceptance statement
word for word, because the distinction is what makes the manifest point land.

## Verification

- Swept `docs/`, `sips/`, `README.md`, `CLAUDE.md` for every cut-gate/cut-bar
  statement — this was the **only** stale one. The plan (`1-4-evidence-arc-plan.md`,
  three places) and the strategy doc (`IDEA-Functional-App-Obstacles-and-Roadmap.md`)
  already said ≥3.
- `tests/unit/architecture/test_docs_version_sync.py` — 13 passed.

Docs only; no behavior change.

## Source of truth

`docs/plans/1-4-evidence-arc-plan.md` lines 250–257 (cut gate + manifest-mode
restriction) and line 26 (rev-2 bar raise). The roadmap now summarizes it without
contradicting it.

---

## Update — two more accuracy gaps folded in

Same class of problem, same file set, so they ride here rather than in separate PRs.

### CHANGELOG `[Unreleased]` was empty against 133 merges

Everything since 1.3.1 — the whole 1.4 arc — was unrecorded. Now reconstructed and
grouped by arc rather than listed per-PR:

- the golden-path scaffold stack (SIP-0098 / 0099 / 0100 / 0101)
- verification evidence integrity (SIP-0096 Phases 1–3)
- correction and repair convergence — the bulk of the volume, and the work that had
  to land before any scaffold claim could be measured
- the scaffold's own contract, acceptance checks and emission parsing, plan authoring,
  runtime and ops, removals, SIP lifecycle

Bare `#NNN` in that section are PRs; a note at the top says so, since the existing
sections mix issue and PR references.

### ROADMAP's v1.4 entry predated three of its own SIPs

Written 2026-07-14. SIP-0098 (accepted 07-16), SIP-0100 (07-22) and SIP-0101 (07-25)
appeared **nowhere** in the file, and Contract-First Build Scaffolding was named only in
prose, never as SIP-0099. The entry now names all four with their current implementation
state, and says plainly that Lane S has not started and is the arc's critical path —
which is the single most useful fact about 1.4 right now.

### README had two different stale test counts

"4,700+" in the overview, "3,030+" in the directory tree, in the same file. Actual is
5,580; both now read "5,500+".

Architecture guards pass (19) — includes docs version-sync and doc-ref existence.


jasonLadd (commented):
reviewed

jasonLadd (commented):
reviewed

PR #612, opened 2026-07-26
fix(plan): prove frozen-file checks can pass before dispatching the run

jasonLadd:
## What broke

pf-42's plan was gated on human review, passed every existing net, and was
**unwinnable before it started.**

Two dev tasks carried error-severity typed checks aimed at files the scaffold owns:

| Task | Check | Reality |
|---|---|---|
| 0 | `field_present` — `RunEvent.meeting_location` in `backend/models.py` | the manifest declares it `location` |
| 2 | `import_present` — `backend.routes` in `backend/main.py` | the frozen file imports `from .routes` |

Both files are frozen. Scaffold ownership enforcement restores a frozen-file emission
before the check re-runs, so the repair loop's only move is undone every time. The tasks
fail identically on all five correction attempts, and a three-hour roll is lost to a
plausible-looking field name.

## What this adds

Frozen files aren't a guess — the manifest expands to them deterministically before any
agent runs. So a check aimed at one is **decidable at authoring time**: materialize the
skeleton, run the check, see.

`frozen_check_violations()` does exactly that, wired into the gate-time plan validation
that records a graceful system rejection (#473), beside the criteria-refs and
qa-ownership nets it belongs with. With `framing_max_rerolls=2` configured, a rejection
re-rolls framing instead of burning the afternoon.

Replaying the **real stored pf-42 plan** through it:

```
REPLAY OF THE REAL pf-42 PLAN -> 2 violation(s)

 * Task 0 (backend run model): field_present on 'backend/models.py' can never pass —
   that file is frozen (scaffold-owned), and the skeleton it will actually contain
   fails this check (fields_missing; missing ['meeting_location'], file declares
   ['datetime', 'distance', 'id', 'location', 'pace_target', 'participants',
   'route_notes', 'title']). No repair can fix it: a frozen-file emission is restored
   before the check re-runs, so this fails identically on every correction attempt
   until the budget is spent. Assert what the scaffold declares, or drop the check

 * Task 2 (backend app main): import_present on 'backend/main.py' can never pass — ...
```

The message names the missing field *and* what the file actually declares, so the fix is
obvious rather than a hunt.

## Two deliberate limits

**Only `severity=error`.** A warning cannot block a build (RC-9), so a failing warning
costs nothing and is not worth rejecting a plan over.

**Only `status=failed` rejects.** An evaluator that *errors* has proved nothing about the
plan, and rejecting on it would let a single broken checker block every plan in the
system. Those are logged and skipped — the same check still runs for real at task
verification, where fail-closed applies. This is the one place where failing open is the
right call, and it is scoped to "we could not decide", never to "we decided it fails".

## Why this shape and not another detector

The recurring defect all week has been: *the system holds the authoritative answer and
never puts it in front of the thing that needs it.* A detector that merely rejects the
guess doesn't supply the answer, so the next attempt guesses again. Here the authoritative
answer is the frozen file itself, and the rejection quotes it back.

## Drive-by: one materializer, not three

Two copies of the expand-and-write loop had grown in `scripts/dev/`, and **only one
carried the chroot guard**. `scaffold.materialize()` is now the single writer; both
scripts delegate to it, so the safety check can't be the thing a third caller forgets.
The traversal test moves onto the canonical helper. (A third copy lives in
`tests/unit/cycles/test_verification_contract_runner.py`; left alone to keep this diff
honest.)

## Verification

- **Replay** against pf-42's real stored plan artifact — both violations, no false ones.
- 8 new tests, including both real failing specs, plus the cases that must still be
  admitted: the same check with the correct field name, checks on fill slots (empty at
  authoring time — judging them would reject every plan), warning severity, prose
  criteria, and an unparseable manifest.
- Regression **5588 passed**, 1 skipped. ruff clean.
- Contract gates re-run in-container against this tree: **lint GREEN, bare-skeleton 6/6,
  reference-fill 6/6**.
- **Contract emission unchanged** — `content_hash` stays `26503867…`, so the seeded v6
  contract remains valid and **no re-seed is needed** before the next roll.

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -37,6 +37,7 @@ from squadops.cycles.acceptance_evaluation import resolve_check_stack
 from squadops.cycles.agent_config import build_agent_resolver
 from squadops.cycles.build_completeness import compute_missing_required_files
 from squadops.cycles.checkpoint import RunCheckpoint
+from squadops.cycles.frozen_check_validation import frozen_check_violations
 from squadops.cycles.models import (
     ArtifactRef,
     Cycle,
@@ -2656,6 +2657,18 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                         f"verification_contract: {e}"
                         for e in parsed_plan.validate_qa_artifact_ownership(contract)
                     )
+                    # pf-42: a typed check aimed at a frozen file is decidable right
+                    # now — the skeleton those files will contain is deterministic.
+                    # A failing one makes the plan unwinnable (frozen emissions are
+                    # restored, so the repair loop can never converge), and it costs
+                    # milliseconds to prove instead of a three-hour roll.
+                    if binding_content is not None:
+                        errors.extend(
+                            f"verification_contract: {e}"
+                            for e in await frozen_check_violations(
+                                parsed_plan, contract, binding_content
+                            )
+                        )
 
         # Soft (warning/info-severity) structural violations are tolerated, not
         # rejected — but logged so the pass is never silent (a warning check can't
--- a/scripts/dev/contract_gate.py
+++ b/scripts/dev/contract_gate.py
@@ -45,7 +45,7 @@ import tempfile
 from pathlib import Path
 
 from squadops.capabilities.handlers.probe_runner import run_probes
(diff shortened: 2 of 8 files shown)


jasonLadd (commented):
reviewed

PR #613, opened 2026-07-26
fix(plan): show the plan author what the frozen files declare

jasonLadd:
Companion to #612. That one proves a frozen-file check can pass before the run is
dispatched; this one means the author never had to guess in the first place.

## The root cause, which is narrower than "the model got it wrong"

Here is the **entire** file index a bind-mode plan author receives today:

```
- backend/routes.py: bind vc-routes-endpoints, vc-routes-apierror, vc-routes-compiles
- frontend/src/views/RunsListView.jsx:  contract-owned (no per-file typed criteria)
- frontend/src/views/CreateRunView.jsx: contract-owned (no per-file typed criteria)
- frontend/src/views/RunDetailView.jsx: contract-owned (no per-file typed criteria)
```

Four fill slots. The group_run skeleton has **seventeen** files — the other thirteen are
frozen, and not one of them appears anywhere in the authoring prompt.

So pf-42's author wrote `field_present` on `backend/models.py` — a file it had never been
told exists, let alone what it declares. It guessed `meeting_location` because guessing
was the only move available. Same for `import_present` on `backend.routes` against a
module that writes `from .routes`.

There is a second edge: the existing appendix *does* say "do not author typed checks for
a **covered** file". But covered means the four fill slots. Frozen files aren't covered,
so that instruction never reached the files that caused this. **The rule was scoped to
the wrong surface.**

## What the author now sees

```
- `backend/main.py` — functions health; imports `.errors`, `.routes`, `fastapi`, ...
- `backend/models.py` — defines Participant(id, name), RunEvent(id, title, datetime,
  location, distance, pace_target, route_notes, participants), RunEventCreate(...),
  ParticipantName(name); imports `pydantic`
- `backend/errors.py` — defines ApiError; functions register_error_handlers; ...
- `conftest.py` — functions client; imports `backend.main`, `fastapi.testclient`, ...
- `frontend/vite.config.js`
...
```

Both pf-42 defects are answered on their face: `RunEvent` lists `location` and no
`meeting_location`; `main.py` imports `` `.routes` `` and not `backend.routes`. Imports
render **as written**, which is the point — the relative form is what makes the mismatch
visible.

Plus the rule, in the managed asset: don't assign a frozen file as `expected_artifacts`,
and don't write a check against one unless the index proves it passes — because a check
on a frozen file either can't fail (verifies nothing) or can't pass (kills the plan).

## Derived, not described

The index is produced by parsing the **expanded skeleton itself**. Nothing is
hand-maintained, so it cannot drift: change a template and the index changes with it.
There's a test that pins exactly this by swapping the expander and asserting the index
follows.

## Shape

Mirrors the bind-criteria section it sits beside, so there's no new pattern to learn:

- executor injects **data only**; instruction prose lives in a managed prompt asset (#448)
- bind mode only, dev/qa proposers only (strategy proposes guidance, authors no checks)
- author mode and non-scaffold stacks render **byte-identically** to today

One deliberate deviation worth review: the manifest is read from the **operator-seeded**
rail rather than `_load_interface_manifest_for_run`, which returns `None` for framing runs
by design (#496) — and framing is exactly when the plan is authored. #496 exists so a
framing run does not expand or carry skeleton *files*; describing the interface in a
prompt materializes nothing, so this stays inside that rule. Flagging it explicitly
because it looks like a bypass at a glance.

## Verification

- 17 new tests: the index answers both pf-42 questions, tracks the expander rather than a
  hand-written description, covers every frozen file and no fill slot, and degrades
  cleanly (no manifest, non-Python files, unparseable source).
- A **live render** of the real managed asset, asserting the field names and relative
  imports actually survive into the prompt — not just that the function returns them.
- Regression **5610 passed**, 1 skipped. ruff clean.
- Contract gates in-container: **lint GREEN, bare-skeleton 6/6, reference-fill 6/6**.
- **Contract emission unchanged** — `content_hash` still `26503867…`, so the seeded v6
  contract stays valid and **no re-seed** is needed before pf-43.

## What this does not do

It cannot guarantee the author reads the index — that's only knowable from a live framing
run, which is exactly why #612 exists as the deterministic backstop. Supply the fact,
verify the result; neither half replaces the other.

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -282,12 +282,29 @@ class DispatchedFlowExecutor(FlowExecutionPort):
             # TypedChecks. Absent contract = author mode = today's behavior.
             verification_contract = await self._load_contract_for_run(cycle, run)
 
+            # pf-42: the proposer binds criteria for the fill slots but is told nothing
+            # about the frozen files, so a check it wants on one is written against an
+            # invented interior. The manifest is what those files expand from, so it is
+            # the authority on their contents.
+            #
+            # Read from the OPERATOR-SEEDED rail, not `_load_interface_manifest_for_run`
+            # — that one returns None for a framing run by design (#496), and framing is
+            # exactly when the proposer needs this. The #496 rule is that a framing run
+            # must not expand or carry skeleton FILES; describing the interface in a
+            # prompt materializes nothing, so it stays inside that rule. Bind mode
+            # already requires a seeded manifest (#494), so this is the same document
+            # the gate hash-checks and the skeleton later expands from.
+            interface_manifest = None
+            if verification_contract is not None:
+                interface_manifest = await self._seeded_manifest_for_authoring(cycle)
+
             plan = generate_task_plan(
                 cycle,
                 run,
                 profile,
                 plan=implementation_plan,
                 contract=verification_contract,
+                interface_manifest=interface_manifest,
             )
             participating_agent_ids = {e.agent_id for e in plan}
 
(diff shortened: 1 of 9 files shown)


jasonLadd (commented):
reviewed

PR #615, opened 2026-07-26
fix(cycles): let the framing re-roll supersede a completed run (#522 never fired live)

jasonLadd:
Root-causes why the #522 framing re-roll **has never fired in production**, using the
deterministic repro that #612 handed us.

## The bug, in one sentence

The re-roll's first act is `cancel_run()` on the superseded framing run — but
inter-workload gates are decided on **COMPLETED** runs by design, and
`COMPLETED -> CANCELLED` was not a legal transition, so the cancel raised and killed the
re-roll before it created the replacement run.

`lifecycle.py` says the quiet part out loud, two lines below the transition table:

```python
# States that reject gate decisions — COMPLETED is intentionally excluded
# because inter-workload gates are decided on completed runs (SIP-0083 D15).
GATE_REJECTED_STATES = frozenset({RunStatus.FAILED, RunStatus.CANCELLED})
```

So the run is *always* terminal when a plan rejection lands. The re-roll could never have
worked live. It passes its harness because the harness drives a still-RUNNING run, where
the cancel is legal — a textbook harness-vs-live divergence.

## Evidence from pf-43 (`cyc_e83829c268a6`)

| time | event |
|---|---|
| 16:47:10.65 | framing run finishes → `completed` |
| 16:47:10.73 | REJECTED gate decision recorded by `system:plan_validation` |
| — | no re-roll log, no `WORKLOAD_ADVANCED` event, no second run |

`framing_max_rerolls` was 2 and zero were used. The guard's conditions were all
satisfiable — I confirmed `workload_entry["type"] == "framing"` straight from the cycle
row and that the re-roll branch is present in the **loaded** executor code, not just on
disk — so the failure had to be inside the branch. It was the first statement in it.

## The fix

One named edge: `("supersede", COMPLETED, CANCELLED)`.

Cancelling the superseded run is what preserves the positional run↔workload invariant
(#257/D14: exactly one non-cancelled run per position), so the re-roll needs it; the state
machine simply didn't model a real thing that happens.

There is precedent in the same table — `resume_from_failed` already gives FAILED an
outgoing edge — so "terminal" was never absolute here.

## ⚠️ This reverses a deliberate prior boundary

Two tests asserted `COMPLETED -> CANCELLED` was illegal. **Flagging explicitly rather than
quietly editing them.** Both are re-expressed rather than deleted, and the remaining
boundary is pinned harder than before:

- COMPLETED now has **exactly one** outgoing edge and no more (parametrized over every
  status)
- CANCELLED still has **none**

If you'd rather model supersession as its own status than widen this edge, that's a
reasonable alternative and I'd take the note — it's a bigger change (DB enum, DTOs,
console) which is why I didn't reach for it.

## Operator-facing cancel is unchanged

The edge is for the orchestrator. Cancelling a run that already finished is not a cancel,
it's a rewrite of what happened. The API cancel route now refuses terminal runs
explicitly, so `squadops runs cancel` on a completed run still returns the same conflict
it does today.

## Verification

- 6 new tests, including a replay of pf-43's exact state sequence (queued → running →
  completed → gate rejection → supersede).
- The two reversed assertions re-expressed, plus the "exactly one edge" pin.
- Regression **5617 passed**, 1 skipped. ruff clean.

## Why it matters now

Without this, #612 converts a three-hour loss into a 45-minute loss but the cycle still
dies. With it, a rejected plan re-rolls framing up to `framing_max_rerolls` times and the
cycle continues — which is what #522 intended and what the measurement rolls need.

### DIFF
--- a/src/squadops/api/routes/cycles/runs.py
+++ b/src/squadops/api/routes/cycles/runs.py
@@ -17,13 +17,18 @@ from squadops.api.routes.cycles.dtos import (
 from squadops.api.routes.cycles.errors import handle_cycle_error
 from squadops.api.routes.cycles.mapping import compute_workload_progress, run_to_response
 from squadops.auth.models import Scope
-from squadops.cycles.lifecycle import compute_config_hash, resolve_cycle_status
+from squadops.cycles.lifecycle import (
+    TERMINAL_STATES,
+    compute_config_hash,
+    resolve_cycle_status,
+)
 from squadops.cycles.models import (
     Cycle,
     CycleError,
     CycleStatus,
     GateDecision,
     GateDecisionValue,
+    IllegalStateTransitionError,
     Run,
     RunStatus,
     RunTerminalError,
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #616, opened 2026-07-26
fix(scaffold): never append a status_code the decorator already carries

jasonLadd:
pf-44 got further than any roll since the first green — clean plan, no frozen-file
targeting, the dev agent wrote working code — and **this enforcement broke it on the way
through.**

## What happened

The dev agent emitted:

```python
@router.post("/runs", status_code=status.HTTP_201_CREATED)
```

Correct, idiomatic, and exactly what the scaffold declared. But `_decorator_kwargs` only
recognised an int literal, so a symbolic value parsed as `None` — **indistinguishable from
"no keyword at all"**. The restorer concluded the status code had been stripped and
appended its own:

```python
@router.post("/runs", status_code=status.HTTP_201_CREATED, status_code=201)
```

Duplicate keyword argument → `SyntaxError` → `backend/routes.py` will not import → pytest
aborts at collection with exit 4 → **not one test ran** → every check downstream failed.

The enforcement log stated the false premise plainly:

> restored POST /runs: **status_code=201 declared by the scaffold, emitted as absent**

It was not absent. It was right there, spelled differently.

## Why the roll could not recover

Neo's repair failed for the same reason it was called. Whatever it emitted, the restorer
re-appended `status_code=201` and re-created the duplicate. Traced across every stored
version of the file:

| artifact | line 27 |
|---|---|
| scaffold seed | `@router.post("/runs", response_model=RunEvent, status_code=201)` |
| dev emission | `status_code=status.HTTP_201_CREATED, status_code=201` |
| repair emission | `status_code=status.HTTP_201_CREATED, status_code=201` |

**The mechanism regenerated the corruption it was called to fix.** A guard that turns
correct code into unparseable code is worse than no guard.

## The fix

`status_code` is now read as a *keyword* rather than a *value*, so presence and
literal-ness are separate facts. Three cases where there used to be one:

| emitted | action |
|---|---|
| absent | **append** — the pf-40 behaviour, preserved exactly |
| literal, wrong value | **replace in place** |
| expression (`status.HTTP_201_CREATED`) | **leave alone, report only** |

The middle case matters beyond tidiness: appending onto `status_code=200` would have
produced the identical duplicate, so that was a **latent second instance** of this bug
waiting for a producer to write the wrong literal.

The third case is deliberately hands-off. A non-literal cannot be evaluated from here,
`status.HTTP_201_CREATED` is the idiomatic spelling of exactly what was asked for, and the
behavioral probe already checks the status the app actually returns. Reporting is the
honest move; rewriting would be guessing.

## Verification

- **Replayed against pf-44's real stored artifact.** Recovered the producer's emission from
  the corrupted file, ran it through the fixed restorer: parses before, parses after,
  exactly one `status_code=`, divergence reported rather than forced. The same input
  produced the SyntaxError above on the deployed build.
- 5 new tests: symbolic left alone and still parseable (the regression), wrong literal
  replaced not appended, correct literal untouched, absent still restored, plus a sweep
  over six decorator spellings asserting the invariant that actually matters — **never emit
  unparseable source, never more than one `status_code`**.
- Regression **5624 passed**, 1 skipped. ruff clean (splice application extracted to
  `_apply_splices` for C901 rather than exempted).
- Contract gates in-container: lint / bare-skeleton / reference-fill all **GREEN**.
- Contract emission unchanged (`content_hash` 26503867) — **no re-seed**.

## Worth noting for review

Every guard in `_apply_splices` abandons an edit rather than forcing it. That is the
lesson generalised: this module's job is to hand back source that still parses, so a
splice that does not land exactly where the AST said it would is dropped, not applied on a
best guess.

### DIFF
--- a/src/squadops/cycles/fill_slot_integrity.py
+++ b/src/squadops/cycles/fill_slot_integrity.py
@@ -72,7 +72,7 @@ class DecoratorDivergence:
 class _Route:
     method: str
     path: str
-    status_code: int | None
+    status_kw: _StatusKw | None
     response_model: str | None
     func_name: str
     arg_names: tuple[str, ...]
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #618, opened 2026-07-26
fix(correction): a work_product rewind is escalated to patch (pf-45)

jasonLadd:
## What killed pf-45

Not the bug — the decision about the bug.

pf-45 got further than any roll since pf-38: app booted, probe answered (a real HTTP 500,
not a connection failure), test suite collected and ran. The defect was **one token** —
`pace` where the frozen model declares `pace_target` — squarely in a fill slot, correctly
diagnosed by the analyzer, correctly classified `work_product`. The repair prompt would
have carried the authoritative model surface. This is the exact failure the repair path
was built for.

The lead called it "systemic contract violations" and chose **rewind**. Rewind is
implemented as run death (`Run failed: Rewinding to checkpoint after qa.test failure`).
One correction attempt used, **four discarded**, roll over. Same shape as pf-32 — noted
then, never fixed.

## The fix

A second anchor on the existing #447 deterministic policy seam (`correction_policy.py` —
no new pattern, same guard that already bounds `continue`):

> a `rewind` whose classification is `work_product` is escalated to `patch` while the
> chain's repair slot is unspent.

A narrative escalation of severity cannot outvote a classification that says the code is
reachable by a patch.

**Deliberately narrow:**
- Non-work_product classifications (environment / infrastructure / unknown / absent)
  **keep their rewind** — patching correct code against a broken world is the opposite
  failure, and the locus×mode direction on the arc says infrastructure never patches.
- `abort` remains never overridden — a deliberate hard stop stays one.
- The model's original rationale stays intact in the decision artifact; the override is
  disclosed in the CORRECTION_DECIDED payload and the log line names which anchor fired
  (`override_reason` now rides the resolution instead of a hardcoded string at the call
  site).

## Why bound the decision instead of fixing rewind semantics

Rewind-as-run-death is arguably its own defect, but making rewind actually rewind is
executor surgery mid-campaign. Under current semantics, accepting a work_product rewind
is *strictly worse* than patching while budget remains — if patches exhaust, the run
fails anyway, so rewind buys nothing and costs everything. The guard encodes exactly
that dominance argument, nothing more.

## Verification

- **Replayed pf-45's real stored artifacts**: the exact (`rewind`, `work_product`) pair
  from `failure_analysis.md` + `correction_decision.md` resolves to `patch` with full
  provenance.
- 6 new tests: the escalation, all four non-work_product classifications stand,
  abort untouched, patch passthrough, anchor-1 provenance on the new field.
- Regression **5636 passed**, 1 skipped. ruff clean.
- No contract/scaffold surface touched — no re-seed.

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -716,13 +716,23 @@ class CorrectionRunner:
             decision_outputs.get("correction_path", "abort"),
             failure_evidence,
             cycle.applied_defaults,
+            # pf-45: the rewind anchor keys on the analyzer's classification — a
+            # work_product rewind dies as a run failure with the repair budget unspent,
+            # so the guard substitutes the patch the classification says is possible.
+            classification=str(analysis_outputs.get("classification", "")),
         )
         correction_path = resolution.path
         if resolution.overridden_from:
             logger.warning(
-                "correction_policy_override: %s -> patch (executed-failed required checks: %s)",
+                "correction_policy_override: %s -> %s (%s%s)",
                 resolution.overridden_from,
-                ", ".join(resolution.failed_required_checks),
+                resolution.path,
+                resolution.override_reason,
+                (
+                    "; checks: " + ", ".join(resolution.failed_required_checks)
+                    if resolution.failed_required_checks
+                    else ""
+                ),
             )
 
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #619, opened 2026-07-26
fix(dev): the initial author gets the model surface, field-level (pf-45)

jasonLadd:
Companion to #618 — that one stops the correction loop from throwing away its budget on
pf-45's failure class; this one stops the class from happening.

## The defect, and why no check caught it early

pf-45's dev wrote `pace=data.pace` against a frozen model declaring `pace_target`. One
token. A wrong field name passes import and compile checks — it raises at *request*
time — so every POST /runs became a 500 and the suite failed downstream of creation.

The repair path has carried the authoritative model names since #604. **The author of
the first emission never has.** pf-38's dev guessed the fields right, pf-45's guessed
wrong; that variance is the whole difference between those rolls. Same shape as #588's
error-contract fix: same data, same transport, one step earlier.

## Two upgrades through one seam

Both consumers of `model_surface_instructions` — the initial fill prompt *and* the
repair prompt — get these together:

**Field-level signatures.** The block previously named classes only, so it could not
have prevented pf-45 even on the repair path — it said `RunEvent` but never
`pace_target`. Now:

```
`RunEvent(id, title, datetime, location, distance, pace_target, route_notes, participants)`
`RunEventCreate(title, datetime, location, distance, pace_target, route_notes)`
```

…with an explicit near-miss warning naming this exact failure (`pace` for `pace_target`
raises at request time, not load time).

**The frozen store.** pf-45's second defect: the dev re-declared local store dicts in
the fill slot, shadowing `backend/store.py` — the scaffold's `reset()` then cleared the
*unused* stores, silently breaking test isolation. The block now names
`participant_store` / `run_event_store` / `reset()` and says import, don't re-declare.
Rendered only for in-memory persistence.

## Transport

Mirrors the error seam exactly, no new pattern:
- executor threads **data-only** lines onto `development.develop` envelopes (same
  task-type gate; absent manifest attaches nothing — author mode byte-identical)
- the develop handler renders them through a new managed asset (#448); prose lives in
  `request.development_develop_model_surface_appendix.md`
- the fill-only appendix gains an optional `{{model_surface}}` variable (version 3)

## Verification

- 9 new/updated tests: field-level signatures carry `pace_target`; the store line names
  the real stores; non-memory persistence omits it; the envelope threading with the same
  gating as the error seam (qa.test / no-manifest attach nothing); a live render through
  the real asset reaching the prompt.
- Regression **5639 passed**, 1 skipped. ruff clean.
- **Contract emission unchanged** (`content_hash` 26503867) — prompt-side only, no
  re-seed.

## Honest limit

Same as #613: supplying the fact doesn't compel the model to read it. That's why #618
exists — if the dev still fumbles a field, the correction loop now keeps its budget and
the repair prompt carries the same upgraded surface with the exact token it needs.

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -1700,11 +1700,22 @@ class DispatchedFlowExecutor(FlowExecutionPort):
         # to learn what the scaffold already knew. Data only; the block's prose
         # lives in a managed prompt asset.
         if envelope.task_type in self._ERROR_SEAM_TASK_TYPES:
-            from squadops.capabilities.scaffold import error_seam_instructions
+            from squadops.capabilities.scaffold import (
+                error_seam_instructions,
+                model_surface_instructions,
+            )
 
             error_lines = error_seam_instructions(interface_manifest)
             if error_lines:
                 extra_inputs["error_contract"] = error_lines
+            # pf-45: the model surface, to the INITIAL author on the same transport.
+            # Repairs have carried it since #604; the first fill never did, so the dev
+            # guessed a field name (`pace` for `pace_target`) and every POST /runs
+            # raised into a 500 — a correction spent learning what the scaffold already
+            # knew. Field-level since pf-45 for the same reason.
+            surface_lines = model_surface_instructions(interface_manifest)
+            if surface_lines:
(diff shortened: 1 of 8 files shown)


jasonLadd (commented):
reviewed

PR #623, opened 2026-07-27
SIP-0102 phase 102.2: environment is a pinned, validated contract — live-smoke green

jasonLadd:
## Summary

Phase **102.2** complete, plus a fix for a golden-path-blocking bug found in merged 102.1 code. The phase exit criterion is met on both halves, and the floor has now **executed end-to-end on real docker** — the informal engine-turns-over precursor.

## Live smoke (Mac, docker 29.5.3, canonical image built locally)

Real expander skeleton (17 files) → seed → `install_dependencies` (15s, caches warm) → `build_frontend` (real vite build, 1s) → `start_application` to readiness → `probe_http_endpoint` GET /runs → **501 observed = succeeded** (the bare-skeleton partition: routing works, behavior honestly unimplemented) → converging teardown. Pin verification intact through install/build. The smoke caught two real gaps unit tests missed (below).

## Commits

- **fix(#622)** — 102.1c2's readiness demanded a 2xx from `/health`, re-introducing the exact hidden-product-requirement class #520 fixed: the canonical app declares no /health, so `start_application` would have timed out forever on the golden path. Readiness is now transport-level (any HTTP response). **Closes #622**
- **102.2a+b** — `EnvironmentContract` (content-hashed `contract_id`, import-time validation) with the canonical `fullstack_fastapi_react` contract **coherence-tested against a real expander run**; contract-driven factory (config image = override only); `environment_contract_id` on every result (§7 item 4); **derived-state rule** (`.sandbox-venv`/`node_modules`/`dist`/lockfile are never revision content — §4.7 installs live in the workspace, so without this the first install breaks every pin); checked-in image rendering + build script (local-build decision recorded).
- **102.2c** — one pure `sandbox_environment_decision` rendered by BOTH the cycle-create preflight (SIP-0095 seam) and a new `doctor` **sandbox** category, so they can never disagree. Evidence doctrine per SIP-0095 §6.2: unreachable service / daemon-can't-answer ⇒ warn-and-allow; unknown environment / contract skew / missing image ⇒ block (fix command = the build script). Provider `noop` ⇒ empty decision — dormant on every existing deployment.
- **102.2d** — read-through download caches (pip/npm download caches only — cached deps structurally cannot substitute for undeclared ones, §7 item 13) + the docker-outside-of-docker fix: the containerized service hands workspace paths to the HOST daemon, so compose now uses an identical-string same-path bind instead of 102.1's named volume (which could never work).
- **live-smoke findings** — npm install generates `package-lock.json` (derived state, excluded; evidence capture = the explicit promotion path); the probe result was missing `environment_contract_id` (its base dict has its own shape). Both unit-covered now.

## Exit criterion

- ✅ a deliberately mismatched/missing environment blocks at cycle-create preflight, and doctor reports the same finding (same decision function by construction)
- ✅ every sandbox op result carries image + environment-contract identity
- ✅ (beyond the bar) full floor E2E on real docker

## Inert-to-merge

Route addition is dormant (env-driven; no `SQUADOPS__SANDBOX__*` set ⇒ empty decision, no IO). Doctor category is additive. Compose changes stay profile-guarded. Pure addition elsewhere. Regression **5758 green**.

https://claude.ai/code/session_01GaFdkrgQ8F3UZ1DE7LDEU3

### DIFF
--- a/adapters/sandbox/container_backend.py
+++ b/adapters/sandbox/container_backend.py
@@ -200,6 +200,20 @@ class ContainerBackend(NoOpExecutionSandbox):
             **common, entries=_tail_lines(run.stdout + run.stderr) if run else ()
         )
 
+    async def environment_report(self) -> dict:
+        """Operational facts for /health and the 102.2c preflight
+        reconciliation. ``image_present`` is None when the daemon cannot
+        answer (unverifiable, never a guessed False)."""
+        try:
+            image_present: bool | None = await self._container.has_image(self._image)
+        except ToolContainerError:
+            image_present = None
+        return {
+            "contract_id": self._environment_contract_id,
+            "image": self._image,
+            "image_present": image_present,
+        }
+
     # -- runtime unit (slice c2) ---------------------------------------------
 
     async def _await_ready(self, base_url: str) -> bool:
--- a/adapters/sandbox/container_backend.py
+++ b/adapters/sandbox/container_backend.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 import asyncio
 import time
 from collections.abc import Callable, Mapping
+from pathlib import Path
 
 import httpx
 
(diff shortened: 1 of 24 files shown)


jasonLadd (commented):
reviewed

PR #624, opened 2026-07-27
fix(checks): declare check applicability; behavioral retest decides when no structural check applies

jasonLadd:
Fixes the repair deadlock that killed pf-47 and is killing pf-49 as this PR opens —
two rolls, identical signature, deterministic once triggered.

## The root cause (authoring, not evaluation)

The system pressures the plan author to attach typed checks (the whole #464/#533 arc)
but hands it a **Python-only vocabulary and never says so**. So the QA proposer dresses
frontend test files in Python conventions — pf-49 literally authored `function_defined
name_prefix: "test"` against a Vitest `.jsx` file, where tests are *calls*, not named
functions. Every such check skips at evaluation (correctly, per #607). But repair
acceptance requires an **executed** check to accept a patch, so:

> a task whose checks all skip can never land a repair — zero executed blocking checks
> → unverifiable → fail closed, every attempt, until the budget is gone.

pf-47: 4 repairs rejected unheard, the 5th to the import gate, roll dead. pf-49: same
signature at 10:44 UTC. Meanwhile the **behavioral retest** — which re-runs the actual
failing suite against the patched workspace — sat one step downstream, gated behind a
structural verdict that could never come.

## Fix A — applicability declared once, enforced twice

- `CheckSpec` gains `applicable_extensions`; the five AST checks declare `{".py"}`.
- The authoring vocabulary is **generated from CHECK_SPECS**, so every proposer prompt
  now renders *"Applies to: `.py` files ONLY — never attach to `.js`/`.jsx`"* — no new
  prose asset, no drift possible.
- Dispatch **strips** a file-scoped check whose target the evaluator can never parse
  (one warning per strip). A dead check can no longer reach evaluation, evidence, or
  repair acceptance. Strip (not reject) so a mis-authored plan degrades gracefully
  instead of burning framing re-rolls while the author learns.

## Fix B — behavioral evidence decides when structural evidence cannot exist

- `patch_verification` exports the two *structurally-unevaluable* reasons
  (`no_executed_blocking_checks`, `no_typed_criteria`) as constants.
- At patch acceptance, **exactly those reasons — and only when the failed task carries a
  `test_result` to re-run** — fall through to the #456 retest, whose verdict alone
  decides. The retest is *stronger* evidence than the checks it stands in for: it runs
  the real suite in the QA agent's environment.
- Deliberately NOT unlocked: `evaluator_error:*` (a broken checker proves nothing),
  `unparseable_criteria`, and real check failures — all keep failing closed. §6.2
  survives intact: never accept on absent evidence; substitute stronger evidence for
  inapplicable evidence.

**A and B need each other.** After A strips the nonsense, a frontend QA task
legitimately has *no* structural checks — without B, that re-creates the deadlock
through the honest door (`no_typed_criteria`). B without A leaves meaningless pytest
idioms in plans. Together: the author is taught, the plans are cleaned, and the one
artifact class whose verification is inherently behavioral gets judged behaviorally.

## Verification

- **Replay against pf-49's real stored artifacts**: the exact criteria+emission shape
  that deadlocked returns `unverifiable/no_executed_blocking_checks` and now routes to
  the retest instead of auto-rejection.
- 7 new tests: the strip matrix (`.jsx` AST checks stripped; `.py` checks, doc-regex,
  `command_exit_zero`, and prose survive) + four acceptance-path cases (retest-pass
  breaks the deadlock; retest-fail still rejects; no behavioral evidence stays closed;
  `evaluator_error` stays closed and never reaches the retest).
- Evidence honesty on acceptance: skip rows keep `status: skipped` in the recorded
  checks; the retest rows carry the executed pass.
- Regression **5731 passed**, 1 skipped. ruff clean (strip loop extracted per C901).
- **Contract emission unchanged** (`content_hash` 26503867) — no re-seed.

## Deploy note

Executor + dispatch surface → deploying resets the N=5 baseline (currently 2/5).
Per the tripwire agreed this morning: second deadlock occurred → deploy when pf-49 goes
terminal, take the reset, relaunch on a machine that cannot deadlock this way.

Longer-term (unchanged): per-stack check vocabularies ride the Stack Blueprint work in
1.6 — a React stack eventually brings vitest-aware checks the way FastAPI brings
pytest-aware ones. This PR makes the system honest about the vocabulary it has today.

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -49,6 +49,8 @@ from squadops.cycles.models import (
 from squadops.cycles.naming import flow_run_name
 from squadops.cycles.patch_verification import (
     PATCH_PASSED,
+    PATCH_UNVERIFIABLE,
+    STRUCTURALLY_UNEVALUABLE_REASONS,
     overlay_artifacts,
     verify_patched_artifacts,
 )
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #625, opened 2026-07-27
docs(sip): test-type taxonomy requirement for the Stack Blueprint (pf-47/49 evidence)

jasonLadd:
Adds a **Test-type taxonomy** section to the proposed Stack Blueprint Contract SIP, recording this weekend's three-defects-one-dimension finding as a schema requirement rather than folklore.

**The claim:** test *type* (backend-unit / frontend-component / frontend-build / behavioral-probe) is the first-class classification; *runner* is a property of the type. Each blueprint populates a handling row per type — evidence vocabulary, suite-broken signal, repair role, repair-verification method, execution locus — and the locus classifier, repair router, and patch-acceptance policy become lookups over that row.

**Evidence in the section:** the pf-47/pf-49 repair deadlock (fixed tactically in #624), the pf-49 locus misroute (Vitest transform failure exits 1 → classified subject-broken → the dev role rewrote QA's own test file), and the runner-less `test_result` shape that flattens heterogeneous evidence before anything can reason about it.

**Also sharpens the SIP's acceptance gate:** the schema must be written against at least two test types per stack, or the generalisation repeats the pytest-universal mistake one level up.

Docs only. Near-term seam filed separately as an issue (thread type/runner into `test_result`).

### DIFF
--- a/sips/proposed/SIP-Stack-Blueprint-Contract.md
+++ b/sips/proposed/SIP-Stack-Blueprint-Contract.md
@@ -107,6 +107,53 @@ stack #2.
 3. **After:** migrate the typed checks off their hardcoded `.py` onto the blueprint's
    declared source language.
 
+## Test-type taxonomy (added 2026-07-27 — pf-47/pf-49 evidence)
+
+Three production defects in twenty-four hours traced to the same missing dimension, so it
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #633, opened 2026-07-28
feat(scaffold): frontend test harness is scaffold-owned (#627)

jasonLadd:
## What

Scaffold-owns the frontend test harness — the frontend mirror of SIP-0100's backend conftest move.

## Why

pf-53 (`cyc_86211da4786d`): the scaffolded frontend workspace had **no `test` script, no vitest, no testing-library, no jsdom** — vitest only executed because `npx` ad-hoc-fetched it from the registry. The QA agent either refused to test ("no runner available" — an assumption that *matched the workspace's own signals*) or authored five substantive, distinct test implementations that **all died on the missing environment**. Test-writing capability was never the gap; provisioning was 100% of it.

## How

- **package.json**: `vitest`, `@testing-library/react` (+ its `@testing-library/dom` peer), `@testing-library/jest-dom`, `jsdom` devDependencies; `"test": "vitest run"` script
- **vite.config.js**: `test` key with jsdom environment + `setupFiles` (vitest reads it; `vite build` ignores it)
- **`frontend/src/test-setup.js`** (frozen): registers jest-dom matchers via the **`/vitest` entry** — the bare entry assumes a global `expect` and crashes collection under vitest's default `globals:false`; caught by running the real toolchain, not review
- **`frontend/src/__tests__/harness.test.jsx`** (frozen): renders the app shell under a `MemoryRouter` at a path no route claims — so it passes on the bare skeleton **and** after any fill; it asserts harness wiring, never app behavior, and doubles as the in-workspace idiom example QA is told to follow
- **`task_type.qa.test` fragment** (v0.9.23): new Frontend Tests section — discovery convention, follow-the-harness rule, and *mock `apiFetch`, jsdom has no server* (the exact failure shape of pf-53's fetch-integration repair attempts); the "no runner declared → no frontend tests" scope rule survives for unscaffolded stacks

## Verification (real toolchain, eve container)

- `npm install` clean (174 packages); **harness test passes on the bare skeleton and on the reference fill**; `npm run build` unaffected
- Contract gate `lint` / `bare-skeleton` / `reference-fill` all **GREEN at 7/7 criteria** (including #628's `vc-routes-imports`, first gate run with it live)
- `run_node_tests` early-exits when QA authors no frontend test files (the harness rides as a *source* artifact), so backend-only rolls pay no new install cost
- Full regression: 5794 passed

## Deploy note

[...]

PR #630, opened 2026-07-28
feat(checks): module_imports — runtime-level check static analysis cannot fake (#628)

jasonLadd:
## What

Adds `module_imports` — the first runtime-level typed check — and binds it into the contract as `vc-routes-imports`.

## Why

pf-54 (`cyc_01dbcc1bb3ab`) shipped a routes.py missing the two scaffold header lines (`from fastapi import APIRouter`, `router = APIRouter()`). It passed **all three** bound checks legitimately — `endpoint_defined` (AST finds the decorators), `import_present` (AST finds the import), `py_compile` (syntax is valid) — because the entire vocabulary is static and nothing ever imports the module. First import was the QA conftest → NameError → pytest exit 4 → 5-attempt correction death.

## How

- **Spec** (`acceptance_check_spec.py`): `.py`-scoped, `file` param, vocabulary notes auto-render the semantics.
- **Evaluator** (`acceptance_checks.py`): `python -c "import <module>"` in an isolated subprocess — workspace root as cwd (on sys.path), restricted env, timeout clamp. Module name derived from the file path (`backend/routes.py` → `backend.routes`, `__init__.py` → package). Outcomes:
  - exit 0 → **passed**
  - `ModuleNotFoundError` whose top-level package is **not** in the workspace → **skipped** `missing_tooling` (#462: an environment gap never fails correct code)
  - everything else (incl. missing modules whose top-level IS in the workspace — the app referencing itself wrongly) → **failed** with stderr tail
- **Contract** (`scaffold_contract.py`): `vc-routes-imports` added to the routes fill-file implementation criteria; linter's implementation-class allowlist extended (`verification_contract.py`).

## Verification

- 9 new evaluator tests (NameError fail, clean pass, third-party-dep skip, workspace-internal-miss fail, `__init__` derivation, file-not-found, traversal, frontend-extension skip, missing-then-broken sibling)
- **Bare skeleton passes** the new check (verified by materializing the real group_run manifest — stub bodies import by design, so the criterion is winnable pre-fill)
- **Replay on pf-54's stored artifacts**: the shipped fragment (art_1544d6ba68c9) → `failed / module_import_failed` with the exact production NameError; the repair version (art_3e011ea16382) → `passed`
- Full regression suite: 5742 passed

## Deploy note

Contract emit changes → the next seeded measurement run needs a re-emitted contract (content_hash moves, criteria count 6 → 7). Deliberate — the N=5 freeze is closed; this is step 2 of the 1.4 cut sequence.

Closes #628

[...]

PR #634, opened 2026-07-28
feat(scaffold): blank-input rejection is scaffold-owned AND probe-pinned (#593)

jasonLadd:
## What

Option (c) for the blank-string gap: the scaffold **owns the constraint** and the contract **pins the behavior** with a probe.

## Why

pf-38 volunteered blank-field guards; pf-39 didn't — both went green on identical seeds, because nothing *required* blank rejection and nothing *tested* it. pf-39's suite posted **absent** fields (a different behavior — pydantic's required-field 422 satisfied the narrative criterion), so the gap shipped: `POST /runs {"title":"","datetime":"","location":""}` → 201, blank run created. The N=5 sample was measuring an unenforced behavior.

## How

**Constraint (scaffold model):** required request-shape fields emit as `NonBlankStr = Annotated[str, StringConstraints(strip_whitespace=True, min_length=1)]` — whitespace-only is as blank as empty; optional fields stay unconstrained.

**Probe (contract):** every create probe gains a companion `vc-probe-*-rejects-blank` — all required fields sent blank → expect `422` + `error_code: validation_error` (the envelope the frozen `errors.py` handler shapes). Emitted only alongside an error contract.

**New probe classification — `guards: "scaffold"`:** the constraint fires at the model layer *before any stub or fill body runs*, so this probe passes on the **bare skeleton by design**. That's the probe analog of `frontend_build`'s regression-guard class (vs. `tests_pass`'s fill-behavior class). The `Probe` schema carries the optional field (omitted when unset — pre-#593 probe dicts round-trip byte-identical, wire compatibility proven by the existing exact-dict injection tests), the linter validates its values, and the contract gate expects guards-scaffold probes to pass in **both** modes.

Side alignment with #629: `behavior_expectation_lines` labels 4xx probes "(rejection case)" so the QA-authoring block doesn't render two contradictory-looking lines for the same endpoint (`POST /runs → 201` and `POST /runs → 422`).

## Verification

- Contract gate `bare-skeleton` and `reference-fill` **GREEN at 8/8** on this branch in the eve container — the blank probe passes pre-fill via the constraint alone, and against the **booted reference app** (a real HTTP POST with blank strings answered 422 + the pinned envelope)
- Emitted-model exec tests: `""` and `"   "` rejected, valid input accepted with whitespace stripped
- Lint validates `guards` values; malformed value is a lint error
- Full domain suites green (3494)

## Deploy note

Contract emit changes again (new probe row) — same re-emit + re-seed window as #630/#633. Independent of PR #633 (different regions); either merge order is clean.

Closes #593

### DIFF
--- a/scripts/dev/contract_gate.py
+++ b/scripts/dev/contract_gate.py
@@ -129,13 +129,19 @@ def _record_probes(
 ) -> None:
     """Boot the subject and run every behavioral probe (SIP §6.4/§6.5, 98.4).
 
-    Reference-fill: each probe must ``passed``. Bare skeleton: a probe must NOT pass
-    (its stubs answer 501/nothing) — a probe that passes on the bare skeleton is a
-    false-green admitted at authoring time, exactly what this gate exists to catch."""
+    Reference-fill: each probe must ``passed``. Bare skeleton: a fill-behavior probe
+    must NOT pass (its stubs answer 501/nothing) — a probe that passes on the bare
+    skeleton is a false-green admitted at authoring time, exactly what this gate
+    exists to catch. The exception is ``guards: scaffold`` probes (#593): they pin
+    behavior the skeleton itself provides (e.g. model-layer blank rejection, which
+    fires before any stub body), so they must pass in BOTH modes — the probe analog
+    of the frontend_build regression guard."""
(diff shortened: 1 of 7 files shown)


jasonLadd (commented):
reviewed

PR #635, opened 2026-07-28
fix(sandbox): run operations as the workspace owner — Spark shakedown findings (SIP-0102)

jasonLadd:
## What

SIP-0102 Spark shakedown (handoff steps 1–3 + 5) — executed, two native-Linux divergences found, root-caused, fixed, and the floor smoke is now **GREEN on Spark's docker** end-to-end.

**⚠️ Mac-lane review requested:** this touches `adapters/sandbox/container_backend.py`, a Mac-driven surface per the recorded lane note. This PR is the coordination vehicle — treat it as the shakedown's findings + proposed fix, not a unilateral merge.

## Findings (both masked by Docker Desktop's virtiofs on Mac)

**1. `cap_drop_all` vs host-owned bind mounts.** The env image runs as root; `--cap-drop ALL` strips `CAP_DAC_OVERRIDE`, so on native Linux root cannot write the uid-1000-owned workspace mount — install died at `python -m venv` in 0.3s with `Permission denied`. Proven by three-way isolation: cap-drop+root **fails**, no-cap-drop **works**, cap-drop+`--user 1000:1000` **works**.

**2. Cache mounts assumed root's home.** Once the container runs as uid 1000, npm resolves `$HOME/.npm → /.npm` (unwritable, no passwd entry → `HOME=/`).

## Fix

- `ContainerSpec.user` (inert default `None`; adapter renders `--user` only when set — the pre-0102 byte-identical guarantee holds, pinned by the existing no-flags test)
- The backend runs every operation as the **workspace owner's uid:gid** (`_workspace_user`), keeping the §7 hardening fully intact
- Cache mounts move to user-neutral `/cache/pip` + `/cache/npm`, pinned via env (`HOME=/tmp`, `PIP_CACHE_DIR`, `npm_config_cache`)

## Verification

- **Floor smoke GREEN on Spark**: seed → install → vite build → boot → 501 probe → teardown, pin intact — and the install exercises the post-#633 expander output (vitest/testing-library devDeps) without issue
- New tests: workspace-owner user on operation specs, HOME/cache env pinning, `--user` in the hardened-flag render matrix + absent-by-default; cache-mount test updated to the neutral paths
- Full regression: 5803 passed; C901 on the adapter resolved by extracting `_render_hardening_args` (decompose, not exempt)

## Shakedown status

Steps 1–3 done (image built locally, 361MB). Step 4 (service boot + doctor) deferred until this merges so the service image bakes merged code. Step 5's report is written into the plan-doc handoff section in this PR. Digest-pinning revisit for 1.5 stands.

### DIFF
--- a/adapters/sandbox/container_backend.py
+++ b/adapters/sandbox/container_backend.py
@@ -301,6 +301,11 @@ class ContainerBackend(NoOpExecutionSandbox):
         spec = ContainerSpec(
             image=self._image,
             command=list(command),
+            # Same identity/env posture as the one-shot ops: the runtime unit
+            # executes the UNTRUSTED generated app — it deserves least
+            # privilege most of all, and as root-sans-DAC it could not write
+            # its own workspace (pycache, app state) on native Linux anyway.
+            env=(("HOME", "/tmp"),),
             volumes=((str(self._store.workspace_dir(revision.cycle_id)), _WORKSPACE_MOUNT),),
             working_dir=_WORKSPACE_MOUNT,
             timeout_seconds=self._timeout_seconds,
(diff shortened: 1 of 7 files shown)


jasonLadd (commented):
reviewed

PR #640, opened 2026-07-28
fix(retest): contract probes ride the retest against the patched tree (#639)

jasonLadd:
## What

Closes the pf-50 false-green mechanism: accepted repairs now face the contract probes, executed against the **patched** tree.

## The two halves (both required)

1. **Threading** (`correction_runner.py`): `contract_probes` from the failed task's enriched envelope into `retest_inputs` — it was simply absent, so `_append_contract_probe_rows` no-opped on every retest despite its own "probe evidence rides the retest path too" comment. Same envelope-threading class as the earlier `artifact_contents` instant-fail fix in this file.
2. **Patched-tree overlay** (`qa_test.py`): the retest's probe workspace now overlays the patched files onto dispatch-time sources (last-wins, same order as the retest suite run). Threading alone would have probed the *pre-repair* tree and passed stale — pinned by a test asserting the probe workspace carries the repaired content.

## Effect

pf-50's accepted attempt-2 repair (which fixed the join bug but dropped `status_code=201`) would have produced a failing `vc-probe-runs` row at retest → repair rejected → the loop forced to produce one keeping both the fix and the pin.

## Verification

- Runner tests: probes thread into the retest envelope; probe-less envelopes thread no key
- Handler test: probe workspace carries the patched tree, not the dispatch-time copy (the exact pf-50 false-green shape)
- Full regression 5806; contract emission untouched — no re-seed needed

Pre-window fix per the FAY measurement decision — the window opens on the deploy that includes this.

Closes #639

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -995,6 +995,11 @@ class CorrectionRunner:
             "expected_artifacts": failed_inputs.get("expected_artifacts", []),
             "acceptance_criteria": failed_inputs.get("acceptance_criteria", []),
         }
+        # #639: probes ride the retest or the final verdict carries stale
+        # probe evidence for a tree the repair changed. Presence-keyed, like
+        # the bind-mode injection (probe-less contracts thread no key).
+        if failed_inputs.get("contract_probes"):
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #641, opened 2026-07-28
feat(plan): criterion binding derived from the contract, not transcribed by the author (#509)

jasonLadd:
## What

Deterministic criteria binding: every contract-covered fill file's criterion ids are unioned into the producing task's `criteria_refs` before validation and before dispatch — the contract states the linkage; the plan author no longer transcribes it.

## Why

pf-54's run 2 was auto-rejected (a burned ~45-min framing re-roll, one draw from cycle death) for omitting one ref the contract already declared. Pre-validator, the same variance under-counted coverage (#509's original observation: `criteria_total` 3 → 1 across identical seeded rolls). Either way, criterion linkage measured plan-authoring luck.

## How

- `ImplementationPlan.with_contract_criteria_bound(contract)` — pure, idempotent, order-preserving; returns `(bound_plan, notes)`
- Applied at the inter-workload plan-validation gate (bind → validate → log `criteria_auto_bound`) **and** in `generate_task_plan` — validated plan and executed plan cannot drift
- `validate_criteria_refs` unchanged as the guard: unresolvable refs and authored-on-covered still reject; only the missing-coverage rule becomes structurally vacuous

## Tests

Binding matrix incl. idempotence and authored-order preservation; refless plan gates normally (the flipped pf-54 scenario); unresolvable ref still rejects at the gate and raises at the dispatch net; end-to-end envelope resolution from a refless plan. Full regression 5809.

Pre-window fix for the FAY measurement (removes a framing-rejection class). No contract emission change — no re-seed.

Closes #509

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -2700,6 +2700,15 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                 if binding_content is not None:
                     errors.extend(self._validate_contract_binding(contract, binding_content))
                 if parsed_plan is not None:
+                    # #509: bind the contract's covered-file criteria
+                    # deterministically BEFORE validating — the descoping rule
+                    # then guards against binding bugs instead of taxing every
+                    # roll on the author's transcription. Dispatch applies the
+                    # same normalization (generate_task_plan), so the validated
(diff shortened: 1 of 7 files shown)


jasonLadd (commented):
reviewed

PR #644, opened 2026-07-28
fix(acceptance): typed checks evaluate inside the accepted workspace tree (#643)

jasonLadd:
## What

Fixes the fay-1 window-killer: `module_imports` (and the #591 import pre-gate) evaluated in a workspace containing **only the task's own emitted artifacts**, so any fill file importing its frozen scaffold siblings — which the contract mandates (`vc-routes-apierror` requires `ApiError` from `.errors`) — false-failed on every dev attempt and every repair candidate. Structurally unwinnable in-cycle; the correction chain escalated toward regenerating a frozen scaffold file.

## How

One selection, three consumption sites, all presence-keyed and inert when absent:

- **Executor dispatch** (`dispatched_flow_executor.py`): threads `acceptance_workspace_files` — the full accepted `source`/`config` tree (scaffold included via `by_type`, repair candidates excluded per pf-31 Fix E) — into every build-task envelope. The spec is single-sourced as `_ACCEPTANCE_WORKSPACE_FILTER`; `qa.test`'s prompt filter now references it (it was already this exact selection, #443).
- **Handler typed acceptance** (`base.py`): materializes the tree first, the task's artifacts on top — the candidate supersedes its own slots.
- **Patch verification** (`patch_verification.py` via `_try_accept_patch`): same substrate, passed as a separate parameter so `patched_artifacts` — what an accepted patch re-stores under the task's type (#389 swap) — never absorbs the tree.
- **Retest** (`correction_runner.py`): inherits the key from the failed envelope, same pattern as #639's `contract_probes`.

Deliberately NOT widened: `development.develop`'s `artifact_contents` prompt filter — the verification workspace and the curated prompt context are different concerns (prompt diet).

## Why the gate didn't catch it

The contract gate validates `vc-routes-imports` against the full expanded tree (bare-skeleton / reference-fill modes), and the #630 pf-54 replay was full-workspace too. The in-cycle path was first exercised by fay-1 — the campaign ended before #628/#630 went live.

## Verification

- 10 new tests across the four seams, each pinning the fay-1 mechanism (fails-without / passes-with / candidate-supersedes-slot; scaffold-source inclusion + repair-candidate exclusion; presence-keyed retest threading).
- **Deterministic replay with fay-1's real artifacts**: rejected candidate `art_c1ae4ec3942a` + the run's stored scaffold tree → patch verification `passed` under the fix, still `failed` without it.
- Full regression: 5832 passed, 1 skipped.

Closes #643

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -1000,6 +1000,13 @@ class CorrectionRunner:
         # the bind-mode injection (probe-less contracts thread no key).
         if failed_inputs.get("contract_probes"):
             retest_inputs["contract_probes"] = failed_inputs["contract_probes"]
+        # #643: same presence-keyed threading for the typed-acceptance
+        # workspace — the retest's evaluation needs the scaffold siblings
+        # exactly like the original dispatch did.
+        if failed_inputs.get("acceptance_workspace_files"):
+            retest_inputs["acceptance_workspace_files"] = failed_inputs[
+                "acceptance_workspace_files"
+            ]
 
         retest_envelope = TaskEnvelope(
             task_id=f"retest-{run_id[:12]}-{correction_attempts:02d}-{envelope.task_type}",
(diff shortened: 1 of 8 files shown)


jasonLadd (commented):
reviewed

PR #654, opened 2026-07-29
fix(correction): frontend_build failures widen repair scope to frontend source (#650)

jasonLadd:
Third item of the next-window fix package — the minimal version per the owner call.

**Mechanism**: `_resolve_repair_target` gains one provenance rule: when the failed task's validation carries a `frontend_build` row with `passed: false`, the plan's frontend implementation source (the existing `implementation_artifacts` threading — no new envelope surface) is unioned into the repair target. Applied on both the drift and no-drift branches. Everything else — RC2 package scoping, drift-first ordering, pf-21 own-artifact union — is unchanged.

**Receipt**: fay-8 (`cyc_7f5f1b8b1790`) — four identical `frontend_build` failures across five rounds; every repair emitted backend/test files because the reporting task owns only those; structurally non-convergent. The new test replays exactly that shape (real plan surface) and asserts the views enter the target.

**Guard rails pinned by test**: passing `frontend_build` widens nothing (fay-3's backend-only scoping preserved); skipped/not-executed rows (#306 Node-absent shape) widen nothing; drift co-occurrence keeps drift files first-class.

4 new tests. Full regression 5848 passed.

Closes #650

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -82,6 +82,51 @@ def _scope_to_shared_packages(candidates: list[str], anchors: list[str]) -> list
     return [c for c in candidates if c and _top_level_package(c) in anchor_pkgs]
 
 
+def _frontend_build_failed(failure_evidence: Any) -> bool:
+    """#650 (fay-8): the failed task's validation shows the frontend build failing.
+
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #655, opened 2026-07-30
feat(contract): v8 — chained join/leave probes with response capture (#651)

jasonLadd:
Fourth item of the next-window fix package. **This moves the contract hash — v8 (`6d4f38a3…`) — re-seed rides the next window deploy.**

**Schema**: `Probe.capture` ({var: top-level response key}, omitted-when-empty like `guards`), `{var}` placeholders in request paths, lint rules (placeholder must be captured by an *earlier* probe; capture entries identifier→key). The substitution/capture helpers live on the contract module and are shared by both executors — qa probe runner and the FAY auditor — so semantics can't drift.

**Emission (all manifest-derived, nothing guessed)**: create probe captures the response entity's `id` into the child path parameter; each child POST action probes its declared success status; a declared error code mapped to HTTP 409 yields an immediate duplicate probe pinning the envelope. group_run emits: create(201, capture) → blank-rejection(422, scaffold-guarded) → join(200) → duplicate-join(409 `duplicate_participant`) → leave(200). **11 criteria** now, up from 8.

**Gate proof (run in the eve image)**: lint GREEN; bare-skeleton GREEN 11/11 — the new probes correctly fail on stubs via unresolved placeholders after the create 501s; reference-fill GREEN 11/11 — the full chain executed against the real reference app.

**Receipt**: fay-3 (`cyc_f6bea0a3c1a8`) — broken join, suite caught it, audit passed anyway under v7's probe bar. Under v8 that deliverable audits non-functional, as it should.

9 new tests (emission sequence, lint, capture/resolution semantics incl. the skeleton unresolved case). Full regression 5855 passed.

Closes #651

### DIFF
--- a/scripts/dev/audit_delivered_app.py
+++ b/scripts/dev/audit_delivered_app.py
@@ -44,7 +44,11 @@ import httpx
 from adapters.sandbox.container_backend import ContainerBackend
 from adapters.tools.docker import DockerAdapter
 from squadops.cycles.task_plan import REPAIR_TASK_TYPES
-from squadops.cycles.verification_contract import VerificationContract
+from squadops.cycles.verification_contract import (
+    VerificationContract,
+    capture_probe_values,
+    resolve_probe_path,
+)
 from squadops.sandbox.environment import FULLSTACK_FASTAPI_REACT
 from squadops.sandbox.models import RevisionOrigin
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #676, opened 2026-08-03
fix(runtime): reap stranded runtime_activities — finalize sweep + startup reaper (#672)

jasonLadd:
1.4.1 hardening patch, fix 1 of 5 (`docs/plans/1-4-1-hardening-patch-plan.md`).

## Problem

`runtime_activities` rows stranded by interrupted runs are never reaped. Four rows sit in `state='running'` from July 21–25 cycles (data/eve/max/neo), so every subsequent run's first task per affected agent trips `uq_runtime_activities_one_active_per_agent`. The write is best-effort, so runs proceed — all of FAY windows 2–3 ran over it — but activity tracking is silently dead for those agents and the recurring traceback is dispatch-log noise that can mask a real failure. Leases got release-on-finalize hardening; activities got neither (exact sibling of the #373 class).

## Fix — both halves from the plan sketch

1. **Run-finalize sweep** (`RunCompletion.finalize`): task dispatch for the finishing run is over, so any activity still active for its cycle is stranded — abort them all. The executor threads its activity port into the default `RunCompletion`, so the sweep is live in the real composition. Prevents new strandings on every terminal path (completed/failed/cancelled/paused).
2. **Startup reaper** (runtime-api init): end active activities whose owning cycle is terminal — clears the four historic rows at first deploy, and covers process deaths that skipped finalize.

Machinery lives in a new runtime-domain module, `squadops/runtime/activity_reaper.py`, on a new bounded port read `RuntimeActivityPort.list_active_activities` (at most one active row per agent — D9 — so results are roster-bounded).

## Design notes

- **Per-row best-effort isolation**, same shape as `admission.release_participants`: one bad row (predicate or abort failure) never blocks clearing the rest, since one stranded row blocks all of that agent's future tracking.
- **Race-safe by construction**: both sweeps go through the existing `abort_activity` → `update_state` choke point, whose active-only `WHERE` makes a concurrent terminalization a no-op (`None`, not counted).
- **Layering**: `squadops.runtime` stays free of the cycles domain — cycle-terminal knowledge enters the reaper as an injected async predicate. The composition root builds it via `get_cycle` + `list_runs` + `derive_cycle_status`, the same fetch-and-derive shape the API routes use (no shared helper exists; none added).
- **New reason codes** (event/log-surfaced, not persisted, per the existing contract): `activity_stranded_at_run_finalize`, `activity_stranded_cycle_terminal`.

## Known conservative bounds (intentional)

- The startup reaper keys on derived `CycleStatus` ∈ {completed, failed, cancelled}. The all-runs-cancelled-but-cycle-not-cancelled edge derives `CREATED` and is left alone (can't prove the row is dead cheaply); none of the four historic rows are in it.
- A cancel that never reaches finalize (the #529/#481 orphaned-teardown class) clears at the next restart rather than immediately. In-scope fixes for those paths stay with their issues.

## Tests (15 new)

- `tests/unit/runtime/test_activity_reaper.py` (8): live-cycle rows spared / terminal-cycle rows reaped, cycle-less (duty/ambient) rows untouched, per-row survival of predicate and abort failures, concurrent-terminalization not counted, finalize sweep scoped to its own cycle, quiet no-op on clean runs.
- `tests/unit/runtime/test_runtime_activity.py` (3): new port read filters to active states + oldest-first, cycle filter actually binds, unfiltered read sends no params.
- `tests/unit/cycles/test_run_completion.py` (3): finalize aborts the cycle's active rows with the right reason, survives a sweep failure, no-port composition unchanged.
- `tests/unit/cycles/test_dispatched_flow_executor.py` (1): wiring seam — the executor's default `RunCompletion` receives the activity port (guards the silent-no-op class where the sweep never runs in production while all unit tests pass).

Existing `_FakeActivityPort` fakes extended with the new port method.

## Verification

Per the plan: no artifact replay applies (DB-state fix). Post-deploy checks at the single 1.4.1 deploy window: the four stranded rows are ended, and one fresh cycle logs zero `UniqueViolationError` / `best-effort start_activity failed` lines at dispatch. Regression suite green locally.

Closes #672

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -148,6 +148,7 @@ class DispatchedFlowExecutor(FlowExecutionPort):
             artifact_vault=artifact_vault,
             llm_observability=llm_observability,
             workflow_tracker=workflow_tracker,
+            activity_port=activity_port,
         )
 
         # SIP-0077: Cycle event bus (defaults to NoOp if not provided)
--- a/adapters/cycles/run_completion.py
+++ b/adapters/cycles/run_completion.py
@@ -39,6 +39,7 @@ from squadops.cycles.verification_integrity import (
     aggregate_verification,
 )
 from squadops.events.types import EventType
+from squadops.runtime.activity_reaper import abort_cycle_activities
 
 if TYPE_CHECKING:
     from squadops.cycles.models import Cycle
@@ -46,6 +47,7 @@ if TYPE_CHECKING:
     from squadops.ports.cycles.artifact_vault import ArtifactVaultPort
     from squadops.ports.cycles.cycle_registry import CycleRegistryPort
     from squadops.ports.cycles.workflow_tracker import WorkflowTrackerPort
+    from squadops.ports.runtime.activity import RuntimeActivityPort
     from squadops.ports.telemetry.llm_observability import LLMObservabilityPort
     from squadops.tasks.models import TaskEnvelope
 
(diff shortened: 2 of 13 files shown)


jasonLadd (commented):
reviewed

PR #677, opened 2026-08-03
fix(plan-validation): reject import_present on modules the scaffold surface cannot provide (#671)

jasonLadd:
1.4.1 hardening patch, fix 2 of 5 (`docs/plans/1-4-1-hardening-patch-plan.md`). Validator pair with #673.

## Problem

fay-17's approved framing-2 plan (`cyc_e175cae83a6f`) carried a blocking `import_present: module: app.routes` — the `app` package doesn't exist (scaffold root is `backend`, the pf-26 package-root class). Satisfying the check guarantees a collection-dead suite; omitting it fails acceptance. Contradictory by construction, invisible to the #645-era validators (which cover command executability and artifact shapes, not import-shaped params).

## Fix

New `ImplementationPlan.validate_module_existence(contract)`, wired at both nets like its #645/#658 siblings (gate reject-and-teach → free framing re-roll per #522; dispatch `CycleError` backstop). It proves absence from the contract's **closed** file surface (frozen + fill paths → importable module paths, the same conversion `ModuleImportsCheck` uses):

- **Rejects** a module under a scaffold package root that the surface doesn't contain (`backend.handlers` — the scaffold owns everything under `backend/`, so absence is proof), and a real scaffold leaf under a nonexistent root (`app.routes`, `src.backend.main` — the wrong-root hallucination).
- **Out of scope by design** (zero third-party false positives): relative specs (`.errors` — the contract authors this form itself), dotless specs (`errors` — #441 author-intent-relative semantics), and dotted third-party modules (`fastapi.testclient`).
- Only `severity=error` rejects (RC-9, same split as the #645 rules).

## Scope narrowed vs the issue sketch — entry_modules is exempt (deliberate)

The issue (and the plan doc's verify line) wanted `harness_boundary.entry_modules` existence-netted too, on the fay-16/18 `app.main` receipts. Reading the evaluator killed that premise:

- `HarnessBoundaryCheck` **fails a test that imports an entry module** — `entry_modules` is a *forbidden-import* list, not a "build against one of" list (that reading died with pf-33's wrong-directional expectation line, and the current rendering states the prohibition).
- The scaffold's own convention deliberately seeds nonexistent wrong-guess traps: `_HARNESS_ENTRY_MODULES = ("backend.main", "app.main", "main")`. A nonexistent entry is *protective* — it catches the doomed import at acceptance instead of at collection.
- An existence net there would reject plans for copying the scaffold's own convention.

So fay-16 (which listed `app.main` at three sites and scored **green**) now has a pinned regression test asserting it **passes** — the opposite of the plan doc's sketch, with the receipt in the test docstring. Flagging this explicitly for review since it amends the pre-registered verification expectations.

## Verification — stored-artifact replay

Real fay plans committed as fixtures (`tests/fixtures/fay_plans/`, artifact ids in docstrings):
- fay-17 framing-2 (`art_b12c08f02557`): **trips**, exactly one error, on task 4's `app.routes`.
- fay-16 (`art_4c38a3a19418`): passes — pins the entry_modules exemption.
- fay-19 framing-2 (`art_ff352e740f93`, the window's cleanest green): passes untouched.

Plus unit cases per class, a must-not-over-reject parametrized set, RC-9 warning tolerance, and a dispatch-backstop wiring test. Mechanical note: the contract-gated dispatch chain moved verbatim into `_validate_plan_against_contract` (C901 remedy, same pattern as prior extractions). Regression suite green: 5937 passed (+13).

Closes #671

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -2756,6 +2756,11 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                     # rejection would end the cycle (#522).
                     errors.extend(parsed_plan.validate_command_checks())
                     errors.extend(parsed_plan.validate_expected_artifact_shapes())
+                    # #673: a dual-claimed expected artifact aliases two tasks
+                    # onto one file's fate (repair mis-scoping, last-wins
+                    # emission) — provable plan-wide right here, and a system
+                    # rejection re-rolls framing for free (#522).
+                    errors.extend(parsed_plan.validate_unique_expected_artifacts())
             elif ref.filename == "interface_manifest.yaml" or artifact_type == "interface_manifest":
                 interface_content = content_bytes.decode(errors="replace")
 
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -2829,6 +2829,15 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                         f"verification_contract: {e}"
                         for e in parsed_plan.validate_frozen_artifact_ownership(contract)
                     )
+                    # #671: import_present against a module the closed scaffold
+                    # surface cannot provide is provably unwinnable here, in
+                    # microseconds — a system rejection re-rolls framing for
+                    # free where the doomed roll would burn its correction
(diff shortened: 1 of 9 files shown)


jasonLadd (commented):
reviewed

PR #678, opened 2026-08-03
fix(plan-validation): reject two tasks claiming the same expected artifact (#673)

jasonLadd:
1.4.1 hardening patch, fix 3 of 5 (`docs/plans/1-4-1-hardening-patch-plan.md`). Validator pair with #671.

**Stacked on #677** (same seam and test files) — base is `fix/671-module-existence-validation`, so this diff shows only the #673 change; GitHub retargets to main when #677 merges.

## Problem

Nothing rejected two tasks declaring the same `expected_artifacts` entry. fay-18's approved framing-2 plan (`cyc_42c44ad3af91`) carried it live: qa task 5 declared dev task 4's `backend/tests/test_runs.py` with an explicit "does NOT produce, gates the existing suite" instruction. Benign that roll by dice. The hazard: expected artifacts are load-bearing task identity — write authorization, repair scoping (#650), and missing-artifact locus routing (#665) all key on them — so a failing non-producing claimant aims its repair at a file another task owns, and dual producers hand the shipped suite to last-wins ordering (the #389 class fay-14's repair chain replayed).

## Fix

`ImplementationPlan.validate_unique_expected_artifacts()` — the family's first plan-wide cross-task rule (the ownership rules stay per-task). One error per duplicated path, naming **every** claimant and teaching the legal form: a verification-only task declares `expected_artifacts: []` and binds acceptance via `criteria_refs` (the fay-13 framing-2 receipt). No warn tier — dual-claim has no legitimate use the artifact-less form doesn't cover. Within-task repeats are deliberately not flagged (benign echo, presence-checking is set-shaped — pinned by test). Contract-free, so it fires in author mode too; wired at both nets (gate reject-and-teach, dispatch `CycleError` backstop).

## Verification — stored-artifact replay

- fay-18 framing-2 (`art_d0f7b45ad57d`, committed fixture): **trips**, exactly one error, naming tasks 4 and 5 and the duplicated path.
- fay-19 framing-2 (clean green plan): passes untouched.
- Synthetic: disjoint sets pass; three-way claim → one error naming all three claimants; within-task repeat not flagged; dispatch-backstop wiring test (author mode).

Regression suite green: 5943 passed (+6).

Closes #673

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -2756,6 +2756,11 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                     # rejection would end the cycle (#522).
                     errors.extend(parsed_plan.validate_command_checks())
                     errors.extend(parsed_plan.validate_expected_artifact_shapes())
+                    # #673: a dual-claimed expected artifact aliases two tasks
+                    # onto one file's fate (repair mis-scoping, last-wins
+                    # emission) — provable plan-wide right here, and a system
+                    # rejection re-rolls framing for free (#522).
+                    errors.extend(parsed_plan.validate_unique_expected_artifacts())
             elif ref.filename == "interface_manifest.yaml" or artifact_type == "interface_manifest":
                 interface_content = content_bytes.decode(errors="replace")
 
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #679, opened 2026-08-03
fix(correction): thread the #659 anchor surface into repair and retest envelopes (#667)

jasonLadd:
1.4.1 hardening patch, fix 4 of 5 (`docs/plans/1-4-1-hardening-patch-plan.md`).

## Problem

Third occurrence of the repair-path prompt-threading class (#555 fill-only appendix, pf-34 error-contract seam, now the #659 testid surface): the anchor inventory wired into initial-dispatch prompts never reached repair envelopes. fay-14's receipt (`cyc_42eed09efbec`): neo's **first** fill of RunDetailView honored the manifest convention (`run-detail` root + `participant-list`/`join-form`/... — `art_5ece1244ce22`); four correction rounds later the accepted view (`art_b5890e085e63`) carried only ad-hoc testids — every repair regenerated the view blind. The fill-only appendix v4's `{{testid_surface}}` slot rendered a hole on every repair.

## Fix — pure threading, no new prompt content

**One deliberate divergence from the issue's "copy from the failed task's envelope" sketch:** in the exact fay-14 shape the *failed* task is qa.test (suite fails → SUBJECT locus → dev repair chain), and its envelope carries only `dom_testid_surface` — a same-key copy would miss the receipt case. So the surface is **re-derived from the interface manifest** at `repair_inputs` construction (already a `run_correction_protocol` param; same deriver as initial dispatch, the `_inject_authoritative_evidence` precedent in the same file). Both keys ride every repair envelope, presence-keyed, no task-type branching; each handler reads its own.

- **Dev repair** (`_render_fill_only_section`): now builds the same testid appendix section the develop handler builds at initial dispatch, into the fill-only render's variables — the empty slot fills.
- **qa.test_repair**: gains `_render_dom_anchor_section` (same `request.qa_test_dom_anchor_appendix` asset qa_test renders, same role-gating shape) plus a new `dom_anchor_section` slot in `request.cycle_repair_task` (v1→v2). Without it the re-authored suite asserts invented render details — the fay-6/fay-12 churn class through the repair door.
- **Retest** (`reexecute_repaired_suite`): `dom_testid_surface` presence-keyed like #639's probes and #643's workspace — the retest re-dispatches qa.test which re-authors from scratch (the fay-6 new-dice path) and was equally blind.

## Verification — deterministic replay of the fay-14 shape

- `test_repair_envelope_carries_the_anchor_surface_from_the_manifest`: rebuilds the fay-14 cross-chain repair (failed qa.test → dev repair) with the real manifest fixture; the repair envelope must carry the anchor lines including `run-detail`. Fails without the fix.
- Render side: threaded lines reach the testid appendix and land in the fill-only variables (dev); the dom-anchor block renders from the same asset with role-gating pinned (qa); handle() wiring test guards the renders-but-never-reaches-prompt no-op; template-contract test pins the new slot.
- Boundaries: manifest-less corrections thread no keys; anchor-less retests thread no key; all existing exact-args fill-only assertions unchanged.

12 new tests. Regression suite green: 5958 passed (+15).

**Deploy note (single 1.4.1 window):** template + handler changes ride in agent images (neo/eve) — the standard rebuild-all covers it; no re-seed (no contract/manifest content touched — hash-stability assertion at deploy stands).

Closes #667

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -853,7 +853,10 @@ class CorrectionRunner:
         # the SUBJECT and would point a test re-author at app source files).
         repair_artifacts: list[dict[str, Any]] = []
         if correction_path == "patch":
+            from squadops.capabilities.scaffold import testid_surface_instructions
+
             failed_inputs = envelope.inputs or {}
+            testid_lines = testid_surface_instructions(interface_manifest)
             (
                 failure_locus,
                 repair_expected_artifacts,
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #680, opened 2026-08-03
fix(framing): re-rolls carry the prior rejection — revise, don't re-dice (#669)

jasonLadd:
1.4.1 hardening patch, fix 5 of 5 — the last item on the ledger (`docs/plans/1-4-1-hardening-patch-plan.md`).

## Problem

A #522 framing re-roll treats every system plan-validation rejection as a stochastic fault and grants fresh dice — but threads nothing about what died. The validator's teaching message lands in `gate_decisions` where no model ever reads it. Receipts: fay-10 tripped the same ownership class on all three framings → re-rolls exhausted → cycle death; fay-15's framing-1 #658 rejection named the file, the rule, and the consequence — to nobody. (The record already corrected fay-13's "validator-feedback success" credit: nothing threads; that was luck.)

## Fix — the lightest correct version from the issue

- **Executor re-roll branch**: before re-executing framing, thread `{rejection_reasons, rejected_plan_yaml}` onto the **§6.6 forwarding rail** (the same transport gate outcomes already ride). The rail is rebuilt at the next workload advance, so the context can never leak into the implementation run; a second re-roll replaces the first's context with the latest rejection. The rejected plan loads best-effort from the cancelled run's artifact refs — an unreadable artifact degrades to reasons-only, never blocks the re-roll.
- **`generate_task_plan`**: lifts the context out of `resolved_config` (it's authoring *context*, not config — every envelope's config stays clean) and injects data-only keys onto exactly the four plan-authoring tasks (#657's set: brief author + dev/qa/strategy proposers). The merger is excluded — its normal merge path is deterministic, no prompt to influence.
- **Handlers + assets**: brief author and all three proposers render a `rejection_context_section` via two new managed appendix assets — `request.plan_reroll_rejection_appendix` (authoritative "PRIOR ATTEMPT REJECTED — revise, do not repeat; keep what was sound" + the reasons) and `request.plan_reroll_rejected_plan_appendix` (the fenced prior plan for revision reference). Four templates gain the optional slot (v1→v2). All prose in assets (#448); the handlers move data only.

Net effect: `framing_max_rerolls` converts from a dice budget into a revision budget. The cross-cycle rung (rejection classes as behavioral memories) stays with the agent-memory direction per the issue.

## Verification — fay-15 framing-1 replay shape

- The end-to-end executor test rebuilds the re-roll: system rejection (the real #658 store.py teaching text) → re-roll's `execute_run` receives the context with reasons + plan YAML → **the implementation run's forwarding does not carry it** (leak boundary pinned). Plus the degrade-to-reasons-only path.
- Injection: four authoring task types receive the keys; build/merge/doc tasks never do; first-roll (no context) and reasonless-context inject nothing (a plan without its reasons would be an instruction to copy the rejected shape).
- Render: all four handlers produce the section through the real asset chain; live-render test proves the fay-15 teaching line, the revise instruction, and the rejected plan body survive the templates; handle-level wiring tests guard the renders-but-never-reaches-template no-op, plus a first-roll negative.

Regression suite green: 5974 passed. **Deploy note (single 1.4.1 window):** executor/task_plan ride runtime-api; handlers + templates ride agent images — standard rebuild-all; no re-seed (no contract/manifest content touched).

Closes #669

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -648,6 +648,20 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                     ):
                         framing_rerolls += 1
                         await self._cycle_registry.cancel_run(current_run_id)
+                        # #669: the re-roll must revise, not re-dice — thread
+                        # what died and why into the new framing's authoring
+                        # prompts on the §6.6 forwarding rail. The rail is
+                        # rebuilt at the next workload advance, so the context
+                        # never leaks past framing; a second re-roll replaces
+                        # the first's context with the latest rejection.
+                        rejected_plan_yaml = await self._load_rejected_plan_yaml(run)
+                        rejection_context: dict[str, Any] = {"rejection_reasons": list(plan_errors)}
+                        if rejected_plan_yaml:
+                            rejection_context["rejected_plan_yaml"] = rejected_plan_yaml
+                        forwarding_overrides = {
+                            **(forwarding_overrides or {}),
+                            "framing_rejection_context": rejection_context,
+                        }
                         reroll_run = await self._create_next_workload_run(
                             cycle,
                             run,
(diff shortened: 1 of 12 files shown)


jasonLadd (commented):
reviewed

jasonLadd (commented):
reviewed

Issue #668, opened 2026-07-30
DOM testid contract needs its enforcement layer — qa suites ignore the anchors prompts-only (fay-14)


jasonLadd:
## Class
#659's pre-registered enforcement layer: the DOM testid contract shipped prompts-only ("prompts+anchors load-bearing", enforcement deliberately deferred pending live evidence). First live evidence is in: the qa half was non-compliant from authoring — prompts alone under-delivered on the suite side.

## Receipts
- **fay-14** (`cyc_42eed09efbec`/`run_ee0d98d19f7a`): eve's RTL suites made **zero** TestId queries in every version — first version 16 `getByText` / 3 `getByRole` / 3 `getByPlaceholderText` (art_817a857e754c), final 26/14/14 (art_428bb2c5468c) — while the same deploy's dev half provably complied at first fill (manifest-convention anchors in art_5ece1244ce22). The injection path is verified loaded (qa_test.py `_dom_anchor_section` + appendix asset in the eve image + manifest v4 testids seeded); prompt bundles aren't persisted for these tasks, so the rendered prompt can't be post-hoc proven, but the render path is deterministic from loaded code + seeded manifest.
- DOM channel is now 3-for-3 windows as the sole green-killer when an RTL task exists (fay-6, fay-12, fay-14).

## Honest scope note
Anchors arbitrate WHERE the suite looks, not the data-fetch contract. fay-14's specific failing assertions were an `apiFetch` mock-signature/state-hydration mismatch against the **frozen** `frontend/src/api.js` client — testid-only queries would still have found an empty `<h1>`. Enforcement targets the selector-churn half of the channel (fay-6's multiple-match crash, fay-12's text-expectation mismatches); the mock-vs-frozen-client mismatch is an adjacent salience gap (the client signature is deterministic and knowable — same "system holds the answer, never shows the author" root as #629) and is noted here for the owner, not claimed by this issue.

## Options (owner call — same fork shape as #593)
1. **Typed check on the suite** (e.g. `dom_anchor_queries`): blocking check that a suite covering a contract view queries its root anchor testid; optionally flag text-content queries on covered views. Deterministic, catches at task acceptance, repairs get precise evidence. Risk: over-constraint (some role/label queries are legitimate accessibility practice).
[...]

PR #694, opened 2026-08-03
docs(1.4.2): rule #689's seam + vehicle, correct two stale plan references

jasonLadd:
Three corrections to `docs/plans/1-4-2-correction-aim-patch-plan.md`, all found while grounding the plan against the code before opening the first fix branch. Docs only — no code, no behavior change.

## D0 — new decision, ruled: #689's seam

The plan left D1 (which linter) open but assumed the seam. Grounding it surfaced a decision that has to come *first*, because it decides hash stability.

The natural home for an F821 check is `scaffold_contract.py::_routes_criteria` — that is already where #628's `module_imports` hangs off `backend/routes.py`, for the same reason (static checks passed, the module still blew up). But `_routes_criteria` **generates the verification contract**:

- a new row there moves the contract hash → violates this patch's stated hash-stability constraint (`contract v9 art_4f368ea08799` must not move);
- and bind mode **loads** the pinned contract from `contract_ref` rather than regenerating it, so a contract-authored check would be **invisible in the confirmation shakedown** and would first appear at a re-seed — which this plan explicitly defers to the #668 window.

Ruled: **framework-injected at emission acceptance**, not contract-authored. That keeps the check live under the pinned v9 contract.

## D1 — ruled: pyflakes, imported in-process

Facts that decided it:

- acceptance checks evaluate in the **agent container** — the `frontend_compiles` docstring is explicit that npm runs against "the agent container's warm cache";
- neither `ruff` nor `pyflakes` is in `requirements/agent.lock` today, so *either* choice adds a dep — that was not a differentiator;
- pyflakes is pure-Python with zero transitive deps and is the reference F821 implementation that ruff's F-rules reimplement. ruff would put a ~25MB binary wheel in seven agent images to shell out for one rule;
- hand-rolled AST is rejected — F821 scope analysis (comprehension scopes, class bodies, star imports, conditional definition, `__all__`, builtins) is exactly the wheel pyflakes already is.

**Corollary recorded, and it matters more than the choice:** this check must *not* inherit the skip-on-missing-tooling precedent. #462's skip-never-fail rule and the SIP-0096 `check_tooling.py` seam exist for tooling with **per-role variance** provisioned through `agents/instances/<role>/system-packages.txt` — Node lives in the qa image only, which is why it earns `TOOL_NODE`. A base-lock pip dep is present in every agent image *by construction*, so it is not provisioned tooling in the §6.3 sense and gets no `TOOL_` identifier. An ImportError on pyflakes is therefore a **build defect, not an environment gap**, and must surface as `error`, never `skipped` — otherwise #689 ships as exactly the "looks-enforced-but-isn't" no-op SIP-0096 exists to kill.

Plus a forward-compatibility note: checks run in-agent-container today only because SIP-0102 migration step 3 (routing `test_runner.py` / `CommandExitZeroCheck` / `probe_runner.py` through typed sandbox ops) is a v1.6 S-lane rider. The check should be a registry entry with a single execution call site so step 3 can reroute it without a rewrite.

## Two stale references corrected

- The OUT list called **#668 + #593** "the hash-moving pair." **#593 was closed 2026-07-28 by PR #634** (blank-input rejection made scaffold-owned and probe-pinned) — five days before this plan was written. The seed-roll window is **#668 alone**, still held on the owner's enforce-vs-advisory call.
- The **#114** rider was already discharged (closed 2026-08-03, before this patch opened). Struck from the rider list; the dead `command_check_safelist` key remains the only one.

## Verification

`tests/unit/architecture/test_docs_version_sync.py` — 13 passed (the docs-drift guards that police version markers, SIP target parity, and doc-ref existence).

No `Closes` — this corrects planning docs; the issues it discusses stay open and are fixed by their own PRs.

### DIFF
--- a/docs/plans/1-4-2-correction-aim-patch-plan.md
+++ b/docs/plans/1-4-2-correction-aim-patch-plan.md
@@ -43,11 +43,42 @@ rejected with disclosure in the validation result.
 ### 3. #689 — undefined-name (F821) typed check on `.py` fill slots
 Error-severity, auto-applied to `.py` fill-slot emissions, catching used-but-unimported
 symbols statically at acceptance — the class every current gate missed.
-**Open decision D1 (resolve in the fix PR):** implementation vehicle — `ruff --select
-F821` requires ruff in the *agent* images (acceptance checks evaluate in handler
-context); `pyflakes` is a lighter dep; an AST-based custom check needs no dep but
-re-implements scope analysis. Default lean: pyflakes if the dep is acceptable, else
-ruff-in-image; hand-rolled AST only as last resort.
+**Decision D0 — the seam, RULED (2026-08-03, pre-build):** the check is
+**framework-injected at emission acceptance**, NOT a contract-authored criterion. The
+tempting home is `scaffold_contract.py::_routes_criteria`, where #628's `module_imports`
+already hangs off `backend/routes.py` — but that function *generates the verification
+contract*, so a new row there moves the contract hash, violating this patch's
+hash-stability constraint. Worse, it would be untestable in the confirmation window:
+bind mode **loads** the pinned contract from `contract_ref`
+(`dispatched_flow_executor.py` `_is_bind_mode` / contract loading) instead of
+regenerating it, so a contract-authored check would stay invisible until a re-seed —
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #695, opened 2026-08-04
fix(correction): aim repairs at the fill slot that owns the failing probe (#688)

jasonLadd:
First fix of the 1.4.2 patch line (`docs/plans/1-4-2-correction-aim-patch-plan.md`). Hash-stable: no contract or manifest change.

## What shk-2 lost, and why

`backend/routes.py` used `RunEvent` without importing it → `NameError` at request time → `vc-probe-runs` answered 500. Both repairs emitted `backend/main.py` + `tests/test_api.py` and **never routes.py**, so each attempt reproduced the identical 500 and the correction budget ran out.

The root cause is not "the analyzer guessed" (that's #687). It's that **every path into the repair target was indirect**:

| Path | What it names | shk-2 |
|---|---|---|
| Drift branch (#531/#532) | files the interface-drift evidence named | `backend/main.py` — dev's unauthorized `/health` route. Real, but harmless to the failing probe |
| Failed task's own artifacts (#531) | the qa suite | `tests/test_api.py` |
| Package-scoped union (pf-24/pf-27) | dev source sharing a top-level package with the failed artifacts | **empty** |

That last row is the finding. `_scope_to_shared_packages` anchors on the failed qa task's own artifacts. shk-2's suite was authored at **root-level `tests/`**, so the anchor package was `tests`, and `backend/routes.py` (package `backend`) was filtered straight out. fay-16, fay-17, fay-18 and fay-19 all authored `backend/tests/…`, which matched — so pf-24's and pf-27's fixes worked in those rolls and were silently inert here. **The correction loop's reach into app source depended on where the squad happened to put its tests.**

## The fix

The direct chain the issue specifies: **failed probe row → the probe's declared `METHOD /path` → the contract's endpoint→fill-slot map → the file that owns the endpoint.** Those slots now *lead* the target set in both branches. Drift files and the failed task's own artifacts still ride — they carry real defects too (the pf-21 lesson) — but they can no longer displace the defect site.

- **`VerificationContract.endpoint_owners()`** — the endpoint→slot relation, read off the `endpoint_defined` criteria. The contract is the only artifact that states it: the interface manifest declares the endpoints, the blueprint decides which slot implements them, and only the expander's output joins the two. `Probe.endpoint_token()` renders the joining key (placeholders preserved — it joins against contract endpoints, not a live request).
- **`task_plan`** threads `contract_endpoint_owners` onto qa.test envelopes beside `contract_probes` — data only (#448), and only when probes exist, so probe-less contracts and author mode stay byte-identical.
- **`correction_runner._probe_owned_slots()`** resolves it. Only `failed` rows count: `skipped` means the subject never booted, which indicts no particular endpoint.
- The `"METHOD /path"` token grammar moves from `acceptance_checks` to **`acceptance_check_spec`** — its second reader is no longer an evaluator, so the grammar belongs with the rest of the vocabulary rather than duplicated or reached into privately. `CHECK_ENDPOINT_DEFINED` binds the filter to the `CHECK_SPECS` key so a rename can't silently degrade this to "no endpoints owned" (that would revert the fix to the no-op it exists to end).

## Replay verification

Rerun of target resolution against shk-2's stored artifacts — contract v9 `art_4f368ea08799`, plan `art_4b9606338510`, and the failed check ids from `run_verification_summaries` for `run_da25453895e3` (`vc-probe-runs`, `vc-probe-runs-join`, `vc-probe-runs-join-duplicate`, `vc-probe-runs-leave`, `tests_pass`):

```
endpoint_owners():
  GET /runs                    -> backend/routes.py
  POST /runs                   -> backend/routes.py
  GET /runs/{run_id}           -> backend/routes.py
  POST /runs/{run_id}/join     -> backend/routes.py
  POST /runs/{run_id}/leave    -> backend/routes.py

qa expected_artifacts    : ['tests/test_api.py']
implementation_artifacts : ['backend/routes.py', 'frontend/src/views/...']

OLD target : ['backend/main.py', 'tests/test_api.py']
             (package anchors ['tests'] -> scoped source [])
NEW target : ['backend/routes.py', 'backend/main.py', 'tests/test_api.py']
```

The OLD row reproduces exactly what both shipped repairs emitted (`art_deaec27453ae` main.py + `art_b74127a97e5b` tests/test_api.py) — the historical bug reproduced from stored evidence, then fixed.

## Tests

32 new across both commits. The ones that earn their place: the shk-2 replay (exact target list), a pin on the *mechanism* so a regression is diagnosable rather than mysterious, skipped-and-passing probes indicting nothing, malformed wire rows staying inert (a raise here strands the run), and the trailing-slash join. Added for the fallback: the suite-only case reaching routes.py while frontend source stays out, package scoping still winning when it matches, the mirror frontend case, and mixed-language anchors widening nothing. Regression suite **6021 passed, 1 skipped**.

## Second commit: the package-scoping hole itself (folded in on review)

The chain above is probe-backed by construction, so it doesn't cover a **suite-only** failure — no probes, or probes passed while `tests_pass` failed. On a root-level-`tests/` layout that case still had no route to app source at all.

`_scoped_implementation_surface` keeps package scoping as the primary rule — strictly narrower, unchanged whenever it matches anything — and falls back to the **language boundary** only when it comes back empty, because an anchor that matches nothing was uninformative, not exclusive. The language line is what RC2's stated guarantee actually means ("a backend failure can never regress frontend source"), and unlike the directory tree it holds wherever the suite was authored. Anchors straddling both sides widen nothing: excluding nothing is not scoping.

`FRONTEND_SUFFIXES` moves to `acceptance_check_spec` alongside the route grammar, now that repair targeting reads the same family the evaluators do.

Suite-only failure on the shk-2 layout, before and after:

```
OLD → ['backend/main.py', 'tests/test_api.py']
NEW → ['backend/main.py', 'tests/test_api.py', 'backend/routes.py']
      frontend views stay out — the RC2 guarantee survives the widening
```

Closes #688

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -127,6 +127,77 @@ def _widen_target_for_frontend_build(
     return list(dict.fromkeys([*target, *frontend_source]))
 
 
+def _failed_probe_ids(failure_evidence: Any) -> list[str]:
+    """Ids of the behavioral probes that FAILED, in evidence order (#688).
+
+    Probe rows enter ``validation_result.checks`` from ``probe_check_rows`` with
+    ``check`` == ``criterion_id`` == the probe id and a ``status`` of
+    passed/failed/skipped. Only ``failed`` counts: ``skipped`` means the subject
+    never booted, which indicts no particular endpoint.
+    """
+    if not isinstance(failure_evidence, dict):
+        return []
+    rows = (failure_evidence.get("validation_result") or {}).get("checks") or []
+    return [
+        str(row.get("check"))
+        for row in rows
+        if isinstance(row, dict) and row.get("status") == "failed" and row.get("check")
+    ]
+
+
+def _probe_owned_slots(failure_evidence: Any, failed_inputs: dict[str, Any]) -> list[str]:
+    """Fill-slot files owning the endpoints whose behavioral probes failed (#688).
+
+    The deterministic chain the shk-2 loss chain needed and did not have:
+    failed probe row → the probe's declared ``METHOD /path`` → the contract's
+    endpoint→fill-slot map → the file that owns the failing endpoint.
+
+    shk-2 (cyc_88162ecfd895): ``vc-probe-runs`` answered 500 because
+    ``backend/routes.py`` used ``RunEvent`` without importing it. Both repairs
+    emitted ``backend/main.py`` (named by interface-drift evidence) plus the
+    failed qa task's own suite, and never ``routes.py`` — the target set could
+    not name the defect site, so the loop reproduced the identical 500 and
+    exhausted. This resolution names it from contract data alone, independent of
+    the drift evidence and of where the squad chose to put its tests.
+
+    Empty whenever the inputs are absent (author mode, probe-less contracts,
+    pre-#688 envelopes) or no probe failed — the caller's target is then
+    byte-identical to its prior behavior.
+    """
+    owners = failed_inputs.get("contract_endpoint_owners") or {}
+    probes = failed_inputs.get("contract_probes") or []
+    if not owners or not probes:
+        return []
+    failed_ids = _failed_probe_ids(failure_evidence)
+    if not failed_ids:
+        return []
+
+    from squadops.cycles.verification_contract import Probe
+
+    tokens_by_id: dict[str, str] = {}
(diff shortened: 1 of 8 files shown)


jasonLadd (commented):
reviewed

PR #696, opened 2026-08-04
fix(correction): stop reporting scaffold-frozen invariants as producer drift (#691)

jasonLadd:
Second fix of the 1.4.2 patch line. Hash-stable: no contract or manifest change.

**The issue was rewritten before building it** — its original premise didn't survive the artifact metadata, and the correction is on the issue. Details below.

## What the provenance says

The two artifacts the original filing cited as dev's unauthorized frozen writes:

```
art_1b357644220a  backend/main.py    producing_task_type: scaffold.expand, scaffold_seeded: true
art_4b4a4cb88095  backend/models.py  producing_task_type: scaffold.expand, scaffold_seeded: true
```

Both hash **exactly** to the contract's frozen entries. They are scaffold seeding, not an emission. Dev task 0 emitted one file — `backend/routes.py`, its one authorized slot. **There was no unauthorized dev write.** SIP-0100 held.

## The real defect

`GET /health` is the **scaffold's own** readiness probe — `main.py`'s docstring calls it "the deterministic readiness probe" — and the interface manifest declares only the five business endpoints, by design. Drift detection over a scaffold-seeded frozen file therefore reports a route that was never a producer defect:

```
DRIFT kind=route_drift file=backend/main.py extra=['GET /health']
```

…on a file whose sha256 equals the contract's frozen hash. Systematic, not incidental:

1. `qa.test` resolves `artifact_contents` with the whole-accepted-tree filter, so frozen files **always** enter the correction path's evidence.
2. The finding pins `_resolve_repair_target` to its drift branch and heads the repair target with a file no producer may write — three shk-2 repairs rewrote main.py.
3. It reaches the analyzer labelled *authoritative deterministic evidence*, which is why both failure analyses cited "unauthorized GET /health route in backend/main.py". The analyzer wasn't guessing there — it was correctly reading false evidence.

Every bind-mode cycle carries this. It only bites when a correction happens, which is why the green rolls never surfaced it.

**Fix:** `detect_interface_drift` takes `frozen_paths` and excludes them before any analysis, comparing on the D7 canonical identity so an alias (`./backend/main.py`) can't smuggle a frozen file back in. The correction runner passes the bound record's paths — it already had the record four lines above the call. Omitted (author mode, unbound runs) leaves prior behavior exactly.

## Second finding, same run

Both repair attempts' `backend/main.py` artifacts (`art_deaec27453ae`, `art_7ffa879238ee`) hash to the frozen bytes. SIP-0100 restored each write, then **stored the restored copy under `development.correction_repair`**. So each attempt produced *zero net change* to the file it was aimed at, nothing said so — and that stored duplicate is what fed the phantom drift above.

**Fix:** frozen-path emissions are **dropped with disclosure** rather than restored-and-stored. This does not reopen the 2.4 decision: that rationale was against a **response-atomic reject** ("would break every bind-mode build"). Per-artifact **drop** keeps siblings and is the disposition the QA and builder lanes already use — the same docstring draws exactly that distinction. Dropping cannot lose the file: a bound run's scaffold seeds every frozen path as its own artifact, selected by artifact type in every workspace filter, independent of producer. `restore_frozen_files` remains the D2 restoration authority at the filesystem seam.

## Replay verification (shk-2 stored artifacts)

```
=== Part A: interface drift ===
  OLD (no exclusion)    -> route_drift backend/main.py extra=['GET /health']
  NEW (frozen excluded) -> (none)

=== Part B: repair emitting the frozen backend/main.py ===
  emitted : ['backend/main.py', 'tests/test_api.py']
  accepted: ['tests/test_api.py']
  evidence: frozen_path_emission disposition=dropped path=backend/main.py
            attempted_sha256 == expected_sha256: True
```

That last line is the proof the repair's write was *already* a no-op — it just wasn't disclosed.

## Tests

9 new: the shk-2 drift shape, the exclusion staying surgical (real drift in a writable slot still reported alongside an excluded frozen sibling — a blanket suppression would silently disable the whole diagnosis), canonical-alias matching, prior behavior preserved when omitted, and the **wiring seam** (the runner must actually pass the paths, or the exclusion exists and is never reached).

10 existing tests re-pointed at the stronger invariant — dropped rather than restored. One was reframed rather than deleted: `test_restore_uses_bound_bytes_not_the_live_expander` proved D2 authority through the restored content, which no longer exists at that seam, so it now proves the same thing through the evidence's `expected_sha256`; the filesystem-seam test beside it already covers restoration itself.

Regression suite **6029 passed, 1 skipped**.

Closes #691

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -353,6 +353,7 @@ def _inject_deterministic_evidence(
     interface_manifest: Any,
     artifact_contents: dict[str, str] | None,
     scaffold_enforcement_carry: list[str] | None,
+    bound_record: Any = None,
 ) -> None:
     """Deterministic authoritative-evidence injection for the correction chain.
 
@@ -360,8 +361,11 @@ def _inject_deterministic_evidence(
     never LLM output, and each travels the same failure_evidence →
     authoritative-prompt-block transport. Additive: absent sources inject nothing.
 
-    - ``interface_drift`` (piece 1): exact renamed identifiers vs the manifest.
-    - ``scaffold_enforcement`` (3.4b): prior attempts' frozen-restore instructions.
+    - ``interface_drift`` (piece 1): exact renamed identifiers vs the manifest,
+      with the bound record's frozen paths excluded (#691 — a scaffold-owned file
+      cannot drift, and reporting that it does aims repairs at bytes the producer
+      may not write; shk-2 lost three attempts to exactly that).
+    - ``scaffold_enforcement`` (3.4b): prior attempts' frozen-emission instructions.
     - ``contract_expectations`` (pf-31 Fix A): the failed task's typed criteria
       as exact expectation lines.
     - ``error_contract`` (pf-34): the ApiError raise convention + code→status
@@ -382,7 +386,11 @@ def _inject_deterministic_evidence(
     from squadops.cycles.contract_expectations import expectation_lines
     from squadops.cycles.interface_conformance import detect_interface_drift
 
-    drift = detect_interface_drift(interface_manifest, artifact_contents)
+    drift = detect_interface_drift(
+        interface_manifest,
+        artifact_contents,
+        frozen_paths=bound_record.frozen_paths() if bound_record is not None else None,
+    )
     if drift:
         failure_evidence["interface_drift"] = [
             {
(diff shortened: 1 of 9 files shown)


jasonLadd (commented):
reviewed

PR #697, opened 2026-08-04
feat(acceptance): catch used-but-never-imported names on emitted Python (#689)

jasonLadd:
Third fix of the 1.4.2 patch line. Hash-stable: no contract or manifest change. Both open decisions were ruled before building (PR #694, issue comment on #689) and are implemented as ruled.

## The gap

shk-2's dev task 0 shipped a `backend/routes.py` whose `create_run` called `RunEvent(...)` without importing it — and passed **every** gate:

| Gate | Why it missed |
|---|---|
| `command_exit_zero` (py_compile) | valid syntax |
| `import_present` | checks the imports that ARE there |
| `endpoint_defined` | reads decorators |
| `module_imports` (#628) | the module imports fine — the name resolves only when the handler runs |

First invocation was the qa suite: 500, probe cascade, correction budget spent without reaching the defect.

## D0 — the seam: framework-injected, not contract-authored

The natural home is `scaffold_contract._routes_criteria`, where #628's `module_imports` already hangs off routes.py for the same reason. Two things rule it out: that function **generates the verification contract**, so a row there moves the contract hash; and bind mode **loads** the pinned contract from `contract_ref` rather than regenerating it, so a contract-authored check would be invisible until a re-seed — which this plan defers to the #668 window. The fix would ship unverified.

Injection happens at emission acceptance, scoped to the producer's **own** emitted artifacts, so it can never indict a file the task did not write. An authored row for the same file wins (the author may have chosen a different severity; two rows would double-count in the evidence).

## D1 — the vehicle: pyflakes, in-process

Neither ruff nor pyflakes was in the locks, so either choice added a dependency — not a differentiator. pyflakes is pure Python with **zero transitive dependencies** and is the reference F821 implementation ruff's F-rules reimplement; ruff would put a ~25MB binary wheel in seven agent images to shell out for one rule. Hand-rolled AST was rejected: F821 scope analysis is exactly the wheel pyflakes already is. Added to `requirements/base.txt` (both the agent images and runtime-api evaluate checks — `patch_verification` calls `evaluate_criterion`), pinned exactly because this library's verdicts gate acceptance.

**The corollary matters more than the choice:** a missing pyflakes is an `error`, never a `skipped`. #462's skip-never-fail rule is for tooling with **per-role variance**, provisioned through `agents/instances/<role>/system-packages.txt` and declared via SIP-0096's `check_registry` — Node lives in the qa image alone, which is why it earns `TOOL_NODE`. A base-lock pip dependency is present in every image by construction, so its absence is a build defect. Skipping would ship this as exactly the looks-enforced-but-isn't no-op SIP-0096 exists to kill, on the class that already cost a full correction budget.

`framework_injected` keeps the check out of the rendered proposer vocabulary: a check the system always runs is not a decision to hand an author, and on a bind-mode cycle an authored row is the one form that can't reach the task.

## Verification against the stored artifact store

**Trip** — shk-2's `backend/routes.py` v2 (`art_87442fcf46db`):

```
status=failed  undefined=['RunEvent']
reason=undefined name(s): RunEvent (line 23)
```

**Clean** — every producer-emitted `.py` in the GREEN shk-1 roll (`cyc_b03d203df3f2`): 5/5 pass, including the qa suites.

**Sweep** — the false-positive question answered empirically rather than asserted: **765** producer-emitted `.py` artifacts across every stored cycle, **36 flagged (4.7%)**. I inspected the two flags likeliest to be false positives and both are real defects:

- `monkeypatch` used as a bare name inside a test that never takes the fixture (fixtures bind as parameters — `with monkeypatch.context():` in a `def test_x(client)` is a `NameError`), alongside a `uuid` used without importing it;
- `InMemoryRepository()` in a file whose only import is `pytest`.

One flag is #628's own pf-54 shape — the dropped `router = APIRouter()`.

## Rider

Deleted the dead `command_check_safelist` CRP key: declared at `schema.py:48`, read by nothing (2026-08-03 audit), superseded by the code-owned `COMMAND_SAFELIST` plus the typed vocabulary. A key a profile can set and nothing consumes is the same looks-configured-but-isn't class this patch line keeps closing.

## Tests

25 new. The ones that earn their place: the shk-2 shape (name used only in a function body), all-names-reported (a repair fixes what the evidence names — drip-feeding one of three costs rounds), the missing-analyzer corollary, and a parametrized sweep of scoping shapes that would break real emissions if flagged — pytest fixture parameters, star imports, conditional definition, `global`, comprehensions, class scopes, forward annotations.

12 existing tests re-pointed. Two changes worth naming: the shared `_art` fixture's default content was `"x" * 200`, which is a single undefined name once a `.py` emission is checked, so it is now valid Python; and `test_summary_log_skipped_when_no_typed_checks` now uses a non-Python artifact, because a `.py` emission genuinely does have a typed check to summarize. The builder tests changed for a real reason too: an assembly emitting `.py` now always produces the #114 typed-check evaluation artifact, which is how a green roll will show the check ran at all.

Regression suite **6051 passed, 1 skipped**.

Closes #689

### DIFF
--- a/requirements/agent.lock
+++ b/requirements/agent.lock
@@ -126,6 +126,8 @@ pydantic==2.12.5
     #   langfuse
 pydantic-core==2.41.5
     # via pydantic
+pyflakes==3.4.0
+    # via -r requirements/base.txt
 pylance==0.12.1
     # via lancedb
 python-dotenv==1.0.0
--- a/requirements/api.lock
+++ b/requirements/api.lock
@@ -89,6 +89,8 @@ pydantic==2.12.5
     #   langfuse
 pydantic-core==2.41.5
     # via pydantic
+pyflakes==3.4.0
+    # via -r base.txt
 pyjwt==2.11.0
     # via redis
 python-dotenv==1.0.0
--- a/requirements/base.lock
+++ b/requirements/base.lock
@@ -34,6 +34,8 @@ pydantic==2.12.5
     # via -r base.txt
 pydantic-core==2.41.5
     # via pydantic
+pyflakes==3.4.0
+    # via -r base.txt
 python-dotenv==1.0.0
     # via -r base.txt
 pyyaml==6.0.1
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -12,3 +12,9 @@ httpx>=0.27.0
 
 # Validation
 jsonschema>=4.20.0
+
+# Undefined-name analysis for the `undefined_names` acceptance check (#689).
+# Pure Python, zero transitive deps, and the reference F821 implementation that
+# ruff's F-rules reimplement. Pinned exactly: this library's verdicts gate task
+# acceptance, so a silent upgrade would change what the squad may ship.
+pyflakes==3.4.0
--- a/src/squadops/capabilities/handlers/cycle/base.py
+++ b/src/squadops/capabilities/handlers/cycle/base.py
@@ -15,6 +15,7 @@ from squadops.capabilities.handlers.base import (
     HandlerEvidence,
     HandlerResult,
 )
+from squadops.cycles.acceptance_check_spec import CHECK_UNDEFINED_NAMES, is_check_applicable
 from squadops.cycles.acceptance_checks import CheckOutcome
 from squadops.cycles.acceptance_evaluation import (
     evaluate_criterion,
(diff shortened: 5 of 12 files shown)


jasonLadd (commented):
reviewed

PR #698, opened 2026-08-04
feat(planning): state the plan-shape rules in the authoring prompts (#686)

jasonLadd:
Last fix of the 1.4.2 patch line. Hash-stable: no contract or manifest change. Agent-image surface only (templates + one new asset).

## What shk-1 paid for

Its first framing authored the #673 dual-claim shape with the contract, the interface manifest, and the typed-acceptance vocabulary all present in its inputs — because the plan-shape **rules** appeared in no authoring prompt. #673 caught it at the gate; the teaching arrived one framing re-roll and 45 minutes late. One more authoring defect would have exhausted `framing_max_rerolls` and hard-failed a cycle whose every mechanism worked.

#629's pattern — the system holds the answer and never shows the author — applied to plan authoring.

## The rules

A managed asset, `request.plan_authoring_rules_appendix`, rendered into the four authoring prompts: the brief (which pins the frame the proposers work from) and the three proposers.

**Unconditional**, unlike the sections beside it. There is no input to key on — these rules hold for every plan on every roll — and a section that rendered only sometimes would leave the *first* framing, the one shk-1 lost, exactly as uninformed as before.

Eight rules, each traced to its validator:

| rule id | validator | origin |
|---|---|---|
| `one-file-one-owner` | `validate_unique_expected_artifacts` | #673 |
| `artifacts-are-files` | `validate_expected_artifact_shapes` | fay-9 |
| `qa-owns-only-tests` | `validate_qa_artifact_ownership` | pf-39 |
| `no-frozen-claims` | `validate_frozen_artifact_ownership` | #658 |
| `imports-must-exist` | `validate_module_existence` | #671 |
| `regex-only-on-documents` | `validate_criteria_scope` | #464 |
| `commands-must-run-here` | `validate_command_checks` | #645 |
| `roles-must-exist` | `validate_against_profile` | — |

Each states the rule **and why breaking it is unsatisfiable**, so an author can reason about a case the wording didn't anticipate rather than pattern-match the examples. The dual-claim ban ships with its legitimate alternative (`expected_artifacts: []` plus criteria) — stating the ban alone would push authors to drop verification tasks instead of declaring them artifact-less.

## The part that keeps it true

The issue's real requirement was *"validators and prompts share one source of truth — any future validator lands here the day it ships."* Prose in an asset can't enforce that by itself, so the binding is checked.

`squadops.cycles.plan_authoring_rules` classifies every validator into one of three tables:

- **`AUTHOR_FACING`** → rule id, which must appear in the asset;
- **`COVERED_ELSEWHERE`** → names the asset that already teaches it. `validate_criteria_refs` lives here: the bind appendix teaches "bind, don't author" per-cycle *with the actual criterion ids*, which beats a general restatement and would drift if duplicated;
- **`NOT_AUTHOR_FACING`** → nothing the author writes can trip it.

A test asserts every `ImplementationPlan.validate_*` appears in exactly one table, that every author-facing rule id appears in the asset, and — the reverse direction — that the asset states no rule the table doesn't claim, since prose for an unenforced rule spends author attention on something the system won't check. A new validator now cannot ship without a decision about whether authors are told. Classification is data; prose is the asset (#448).

## One test refined

`test_every_template_has_required_variables` demanded at least one required variable per template. A prose-only asset can't satisfy that without inventing a variable, so I restated it against the body: **every placeholder used must be declared**. That's the bug its own docstring names ("missing variables produce empty sections instead of errors") and is strictly tighter than a count. Verified against all templates before changing it — the only pre-existing mismatch is in the other direction (`role` declared but unused in `request.cycle_repair_task`), which I left alone rather than pull an unrelated fix into this PR.

## Tests

10 new: the three-way binding guard, an unconditional render from all four handlers, a live render through the real template (a frontmatter typo would otherwise render *nothing* into the prompt meant to carry the teaching, with every handler-side mock test still green), and a per-template check that the slot is both declared and placed — declaring without placing is silently inert.

Three existing `test_propose_plan_tasks` assertions moved from `render.assert_called_once()` to asserting the main template is the last render, since counting renders now just counts how many appendices happen to exist.

Regression suite **6070 passed, 1 skipped**.

**Live proof is negative and comes at the deploy window:** shk-1's dual-claim class should not recur in the confirmation cycle, and if framing authors it anyway, the #669 re-roll receipt shows whether the rules section was present in the authoring prompt.

Closes #686

### DIFF
--- a/src/squadops/capabilities/handlers/planning_tasks.py
+++ b/src/squadops/capabilities/handlers/planning_tasks.py
@@ -138,6 +138,20 @@ class _PlanningTaskHandler(_CycleTaskHandler):
             variables["time_budget_section"] = budget_section
         return variables
 
+    async def _authoring_rules_section(self, renderer: Any) -> str:
+        """The plan-shape rules every deterministic validator enforces (#686).
+
+        Unconditional — unlike the sections around it there is no input to key on,
+        because these rules hold for every plan on every roll. shk-1's first framing
+        authored the #673 dual-claim shape with the contract, the manifest and the
+        typed-acceptance vocabulary all present, because the rules themselves appeared
+        in no prompt: the validator knew, the author never did (#629's pattern). Prose
+        is the managed asset; which validators are author-facing is the data table in
+        ``squadops.cycles.plan_authoring_rules``, and a test binds the two (#448).
+        """
+        rendered = await renderer.render("request.plan_authoring_rules_appendix", {})
+        return rendered.content
+
     async def _rejection_context_section(self, renderer: Any, inputs: dict[str, Any]) -> str:
         """The prior-rejection appendix on a framing re-roll, or "" (#669).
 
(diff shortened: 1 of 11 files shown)


jasonLadd (commented):
reviewed

PR #699, opened 2026-08-04
docs(sip): record #571 as a hard prerequisite for Cross-Cycle Memory

jasonLadd:
Verified `adapters/memory/lancedb.py` today while triaging the backlog. Two defects in the SIP-042 recall path that Cross-Cycle Memory Phase 1 would build directly on top of:

- **line 130** — `.limit(query.limit)` is applied *before* namespace and tag filtering, which happens in Python at 138–147. A query whose nearest neighbours all sit in another namespace returns **zero** results while matching entries exist.
- **line 136** — `score = 1.0 - _distance` against LanceDB's default **L2** metric. That is not a `[0, 1]` similarity; L2 is unbounded above, so scores go negative and `query.threshold` stops meaning what its name says.

Namespace-scoped recall and confidence gating are exactly §5's Phase-1 filter chain, so both defects sit directly under it.

Neither is live today — nothing leverages memory yet — and that is precisely the argument for fixing them *first*. §9's gate is a **measured recurrence-rate drop**. A starving recall path would report "memory doesn't help" when the truth is "recall never returned the memory," and that is an expensive wrong conclusion to bank.

Adds the prerequisite to §12 and amends the sequencing bullet. Docs only; the fix itself is #571, which is small and independent.

### DIFF
--- a/sips/proposed/SIP-Cross-Cycle-Memory.md
+++ b/sips/proposed/SIP-Cross-Cycle-Memory.md
@@ -403,9 +403,22 @@ Per the ratified post-1.4 reshuffle (`docs/plans/post-1-4-roadmap-reconciliation
   *learned-experience* leg of the capability-backed agent equation — **agent = identity
   (SIP-0088/0089) + capability (packs) + memory (this SIP) + policy + evidence
   history** — which is why it precedes the umbrella rather than riding inside it.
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #700, opened 2026-08-04
docs(plan): 1.4.3 patch line — the loop can't strand or hide

jasonLadd:
Successor to `docs/plans/1-4-2-correction-aim-patch-plan.md`, same discipline. Docs only.

**Character:** a cycle can neither strand the next one nor fail silently. Where 1.4.2 aimed the correction chain, this line closes the two ways the machinery stops being usable without saying so — a run that leaks state the next cycle blocks on, and a check that reports an environment gap as an outcome.

Every item was verified against the code during the 2026-08-04 backlog triage rather than read off a title, and **two changed classification on inspection**.

## The four fixes

**#373 + #529 — stranded focus leases.** The one that pays for itself immediately, and its shape is already proven: `activity_reaper.py`'s own first line reads *"Stranded RuntimeActivity hygiene (#672 — the FocusLease #373 class, applied to activities)."* We built the analogue in 1.4.1 and never built the original. #529 is the live symptom — a cancelled run leaks its leases, the next cycle deadlocks acquiring them — currently mitigated by a **manual** `select count(*) from focus_leases where released_at is null` before every launch. `FocusLeasePort` has request/renew/release/revoke/get and no listing operation, so the fix mirrors #672 exactly: port read, domain reaper, finalize sweep, startup reap.

**#498 — bare `python` in emitted argv.** Fix recorded as **evaluator-side, not emission-side**, and the reason is the hash constraint doing real work: emitting `python3` changes `scaffold_contract`'s output and moves the contract hash — inadmissible on a patch line — *and* would leave every already-seeded contract, including the v9 this window must not re-seed, still carrying the bare name. Resolving `python` → `sys.executable` at spawn fixes the pinned contracts too. Reproduced twice during 1.4.2 development.

**#572 — queue advertises a delay it can't honor.** TTL set with no DLX declared, while `capabilities()` reports `"delay": True`. **Nothing is dropped today** — the only caller passes `0` — so the defect is the declaration, which is a contract with future callers.

**#573 — literal `|` in the redaction email class.** `[A-Z|a-z]`. **Does not leak**: both ranges are still covered. One character, in a surface where a reader who mistakes intent for behavior may build on the wrong assumption.

**#605 rides as a close-with-evidence** — every AST-parsing check already calls `_unparseable_source_skip`; the four that don't (`regex_match`, `count_at_least`, `command_exit_zero`, `frontend_compiles`) never parse Python. Closed with a registry-wide test so the property is enforced rather than re-verified by hand.

## Deploy window

Notably it includes a **cancel probe** — a deliberate cancel rather than a green roll — because #373/#529's proof cannot be obtained from a passing cycle. Then the usual unfiltered confirmation shakedown, then the bump.

## Deliberately out

The **false-green family** (#423, #424, #426, #427, #571) is held for 1.4.4. Those are bugs that report success that isn't; they form a coherent line and splitting them across two windows would blur both. Also out: #668 (hash-moving, held on your #670 ruling), #687 (cross-component → 1.5, the fourth of shk-2's four causes), the structural refactors, and packaging.

### DIFF
--- a/docs/plans/1-4-3-patch-plan.md
+++ b/docs/plans/1-4-3-patch-plan.md
@@ -1,0 +1,159 @@
+# 1.4.3 Patch Plan — The Loop Can't Strand or Hide
+
+**Established:** 2026-08-04 · successor to `docs/plans/1-4-2-correction-aim-patch-plan.md`
+(same discipline: one PR per issue with `Closes`, hash-stable throughout, targeted
+verification per fix, ONE deploy window, bump only after live confirmation).
+**Opens:** after the v1.4.2 tag lands. Sequencing, not dates.
+
+## Character
+
+One coherent claim: **a cycle can neither strand the next one nor fail silently.** Where
+1.4.2 aimed the correction chain, this line clears the two ways the machinery stops being
+usable without saying so — a run that leaks state the next cycle blocks on, and a check
+that reports an environment gap as an outcome.
+
+Every item was verified against the code during the 2026-08-04 backlog triage, not read
+off an issue title. Two of the five changed classification on inspection: #605 appears
+already fixed, and #498's fix belongs somewhere other than where the issue first
+proposed.
+
+**Hash-stable by construction:** no fix touches the verification contract or interface
+manifest. This is a live constraint here, not a formality — see #498, where one of the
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #701, opened 2026-08-04
docs(plan): amend 1.4.3 — add #561, retitle the #605 rider, declare pyflakes

jasonLadd:
Follow-up to #700 (merged), from the full-backlog sweep that ran after it. Docs only.

## #561 joins as fix 2

Narrowing it during the sweep turned it into a patch item. It is now precisely the residue #672 deliberately left behind — `activity_reaper.py`'s own docstring says activities without a `cycle_id` (duty/ambient postures, SIP-0089) have no owning cycle to consult and are left alone. Cycle-owned rows heal; those don't, and one stranded row blocks all of that agent's future activity tracking.

Sequenced immediately after #373 because it is the same module, the same startup/finalize seams, the same reviewer context, and the same cancel probe. One extra PR closes a third stranding issue in a window already about stranding.

Its one design question — there is no cycle to ask, so it needs a different terminal predicate (lease expiry? heartbeat-age bound?) — is flagged for resolution **in the fix PR**, the way #689's D0/D1 rulings were recorded before building.

## #605 rider retitled

The issue is **closed** (2026-08-04, evidence on the issue): every AST-parsing check already calls `_unparseable_source_skip`, and the four that don't — `regex_match`, `count_at_least`, `command_exit_zero`, `frontend_compiles` — never parse Python. What remains is that the property is true by construction and asserted nowhere, so the rider is now the registry-wide pinning test rather than the close.

## New rider: declare `pyflakes` in `tests/requirements.txt`

1.4.2 added pyflakes to `requirements/base.txt` and the three locks, so every container has it. The **test harness** gets it only transitively, through `flake8>=6.0.0` at `tests/requirements.txt:92`.

CI and contributors are fine today *for that reason alone*. They stop being fine the day anyone retires flake8 — which is a live prospect, since ruff's own config comment in `pyproject.toml` records that its `F` rules **are** pyflakes. #689's `undefined_names` would then return `error: missing_analyzer` across the suite, with nothing pointing at why. One line, declaring a load-bearing dependency as load-bearing.

## Net

Five fixes, two riders, **six issues closed** in one hash-stable window. Character unchanged: everything added is still state that strands, or a dependency that lies about being there.

### DIFF
--- a/docs/plans/1-4-3-patch-plan.md
+++ b/docs/plans/1-4-3-patch-plan.md
@@ -13,16 +13,22 @@ usable without saying so — a run that leaks state the next cycle blocks on, an
 that reports an environment gap as an outcome.
 
 Every item was verified against the code during the 2026-08-04 backlog triage, not read
-off an issue title. Two of the five changed classification on inspection: #605 appears
-already fixed, and #498's fix belongs somewhere other than where the issue first
-proposed.
+off an issue title. Two changed classification on inspection: #605 was already fixed
+(**closed 2026-08-04**, evidence on the issue), and #498's fix belongs somewhere other
+than where the issue first proposed.
+
+**Amended 2026-08-04** after the full-backlog sweep that followed: **#561 joins as fix 2**
+(the sweep narrowed it to a residue that lands in the same module as #373, so it costs one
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #702, opened 2026-08-04
docs(roadmap): record the #670 fork-1 ruling; correct #689's stated scope

jasonLadd:
Two changes, one cause: the shk-3 confirmation roll showed #689 stopping at the qa boundary, which both settles #670 and reveals that I described #689's reach too broadly.

## #670 — ruled fork 1, slated v1.5

The ROADMAP carried a conditional ("the qa-enforcement fork **if it grows structural**"). It's now a decision with its grounds recorded, and the full rationale is on the issue.

Two things changed since #670 was filed, and together they flip it:

**Its stated prerequisite shipped.** #670 made fork 1 conditional on check-param validation against the scaffold surface — *"or the validator hands authors a loaded gun."* **#671 landed that in v1.4.1**: an error-severity `import_present` naming a module the scaffold provably cannot provide is rejected at the gate, wrong-root hallucinations included. That is fay-17's exact doom (`app.routes`). Under today's code that plan never reaches execution, so the central objection — "fork 1 would have turned fay-17's green into a loss" — no longer describes the system.

*Caveat kept explicit:* #671 shipped with `harness_boundary` `entry_modules` deliberately exempt, because the scaffold seeds nonexistent entry modules as traps on purpose. The `import_present` half is disarmed; that half isn't, and wants re-examination during implementation.

**Fork 2's price rose, and shk-3 measured it.** #689's `undefined_names` is framework-*injected* inside `_evaluate_typed_acceptance`, so a handler outside that seam receives no framework check at all. On shk-3, qa task 4 emitted `backend/tests/test_runs.py` — a Python file, exactly the artifact class #689 protects — and produced **no** `typed_check_evaluation_task_4.json`, because zero typed checks evaluated. Its four authored checks were render-only *and* the injection never arrived. Declaring advisory would have made that permanent and growing: every future framework check silently skipping qa.

So the scope is larger than "evaluate authored checks" — it's **bring `qa_test.py` into the seam**, which carries authored checks and framework injections alike.

## #689 — scope wording corrected

The spec note read *"Applied by the framework to every emitted .py artifact"* and the injection docstring *"every emitted `.py` file."* Both overstate it: `qa_test.py:73` overrides `_validate_output` and never calls the seam, and its own docstring says so (*"Typed acceptance evaluation in the QA path is out of scope for M1.3"*).

Corrected to name dev and builder explicitly, with a pointer to #670 as the ruling that retires the exclusion. **Behavior unchanged** — this is the code saying what it does rather than what I claimed for it.

Regression **6070 passed, 1 skipped**.


jasonLadd (commented):
reviewed

PR #704, opened 2026-08-04
fix(runtime): reclaim stranded focus leases at cancel, finalize, and startup (#373, #529)

jasonLadd:
First fix of the **1.4.3** line (`docs/plans/1-4-3-patch-plan.md`) — *a cycle can neither strand the next one nor fail silently.*

## The bug

A cycle FocusLease is created with `expires_at = NULL` and released only by the executor's finalize path. Anything that skips that path leaves the lease held forever, and post-#288 (which correctly hard-rejects a same-mode `cycle` request from a different owner) that is a permanent block: every later cycle recruiting those agents pauses at admission with `focus_lease_conflict` before dispatching a single task. Recovery was raw SQL — the pre-launch `select count(*) from focus_leases where released_at is null` check is in my own runbook.

Two skip paths, one class:
- **#529** — cancellation. `cancel_run` writes `cycle_runs.status`; `cancel_cycle` writes only the cycle's `cancelled` flag. Neither touches leases, and neither goes through the executor's finalize.
- **#373** — process death. Nothing ever reclaims a lease whose owner died.

We shipped this reaper's analogue in 1.4.1 and never built the original. `activity_reaper.py`'s first line says so: *"Stranded RuntimeActivity hygiene (#672 — the FocusLease #373 class, applied to activities)"*. This mirrors it.

## The fix

**`FocusLeasePort.list_active_leases(owner_ref=...)`** — the sweep read, narrowed in SQL. The agents a dead or cancelled run recruited are exactly what a caller cannot enumerate; `get_current_lease` only answers the inverse question. Backed by the existing `WHERE released_at IS NULL` partial index, so no migration.

**`runtime.lease_reaper`** — `release_owner_leases` (by owner) and `reap_stale_leases` (predicate-driven), mirroring `activity_reaper`'s two-sweep shape, both best-effort and per-row isolated.

Both drive the **coordinator**, not the lease port. This is the load-bearing decision: a stranded lease comes with a stranded `mode = cycle`. Releasing only the lease leaves the agent pinned in `cycle`, where the next recruitment hits the #288 same-mode path, finds no conflicting lease, and *idempotently skips* — admitted without ever acquiring, so finalize releases nothing and the agent is lost mid-run to the next owner. `request_transition(…, "ambient")` writes the mode and releases the lease in one unit of work and emits the canonical events. A lease the transition cannot clear — agent already `ambient`, which the null-UoW path produces when a mode write commits and its release then fails — is released through the port afterwards.

**Four call sites:**

| Site | Sweep |
|---|---|
| `runs/{id}/cancel` | by `owner_ref = run_id` |
| `cycles/{id}/cancel` | every run in the cycle — a lease under an *already-terminal* run is #373's case and blocks just as hard |
| executor `finally` | by `owner_ref = run_id`, after `release_participants` |
| runtime-api startup | every held cycle lease |

`cancel_cycle` also transitions its non-terminal runs now — #529's second half, and without it those runs read `running` forever.

The executor sweep is not redundant with `release_participants`: that clears the agents *one admission call recorded*, and a recruitment replay (#288 idempotent-skip on a resumed run) holds leases under the run that were never returned in `recruited_agent_ids`.

## Two judgment calls worth review

**1. Startup releases every held cycle lease, not just terminal-owned ones.** Runs execute in-process (the executor dispatches from inside runtime-api), so a process that has not begun serving owns no in-flight run — the same single-instance constraint the coordinator's single-writer rule already depends on. Gating on a terminal run status would have missed #373's *own evidence*: a SIGKILL leaves the run row `running` forever, so the leases under it would never clear and manual SQL would stay mandatory for exactly the reported case. The run status is looked up for the **log**, not the decision — a lease under a non-terminal run gets a WARNING naming the run, because "no error logged, so it looks like a silent hang" was the complaint. If runtime-api is ever scaled past one instance this needs leader election, which `scheduler_bootstrap` already documents as a precondition for the coordinator itself.

**2. Only `cycle` leases are swept.** A `duty` lease's `owner_ref` is an assignment id, not a run id, so a run-shaped predicate would answer "no such run → orphaned" and revoke a live duty window. Filtered before the predicate is even asked.

## Deliberately not done

#373 lists four candidate fixes. Two are in; two are not:
- **A non-null `expires_at` on `ttl` leases** — a TTL introduces a new failure mode (a long cycle whose lease expires mid-run). Reaping on owner-liveness is strictly safer and covers the same ground.
- **A CLI/doctor affordance to list and release leases** — the ask behind it was "recovery shouldn't be raw SQL". Self-healing removes the need for recovery at all. Say the word if you still want the operator surface and I'll file it.

## Structural note

Both startup reaps moved out of `main.py` into `api.runtime.startup_reaps` — `_init_cycle_subsystem` was already at the complexity ceiling and adding the lease block tripped it (C901 17 > 12). Same rationale as `scheduler_bootstrap`: the wiring gate and the terminal predicate become unit-testable without standing up FastAPI. The #672 activity reap is behavior-identical, just relocated.

## Verification

- **58 new tests**; full regression **6107 passed, 1 skipped**.
- Reaper unit tests cover: live-owner leases spared, duty leases never touched, agent returned to ambient (not lease-only), the already-ambient residue case, per-row isolation on predicate/transition/release failure, owner narrowing, and a slot re-acquired mid-sweep counting as cleared without stealing the successor's lease.
- Route tests reproduce #529 directly, plus best-effort degradation (lease-sweep failure must not turn a cancel into a 500) and the pool-less path.
- A factory-forwarding test for the same silent-no-op class #672's PR guarded — a dropped kwarg would leave the finalize sweep permanently inert with every `lease_reaper` unit test still green.
- **Hash-stable**: no contract or manifest surface touched; contract v9 `art_4f368ea08799` / manifest v4 `art_8becd104e9fc` do not move.

The live proof is the **cancel probe** in the 1.4.3 deploy window — launch, cancel mid-implementation, assert zero unreleased leases with no restart, relaunch and confirm immediate acquisition. #373/#529's proof cannot come from a passing cycle.

Closes #373
Closes #529

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -60,7 +60,9 @@ from squadops.cycles.task_plan import generate_task_plan
 from squadops.cycles.verification_normalize import normalize_task_checks
 from squadops.events.types import EventType
 from squadops.ports.cycles.flow_execution import FlowExecutionPort
+from squadops.runtime import reasons
 from squadops.runtime.admission import admit_participants, release_participants
+from squadops.runtime.lease_reaper import release_owner_leases
 from squadops.runtime.recruitment import reserve_buffer_decision
 from squadops.tasks.models import TaskEnvelope, TaskResult, TaskResultStatus
 from squadops.telemetry.context import use_correlation_context
@@ -78,6 +80,7 @@ if TYPE_CHECKING:
     from squadops.ports.events.cycle_event_bus import CycleEventBusPort
     from squadops.ports.runtime.activity import RuntimeActivityPort
     from squadops.ports.runtime.assignments import AssignmentPort
+    from squadops.ports.runtime.focus_lease import FocusLeasePort
     from squadops.ports.telemetry.llm_observability import LLMObservabilityPort
     from squadops.runtime.coordinator import RuntimeCoordinator
 
@@ -109,6 +112,7 @@ class DispatchedFlowExecutor(FlowExecutionPort):
         assignment_port: AssignmentPort | None = None,
         activity_port: RuntimeActivityPort | None = None,
         coordinator: RuntimeCoordinator | None = None,
+        focus_lease_port: FocusLeasePort | None = None,
         run_completion: RunCompletion | None = None,
         correction_runner: CorrectionRunner | None = None,
         pulse_boundary_runner: PulseBoundaryRunner | None = None,
@@ -139,6 +143,12 @@ class DispatchedFlowExecutor(FlowExecutionPort):
         # falls back to §2.5-only — a cycle never hard-fails for missing runtime
         # infra.
         self._coordinator = coordinator
+        # #373: opt-in FocusLeasePort for the finalize-time stranded-lease sweep.
+        # `release_participants` below clears the leases *this call* recorded;
+        # this clears the ones the *run* holds — the difference is a recruitment
+        # replay (#288 idempotent-skip, e.g. a resumed run), whose leases are
+        # owned by this run but were never returned in `recruited_agent_ids`.
+        self._focus_lease_port = focus_lease_port
         self._cancelled: set[str] = set()
         # SIP-0097 §6.4: run-completion collaborator (terminal path). Plain
         # injected collaborator with a default composed from this executor's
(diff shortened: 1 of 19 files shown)


jasonLadd (commented):
reviewed

PR #705, opened 2026-08-04
fix(runtime): self-heal stale activity rows and end them at cancel and startup (#561)

jasonLadd:
Second fix of the **1.4.3** line. Same character as #373/#529 — *a cycle can neither strand the next one nor fail silently* — one layer down.

## The bug

`runtime_activities` allows one active row per agent (D9, the §4.2 partial unique index). A task that ends without terminalizing its row leaves that row active forever, so **every** later `start_activity` for that agent raises `UniqueViolationError`. The §4.4 instrumentation is best-effort by design, so `task_dispatcher` swallows it — runs proceed, a traceback logs on every dispatch, and the agent's activity tracking is dead until someone runs SQL by hand.

The issue's live evidence is four agents with exactly one open row each — the D9 signature — two of them dead for **three days spanning the entire green-roll measurement arc**. The "best-effort, never break dispatch" guard is correct, and it is exactly what hid that the feature had been off for days.

## The fix — three ways in, closed in the order the row's life runs

**1. Self-heal on conflict** (the issue's headline recommendation). `start_activity` catches `UniqueViolationError` *specifically*, supersedes the conflicting active row, and retries the same INSERT once.

The justification for auto-healing is the issue's own: under D9 an agent holds at most one activity, so a **new dispatch for that agent is itself the evidence** the old one is no longer live. Dispatch is sequential within a run, and #288 stops a second cycle from recruiting an agent another cycle holds. I deliberately did **not** add the age threshold the issue floats as an alternative — it would be a tunable with no principled default, and `feedback_externalize_config_verify_seam` says invent those only when something real needs tuning. When nothing stale can be cleared (`fetchval` returns None) the original violation surfaces rather than retrying into a second failure.

This is the one that matters most: it heals at the moment of the conflict, without waiting for a restart.

**2. Cancel teardown.** Cancellation bypasses `RunCompletion.finalize` — the identical bypass #529 exposed for leases — so both cancel routes now end the cycle's open activities alongside the lease sweep. This is the issue's second acceptance criterion.

**3. Startup.** `reap_stale_activities` gated on a terminal *cycle* status, and **a cycle whose runs were killed derives as ACTIVE forever** — so it spared exactly the rows a crash strands. Same correction as #373's: nothing dispatches in a process that has not begun serving, so every active row at startup is residue. The derived status is now logged instead of gating.

The predicate is also asked about `cycle_id`-less rows now. Skipping them *before* the predicate made them unreachable by any sweep at all — a row nothing can ever ask about is a row nothing can ever clear.

## Correcting the plan's premise

The 1.4.3 plan narrowed #561 to "the residue #672's reaper docstring explicitly leaves alone" — duty/ambient rows with no `cycle_id` — and posed the open design question as *which terminal predicate replaces "owning cycle is terminal"*.

That premise doesn't survive contact with the code. **`task_dispatcher.py:240` is the only `start_activity` caller in the repo, and `TaskEnvelope.cycle_id` is a plain `str`** — so no cycle_id-less row can currently be created. I wrote that narrowing off the reaper's docstring caveat rather than off who actually writes rows.

The rows that actually strand are cycle-owned ones under a cycle that never reaches a terminal status. So the plan's open design question resolves to **no new predicate** — not lease expiry, not a heartbeat-age bound. Both would have been machinery invented for a population that doesn't exist, while missing the one that does.

## Structural note

`deps.set_lease_sweep_ports` becomes `set_cancellation_ports` now that a third port joins it. The seam's concern is "runtime state a cancel has to tear down", not leases specifically; renaming a one-PR-old private seam is cheaper than letting the name go stale.

`RuntimeActivityPort.start_activity`'s contract changed (raise → supersede), so the port docstring changed with it.

## Verification

- **7 new/rewritten tests**; full regression **6115 passed, 1 skipped**.
- Adapter tests cover the supersede path end to end: the stale row is *aborted* (not deleted), scoped to that agent's active row, the retry replays the **same** statement rather than a divergent copy, an unclearable slot re-raises with no retry attempted, and a successful insert never touches the supersede path — that last one guards against a blanket pre-clear aborting a live activity on every dispatch.
- Route tests cover cancel teardown on both routes, plus degradation (a teardown failure must not turn a cancel into a 500) and the pool-less path.
- Reaper tests now assert the predicate is *asked* about cycle-less rows, and the startup test asserts an ACTIVE-deriving cycle's row is ended and logged.

The reason-code rename (`activity_stranded_cycle_terminal` → `activity_stranded_at_startup`) needs no data migration — reason codes are event-surfaced, not persisted as a column in v1.1.

Live proof rides the same **cancel probe** as #373/#529 in the deploy window, from the other side: assert zero stranded activities after a mid-implementation cancel with no restart.

Closes #561

### DIFF
--- a/adapters/cycles/task_dispatcher.py
+++ b/adapters/cycles/task_dispatcher.py
@@ -232,7 +232,9 @@ class TaskDispatcher:
         Best-effort: returns the activity id, or None when disabled/failed — it
         must never raise, so observability can't break dispatch. D9 (one active
         activity per agent) is enforced by the partial unique index; a leftover
-        active row from a prior crashed task makes this conflict, which we swallow.
+        active row from a prior crashed task is superseded by the adapter and the
+        insert retried (#561), so the swallow below no longer hides a permanently
+        dead tracking surface — it catches genuine failures only.
         """
         if self._activity_port is None:
             return None
--- a/adapters/persistence/runtime/activity_postgres.py
+++ b/adapters/persistence/runtime/activity_postgres.py
@@ -5,11 +5,14 @@ registry and the other runtime adapters. All rows hit `runtime_activities`
 (migration `1130_runtime_activities.sql`).
 
 `start_activity` inserts a `running` activity; the §4.2 partial unique index
-enforces the single-active-activity invariant (D9), so a second active start for
-the same agent raises `UniqueViolationError` — the §4.4 instrumentation treats
-activity calls as best-effort and swallows it so observability never breaks a
-real task. `update_state` manages the lifecycle timestamps; the terminal helpers
-are thin wrappers over it (their reason/evidence is event-surfaced, not stored).
+enforces the single-active-activity invariant (D9). A conflicting *stale* row
+(one whose task died without terminalizing it) is superseded and the insert
+retried once — see :meth:`PostgresRuntimeActivity.start_activity`. Without that,
+a single leftover row silently killed activity tracking for that agent forever,
+because the §4.4 instrumentation treats activity calls as best-effort and
+swallows the conflict so observability never breaks a real task (#561).
+`update_state` manages the lifecycle timestamps; the terminal helpers are thin
+wrappers over it (their reason/evidence is event-surfaced, not stored).
 
 JSONB columns (`completion_conditions`/`evidence_requirements`) follow the cycle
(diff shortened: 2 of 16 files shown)


jasonLadd (commented):
reviewed

PR #706, opened 2026-08-04
fix(checks): resolve a bare `python` at spawn instead of through PATH (#498)

jasonLadd:
Third fix of the **1.4.3** line, plus both of the line's riders.

## The bug, reproduced

`scaffold_contract` emits `vc-*-compiles` as `argv: ["python", "-m", "py_compile", …]`, and `_restricted_env()` inherits `PATH`. Ubuntu ships only `/usr/bin/python3` — there is no `python` — so the spawn raises `FileNotFoundError` and the check degrades to `skipped (missing_tooling)`.

On this box, before the change:

```
$ env PATH=/usr/bin:/bin .venv/bin/python -m pytest tests/unit/cycles/test_verification_contract_runner.py
FAILED test_every_structural_criterion_passes_on_bare_skeleton - AssertionError: vc-routes-compiles -> skipped (missing_tooling)
FAILED test_structural_criteria_still_pass_with_reference_fill  - AssertionError: assert 'skipped' == 'passed'
2 failed, 4 passed
```

After: **6 passed**. CI only ever passed because `setup-python` happens to leave a `python` shim on `PATH`.

The tests are the visible symptom; the shipped consequence is that whether `vc-*-compiles` executes or silently skips in a live cycle depends on the evaluating container having a `python` alias. Today's agent images do, so it is latent — but "structural criteria always execute" is the premise the bare-skeleton CI gate encodes (SIP-0098 §7), and a criterion that can quietly become `skipped` on an env detail is not that.

## Fixed in the evaluator, not the emission

Per the ruling recorded in `docs/plans/1-4-3-patch-plan.md` before any code. The issue offers both sites; the emission is the wrong one on a patch line:

- Emitting `python3` **moves the contract hash** — inadmissible here, and this window must not re-seed.
- It also would not help any **already-seeded** contract, v9 included, which keeps the bare name forever.

Resolving `python` → `sys.executable` at spawn fixes the pinned contracts too and is hash-stable by construction. `module_imports` (#628) already spawns `sys.executable`, so this conforms to the pattern rather than inventing one — and it makes `_CHECK_ENV_EXECUTABLES`' claim that the check environment provides `python` true *by construction* instead of by image accident.

**Ordering is load-bearing:** resolution runs strictly *after* the safelist gate. The safelist governs the name a contract may ask for; resolution is a spawn detail. Reversing it would mean `argv[0]` is already an absolute path when the patterns run, so either everything errors or the gate has to be loosened to accept arbitrary interpreter paths. There is a test pinning that order. `argv` also stays the authored form in evidence, so a genuine `missing_tooling` skip still reports `node`, not a path no reader can match back to the contract.

## Riders

**1. Pin the #605 property registry-wide.** Every Python-parsing check must *skip* a non-Python target rather than error — an `error` becomes `evaluator_error:<check>` and the patch lands UNVERIFIABLE, neither accepted nor rejected. pf-41 lost a roll to exactly that; pf-40 is the control (same bad repairs, but verification returned a verdict and rejected them).

#605 itself was already fixed — the property held by construction and was asserted **nowhere**. The test derives its parametrization from each spec's `applicable_extensions`, so it currently covers 7 checks × 2 non-Python targets, and a Python check added later inherits it instead of quietly opting out.

**2. Declare `pyflakes` in `tests/requirements.txt`.** 1.4.2 pinned it in `requirements/base.txt` and the three locks, so every container has it — but the harness got it only **transitively, via `flake8>=6.0.0`**. Ruff's own config comment records that its `F` rules *are* pyflakes, so retiring flake8 is a live prospect; the day that happens, every `undefined_names` check returns `error: missing_analyzer` with nothing pointing at why.

## Not fixed here — worth a look

While tracing the safelist I found `_CHECK_ENV_EXECUTABLES` (plan validation, #645) and `COMMAND_SAFELIST` (authoring lint + runtime gate) are two overlapping allowlists that **disagree in both directions**:

- `python3`, `pytest`, `npm`, `npx` are advertised by the validator's error message but have no safelist pattern, so a plan taking that advice is rejected by the very next gate.
- `ruff`, `tsc`, `eslint` are safelisted but absent from `_CHECK_ENV_EXECUTABLES`, so a plan authoring `ruff check x.py` is rejected as "the check environment does not provide ruff".

Neither is reachable as a live failure today (the two gates agree on the intersection actually in use, `python` and `node`), and reconciling them changes what plans may author — out of scope for a patch line. Flagging rather than filing.

## Verification

- The issue's exact repro now passes: **6 passed** with `.venv/bin` off `PATH`.
- Whole `tests/unit/cycles/` + `tests/unit/capabilities/` in that same env: **3757 passed, 2 skipped**.
- Full regression: **6138 passed, 1 skipped**.
- New tests: bare `python` runs with nothing on `PATH`; a real compile failure still **fails** (guards against resolving the check into a no-op); non-interpreter argv untouched and reported under its authored name; safelist-before-resolution ordering; and the resolver's rewrite/no-rewrite matrix — including `python3`, which is not authorable today and is covered so a later safelist entry can't silently reintroduce the PATH dependency.

Also renamed `test_compile_ok_passed`, which asserted a safelist rejection and had nothing to do with a passing compile.

Closes #498

### DIFF
--- a/src/squadops/cycles/acceptance_checks.py
+++ b/src/squadops/cycles/acceptance_checks.py
@@ -159,6 +159,38 @@ def _restricted_env() -> dict[str, str]:
     return {k: v for k, v in os.environ.items() if k in keep}
 
 
+# Bare interpreter names a contract may emit for `command_exit_zero`.
+_BARE_INTERPRETERS = frozenset({"python", "python3"})
+
+
+def _resolve_interpreter(argv: list[str]) -> list[str]:
+    """Substitute the running interpreter for a bare python name in ``argv[0]``.
+
+    #498: ``command_exit_zero`` argv is authored (or scaffold-emitted) as a
+    *name*, and resolving that name through the inherited ``PATH`` at spawn made
+    the outcome depend on an environment detail. Ubuntu ships only
+    ``/usr/bin/python3`` — there is no ``python`` — so ``python -m py_compile``
+    raised ``FileNotFoundError`` and the check degraded to
+    ``skipped(missing_tooling)``. A structural criterion that silently stops
+    executing on a host detail is exactly what the bare-skeleton gate exists to
+    rule out (SIP-0098 §7).
+
+    The evaluator *is* a Python process, so ``sys.executable`` is by definition a
+    working interpreter — which makes ``_CHECK_ENV_EXECUTABLES``' claim that the
+    check environment provides ``python``/``python3`` true by construction
+    instead of by image accident. It also fixes contracts already seeded with the
+    bare name (v9 among them), which emitting ``python3`` could not do — and that
+    would have moved the contract hash besides.
+
+    Runs strictly *after* the safelist gate: the safelist governs what a contract
+    may ask for (a name), and resolution is a spawn detail. Reversing the order
+    would force every safelist pattern to know about absolute interpreter paths.
+    """
+    if not argv or argv[0] not in _BARE_INTERPRETERS or not sys.executable:
+        return argv
+    return [sys.executable, *argv[1:]]
+
+
 # ---------------------------------------------------------------------------
 # Base + registry
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #708, opened 2026-08-04
fix(comms): report the delay and priority capabilities the queue actually has (#572)

jasonLadd:
Fourth fix of the **1.4.3** line.

## The bug

`publish()` set a per-message `expiration` (TTL) with **no dead-letter exchange declared anywhere** — no `x-dead-letter-exchange` or `x-dead-letter-routing-key` on any queue. In RabbitMQ semantics an expired message is *discarded*, not delivered later. So a "delayed" message was silently dropped, while `capabilities()["delay"]` reported `True` with the comment *"Supported via TTL+DLX"*.

**Nothing is being dropped today.** `delay_seconds` has exactly one live caller — `adapters/capabilities/aci_executor.py:162` — and it passes `0`. The defect is the *declaration*: a capability flag is a contract with future callers, and this one promised a feature that loses messages.

## The fix

Took the issue's **option 3** — delete the TTL path, report the truth — per the 1.4.3 plan's stated preference for the honest declaration. There is no delayed-delivery consumer to serve, and wiring a real wait-queue + DLX (option 1) or adopting `rabbitmq-delayed-message-exchange` (option 2) is building a feature nothing asks for, on a patch line.

A non-zero `delay_seconds` now raises `NotImplementedError`. Both alternatives are wrong: delivering immediately is not what the caller asked for, and setting a TTL loses the message. Refusing is the only honest option, and it is loud.

**One detail worth review:** `retry()` wraps failures in `QueueError`, and its `except Exception` would have caught the refusal and laundered it into one. Callers treat `QueueError` as transient and retry — so an impossible delay would be attempted forever instead of surfacing once. There is now an explicit `except NotImplementedError: raise` ahead of the general handler, and a test pinning it. The original message is left unacked, so the broker still owns it.

## `priority` is the same lie, one line down

Found while verifying the change rather than from the issue. `"priority": True` claims support that needs `x-max-priority` on the queue declaration plus a `priority` property per message — `_get_queue` declares neither and `publish` sets neither, so the broker ignores priority entirely. A caller ordering work by it would get plain delivery and never know.

Fixing `delay` and leaving `priority` would be precisely the half-state the issue says not to keep, so both now report `False`.

## Both NoOps

There are **two** `NoOpQueuePort` classes — `adapters/noop/ports.py:188` says in its own docstring that it "replicates existing pattern from ports/comms/noop.py". Both advertised `delay`/`priority` as `True`.

Aligned both, because a null object advertising an operation every real provider refuses sends a test down the branch production never takes — a future caller writing `if queue.capabilities()["delay"]: … else: …` would have had its tests exercise the wrong half. They also now refuse a non-zero delay, so the port contract holds for every implementation rather than just the real one.

`fifo` is deliberately left as-is (NoOp `True`, RabbitMQ `False`). It gates no operation, so nothing can be built on it — unlike delay, which is a parameter you can pass.

The duplicated null object is a separate smell; not touched here.

## Verification

- **Against the live broker**: 11 integration tests pass, including a new one asserting the refusal reaches the caller intact rather than as a `QueueError`, and the existing retry test upgraded from "we verify retry succeeded" (which asserted nothing) to actually consuming the republished message back.
- Unit: capabilities parity for both flags; a non-zero delay never reaches the transport (`default_exchange.publish` not awaited); `None`/`0` unaffected, since that is the one live caller's path; retry's refusal not laundered and the original left unacked; retry at `delay_seconds=0` still republishes and acks.
- Full regression: **6147 passed, 1 skipped**.

Closes #572

### DIFF
--- a/adapters/comms/rabbitmq.py
+++ b/adapters/comms/rabbitmq.py
@@ -155,24 +155,32 @@ class RabbitMQAdapter(QueuePort):
         Args:
             queue_name: Name of the queue
             payload: Message payload (must be ACI TaskEnvelope JSON)
-            delay_seconds: Optional delay before message becomes available (None = immediate)
+            delay_seconds: Optional delay before message becomes available.
+                Only ``None``/``0`` (immediate) is supported — see Raises.
 
         Raises:
             QueueError: If publishing fails
+            NotImplementedError: If a non-zero ``delay_seconds`` is requested.
+                This adapter cannot delay delivery (#572), and the previous
+                behavior — a per-message TTL with no dead-letter exchange —
+                made the broker *discard* the message when the TTL expired.
+                Failing loudly beats losing the message: a caller that wanted
+                a delay has to know it did not get one.
         """
+        if delay_seconds is not None and delay_seconds > 0:
+            raise NotImplementedError(
+                f"Delayed delivery is not supported by this queue provider "
+                f"(delay_seconds={delay_seconds}); capabilities()['delay'] is False. "
+                f"Delivering it immediately would be wrong and setting a TTL would "
+                f"drop it — declare a dead-letter exchange or adopt "
+                f"rabbitmq-delayed-message-exchange before relying on this."
+            )
+
         # Build the message once; only the transport send is retried. A bad
         # payload fails here, before the loop, so it is never retried.
         message_properties: dict[str, Any] = {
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #709, opened 2026-08-04
fix(telemetry): a literal `|` in the email TLD class swallowed the next log field (#573)

jasonLadd:
Fifth and final fix of the **1.4.3** line.

## Not cosmetic

`[A-Z|a-z]` is "uppercase, a literal pipe, or lowercase" — inside a character class `|` is a literal, never alternation. Both letter ranges were still covered, so every genuine address kept redacting, which is why the issue reads as cosmetic-adjacent.

Measuring it says otherwise. Accepting `|` as a TLD character let the match run **past the address into the next pipe-delimited field**:

```
input : email=alice@example.com|status=ok|retry=3
before: email=[REDACTED-PII]=ok|retry=3          <- ate `|status`, left a corrupted `=ok`
after : email=[REDACTED-PII]|status=ok|retry=3
```

Pipe-delimited fields are everywhere in log output, so in strict mode this quietly destroyed the surrounding context of the very record it was making safe. That is the shape that actually bit — not "the class also matches `|` in a TLD position".

**No leak in either direction.** The only other behavioral delta is that `user@example.c|m` — not a valid address — stops matching, which is over-redaction going away. Verified across a matrix of real and malformed addresses; every genuine form still scrubs.

## The sibling sweep

The issue asked for "a quick eyeball of the sibling patterns for the same class of typo". Rather than eyeball it, I swept the **whole repo** for the shape (`[...]` containing an unescaped `|`, in a line that looks like a regex).

One other hit: `src/squadops/capabilities/handlers/wrapup_tasks.py:64`, `[:=|]`. That literal pipe is **deliberate** — the pattern parses `suggested_owner` out of "markdown tables or YAML-like content", where `|` is the table delimiter. Correct usage; left alone.

## Guard

The sweep result is pinned by a parametrized test over every pattern in the redaction module — API-key, password, and PII lists — asserting none contains a literal pipe inside a character class. Scoped to this module deliberately: a literal pipe is legitimate elsewhere (see `wrapup_tasks`), but in a scrubber it means the class silently accepts a character it was never meant to and the match overruns its intended end.

Confirmed the guard has teeth: it flags the old pattern (`[A-Z|a-z]`) and passes the new one.

## Verification

- New tests: the pipe-delimited log line redacts to exactly the expected string (exact-equality, not a substring check); `user@example.c|m` is left alone; four real address shapes still scrub via parametrize; the pattern-hygiene guard.
- Full regression: **6162 passed, 1 skipped**.

## Line status

This closes the 1.4.3 code. All five fixes merged or open: #373+#529 (PR #704), #561 (#705), #498 + both riders (#706), #572 (#708), #573 (this one). Six issues closed, two riders landed, plus #707 filed to 1.5 from the #498 sweep.

Next is the deploy window — rebuild + verify-LOADED on both evaluator surfaces, assert contract v9 / manifest v4 unchanged, the **cancel probe** (#373/#529/#561's only real proof: launch, cancel mid-implementation, assert zero unreleased leases and zero stranded activities with no restart, then relaunch and confirm immediate acquisition), one unfiltered confirmation shakedown, then the 1.4.3 bump and tag.

Closes #573

### DIFF
--- a/adapters/telemetry/langfuse/redaction.py
+++ b/adapters/telemetry/langfuse/redaction.py
@@ -45,8 +45,13 @@ _PASSWORD_PATTERNS: list[re.Pattern[str]] = [
 
 # PII patterns (strict mode only)
 _PII_PATTERNS: list[re.Pattern[str]] = [
-    # Email addresses
-    re.compile(r"\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b"),
+    # Email addresses. The TLD class is `[A-Za-z]`, NOT `[A-Z|a-z]` (#573): a `|`
+    # inside a character class is a literal pipe, not alternation, so the old
+    # form accepted `|` as a TLD character and ran the match past the address
+    # into the next pipe-delimited field —
+    # `email=a@b.com|status=ok` redacted to `email=[REDACTED-PII]=ok`, destroying
+    # the surrounding context instead of just the address.
+    re.compile(r"\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}\b"),
     # Phone numbers (US/international)
     re.compile(r"\b(?:\+?1[-.\s]?)?\(?\d{3}\)?[-.\s]?\d{3}[-.\s]?\d{4}\b"),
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #711, opened 2026-08-04
fix(runtime): reap agents stranded in cycle mode with no lease at startup (#710)

jasonLadd:
Sixth fix of the **1.4.3** line, added after the pre-deploy state capture turned up a blocker for the window's own verification.

## What the box was actually doing

```
 agent_id   |  mode   |          focus_leases where released_at is null: 0
------------+---------+          most recent lease row of any kind: released 2026-07-21
 bob        | cycle   |          cycles created since then: 64
 data       | cycle   |          focus_lease / mode-transition log lines in those cycles: 0
 eve        | cycle   |
 max        | cycle   |
 nat        | cycle   |
 neo        | cycle   |
 comms-agent| ambient |
```

Sixty-four cycles over two weeks — **including shk-3, the green 1.4.2 confirmation roll** — and not one acquired a focus lease. SIP-0089 focus arbitration has been inert, and with it #288's protection against two cycles sharing an agent.

## Why it never surfaced

Every step is individually correct, and the loop is silent:

1. An agent is left in `mode = cycle` with no held lease.
2. Recruitment calls `request_transition(agent, "cycle", …)`. `current == target_mode`, so it takes the #288 same-mode branch (`coordinator.py:235`), reads `get_current_lease` → `None`, finds nothing to conflict with, returns `idempotent_skip`.
3. `admit_participants` treats that as admitted but — correctly — does **not** add the agent to `recruited_agent_ids`.
4. Nothing acquired, so finalize releases nothing and the mode stays `cycle`. Back to 1, forever.

The same-mode branch emits no event; admission logs nothing on a skip. No test could have caught it either — every unit under it behaves correctly in isolation. Same discovery shape as #689's qa-boundary gap on shk-3: found by looking at what actually happened, not by the suite.

## Why it blocks the deploy window

The 1.4.3 **cancel probe** asserts zero unreleased leases after a mid-implementation cancel. On this box that passes trivially, because zero leases are ever acquired. #373/#529/#561 would have shipped reported-verified and actually unverified — precisely the false-green class this release line exists to close.

And the #373 reaper cannot clear it. `reap_stale_leases` iterates **held leases**; with none held it has nothing to iterate and the modes survive. #373's own documented recovery SQL has two statements for exactly this reason — one for `focus_leases`, one for `agent_runtime_state` — and what shipped only performs the second as a side effect of the first.

## The fix

A third startup sweep, on the same justification as the other two: nothing dispatches in a process that has not begun serving, so no agent can legitimately hold cycle focus yet. It runs *after* the lease sweep, so what it sees is the residue with no lease at all.

- `RuntimeStatePort.list_states(mode=…)` — the port exposed only `get_state(agent_id)`, the same gap `list_active_leases` filled for #373. An agent stranded by a dead process is one nobody thought to ask about, so a per-agent read cannot find it. Narrowed in SQL.
- `duty` is never swept — a duty agent may be legitimately mid-window with the scheduler about to re-evaluate, and duty leases key to assignments rather than runs.
- Counts *applied* transitions, not attempts: the reap total is the operator's only signal that the box healed, so a rejected transition must not inflate it.

## Rename

`runtime/lease_reaper.py` → `runtime/focus_reaper.py`. An agent's focus is two coupled facts — the `FocusLease` and the `RuntimeMode` that rides with it, written atomically since #244/D25 — and the module now sweeps both. Naming it for one half is a fair part of why the other half went unnoticed for two weeks. Mechanical: 4 import sites plus the test module.

## Likely origin

Probably historical residue rather than something current code reproduces: since #244/D25 the coordinator writes mode and lease inside one `RuntimeTransactionPort.begin()` in both directions, so `cycle` with no lease should no longer be reachable through the normal path. That is a hypothesis, not a proof — and a state that cannot be *created* is still worth being able to *clear*, especially one this quiet.

## Verification

- 8 new tests. The headline one asserts the lease sweep returns **0** on this exact state before showing the mode sweep clears it — the structural blindness, pinned.
- Also covered: `duty`/`ambient` untouched; per-agent isolation on a failed transition; a rejected transition not counted; `list_states` called with `mode="cycle"` so `duty` is never even enumerated; all three wiring-gate permutations; a port failure never blocking startup.
- Four existing `_FakeStatePort` doubles gained `list_states`, and `_FakeCoordinator` now returns a real `TransitionOutcome` instead of `None` — which makes the existing lease tests more faithful too.
- Full regression: **6172 passed, 1 skipped**.

Live proof arrives at the deploy: the first runtime-api boot should log the mode reap clearing 6 agents, after which a launched cycle acquires real leases and the cancel probe becomes meaningful.

Closes #710

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -62,7 +62,7 @@ from squadops.events.types import EventType
 from squadops.ports.cycles.flow_execution import FlowExecutionPort
 from squadops.runtime import reasons
 from squadops.runtime.admission import admit_participants, release_participants
-from squadops.runtime.lease_reaper import release_owner_leases
+from squadops.runtime.focus_reaper import release_owner_leases
 from squadops.runtime.recruitment import reserve_buffer_decision
 from squadops.tasks.models import TaskEnvelope, TaskResult, TaskResultStatus
 from squadops.telemetry.context import use_correlation_context
--- a/adapters/persistence/runtime/state_postgres.py
+++ b/adapters/persistence/runtime/state_postgres.py
@@ -43,6 +43,24 @@ class PostgresRuntimeState(RuntimeStatePort):
             )
         return _row_to_state(row) if row else None
 
+    async def list_states(
+        self, *, mode: str | None = None, conn: Any = None
+    ) -> tuple[AgentRuntimeState, ...]:
+        query = (
+            "SELECT agent_id, mode, runtime_status, focus, "
+            "current_runtime_activity_id, interruptibility, "
+            "last_heartbeat_at, current_assignment_ref "
+            "FROM agent_runtime_state"
+        )
+        args: list[Any] = []
+        if mode is not None:
+            query += " WHERE mode = $1"
+            args.append(mode)
+        query += " ORDER BY agent_id"
+        async with acquire(self._pool, conn) as conn:
+            rows = await conn.fetch(query, *args)
+        return tuple(_row_to_state(row) for row in rows)
+
     async def upsert_state(
         self, state: AgentRuntimeState, *, conn: Any = None
     ) -> AgentRuntimeState:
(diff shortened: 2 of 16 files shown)


jasonLadd (commented):
reviewed

PR #713, opened 2026-08-04
fix(runtime): owner-check the lease release on ambient entry (#712)

jasonLadd:
Fix 7 of the **1.4.3** line — the F1 finding from the pre-deploy code review, fixed before the rebuild so the cancel probe proves what it claims to.

## The race

The #529 fix makes cancel→relaunch possible, and thereby exposes a hole underneath it: **the coordinator's ambient-entry path releases whatever leaving-mode lease the agent holds without checking who owns it** (`coordinator.py` `_arbitrate_lease`, ambient branch — `holds_leaving_lease` compares `owner_type` only, never `owner_ref`).

Cancel detection is a dispatch-boundary registry poll (`dispatched_flow_executor.py:3052`), so a cancelled run's executor keeps unwinding until its in-flight task replies — which can be minutes. Sequence:

1. Cancel route releases the run's leases, agents → ambient *(intended)*.
2. Operator relaunches; the new cycle recruits the same agents — new leases, `owner_ref = new_run` *(intended)*.
3. The old executor's in-flight task replies → `_CancellationError` → `finally:` → `release_participants(recruited_agent_ids, owner_ref=old_run)`.
4. The coordinator flips the agent to ambient and **releases the new run's lease**.

The relaunched run then executes with no focus protection — a third cycle can double-book its agents (the #288 hole reopened), the duty scheduler could pull one via a legal `ambient → duty`, and #561's self-heal would let an interloper supersede the live activity. All silent. Pre-1.4.3 the race was unreachable only because cancel never released leases: the deadlock #529 complains about was an accidental guard. The 1.4.3 cancel probe — cancel, relaunch immediately — constructs exactly this window, which is why this ships before the rebuild rather than after.

## The fix

The symmetric half of #288: **acquire is owner-checked, release now is too.** Entering ambient while the leaving-mode lease belongs to a different `owner_ref` is refused with the new typed reason `focus_lease_held_by_other_owner` — no mode write, no release — and surfaces through the existing `FOCUS_LEASE_REJECTED` event, so a stale finalize is *visible* in the event stream instead of silently stealing.

Checked against every ambient-entry call site:

| Caller | owner_ref vs held lease | Result |
|---|---|---|
| Normal run finalize (`release_participants`) | match | proceeds, unchanged |
| Stale cancelled-run finalize | **mismatch** | **refused — the fix** |
| Cancel-route sweep / startup lease reap (`_return_to_ambient`) | passes `lease.owner_ref`, match by construction | proceeds |
| #710 mode sweep | agent holds **no** lease → guard bypassed | proceeds |
| Scheduler duty-close | assignment id the lease was acquired under, match | proceeds |
| Partial-admission rollback | just-recruited, match | proceeds |

## Riders from the same review

- **F2**: the #710 mode sweep passed `reasons.MODE_STRANDED_AT_STARTUP` — a reason-code string — as `owner_ref`. Now a proper `STARTUP_REAP_OWNER_REF` sentinel; owner_ref semantics are load-bearing with this guard in place.
- **F3**: the startup lease reap's log claimed "its executor died mid-run" even for a run that was still `queued`; reworded.

## Verification

- 3 new coordinator tests: the direct refusal (mode unwritten, lease intact, `FOCUS_LEASE_REJECTED` carries the new reason); the **full four-step stale-finalize narrative** ending with the relaunched run still holding both mode and lease; and the #710 no-lease path still transitioning — the guard must not re-strand what the mode sweep exists to clear.
- Existing owner-matched tests (`duty-1`/`duty-1` etc.) pass untouched — the guard is invisible to every legitimate caller.
- Full regression: **6175 passed, 1 skipped**.
- **Hash-stable**: no contract or manifest surface touched; no migrations.

With this merged, the review's findings are all resolved or placed: F1 fixed here, F2/F3 ride along, F4 was a no-op (additive response keys), F5 (plan-doc amendments) rides the bump PR, #707 filed to 1.5. Next: rebuild + deploy, verify-LOADED, seed check, **cancel probe**, shakedown, bump.

Closes #712

### DIFF
--- a/src/squadops/api/runtime/startup_reaps.py
+++ b/src/squadops/api/runtime/startup_reaps.py
@@ -85,8 +85,8 @@ async def reap_stranded_leases(
             if RunStatus(owning_run.status) not in TERMINAL_STATES:
                 logger.warning(
                     "Releasing focus lease held by run %s still marked %r — no run "
-                    "can be executing at startup, so its executor died mid-run and "
-                    "the run status is stale.",
+                    "can be executing at startup, so its owning process exited "
+                    "without finalizing it and the run status is stale.",
                     run_id,
                     owning_run.status,
                 )
--- a/src/squadops/runtime/coordinator.py
+++ b/src/squadops/runtime/coordinator.py
@@ -430,6 +430,15 @@ class RuntimeCoordinator:
         Never writes `mode` (lease ≠ mode). All reads/writes run on the caller's
         unit-of-work `conn` (§4.5/D25) so they roll back with the transition on a
         later failure. Resolution by case:
+        - entering ambient with the leaving-mode lease held by a DIFFERENT owner
+          → not ok (#712). Release is owner-checked, symmetric with #288's
+          owner-checked acquire: a cancelled run's finalize fires only at its
+          next dispatch boundary, which can be minutes after the cancel route
+          already released its leases — and a relaunched run may have
+          re-recruited the agent in between. Without this check the stale
+          finalize would flip the agent to ambient and release the NEW run's
+          lease, leaving that run to execute with no focus protection (the #288
+          double-booking hole, reopened silently);
         - entering ambient → no acquire; defer releasing the leaving-mode lease
           to after the mode write;
         - entering cycle/duty with no conflict → acquire (`granted`);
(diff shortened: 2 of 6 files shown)


jasonLadd (commented):
reviewed

PR #717, opened 2026-08-05
fix(runs): persist the terminal failure reason on the run row (#427)

jasonLadd:
1.4.4 fix 1 (plan: `docs/plans/1-4-4-patch-plan.md`). The persistence half of #427 — the logging half shipped earlier (`configure_logging` in the API entrypoint, verified live through the whole 1.4.3 deploy window).

## What

A failed run's terminal exception now lands on the run row, atomically with the FAILED transition, and surfaces everywhere a triager looks:

- **Migration 1010** (the line's only one, per the merged plan's design decision): nullable `failure_reason TEXT` on `cycle_runs`. Additive + idempotent; the exact ALTER and the new terminal UPDATE were validated against live Postgres in a rolled-back transaction (zero residue).
- **`TerminalOutcome.failure_reason`**: the FAILED mappings in `resolve_terminal_outcome` carry it; CANCELLED/PAUSED deliberately do not (they have their own affordances). `_ExecutionError` persists its message alone — the wrapper class name is executor plumbing; unexpected exceptions persist `TypeName: message` (there the type IS the diagnosis). Bounded at 2000 chars.
- **Port**: `update_run_status(..., *, failure_reason=None)` — reason commits with the terminal status in one UPDATE; `COALESCE` on the value side so a reason-less transition never nulls a recorded reason. Postgres + memory registries in parity.
- **Both executors** thread it through `_safe_transition` (dispatched via `TerminalOutcome`, in-process at its two FAILED sites).
- **Surfaces**: `run_report.md` gains a `Failure Reason` metadata line (the report builder reads the run row finalize just wrote); `RunResponse` carries the field, so `runs show` renders it with zero CLI changes (`print_detail` is schema-driven).

## Verification

- 12 new tests: FAILED-mapping reason content (wrapper vs unexpected), no reason on cancel/pause, 2000-char bound, messageless fallback, registry round-trip + non-terminal-ignores-reason + completed-carries-none, report line present/absent, DTO carried/None, and the executor wiring assertion (the FAILED transition call carries the diagnosis).
- Full regression: **6187 passed, 1 skipped** (was 6175).
- Live-proof lands in the deploy window's designed-failure probe (per the plan).

Closes #427

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -470,7 +470,9 @@ class DispatchedFlowExecutor(FlowExecutionPort):
             # former per-class handlers).
             outcome = resolve_terminal_outcome(exc, run_id)
             terminal_status = outcome.terminal_status
-            await self._safe_transition(run_id, outcome.run_status)
+            await self._safe_transition(
+                run_id, outcome.run_status, failure_reason=outcome.failure_reason
+            )
             self._cycle_event_bus.emit(
                 outcome.event_type,
                 entity_type="run",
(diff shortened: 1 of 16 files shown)


jasonLadd (commented):
reviewed

PR #718, opened 2026-08-05
fix(planning): builder offer and gate net both key off configured build_profile (#426)

jasonLadd:
1.4.4 fix 2 (plan: `docs/plans/1-4-4-patch-plan.md`). Closes the mismatched-sources-of-truth defect where a builder-carrying squad on a profile without `build_profile` authored plans that passed the gate and died at implementation start (`cyc_23c6d9a7363a`: 8ms failure after a full framing).

## What

Three layers, one source of truth:

- **`Cycle.resolved_config()`** — the single definition of the `{**applied_defaults, **execution_overrides}` merge. `generate_task_plan`, the build-completeness check, and the new gate net all read it; two of those previously open-coded the merge, which is exactly how offer-vs-guard drift happens.
- **The offer keys off config**: `produce_plan` computes `offer_builder = has_builder AND build_profile`, and the *entire* builder prompt bundle keys off it — the task-type vocabulary and the guideline/example prose together (dropping only the vocabulary would leave prose teaching a forbidden task type). `planner_build_task_types`'s kwarg renamed `has_builder` → `offer_builder` to say what it now means.
- **The gate net**: `ImplementationPlan.validate_build_config(resolved_config)` rides the existing #295/#464 fail-fast seam in `_reject_unsatisfiable_plan_at_gate` — a builder task without a configured `build_profile` is rejected at the gate (minutes) instead of at dispatch (a full admitted run).

The #686 classification test forced the author-teaching decision on the spot: `validate_build_config` is recorded **NOT_AUTHOR_FACING** with the reasoning — the offer removes the option upstream, so the inline "use ONLY these task types" vocabulary is the teaching; restating the rule to authors who can't be offered builder tasks would be noise.

**Recorded no-change decision** (per the merged plan): the `validation` request profile does not gain a `build_profile` — it measures plan/validation machinery, not assembly.

## Verification

- New tests: the offer pair (builder squad without `build_profile` → no `builder.assemble` in vocabulary and empty guideline; with it → both present), `validate_build_config` (builder task + no profile → named rejection; with profile → clean; builderless plan → clean), `Cycle.resolved_config` override-wins, and the executor gate-net pair — including the polarity guard: a `build_profile` arriving via `execution_overrides` must pass, or the net kills valid cycles.
- Two pre-existing tests updated where the old role-only premise was the very behavior removed.
- Full regression: **6195 passed, 1 skipped** (was 6187).
- Live-proof lands in the deploy window's gate-rejection replay (per the plan).

Closes #426

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -1516,7 +1516,7 @@ class DispatchedFlowExecutor(FlowExecutionPort):
         # roll-up reads is emitted at the builder task itself (#399), so an
         # in-loop builder failure is disclosed even though this gate is only
         # reached when every task succeeded.
-        resolved_config = {**cycle.applied_defaults, **cycle.execution_overrides}
+        resolved_config = cycle.resolved_config()
         deficiency = compute_missing_required_files(plan, stored_artifacts, resolved_config)
         if deficiency is not None:
             profile_name, missing = deficiency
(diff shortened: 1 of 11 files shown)


jasonLadd (commented):
reviewed

PR #719, opened 2026-08-05
fix(plan-validation): reject qa.test tasks whose artifacts cannot satisfy required tests_pass (#715)

jasonLadd:
1.4.4 fix 3 (plan: `docs/plans/1-4-4-patch-plan.md`). The unwinnable-plan class from the 1.4.3 deploy window, caught where it's statically visible: authoring time.

## What

- **`ImplementationPlan.validate_check_applicability(resolved_config)`**: when `tests_pass` is in the config's `required_checks`, every `qa.test` task that declares expected artifacts must include at least one pytest-discoverable file (`test_*.py` / `*_test.py`). Verification-only tasks (`expected_artifacts: []`) are exempt; the rule is inert when `tests_pass` isn't required (author-mode cycles keep full freedom).
- **Wired at both gate seams**: the workload-sequence seam (errors returned → system-REJECTED gate decision → free framing re-roll with #669 revision context) and the mid-run gate seam (parity).
- **The author is taught the rule, not just rejected** (#686 discipline): classified AUTHOR_FACING as `qa-tests-pytest-discoverable`, with the rule prose added to `request.plan_authoring_rules_appendix` (bumped to v2) — the shape, not the work, is the failure, and the appendix now says so with the fix (a pytest wrapper, or a verification-only task).

## Also: closes a #718 seam gap found while wiring

The executor has two gate seams, and multi-workload cycles traverse **only** `_reject_invalid_plan_before_workload_gate` (its own docstring: the mid-run seam "never fires here"). #718's `validate_build_config` net landed only on the mid-run seam — so the exact #426 repro shape (framing run → gate → implementation run) still fell through to the #291 dispatch guard. No regression (the dispatch guard always backstopped it), but the gate net wasn't doing its job on the path that mattered. Both the #426 and #715 nets now run on both seams, with a regression test pinning the workload-seam wiring for each.

## Verification

- Validator units: the shk-4 `.js`-only shape rejected with a teaching message; not-required → inert (polarity guard); `test_*.py` / `*_test.py` / mixed `.js`+`.py` all pass (the mixed shape is shk-4's off-plan escape, now legal to declare up front); verification-only exempt; dev tasks not subject; `conftest.py`/`tests.py` lookalikes correctly non-discoverable.
- Seam tests: workload seam returns the #715 error and the #426 error (gap closure), passes the polarity case; mid-run seam raises with the same named finding.
- Full regression: **6207 passed, 1 skipped** (was 6195).
- Live-proof lands in the deploy window's gate-rejection replay of the stored shk-4 plan (per the plan doc).

Closes #715

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -2823,6 +2823,15 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                     # emission) — provable plan-wide right here, and a system
                     # rejection re-rolls framing for free (#522).
                     errors.extend(parsed_plan.validate_unique_expected_artifacts())
+                    # #715: a qa.test task whose declared artifacts can never
+                    # satisfy required tests_pass fails on any content — shk-4
+                    # burned three correction rounds on one. #426: builder
+                    # tasks without a configured build_profile die at the #291
+                    # dispatch guard — this seam (the one multi-workload
+                    # cycles actually traverse) rejects both for a free
+                    # framing re-roll.
+                    errors.extend(parsed_plan.validate_check_applicability(cycle.resolved_config()))
+                    errors.extend(parsed_plan.validate_build_config(cycle.resolved_config()))
             elif ref.filename == "interface_manifest.yaml" or artifact_type == "interface_manifest":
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #720, opened 2026-08-05
fix(verification): split skipped typed checks into benign vs evidence-gap — gaps never read passed (#423)

jasonLadd:
1.4.4 fix 4 (plan: `docs/plans/1-4-4-patch-plan.md`). The false-green half of the verification-integrity family: every skipped typed check read `passed: true` regardless of cause (`handlers/cycle/base.py:455`).

## What

Skip causes now split into two accounting classes:

- **Benign** — config-off (`typed_acceptance_disabled`, `command_acceptance_checks_disabled`: the only two reasons an actual config flag produces) and framework-injected checks. Behavior unchanged.
- **Evidence gap** — an *authored* error-severity check the evaluator could not run. The row reads `passed: false` + `evidence_gap: true`; the SIP-0096 roll-up discloses it as unverified with an `evaluator_gap:<reason>` prefix; and when **contract-bound** (`criterion_id` set) it is required-level unverified → verdict `blocked_unverified` — a contract criterion can never again be silently unenforceable. Unbound gaps stay advisory (author-mode plans keep their freedom). This implements the RC-9 decision ratified at the plan review.

Two deliberate design points:

1. **A gap never fails the task or missions a correction** — no code repair closes an evaluator limitation; blocking happens at the roll-up. (The develop handler's all-passed rule excludes gap rows; builder blocks only via `missing`, untouched.)
2. **`frontend_acceptance_checks_disabled` is classified as a gap, not config-off** — despite the name there is no such config flag; it marks the unimplemented JS/TS analyzer, and an authored AST check on `.tsx` was exactly #423's measured exhibit (7 of 14 evaluations skipped-yet-passed). The classification comment documents the naming trap.

**Premise correction found while building** (the aggregation is healthier than the issue's era): since SIP-0096 landed, skipped typed rows already reach `cycle_outcome.unverified` non-creditably — the surviving false greens were the per-task `passed` flag (feeding the task-acceptance rule and the #114 gate-evaluator artifacts) and the missing required-escalation for contract-bound gaps. Both closed here. The #422 polarity note is reconciled as a side effect: unsupported command → executed error; unsupported extension → unverified; neither silently passes.

## Guardrails honored

- The **#605 registry-wide skip property is untouched** — evaluator outcomes don't change, only their accounting (the suite's own test proves it).
- §6.6.3 "never dropped": benign skips remain *disclosed* in unverified (required=False) — the plan's "out of the roll-up" is implemented as never-required, not undisclosed.
- The architecture guard caught my `== "skipped"` literal; it now compares `ResultStatus.SKIPPED`.

## Verification

New tests: the `.tsx` exhibit (gap row honest, task still green, no correction mission), warning-severity and config-off polarity guards, normalizer threading, and the three aggregation cases (contract-bound gap → `blocked_unverified` with prefixed reason; unbound gap → advisory; benign skip → byte-identical behavior). Full regression: **6214 passed, 1 skipped** (was 6207).

Closes #423

### DIFF
--- a/src/squadops/capabilities/handlers/cycle/base.py
+++ b/src/squadops/capabilities/handlers/cycle/base.py
@@ -15,7 +15,11 @@ from squadops.capabilities.handlers.base import (
     HandlerEvidence,
     HandlerResult,
 )
-from squadops.cycles.acceptance_check_spec import CHECK_UNDEFINED_NAMES, is_check_applicable
+from squadops.cycles.acceptance_check_spec import (
+    CHECK_UNDEFINED_NAMES,
+    CONFIG_OFF_SKIP_REASONS,
+    is_check_applicable,
+)
 from squadops.cycles.acceptance_checks import CheckOutcome
 from squadops.cycles.acceptance_evaluation import (
     evaluate_criterion,
@@ -24,6 +28,7 @@ from squadops.cycles.acceptance_evaluation import (
 )
 from squadops.cycles.implementation_plan import TypedCheck
 from squadops.cycles.patch_verification import materialize_artifacts
+from squadops.cycles.verification_integrity import ResultStatus
 from squadops.llm.exceptions import LLMError
 from squadops.llm.model_registry import get_model_spec
 from squadops.llm.models import ChatMessage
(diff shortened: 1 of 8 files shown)


jasonLadd (commented):
reviewed

PR #723, opened 2026-08-05
fix(memory): LanceDB search prefilters in-query and scores with an explicit cosine metric (#571)

jasonLadd:
1.4.4 fix 7 of 7 (plan: `docs/plans/1-4-4-patch-plan.md`). Mechanism in the commit message.

- Prefiltered `.where(namespace = ... AND array_has(tags, ...), prefilter=True)` + explicit `.metric("cosine")`; threshold deliberately stays a post-filter (quality floor on the now-valid score).
- Tests against a **real** LanceDB store with deterministic vectors — the starvation repro (nearest-N all wrong-namespace → old code returns zero, fixed code returns the matches), the tag variant, the scaled-same-direction cosine pin, and the threshold floor. **Revert-checked**: 3 of 4 fail on the pre-fix adapter.
- Dev venv runs lancedb 0.29.2; agent images pin 0.8.2 — the API surface used (`metric`, `prefilter`, `array_has`) exists in both, and tonight's integration deploy window includes an in-container recall smoke against the pinned version.
- Full regression: **6207 passed, 1 skipped** (the regression gate excludes the memory domain; the 4 new domain_memory tests run standalone — 4 passed, revert-checked).

Closes #571

### DIFF
--- a/adapters/memory/lancedb.py
+++ b/adapters/memory/lancedb.py
@@ -126,26 +126,33 @@ class LanceDBAdapter(MemoryPort):
             # Generate query embedding via port
             query_embedding = await self._embeddings.embed(query.text)
 
-            # Search
-            results = self._table.search(query_embedding).limit(query.limit).to_list()
+            # #571: filters are pushed into the query with prefilter=True —
+            # filtering in Python AFTER .limit(N) returned the N nearest rows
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #735, opened 2026-08-05
feat(replay): SIP-0101 Slice 1 — replay evidence rails (inert)

jasonLadd:
## What

Slice 1 of the SIP-0101 minimum path (1.5 Gate-1 Track C; `docs/plans/SIP-0101-cycle-replay-harness-plan.md`): the evidence rails, landed **before** the mechanism per the plan's sequencing rule — nothing sets the attribute yet, so this is inert and independently mergeable, and an unlabelled replayed run can never come to exist.

- `ReplayProvenance` (source_run_id, boundary_index, compatibility_set) — placed in `verification_integrity.py` beside its evidence-disclosure siblings `UnverifiedCheck`/`WaivedCheck`, keeping that module's zero-squadops-imports purity guard intact; `cycles/replay.py` re-exports it (the SIP-named path) and owns the single-sourced disclosure wording.
- `CycleOutcome.replay: ReplayProvenance | None = None` + DTO/mapping threading on the cycle detail GET.
- Run report leads with the disclosure when provenance is passed; `cycles show` renders a yellow banner off the response's replay block.
- Tests cover both §4 halves: a replayed outcome can't render unmarked, a normal outcome can't render marked; DTO boundary survival; wording single-sourced.

Regression: **6235 passed**, quality lint clean.

## Decision needed before Slice 2 (flagging per the plan's own rule)

Slice 2 (boundary retention — checkpoints worth replaying currently get pruned by `max_keep=5` mid-run) carries an **additive `retained` column migration**. The 1.5 plan says "#682 is the line's only expected migration; any second is a design-review stop." This one was in the accepted SIP-0101 plan all along — the 1.5 plan's premise was incomplete, not the migration new. **Recommendation:** accept it (additive, nullable-default-false, same 1010-pattern compatibility; range 1000–1099) and amend the plan line to "two expected migrations (#682, SIP-0101 retention)." Say the word with the merge and Slice 2 proceeds.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01LtPXzvgnwxqqwswMcsCTWv

PR #736, opened 2026-08-05
feat(replay): SIP-0101 Slice 2 — boundary retention (migration 1020)

jasonLadd:
## What

Slice 2 of the SIP-0101 minimum path (1.5 Gate-1 Track C): **workload-boundary checkpoints survive pruning**. Today `max_keep=5` deletes a long run's post-framing/post-dev boundaries before the run finishes — the exact state a replay restores from.

- **Migration 1020** (`retained BOOLEAN NOT NULL DEFAULT FALSE`) — additive, 1010-pattern compatibility, forward-only; the second of the line's two owner-accepted migrations (plan amended in this PR per the #735 ruling).
- Port: `save_checkpoint(..., retain: bool = False)`; postgres prune `DELETE` excludes retained rows; memory adapter parity (same all-rows survival window, so retention widens survival, never narrows it).
- Executor: `_is_phase_boundary` (last task or role transition) marks the checkpoint retained at the #643 collect-and-checkpoint site.
- **Premise delta vs the SIP plan's task 2.4** ("mark at the two write sites"): the correction-runner site checkpoints correction-chain *iteration state*, not replay boundaries — it stays deliberately unretained (commented in place) so a correction storm can't pin unbounded rows past the prune window.

## Tests

Retained-survives-overflow + unretained-pruning-unchanged + latest-ignores-retention (real memory store); boundary predicate (last task / role transition / mid-phase); postgres INSERT+DELETE SQL patterns; executor threads `retain` on every save with the final checkpoint retained. Regression **6246 passed**.

Next: Slice 3 (execution_mode + checkpoint-sourced restore + strict-equality compatibility gate) — closes Gate 1.

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -552,6 +552,9 @@ class CorrectionRunner:
             plan_delta_refs=tuple(plan_delta_refs),
             created_at=datetime.now(UTC),
         )
+        # SIP-0101 Slice 2: correction-chain checkpoints are iteration state, not
+        # replay boundaries — deliberately unretained (retain defaults False) so a
+        # correction storm cannot pin unbounded rows past the prune window.
         await self._cycle_registry.save_checkpoint(new_checkpoint)
         self._event_bus.emit(
             EventType.CHECKPOINT_CREATED,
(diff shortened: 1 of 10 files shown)


jasonLadd (commented):
reviewed

PR #739, opened 2026-08-05
fix(correction): #687 — thread the app traceback into failure_evidence

jasonLadd:
## What

First half of the Gate-2 #687+#431 pair (plan A3): **the analyzer stops guessing at behavioral 500s** — the app's own stack trace reaches `failure_evidence` as structured data.

- Probe runner: each probe is bracketed with its stderr-spool position; a FAILED probe reads its own delta with a short grace poll (measured: uvicorn flushes the traceback ~50ms *after* the 500 response) and carries the **last traceback block, tail-kept** under a 2000-char bound — the live test caught that a head-kept cap preserved anyio middleware frames and dropped the one line naming the defect.
- `build_failure_evidence` hoists tracebacks top-level (bounded 3) beside the rows, so `data.analyze_failure`'s evidence leads with the actual cause.
- **The A3 shared taxonomy lands here**: `FailureEvidenceCategory` (seven categories) + `derive_failure_category` stamped on every payload, with diagnosis-integrity precedence — infrastructure/emission facts invalidate downstream product signals (the #431 lesson made deterministic). `EXTRACTION_LOSS` is declared with its signal arriving in the #431 half, next PR.

## The exhibit, live

shk-2 (`cyc_88162ecfd895`): `NameError: name 'RunEvent'` at request time → 500 → two of five correction attempts burned on guessed causes while the traceback sat in the sandbox spool. The headline test reproduces the shape end-to-end — a NameError route under real uvicorn — and asserts the captured evidence names the undefined symbol.

## Verification

Regression **6291 passed** (enum-shadow guard caught three raw literals in my first pass — now `ResultStatus`). Replay preservation (A3 criterion): evidence is persisted data on analysis inputs/artifacts, so stored-artifact replays carry the category and tracebacks by construction.

Next: the #431 half — per-attempt emission stats + the extraction-loss gap rule feeding `EXTRACTION_LOSS` and the locus classifier.

Closes #687

### DIFF
--- a/src/squadops/capabilities/handlers/probe_runner.py
+++ b/src/squadops/capabilities/handlers/probe_runner.py
@@ -25,6 +25,7 @@ qa.test handler wraps it in a thread.
 from __future__ import annotations
 
 import contextlib
+import dataclasses
 import os
 import socket
 import subprocess
@@ -42,6 +43,7 @@ from squadops.cycles.verification_contract import (
     capture_probe_values,
     resolve_probe_path,
 )
+from squadops.cycles.verification_integrity import ResultStatus
 
 # The one subject the default profile knows how to boot: a FastAPI backend. A probe whose
 # subject the active profile cannot boot is reported skipped (not-executed), never a false pass.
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #741, opened 2026-08-05
fix(correction): #435 — progress-aware termination (the A4 bounded lever)

jasonLadd:
## What

Gate-2 A4: correction termination becomes **progress-aware** — the one behavioral change the odd-minor quarantine exists for, built to the decision table posted on #435 before code (the design-gate artifact).

- **A4.1** `correction_signature.py` (pure): a round's signature = failing product rows' `(check_id, subject_file, reason_token)`. Evidence text never participates (a re-described identical failure still repeats); infra rounds (A3's `extraction_loss`/`emission_failure`) have no signature and **clear** chain adjacency.
- **A4.2** movement classes: partial reduction is PROGRESS and resets; expansion is recorded but deliberately not a repeat (exact equality only).
- **A4.3** the rule: two **adjacent** rounds, **same** normalized complete signature, **both** decisions carrying a non-`none` structural candidate → the chain terminates `plan_defect` after the decision step, **before any repair dispatch**. Bound by the M2→M3 gate evaluation: candidates are pervasive (89% of deltas), so all selectivity comes from the signature repeat — shk-4 is the canonical true positive (three identical rounds, ~2h; this rule names it at round 1), single-round converged cycles the false-positive population.
- **The typed owner**: `CorrectionTermination` + `CorrectionTerminationReason` in `task_outcome.py` beside the SIP-0079 §7.7 vocabulary; persisted as a `correction_termination` artifact; `failure_reason` (#427) names it; cycle status stays derived; #683's wrap-up consumer picks it up when it lands; replay preserved by artifact persistence. Only `PLAN_DEFECT` is produced — the other reasons are declared vocabulary for #557's SIP and the exhaustion path.

Chain state is executor-owned, run-lived beside `correction_counter`, keyed by failed task id; callers without it are byte-identical to today (tested).

## Out of scope (per the 1.5 plan)

Any plan mutation — SIP-0092 M3 remains the 1.6 decision on the gate evaluation's evidence.

## Verification

Regression **6312 passed**. Pure-rule tests enforce every design-gate binding; runner-level tests prove the two-round termination (artifact stored, abort before repair dispatch) plus the none-candidate and no-state controls. Per the evidence matrix, the **designed-failure probe** (a plan-planted unwinnable task tripping the terminate row) and the **no-false-termination check on a converging roll** ride the Gate-2 exit shakedown at the deploy window.

Closes #435

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -38,11 +38,19 @@ from datetime import UTC, datetime
 from hashlib import sha256
 from typing import TYPE_CHECKING, Any
 
+from adapters.cycles.execution_errors import _ExecutionError
 from squadops.cycles.agent_config import resolve_agent_config
 from squadops.cycles.checkpoint import RunCheckpoint
+from squadops.cycles.correction_signature import (
+    classify_movement,
+    failure_signature,
+    render_signature,
+    should_terminate_plan_defect,
+)
 from squadops.cycles.failure_evidence import build_failure_evidence, compose_failure_trigger
 from squadops.cycles.models import ArtifactRef
 from squadops.cycles.plan_delta import PlanDelta
+from squadops.cycles.task_outcome import CorrectionTermination, CorrectionTerminationReason
 from squadops.events.types import EventType
 from squadops.tasks.models import TaskEnvelope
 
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #742, opened 2026-08-05
feat(verification): #682 — the gate-waiver slice (SIP-0096 §6.5/AC#12)

jasonLadd:
## What

Gate-2 A2, first of the SIP-0096 completion trio: **accept-with-waiver**, end to end. Before this, only the `WaivedCheck` shape existed (the audit's finding) — `CycleOutcome.waived` was permanently empty and a `blocked_unverified` gate had no recorded waiver path.

- **`GateDecision.waived_checks` + `waiver_reason`** (additive defaults; every existing call site unchanged).
- **Migration 1030** — the line's second and final expected migration, under A2's migration gate: additive + nullable (NULL = no waiver, distinguishable from a recorded one), old code ignores the columns, pre-migration rows assemble to empty defaults (tested), forward-only policy in the header. Replay of pre-migration artifacts produces the same roll-up (aggregation is unchanged when no waivers exist).
- **Validation** (`validate_waiver_request`, homed in `models.py` beside `GateDecisionValue` — the enum-shadow guard and the aggregation module's purity guard were mutually exclusive, so the validator lives with the enum): approved-only (accept-with-waiver), non-empty reason (**a waiver is never implicit**), and waived ids ⊆ the run's own unverified disclosure via a new `get_run_verification_summary` port getter — **a waiver can never touch a check that ran**.
- **Roll-up**: `resolve_cycle_outcome` collects waivers off recorded gate decisions into `CycleOutcome.waived`; the verdict remains the **un-waived** evidence verdict, tested (a waived `blocked_unverified` still reads `blocked_unverified`, with the waiver beside it).
- **Surface**: `GATE_DECIDED` payload, response DTOs, and `squadops runs gate --approve --waive <check_id> --waiver-reason "..."`.

## Verification

Regression **6325 passed**. Migration 1030 boot-apply + a live waiver exercise ride the Gate-2 deploy window (the SIP's §15 E2E case: required-check `blocked_unverified` → waiver path recorded).

Next in the trio: #683 (wrap-up consumes `CycleOutcome`) — schema before reader, per the plan's concurrency rule.

Closes #682

### DIFF
--- a/adapters/cycles/memory_cycle_registry.py
+++ b/adapters/cycles/memory_cycle_registry.py
@@ -236,6 +236,10 @@ class MemoryCycleRegistry(CycleRegistryPort):
             raise RunNotFoundError(f"Run not found: {run_id}")
         self._verification_summaries[run_id] = summary
 
+    async def get_run_verification_summary(self, run_id: str) -> RunVerificationSummary | None:
+        """One run's persisted verification roll-up, or None (#682)."""
+        return self._verification_summaries.get(run_id)
+
     async def list_run_verification_summaries(self, cycle_id: str) -> list[RunVerificationSummary]:
         """Return a cycle's persisted per-run verification summaries, by run_number."""
         runs = sorted(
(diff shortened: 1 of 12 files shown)


jasonLadd (commented):
reviewed

PR #744, opened 2026-08-05
feat(verification): #684 SIP-0096 §9 inert-check detection — CycleOutcome.inert populated

jasonLadd:
Closes #684 — the last of the SIP-0096 trio (#682 → #683 → #684). `CycleOutcome.inert` stops being permanently empty.

Design decisions ratified via the decision table posted to #684 before code (D1–D7). The load-bearing one: **detection is derived on read from the persisted per-run `RunVerificationSummary` rows — no counter table, no migration** (the line's policy allows exactly the two landed migrations; a materialized counter would have been a design-review stop). The "counter" is a streak computed by walking the perspective cycle's prior same-project/profile cycles newest→oldest at `resolve_cycle_outcome` time — the same derive-on-read philosophy as the roll-up itself — so reset-on-real-execution falls out of the walk and both existing consumers (API cycle detail, #683 wrap-up evidence injection) get the population for free.

## What landed

- **`src/squadops/cycles/inert_detection.py`** (new, pure): `detect_inert_checks` — per stable check, walking newest→oldest: execution resets, a not-executed report increments, absence **pauses without resetting** (§9: the counter "resets only when the check evaluates the real subject — not when the check disappears, is renamed, or is reclassified optional"); inert iff streak ≥ N. `cycle_check_state` derives one cycle's reconciled `(executed, not_executed)` via `aggregate_cycle_outcome`, so detection and disclosure can never disagree about what a cycle proved.
- **Identity domain** = `check_registry.framework_check_ids()` — the §6.3 single source of truth (its docstring already rules pulse suites a separate axis; typed-acceptance checks have per-cycle identity). Rename survival holds by construction: these ids are the canonical strings producers emit.
- **Series scope**: strict `(project_id, squad_profile_id, request_profile)` equality, cycles created strictly before the perspective cycle, bounded at 10 prior cycles — cross-profile history can't accrue streaks against checks with different applicability, and a streak not resolvable within the window is never guessed inert.
- **N is config-addressable**: `SQUADOPS__CYCLES__INERT_CYCLE_THRESHOLD` (CyclesConfig, default 3, ge=1), drift-tested against the module constant; `resolve_cycle_outcome` takes an explicit `inert_threshold` override for tests.
- **Containment**: disclosure-only enrichment — a history-read failure logs a warning and yields `inert=()`, never a failed roll-up; the verdict and #683 confidence ceiling never depend on it.
- **Threading**: `CycleOutcomeDTO.inert` + mapping; one `inert:` line in the wrap-up `verification_evidence_summary` (tolerates pre-#684 outcome dicts).
- **Premise-delta fix (first commit)**: `list_cycles` had no ordering contract — postgres returns `created_at DESC`, the memory adapter returned insertion order and sliced the *oldest* page under `limit`, so a bounded read handed different windows per adapter. The port docstring now states newest-first, the memory adapter conforms, and a parity test enforces it.

## Verification

- Pure-walk unit tests: threshold/reset/pause semantics, perspective-cycle execution wins, chronic-but-absent-now stays inert, non-stable ids invisible, invalid threshold.
- **Multi-cycle real-store sequence** (the evidence matrix's row for A2 #684, not mocks): through `resolve_cycle_outcome` against the memory registry — three chronic cycles flag inert / a real execution clears it / cross-profile cycles are outside the series / history-read failure contained.
- Full regression: **6360 passed, 1 skipped**.
- Live proof rides the Gate-2 exit confirmation shakedown, per the line's pattern.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -75,6 +75,9 @@ SQUADOPS_PROFILE=dev  # Options: dev, local, lab, cloud
 # SQUADOPS__TELEMETRY__OTLP_ENDPOINT=http://localhost:4317
 # SQUADOPS__TELEMETRY__PROMETHEUS_PORT=8888
 
+# Cycle Verification (SIP-0096)
+# SQUADOPS__CYCLES__INERT_CYCLE_THRESHOLD=3   # §9 N: consecutive not-executed cycles before a check is inert
+
 # Agent Lifecycle Configuration
 # SQUADOPS__AGENT__INSTANCES_FILE=agents/instances/instances.yaml
 # SQUADOPS__AGENT__HEARTBEAT_TIMEOUT_WINDOW=90
--- a/adapters/cycles/memory_cycle_registry.py
+++ b/adapters/cycles/memory_cycle_registry.py
@@ -83,6 +83,11 @@ class MemoryCycleRegistry(CycleRegistryPort):
                 if derived != status:
                     continue
             results.append(cycle)
+        # Port ordering contract: newest first, so `limit` pages the most recent
+        # cycles — matching the postgres adapter's ORDER BY created_at DESC
+        # (#684: the inert-detection history walk depends on which window a
+        # bounded read returns).
+        results.sort(key=lambda c: c.created_at, reverse=True)
         return results[offset : offset + limit]
 
     async def cancel_cycle(self, cycle_id: str) -> None:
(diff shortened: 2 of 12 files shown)


jasonLadd (commented):
reviewed

PR #729, opened 2026-08-05
docs(1.5): typed-check governance design — the A5 curated menu

jasonLadd:
## What

Gate-1 deliverable A5 of the 1.5 line: `docs/plans/1-5-typed-check-governance-design.md` — the curated typed-check menu, designed as a **governance extension of the existing `CHECK_SPECS` registry** (ownership-before-extension: the machine-readable registry the plan asks for already exists in `acceptance_check_spec.py`; this adds the six missing governance attributes with backfill + drift test + generated docs).

## Rulings requested at this review (recommendations inline)

- **D1 (#504):** restore nothing further; promote the fill-slot divergence report to a blocking injected check `fill_slot_signature` — detection ≠ safe restore.
- **D2 (#629 L2):** two entries split by determinism — `contract_assertions_match` blocking (bind mode), `plan_prose_contract_divergence` advisory with a promotion rule.
- **D3 (#668):** `testids_present` static view check = capacity-bound 1.5; qa-suite-assertion half → 1.6 (trigger: SIP-0102 step 4).
- **D4 (#598 check-half):** `package_builds` declared-unbuilt in the registry (trigger: Stack Blueprint).

Also records the `command_exit_zero` → #707 dependency (failure_ownership untrustworthy until the allowlist precedence ruling exists).

Merging this PR = ratifying the four rulings as recommended (or as amended in review).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01LtPXzvgnwxqqwswMcsCTWv

PR #745, opened 2026-08-05
feat(verification): #629 contract-expectation enforcement, split by determinism (A6/D2)

jasonLadd:
Closes #629 — the enforcement half (layer 2), executing the ratified D2 ruling (governance design, PR #729) and plan A6: **two identities, split by determinism**. Design decision table posted to #629 before code. Layer 1 (authoring injection) had already shipped; this is the backstop for when statement is not enough — pf-54's suite asserted 200-on-create against a pinned 201 **five times**, and five dev-chain repairs of a contract-correct app burned the full correction budget against a suite the contract says is wrong.

## The blocking half — `contract_assertions_match`

- **Injection** (`task_plan.py`, beside `_harness_boundary_criteria` — the exact precedent): bind-mode qa.test tasks get one row per expected `.py` suite file in the stack's QA namespace, `id=contract-assertions:<file>`. Params are self-contained (`METHOD /path STATUS` tokens + suite-wide allowed error statuses), so the evaluator needs no contract access. Author mode / non-qa / probe-less contracts inject nothing — byte-identical plans there.
- **Derivation** (`VerificationContract`, beside `behavior_expectation_lines()`): `pinned_endpoint_statuses()` — a rejection-case probe unions into the same endpoint's set (`POST /runs → {201, 422}`), so asserting 422 after a blank-input POST is contract-correct, never flagged; `allowed_error_statuses()` — every 4xx/5xx a probe pins plus statuses named in `coverage_expectations`. One derivation, three readers.
- **Evaluator** (`acceptance_checks.py`): AST-extracts `(method, path, asserted_status)` triples — assigned-response and direct-call forms, `await` unwrapped, either operand order, `Eq` only, latest binding before the assert by source position. A violation requires an exact pinned `(METHOD, path)` asserted outside pinned ∪ allowed-error, **or the prefix rule**: a non-pinned request path that reaches a pinned path through an undeclared prefix (`/api/runs` vs pinned `/runs` — 3 of pf-54's 5 versions, invisible to the status rule alone while every call 404s against the harness client). Unextractable assertions and off-contract paths are out of scope by design — **zero false positives is the bound**, because a false positive in a blocking check recreates the unwinnable loop this kills. Syntax errors skip (`unsupported_stack_or_syntax`, #605 sibling semantics); the #605 registry-wide skip property covers the new check automatically.
- **Repair routing** (`failure_evidence.py`): a failed row → `FailureLocus.OWN_ARTIFACT` (eve re-authors the suite), with **precedence over the `tests_pass` scan regardless of row order** — when the suite both contradicts the contract and fails, the app must not be the repair target. Safe from the test-gaming guard: the signal comes from the frozen contract, not from the app failing the suite.

## The advisory half — `plan_prose_contract_divergence`

The status-code rule the pf-31 prose lint never had (#590, folded into this issue): **comparable prose** = a `METHOD /path` token followed on the same line — before any next endpoint token — by an explicit 3-digit status; **the contract is authoritative**. Warning rows only on the existing never-rejects channel (the reverted-#552 rule bars hard gates on prose), carrying their own identity so their evidence quality never launders into the blocking check. Promotion to blocking requires a later registry change with proof of deterministic representation, per the ruling.

## Scope notes

- #730 (CheckSpec governance-attribute extension) is untouched: both identities land against today's registry-of-record; `failure_ownership=suite` and the advisory marking ride #730's backfill.
- `plan_prose_contract_divergence` is deliberately NOT a `CHECK_SPECS` entry — that would make it plan-authorable.

## Verification

Evidence-matrix row A6 ("deterministic-side exhibit; prose side ships advisory"): the pf-54 exhibit as tests — wrong create status fails naming the exact row; the rejection assertion passes; the `/api`-prefixed variant fails via the prefix rule; direct-call/reversed/await/rebinding forms extract; off-contract and unextractable assertions never flag; locus routes at the suite with precedence; the prose side proves warning-not-rejection with contract-correct and incomparable prose silent. Full regression: **6395 passed, 1 skipped**. Live proof rides the Gate-2 exit confirmation shakedown.

### DIFF
--- a/src/squadops/cycles/acceptance_check_spec.py
+++ b/src/squadops/cycles/acceptance_check_spec.py
@@ -247,6 +247,18 @@ CHECK_ENDPOINT_DEFINED = "endpoint_defined"
 # above — a literal there would silently inject nothing after a rename.
 CHECK_UNDEFINED_NAMES = "undefined_names"
 
+# #629 (1.5 A6/D2): the blocking suite-vs-contract check. Multi-reader constant
+# (injection in task_plan, locus routing in failure_evidence, the evaluator,
+# tests) — same single-source rule as the two above.
+CHECK_CONTRACT_ASSERTIONS = "contract_assertions_match"
+
+# #629 (1.5 A6/D2): the ADVISORY prose-vs-contract identity. Deliberately NOT a
+# ``CHECK_SPECS`` entry — that would make it plan-authorable; it names the
+# warning rows the plan-prose lint emits (``implementation_plan``), keeping its
+# lower evidence quality from laundering into the blocking check above. The
+# full governance-registry entry rides #730's attribute backfill.
+PLAN_PROSE_CONTRACT_DIVERGENCE = "plan_prose_contract_divergence"
+
 
 # The `"METHOD /path"` endpoint-token grammar. `endpoint_defined.methods_paths`
 # is authored in it and the evaluator matches route decorators against it; #688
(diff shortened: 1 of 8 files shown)


jasonLadd (commented):
reviewed

PR #746, opened 2026-08-05
fix(observability): #506 in-flight task runs visible — transport owns the task-run lifecycle

jasonLadd:
Closes #506. Design decision table posted to the issue before code, including a root-cause correction established against the **live Prefect store** rather than the issue's framing.

## The actual mechanism (premise delta)

The issue says "task runs materialize only at completion — create at dispatch"; the disposition guessed the fix "touches the workflow-tracker port and its adapter." Neither held: dispatch-time creation has existed since SIP-0087 (April, deployed during the July-19 observation), and **the port and adapter need no changes**.

Reading the July-19 timeline off the live server's `task_runs/filter` nailed it: `develop[2/7]` failed at 02:12:22 and its task run went terminal FAILED. `dispatch_with_retry`'s `while True` loop then re-dispatched attempts 2 and 3 with the **same task_run_id** — and `dispatch_task` only set RUNNING on ids it created itself, so the already-terminal row never re-entered RUNNING and no new row appeared. The 02:28→02:45 develop task the operator watched heartbeat "with no row" was attempt 3; `_dispatch_protocol_step` steps in the same window (validate_repair, round-2 analyze/decision/repair) all had correct dispatch-time rows. The "retroactive" appearance was the old row's state flipping when the invisible attempt ended. A second, latent hole: an id created *inside* `dispatch_task` never escaped, so `dispatch_with_retry`'s terminal events carried `task_run_id=""`.

## The fix — transport owns the lifecycle end to end

- **`dispatch_task` (re-)enters RUNNING on every dispatch**, including a supplied id — a retry re-attempt arrives with a prior attempt's terminal id and is now visible in flight. Prefect's own retry model: one task run, state history carries the attempts (no row proliferation; SIP-0087 logs/heartbeats keep one pane per task).
- **`dispatch_task` sets the terminal state from the result** (COMPLETED iff SUCCEEDED; FAILED otherwise, including the raise path), placed exactly beside `_start_task_activity`/`_finish_task_activity` — completing SIP-0097's stated rule that per-task observability starts *and finishes* in the transport, and closing the internal-id hole without any port change. Best-effort throughout: a tracker failure logs and never touches dispatch.
- **`WorkflowTrackerBridge` drops its task-state map** — with the transport owning terminal states it was a second writer of the same state (whose context id was blank on exactly the paths that needed it). Flow-run lifecycle stays bridge-owned; the `TASK_*` events keep flowing for other consumers.

Every dispatch seam (sequential, fan-out, correction, pulse-repair) flows through `dispatch_task`, so the consolidation covers all paths at once — the July asymmetry was per-path only because the retry loop was the one seam that skipped the transport's create step.

## Verification

Dispatcher unit tests: supplied id → RUNNING before publish + terminal from result (the invisible-attempt regression test); FAILED reply → terminal FAILED; internally-created id transitions terminally; tracker failure never breaks dispatch; tracker/flow_run_id absent → zero calls. Bridge + parity tests now pin the single-writer rule (a reintroduced bridge transition fails the suite). Executor observability test updated to the new contract (5 creates, 10 RUNNING, 5 terminal). Full regression: **6394 passed, 1 skipped**.

**Live proof rides the Gate-2 exit shakedown**: mid-cycle `POST /api/task_runs/filter` must show the in-flight task RUNNING — the exact query used in this diagnosis, now the reproducible check.

### DIFF
--- a/adapters/cycles/task_dispatcher.py
+++ b/adapters/cycles/task_dispatcher.py
@@ -107,8 +107,24 @@ class TaskDispatcher:
         task_run_id = await self._workflow_tracker.create_task_run(
             flow_run_id, envelope.task_id, task_name
         )
-        await self._workflow_tracker.set_task_run_state(task_run_id, "RUNNING", "Running")
+        await self._set_task_run_state(task_run_id, "RUNNING", "Running")
         return task_run_id
+
+    async def _set_task_run_state(
+        self, task_run_id: str | None, state_type: str, state_name: str
+    ) -> None:
+        """Best-effort task-run state transition; a tracker failure never touches dispatch."""
+        if self._workflow_tracker is None or not task_run_id:
+            return
+        try:
+            await self._workflow_tracker.set_task_run_state(task_run_id, state_type, state_name)
+        except Exception:
+            logger.warning(
+                "best-effort set_task_run_state(%s, %s) failed",
+                task_run_id,
+                state_type,
+                exc_info=True,
+            )
 
     async def _task_heartbeat(
         self,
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #747, opened 2026-08-05
refactor(prompts): #452 build-profile narratives externalized byte-identically

jasonLadd:
Closes #452. Design decision table posted to the issue before code — including a **scope reconciliation you should review** (D1), since it reclassifies two of the disposition's six counted blocks.

## What moved (the live-path payload)

The four `system_prompt_template` literals in `build_profiles.py` — composed into the builder system prompt on **every** builder task via `system_prompt_for_files` — now live as `src/squadops/prompts/profile_narratives/<profile>.md`, loaded once at import. Prompt prose is reviewable and editable as prose under the prompts seam (#448's rule in substance); `BuildProfile` stays a frozen dataclass and the builder handler is untouched. The profile dimension stays deliberate contract data (the issue's own "keep — and say so" arm); extending the fragment model with a non-role dimension remains the deferred SIP-sized question. The files ride into agent images via the existing `COPY src/` path, same as fragments and request templates.

## The hard acceptance gate (plan §F: byte-identical or it didn't happen)

`test_profile_narratives.py` pins the **pre-move sha256 of every narrative AND every composed `full_system_prompt`** (captured from main before the move). Any byte drift fails the suite — a deliberate prompt change must update the pins in the same PR, so it reads as a prompt change, never a refactor. Also pinned: every profile must be in the byte-equivalence regime (a new profile can't slip past it); the loader strips exactly one trailing newline (files stored newline-terminated, so editor hygiene can't move prompt bytes); a missing narrative file raises at import — loud, never a silently-empty prompt.

## Premise delta for your review (documented on the issue)

The disposition counted six blocks. Two — `cycle/builder.py` (1) and `impl/repair_handlers.py` (1) — turn out to sit on the **renderer-absent fallback branch** (SIP-0084 dual-path): their live paths already render the governed assets `request.builder_assemble.build_assemble` and `request.cycle_repair_task`, and the inline text renders only when renderer construction failed in the agent entrypoint (degraded mode). That is the exact class the disposition itself excluded ("fallback-path text for when no renderer is injected, not managed-asset candidates") — the two excluded `"Please provide your {role}..."` lines, just longer. Fallback text for asset-system absence cannot itself live in the asset system, so these stay inline, deliberately. **With that, live-path inline prompt prose in handler source is zero.**

Residue, named: `_plan_authoring_service` typed-acceptance doctrine and capability-dimensioned `test_prompt_supplement`s were outside the disposition's six-block count and stay put; the fragment-dimension design is the deferred SIP-sized item.

## Verification

Byte-equivalence suite (10 tests) + existing build-profile and builder-assemble suites green; full regression: **6405 passed, 1 skipped**. No live render path changed by construction, so no shakedown scenario is added for this one.

### DIFF
--- a/src/squadops/capabilities/handlers/build_profiles.py
+++ b/src/squadops/capabilities/handlers/build_profiles.py
@@ -11,11 +11,37 @@ narrative `system_prompt_template` cannot drift away from what the validator
 will accept. Adding a file to `required_files` automatically adds it to the
 prompt; the file list cannot be edited in the prompt without also editing
 the validator's `required_files`.
+
+#452: the narrative TEXT lives in ``src/squadops/prompts/profile_narratives/``
+(one ``.md`` per profile, loaded once at import) — prompt prose is reviewable
+and editable as prose under the prompts seam, never a Python literal (#448).
+The profile dimension stays deliberate contract data rather than a fragment
+layer; extending the fragment model with a non-role dimension is the deferred
+SIP-sized question. Byte-equivalence with the pre-move literals is pinned by
+test (the #452 hard acceptance gate).
 """
 
 from __future__ import annotations
 
 from dataclasses import dataclass, field
+from pathlib import Path
+
+import squadops.prompts as _prompts_pkg
+
+_NARRATIVES_DIR = Path(_prompts_pkg.__file__).parent / "profile_narratives"
+
+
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #755, opened 2026-08-06
fix(#481): startup sweep surfaces cycles stranded between workloads (closes #481)

jasonLadd:
## #481 — startup sweep surfaces cycles stranded between workloads (1.5 Gate-3 B3)

Executes the design table posted on the issue. The attempt-2 incident class (`cyc_9d775e1aebeb`): `execute_cycle` dies between workloads, and the cycle sits silently in the window `resolve_cycle_status` rule 3 names in its own docstring — "between gate approval and next-Run creation" — until a human notices. Recovery has been complete since #433/#434; this adds the missing half, detection.

### The domain predicate (D1/D2)

`classify_workload_stranding(workload_sequence, runs, cycle_cancelled)` in `cycles/lifecycle.py`, beside the status derivations — a pure function of the positional D14 run↔workload invariant plus the latest run's own `gate_decisions`:

- **`STRANDED`** — latest non-cancelled run COMPLETED at a non-final position, boundary approved or ungated, no successor. The only actionable class.
- **`GATE_PENDING`** — no decision recorded: visibility only (the gate *poller* died too, so approval alone won't resume it — the log says approve **then** retry).
- **`GATE_REJECTED`** — rejection recorded but the supersede/re-roll died with the process: visibility only.
- **`None`** — everything else, including two cases the survey settled against the executor's #466 exhaustive dispatch: `APPROVED_WITH_REFINEMENTS` **advances**, so its absence strands (⊂ STRANDED); `RETURNED_FOR_REVISION` is a *deliberate, recorded* stop awaiting manual retry — classifying it stranded would aim recovery at an operator-owned state. Unrecognized future decision values are likewise never stranding. *(One refinement vs the posted table: the table's GATE_PENDING is tightened to strictly "no decision rows" — a revision-returned gate is not "pending".)*

### The sweep (D3/D5/D6)

`detect_stranded_cycles` — 4th sweep in `api/runtime/startup_reaps.py`, wired after the three reaps, same best-effort contract (failure logged, never blocks boot). **READ-ONLY**: idempotent by construction across boots; racing a concurrent operator `runs retry` costs at most one stale warning. Each STRANDED cycle gets one structured warning naming position, workload type, gate state, and the exact recovery command (`squadops runs retry <project> <cycle>`).

Trap the survey confirmed, now load-bearing in both code and tests: the sequence-unaware `derive_cycle_status` calls the stranded state **COMPLETED**, so any status pre-filter at the port would miss every stranded cycle — the sweep classifies runs + sequence directly. Per-project enumeration bound (200 newest, #684 ordering) is logged when hit — no silent caps.

### Escalation ruling (D4 — merging ratifies)

Log-only ships and closes the issue (its own sequencing: "log-only is safe to ship first"). Auto-resume remains a possible follow-up requiring separate owner opt-in — not filed; the classification enum is exactly the seam it would consume.

### Tests

- 16 predicate tests: every classification, the #466 parity cases (refinements-approval strands; revision-return doesn't), contradictory-decision conservatism, the cancelled-successor-reopens-position case, and each silent class.
- 6 sweep tests: counted-and-names-recovery, pending-is-visibility, **read-only assertion on the registry mock**, failure-never-blocks-boot, unwired no-op, terminal/legacy silence.
- Full regression: **6459 passed, 1 skipped** (ruff clean).

Closes #481

### DIFF
--- a/src/squadops/api/runtime/main.py
+++ b/src/squadops/api/runtime/main.py
@@ -382,6 +382,7 @@ async def _init_cycle_subsystem(config, pool) -> None:
         # (#373) blocks recruitment outright, a live activity row (#672) kills
         # that agent's activity tracking.
         from squadops.api.runtime.startup_reaps import (
+            detect_stranded_cycles,
             reap_stranded_activities,
             reap_stranded_leases,
             reap_stranded_modes,
@@ -392,6 +393,9 @@ async def _init_cycle_subsystem(config, pool) -> None:
         await reap_stranded_leases(cycle_registry, _runtime_coordinator, focus_lease_port)
         await reap_stranded_modes(_runtime_coordinator, state_port)
         await reap_stranded_activities(cycle_registry, activity_port)
+        # Read-only, last: a post-crash boot immediately names every cycle
+        # stranded between workloads and its recovery command (#481).
+        await detect_stranded_cycles(cycle_registry, project_registry)
 
         flow_executor = create_flow_executor(
             "dispatched",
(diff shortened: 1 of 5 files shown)


jasonLadd (commented):
reviewed

PR #757, opened 2026-08-07
feat(#730): typed-check governance — CheckSpec registry extension (A5 PR 1)

jasonLadd:
## #730 — typed-check governance registry extension (1.5 A5, PR 1)

First PR of the ratified design (`docs/plans/1-5-typed-check-governance-design.md`, PR #729): extend `CheckSpec` with the six curated-menu attributes. **Pure enablement — the attributes are read by nothing yet**; zero behavior change (proposer vocabulary rendering untouched, all existing suites pass unmodified).

### The extension

`CheckSpec` gains `failure_ownership` · `qa_available` · `signature_participation` · `outcome_contribution` · `replayable` · `blocking_default` as **required keyword-only fields with no defaults** — the constructor itself forces every present and future entry to declare its governance explicitly, which is stronger than any presence drift-test could be (and pinned by a test so a refactor can't quietly reintroduce defaults). Vocabulary validated at load (`__post_init__`, the `check_registry` load-time pattern); ownership values are constants per the strings-boundary rule.

### Backfill — 12 entries (the design's 11 + `contract_assertions_match`, riding here as #629 planned)

Per the design's enumeration, with the judgment calls it left open resolved as:

- **ownership**: `harness_boundary` + `contract_assertions_match` → `suite`; everything else → `product`. `command_exit_zero` records the **#707 dependency in a comment beside the entry** (ownership is per-command in truth; untrustworthy until the allowlist inventory + precedence ruling — recorded, not solved, exactly as the design specifies).
- **qa_available**: False only for `frontend_compiles` (qa emits suites, not views); True elsewhere — including `regex_match`/`count_at_least` (authorable against qa-task-emitted targets; `count_at_least`'s canonical example *is* a qa glob) and `module_imports`.
- **signature_participation / replayable**: False together for the two environment-variant checks (`command_exit_zero`, `frontend_compiles`) — a failure that doesn't reproduce deterministically must not key A4 chain-termination identity. This pairing is stated as a **registry law** and enforced by test, so flipping it requires a ruled amendment, not an edit.
- **outcome_contribution**: True across the board (every evaluable check feeds the SIP-0096 roll-up today); **blocking_default**: `error` everywhere (D3's `testids_present` will be the first `warning`-windowed entry when it lands).

### D4 — declared-unbuilt, without the authorability trap

[...]

PR #758, opened 2026-08-07
feat(#730): fill_slot_signature — blocking injected check (D1; closes #730, #504)

jasonLadd:
## #730 D1 — `fill_slot_signature`: the pf-40 report promoted to blocking (closes #730, #504)

Executes the ratified D1 ruling: **restore nothing further — make the report cost.** SIP-0100 restores only the body-independent elements (`status_code`, the router assignment); everything else on the scaffold-owned signature surface was report-only, so pf-40's drift class (dropped `response_model`, renamed handler, renamed params) was free. Now any reported divergence fails acceptance with the divergence list as evidence, routing repair at the producer — the framework still never rewrites producer code.

### The pieces

- **`fill_slot_integrity` public pair** (the reporting logic, reused, not reimplemented): `declared_route_signatures(seed)` extracts the reported-not-restored surface as self-contained check params — `status_code` deliberately absent, so the check never fights its own restore layer (an emission the enforcement silently fixes must not fail acceptance for it). `signature_divergences(declared, emitted)` diffs by the same normalized route key the restore path uses, so the pf-31 path-param rename class still *matches* its scaffold counterpart — and surfaces as a parameter divergence rather than a silent non-match. Dropped routes contribute nothing (that's `endpoint_defined`'s job; one defect, one report).
- **Registry entry + evaluator** (the pairing forced by `assert_registry_complete`): injected, `.py`-scoped, governance per the extension — `product` ownership, `qa_available=False` (fill slots are dev-lane emissions), signature-participating, replayable, blocking `error`. Syntax-broken emissions **skip** (the syntax gate owns them); an empty `routes` declaration is an **injection bug (`error`)**, never an app gap.
- **Plan-time injection** beside the other criteria helpers: per claimed `.py` fill slot on `development.develop`, seeds derived from the manifest via the same expander the bound record uses (one source of truth). The `.jsx` view slots stay out — #668/D3's territory; a `.py`-scoped check there could only ever skip (the pf-47/pf-49 dead-weight class). Repairs inherit the check automatically through `acceptance_criteria` forwarding and `verify_patched_artifacts`.

### Deliberate regens (disclosed per protocol)

- Plan golden: `implementation_bind`'s dev task gains exactly one check carrying the group_run seed's 5-route declared surface.
- `typed-check-menu.md`: one new row (13 evaluable checks).

### Tests

Extraction shape (reported surface only), each divergence class individually named, path-param rename match, dropped-route non-report, unparseable-emission non-report; evaluator fail/pass/skip/error/file-not-found; injection presence + the three absence guards (no manifest, non-slot `.py`, `.jsx` slot). One planned test was dropped as unreachable: a qa task claiming a fill slot is rejected by the qa-artifact-ownership net before injection could ever run — recorded as a comment where the test would have been.

### Verification

Full regression: **6487 passed, 1 skipped** (ruff clean).

### Issue closure

- **Closes #504** — its remaining scope was exactly this promotion (restore-vs-report ruled: nothing further restored).
- **Closes #730** — PR 1 (registry extension, #757) + this D1 entry complete the issue's in-1.5 scope. The two remaining threads have homes: **D3-half-a `testids_present`** stays tracked by #668 (narrowed to the suite half for 1.6; the static-view half is capacity-bound — buildable in 1.5 if capacity allows, on #668); **D4 `package_builds`** is recorded declared-unbuilt in the registry with its named 1.6 trigger.

### DIFF
--- a/docs/architecture/typed-check-menu.md
+++ b/docs/architecture/typed-check-menu.md
@@ -14,6 +14,7 @@ Regenerate: `UPDATE_CHECK_MENU=1 pytest tests/unit/cycles/test_check_governance.
 | `count_at_least` | authored | product | yes | yes | yes | yes | error |
 | `endpoint_defined` | authored | product | yes | yes | yes | yes | error |
 | `field_present` | authored | product | yes | yes | yes | yes | error |
+| `fill_slot_signature` | injected | product | no | yes | yes | yes | error |
 | `frontend_compiles` | authored | product | no | no | yes | no | error |
 | `function_defined` | authored | product | yes | yes | yes | yes | error |
 | `harness_boundary` | authored | suite | yes | yes | yes | yes | error |
--- a/src/squadops/cycles/acceptance_check_spec.py
+++ b/src/squadops/cycles/acceptance_check_spec.py
@@ -312,6 +312,11 @@ CHECK_UNDEFINED_NAMES = "undefined_names"
 # tests) — same single-source rule as the two above.
 CHECK_CONTRACT_ASSERTIONS = "contract_assertions_match"
 
+# #730 D1 / #504: the fill-slot signature surface, promoted from report to
+# blocking injected check. Constant beside its CHECK_SPECS key so a rename
+# moves both together.
+CHECK_FILL_SLOT_SIGNATURE = "fill_slot_signature"
+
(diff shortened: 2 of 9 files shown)


jasonLadd (commented):
reviewed

PR #767, opened 2026-08-07
docs(roadmap): post-1.5 reconciliation — 1.7's identity, 1.8's co-headliners, memory rails

jasonLadd:
## What this record settles

Three things changed at the v1.5.0 cut that the post-1.4 reconciliation could not have anticipated. The roadmap reflected none of them.

1. **1.8's stated gate cleared.** The post-1.4 record gated 1.8 on "SIP-0096 implemented (its `CycleOutcome` fields designed for these readers)." That shipped in 1.5.
2. **1.7 acquired a real pool but had no identity** — it was "debt from 1.6," a placeholder. The 1.5 capacity roll moved a named, classified set into it. A populated pool with no claim becomes a junk drawer.
3. **1.5 built two of the seams Cross-Cycle Memory needs**, for unrelated reasons: `plan_rejection_context` in #663's context-assembly registry, and the existing `MemoryPort`. That changes the memory question from *build it or not* to *which half, and do the rails now*.

## Decisions (owner-ratified in session)

**1.7 — "every port is actually a port."** Where 1.5 extracted structure *inside* the machinery, 1.7 fixes where the machinery meets the outside world. Load-bearing rather than cosmetic: the Atlas provider migration cannot happen safely while a vendor's status vocabulary lives in domain objects and the composition root bypasses the factories, and 1.8's scorecard grades over seams that must be stable first. The pool is classified into five groups (boundary/vocabulary leaks, composition root, provider neutrality, wide infrastructure mechanics, pure extractions) with the composition-root cluster flagged **design gate before code**.

**1.8 — co-headliners, ordered.** The Cycle Evaluation Scorecard and Campaign share the release on the 1.4 dual-lane precedent; Campaign was previously billed as sole Lane M headline. **The composition of 1.8 does not change — the ordering inside it does:** grade definitions land before continuation policy.

That ordering is a dependency, not a preference. Without `CycleAssessment`, a continuation policy must derive its stopping rule from raw checks — exactly what the roadmap forbids of self-improvement — and a stopping rule that reduces to "the cycle completed" runs the false-green class unattended, at scale. Sequencing the scorecard first makes Campaign better, not merely later.

The counter-argument is recorded rather than dismissed: grading meaningfully needs many cycles, and hand-launching them has been the human bottleneck through every FAY window and shakedown. That is a real case for Campaign as enabler. It is resolved by ordering, not exclusion.

**Cross-Cycle Memory — a decision point, not a commitment.** Thin Phase 1 in 1.8, or pushed whole to 2.0 beside Phase 2, decided at 1.8 plan time, because two co-headliners are already substantial surface. **The rails ship in 1.8 either way**, per the standing rails-before-mechanism rule (SIP-0101 Slice 1; SIP-0096's inert Phase 1 core), so the choice stays an adapter swap.

## Design intentions carried forward

Five, so that building the scorecard and Campaign does not foreclose memory. The risk is not missing seams — two already exist — it is 1.8 inventing second versions of them.

| # | Intention | Owner |
|---|---|---|
| 1 | One cycle-lineage identity, owned by Campaign's objective envelope; inert detection and memory recall both read it | 1.8 |
| 2 | One failure-class vocabulary — scorecard attribution, rejection classes, `FailureEvidenceCategory`, locus — not parallel taxonomies | 1.8 |
| 3 | Recall extends the existing `plan_rejection_context` contract; never an `if memory_enabled` branch in an authoring handler | 1.8 |
| 4 | Ship the recall port with a NoOp, inert, regardless of the Phase 1 decision | 1.8 |
| 5 | **Record the rejection-class recurrence baseline during 1.6** | **1.6** |

**Intention 5 is the only one whose omission is permanent.** Memory's value claim is that recurrence of the same mistake falls; that needs a pre-memory baseline from the authored-mode window. Once memory is live the baseline is unrecoverable. Cheap while building authored mode, impossible afterward.

## Files

- **New**: `docs/plans/post-1-5-roadmap-reconciliation.md` — amends the post-1.4 progression table, rows 1.7 and 1.8 only.
- `docs/ROADMAP.md` — adds the missing **v1.7** entry (there wasn't one), rewrites v1.8, and extends the ladder line.
- Three SIP headers amended so they don't contradict the roadmap, per the post-1.4 precedent: **Scorecard** gains a Targets block and its two design constraints, **Campaign**'s billing becomes co-headliner, **Cross-Cycle Memory**'s placement becomes the decision point plus rails.

## Not moved by this record

The 1.6 and 2.0 rows stand, including 1.8's remaining gate (authored-mode FAY > 0 — it does not exist yet, which is why 1.8 gets a shape here and not a plan). No item-level issue re-homing — the six premise-moved issues from the post-cut sweep belong to the backlog pass. No SIP acceptance.


jasonLadd (commented):
reviewed

PR #768, opened 2026-08-07
sips: accept Squad-Authored Manifest as SIP-0103 (v1.6 Lane M headline)

jasonLadd:
## Why now

First step of 1.6 planning. Acceptance is a **design commitment on main**, not an implementation artifact — per the CLAUDE.md contributor workflow, the feature branch starts from a main that already carries the approved spec, and the 1.6 plan's headline row cites a numbered SIP rather than a filename that will move.

## Accepted with the design-review rounds in force

Not the original draft — §5a, §5b, and §5c are part of the commitment. The load-bearing correction is **§5b Correction 1**:

> Contract v9 is a hand-authored artifact, *bound* to the manifest by hash — no `derive_contract(manifest)` exists anywhere in the pipeline.

So squad-authoring the manifest alone would leave the contract as the remaining hand-wired seed, and bind mode without a derived contract is just author mode. Mechanical contract derivation enters scope, split by verified derivability (fully derivable interface layer / largely derivable probe skeletons / authored residue).

## Mechanics

- **Added YAML frontmatter the draft never had.** `update_sip_status.py` reads frontmatter and failed with a bare `Could not extract metadata` until it was present. Raised separately as a papercut — several proposed SIPs are drafted without it, so acceptance fails at the last mile with an unhelpful error.
- Script assigned **103** and moved the file; body H1, `**Status:**`, and `**Targets:**` hand-synced per the standing promotion process (the script's body-status regex expects `**Status:** X`, which this draft's `## Status` heading form didn't match).
- **Repointed live references**: ROADMAP v1.6 entry, the Accepted (Next Up) table (new SIP-0103 row), Stats (7 → 8 accepted); the post-1.4 reconciliation's live pointer. The post-1.4 "moves executed" line keeps its historical verb and gains the new path rather than being rewritten — that line records what a past PR did.
- The post-1.5 reconciliation's "no SIP acceptance" scope note is marked superseded in place, with the reason Stack Blueprint stays proposed.

## Stack Blueprint is deliberately not accepted

`SIP-Stack-Blueprint-Contract.md` carries a section titled *"Why this is not ready to accept"*: every field in its proposed schema is FastAPI-shaped thinking, and generalizing from one instance "produces the FastAPI contract with generic field names — which is worse than no contract, because it looks authoritative and the second stack will quietly bend itself to fit rather than reveal the mismatch."

Its stated acceptance gate is **the existence of a second real stack**. v1.6 builds that stack, so the SIP gets promoted *mid-release*, with its schema written against two instances. Its own sequencing section also prescribes the S lane's opening move: consolidate today's five scattered per-stack facts into one object — **no SIP required**, pure refactor with an exact byte-identical test. The 1.6 plan will carry both as explicit milestones.

## Next

The 1.6 plan (`docs/plans/1-6-0-*.md`) follows in its own PR.


jasonLadd (commented):
reviewed

PR #769, opened 2026-08-07
docs(1.6): the Authorship plan — author the design, prove it was won

jasonLadd:
Opens the 1.6 line off **SIP-0103** (accepted 2026-08-07, PR #768) and the Stack Blueprint SIP's own stated sequencing. Structure follows the 1.5 plan that worked: work classes, gates, protective guards, evidence matrix.

**The claim:** the squad authors the interface design from the PRD, and the release proves it was won rather than asserted.

## Four decisions that were not obvious going in

**1. M0 comes first — contract derivation, proven against the known-good pair.**

SIP-0103 §5b Correction 1 found that **contract v9 is hand-authored**; no `derive_contract(manifest)` exists anywhere in the pipeline. So authoring the manifest alone would leave the contract as the remaining hand-wired seed, and bind mode without a derived contract is just author mode wearing a costume.

> **M0 exit criterion:** `derive_contract(manifest v4)` reproduces contract v9 — byte-equivalent, or with every diff named and justified in a committed record. A diff that cannot be explained is a derivation defect, not a contract improvement.

*Rails before mechanism*, a third time (SIP-0101 Slice 1; SIP-0096's inert Phase 1 core). If derivation cannot reproduce a contract a human wrote for a manifest a human wrote, we learn that against a fixed target instead of inside authoring chaos.

**2. Track S does not open with SIP acceptance.**

The Stack Blueprint SIP argues its own proposed fields are FastAPI-shaped thinking and sets its acceptance gate at the existence of a second real stack. So the lane opens with **S1**, the no-SIP consolidation refactor (exact test: `expand()` byte-identical, contract hashes unmoved, emission gates 6/6); **S2** is the stack-#2 decision (now settled — see below); **S3** promotes the SIP mid-release with the schema written against two stacks.

**3. Two testable guards replace 1.5's "feature-free" device.**

A feature release needs different protection than a stabilization one.

- **Guard 1** — authored mode is a mode of manifest *provenance*, not a second pipeline. Verified by an architecture test that no execution path branches on authoring mode below the framing workload (the `test_plan_gate_seams.py` precedent).
- **Guard 2** — the measurement cannot be tuned by the thing it measures. The FAY window's N, PRD, deploy hash, and scoring discipline are committed **before roll 1**.

**4. Rider creep is called out at plan time.**

The ROADMAP's 1.6 row lists five riders against two headlines — more than any prior release has held. Only the two that serve the claim (SIP-0093 completion, SIP-0102 steps 3–4) are classified **enabling**; SIP-0091 and SIP-0090 Phase 2 are platform work unrelated to authorship and move to **capacity**. Recorded explicitly as a *scope reading, not a scope change* — promoting either back to a commitment is the owner's call with a ROADMAP language update.

## Stack #2 settled: Node/TypeScript *(owner ruling 2026-08-07)*

An Express-or-Fastify API with a typed React frontend. Selection criterion was cost-per-assumption-broken — stack #2 must differ from `fullstack_fastapi_react` along the axes the blueprint SIP names as FastAPI-shaped, but 1.6 already carries two headlines and a measurement window, so the break must not drag the verification machinery into the experiment alongside the schema.

| Assumption | How a TS stack breaks it |
|---|---|
| `analysable_suffix: str` (singular) | `.ts` + `.tsx` |
| `harness_entry_modules` (Python-style import boundary between tests and app) | Node module resolution has no equivalent |
| `qa_test_namespace` (test ownership as directory prefixes) | co-located `*.test.ts` is the convention, not an option |

**HTTP is held constant deliberately** — endpoints, probes, and the derived contract keep working, so the *blueprint schema* is under test rather than the verification stack (which M0 is already changing this release). Cost is low because the toolchain exists: `agents/Dockerfile` carries Node/npm for the frontend build check and vitest (#306, shipped 1.3.1), and the sandbox's `EnvironmentContract` is already parameterized by image plus required tools.

**Rejected, recorded so they aren't relitigated:** **Go** breaks more (single-binary packaging, in-package tests) but needs a new toolchain in the agent image, sandbox image, test runner, and the `command_exit_zero` allowlist #707 already calls untrustworthy — reconsider as **stack #3**, where its packaging break is the point. A **non-HTTP Python stack** breaks the deepest assumption (slots derivable from routes) but puts the probe machinery under test simultaneously with the schema.

**The latent break this forces open:** `analysable_suffix: str` is already wrong today — `.py` backend plus `.jsx` frontend, with typed checks hardcoded to `.py`, which is exactly why #668's `.jsx` half keeps deferring. Stack #2 doesn't create the mismatch; it makes the schema answer for one that already exists.

Two consequences were added to the plan rather than left implicit: a risk row recording that holding HTTP constant leaves the slots-derive-from-routes assumption untested this release, and a **disclosure obligation on S4** — with a TS stack in the tree, "typed checks hardcoded to `.py`" becomes a live gap where `fill_slot_signature` and `undefined_names` would silently skip the second stack. S4 stays capacity, but if S3 lands without it the plan must state that stack #2 ships with a narrower enforced surface than stack #1.

## Also carries B1 — the intention 1.6 owes 1.8

Record the **pre-memory rejection-class recurrence baseline** during the authored-mode window. Classified release-defining because it is the only item on the list whose omission is *permanent*: once memory exists the baseline is unrecoverable and Cross-Cycle Memory's value claim becomes unmeasurable. Read by nothing in 1.6; cheap while building authored mode; impossible afterward.

## Cut gate

Core-claim (M0–M5, S1–S3, B1) · capacity roll to 1.7 · full regression + both guards verified · **authored-mode FAY repeatably > 0**, banked as the baseline 1.8 measures against · confirmation shakedown green.

Non-gating diagnostics are enumerated (structural diff vs the human reference, revision counts, gate-rejection taxonomy). **"Maintainability" and "elegance" metrics are declined** — no deterministic representation exists, and an LLM-graded elegance score is the evidence-quality laundering A6 forbids.

## Open decisions this plan deliberately does not make

- **Authoring decomposition** — single merger-authored vs multi-role proposers (SIP-0103 §6 Q1); §5a recommends single author at design time with multi-reviewer, resolved at the Gate-1 design review.
- **Manifest v5 migration posture** — old manifests migrated or explicitly rejected; decide at Gate 1 and record.

### DIFF
--- a/docs/plans/1-6-0-authorship-plan.md
+++ b/docs/plans/1-6-0-authorship-plan.md
@@ -1,0 +1,324 @@
+# 1.6.0 Authorship Plan — Author the Design, Prove It Was Won
+
+**Established:** 2026-08-07 · **Opens:** now (v1.5.0 tagged 2026-08-06; SIP-0103 accepted
+2026-08-07). Successor to the 1.5 line's discipline (`docs/plans/1-5-0-stabilization-plan.md`):
+one PR per issue with `Closes`, premise re-verified against code at build time, targeted
+verification per fix, live validation before merge for behavior-changing work, bump only
+after a green confirmation shakedown. Sequencing, not dates — **substance gates the cut.**
+
+Roadmap position: `docs/plans/post-1-4-roadmap-reconciliation.md` (the 1.6 row, unchanged)
+and `docs/plans/post-1-5-roadmap-reconciliation.md` (1.7/1.8 rows, plus the design
+intention this release owns).
+
+## Character
+
+An **even minor**: a feature release led by headline SIPs, which gate the version.
+Hardening rides along freely.
+
+One coherent claim: **the squad authors the interface design from the PRD, and the
+release proves it was won rather than asserted.**
+
+Two headlines, one per lane, per the dual-lane precedent set by 1.4:
+
+| Lane | Headline | SIP |
+|---|---|---|
+| **M** (executor / handlers / framing surfaces) | **Squad-Authored Manifest** | **SIP-0103**, accepted 2026-08-07 with §5a/§5b/§5c in force |
+| **S** (test-runner / build-check / agent-image / deploy-infra) | **Generalized Build Capability** | `SIP-Stack-Blueprint-Contract` — **deliberately still proposed**; promoted *mid-release* (see Track S) |
+
+### The two protections this release needs
+
+1.5's protective device was "feature-free, defined behaviorally." A feature release needs
+different guards. Two, both testable:
+
+> **Guard 1 — authored mode is a mode of manifest *provenance*, not a second pipeline.**
+> Post-approval, an authored manifest enters `plan_artifact_refs` exactly as a seeded one
+> does; expander, contract derivation, bind-mode plan validation, and every 1.4/1.5
+> enforcement surface operate identically. **Verification at cut:** no execution path
+> branches on authoring mode below the framing workload — pinned by an architecture test,
+> the `test_plan_gate_seams.py` precedent.
+
+> **Guard 2 — the measurement cannot be tuned by the thing it measures.** The authored-mode
+> FAY window is pre-registered (N, PRD, deploy hash, scoring discipline) before the first
+> roll, unfiltered, on a frozen deploy. The hand-authored reference manifest is excluded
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #771, opened 2026-08-07
docs(1.6): M0 premise correction — the contract deriver already exists, and v9 is its output

jasonLadd:
> ## ⚠️ This plan now contradicts accepted SIP-0103, deliberately
>
> **SIP-0103 §5b Correction 1 states:** *"contract v9 is a hand-authored artifact … no `derive_contract(manifest)` exists anywhere in the pipeline."*
>
> **Both halves are false.** Measured before writing any code.
>
> The SIP is **not** being edited — it is an accepted design commitment on main, not a living document. The correction lives in the plan, where the work is planned, and is flagged prominently in both this PR and the plan header so nobody reconciles them silently later.

## The measurement

Ran against the real artifacts, not a fixture:

```python
emit_contract_dict(InterfaceManifest.from_yaml(art_8becd104e9fc)) == yaml(art_4f368ea08799)
# True — exact dict equality, all six sections, 4 fill_files, 5 probes
```

The deriver is `squadops.capabilities.scaffold_contract.emit_contract_dict`, shipped as **SIP-0098 phase 98.2**, whose own docstring says the criteria are *"derived deterministically from the same interface manifest the skeleton is expanded from, so verification is a fixed property of the scaffold rather than a per-roll LLM lottery."*

**Contract v9 is a pinned emission of manifest v4.** M0's exit criterion — reproduce v9 from v4, byte-equivalent or every diff classified — was met with **zero diffs, before any work started.**

## What's actually missing: wiring, not intelligence

Bind mode loads the pinned contract *by design*. `handlers/cycle/base.py` states it outright: *"bind mode loads the pinned contract from `contract_ref` instead of regenerating it, so a criterion added to the expander would not appear until a re-seed."* Correct for seeded mode; impossible for authored mode, where no pinned artifact exists yet because the manifest was only just written.

M0 collapses to two small deterministic pieces:

- **M0a — standing equivalence guard.** A test over committed fixtures asserting `emit_contract_dict(v4) == v9`, so the deriver can never silently drift from the contract the 1.4 evidence was measured against.
- **M0b — derive-at-seed.** Manifest arrives without a `contract_ref` → derive and pin at seed time, giving authored mode the same frozen-by-hash guarantee seeded mode has. No LLM, no new intelligence, testable offline.

## Second finding: a live latent defect, with a reproduction

The `success_status` rider was a schema-tidiness argument. It's now a defect.

Declared on **1 of 5** endpoints in v4. Where absent, two components default **differently**:

| Component | Behavior when `success_status` is undeclared |
|---|---|
| deriver (`scaffold_contract.py:267`, `:327`) | asserts a status — `or 201` for POST-to-collection, `or 200` for child routes |
| scaffold (`scaffold.py:901`) | omits `status_code=` from the decorator → **FastAPI's own default, 200** |

A POST-to-collection with no declared status produces a contract asserting **201** against a scaffold emitting **200** — an unwinnable contract, authored innocently. Masked today *only* because the single endpoint reaching that branch (`POST /runs`) happens to declare 201. Nothing enforces that coincidence.

Two consequences, both in the plan: it is the **first designed-failure probe for the M3 winnability gate**, available immediately; and it is a **seeded-mode defect too**, not merely an authored-mode hazard. Filing decision pending with the owner — recorded in the plan either way so it can't be lost.

## Also threaded through

- Divergence taxonomy **retained**, not deleted as spent — it now governs M0b and any future deriver change. `v9_defect` → `reference_defect`.
- Risks table: the "derivation can't reproduce v9" row is **retired** and replaced with the real risk the finding exposes — *silent deriver drift* — which M0a covers.
- Gate 1's exit swaps the banked proof for M0a + M0b.
- Class table: M0a/M0b enabling; the derivation proof is noted as already banked.

## One addition beyond the correction — flag for the owner

Track M gains an explicit build order: **gates before author** — `M0a → M0b → M3 → M2 → M6 → M1 → M4 → M5`.

This was **not explicitly requested**; I included it because the original M1-first ordering was a consequence of the wrong M0 premise, and leaving it would have shipped an ordering the finding invalidated. The gates are deterministic and testable against hand-made adversarial manifests today, with no LLM and no deploy. M1 is the only piece whose output is a lottery, so landing it last means a failed authored manifest is attributable to the *author* rather than ambiguous between author and unproven gate. Strike it if you disagree — it's one section.

---

## Added after review: the queue front, and the defaults defect filed

**No 1.5.1.** Asked whether a patch line should precede 1.6 feature work; the population since the cut is five items (#761, #762, #766, #770, #772), none urgent, three of them papercuts — below the bar the 1.4 patch lines set at 5–7 real fixes each. 1.6 also opens with pure offline work (M0a is a fixture test, M0b is seed-time wiring), so a freshly-verified deployed baseline buys nothing yet. The trigger that would flip that ruling is recorded in the plan.

**Instead, three fixes go to the front of 1.6's queue**, ahead of M0a — because 1.6's output is measured in cycle rolls and each of these taxes every roll:

| Item | Cost per roll today |
|---|---|
| **#762** bind mode without `plan_authoring_contributors` → guaranteed plan rejection | cost **three rolls at shk-6** before diagnosis; 1.6 runs an *unfiltered* window where a misconfigured roll is expensive in a way it isn't during development |
| **#766** LangFuse linkage inert + vendor `ERROR` per agent per cycle | degrades trace triage — the instrument this release leans on hardest |
| **#770** `update_sip_status.py` opaque failure | blocks promotion at the worst moment, and 1.6 promotes Stack Blueprint mid-release (S3) |

#762 moves out of the SIP-0093-completion enabling row into the queue front so its preflight block starts paying immediately rather than waiting on the rest of 93's completion.

Build order is now: **queue front → M0a → M0b → M3 → M2 → M6 → M1 → M4 → M5**.

**The defaults disagreement is filed as #772**, homed to 1.6, capacity-bound because it rides M0's schema change. The preferred resolution is recorded as **fix by deletion**: making `success_status` required makes `or 201` and `or 200` both unreachable, so the REST-semantics question — is the deriver presumptuous, or the scaffold under-specified? — is dissolved rather than answered in favor of one side. #772 states explicitly that if the required-field change ever leaves scope, that question comes back live and needs an owner ruling.

### DIFF
--- a/docs/plans/1-6-0-authorship-plan.md
+++ b/docs/plans/1-6-0-authorship-plan.md
@@ -21,6 +21,15 @@ governance**. The review's sharpest finding was that S1 and M0 had proof strateg
 S3 had only a subjective one; the second sharpest was that three separate asks (revision
 reason, memory baseline, window attribution) are one artifact used three ways.
 
+**Revised again 2026-08-07 — M0's premise was wrong.** A pre-build verification pass
+measured the actual artifacts and found that the contract deriver already exists and that
+contract v9 is its exact output for manifest v4. **This plan therefore contradicts accepted
+SIP-0103 §5b Correction 1 on measured evidence**; the SIP stands as the accepted design
+commitment and the correction lives here. Consequences: M0 collapses from "build a deriver"
+to a guard plus wiring; Track M gains an explicit **gates-before-author** build order; and
+the `success_status` rider is upgraded from a schema-tidiness argument to a live latent
+defect with a reproduction.
+
 ## Character
 
 An **even minor**: a feature release led by headline SIPs, which gate the version.
@@ -81,7 +90,7 @@ replay/regression referent — not a legacy path to be retired.
 | Class | Contents |
 |---|---|
 | Release-defining | M1–M6 (SIP-0103's phases + the failure taxonomy) · S1–S3 + S5 (Generalized Build minimum + blueprint governance) · the authored-mode FAY window · **B1** (the memory baseline — see "Owed to 1.8") |
-| Enabling | M0 (contract derivation proof) · SIP-0093 completion (#762, #194 — the authoring pattern SIP-0103 extends) · SIP-0102 steps 3–4 (clean-room verdicts; #376) |
+| Enabling | M0a/M0b (equivalence guard + derive-at-seed; the derivation *proof* is already banked) · SIP-0093 completion (#762, #194 — the authoring pattern SIP-0103 extends) · SIP-0102 steps 3–4 (clean-room verdicts; #376) |
 | Capacity-bound | #668 (both halves) · #761 (A4.1) · #598 structural half · SIP-0092 M3 · #733 Slice B · SIP-0091 · SIP-0090 Phase 2 · SIP-0102 steps 5–7 · agent-comms delivery guarantees · ops riders |
 
 **Scope warning, recorded at plan time.** The ROADMAP's 1.6 row lists five riders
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #773, opened 2026-08-07
fix(preflight): reject bind mode with no plan_authoring_contributors at create (#762)

jasonLadd:
First item of the v1.6 queue front. Design gate (D1–D6) posted to the issue before any code.

Closes #762

## The premise, re-verified against `main` rather than taken from the shakedown write-up

- `bind_criteria_index=True` is set on **exactly two** contracts in `capabilities/context_assembly.py`: `development.propose_plan_tasks` and `qa.propose_plan_tasks`. `governance.merge_plan` is **deliberately** excluded, and the registry says why.
- `build_planning_steps()` adds proposer steps only for roles in `plan_authoring_contributors`; an empty list yields *no proposers* and the merger runs sole-author.
- So nothing in framing ever receives the criteria index → no `criteria_refs` can be authored → bind-mode validation rejects, deterministically, after a full framing workload is spent. shk-6 rolls 1–3 burned three.

## The check

Blocks iff **all three** hold on the resolved config (#724's single merge, which is what dispatch honors):

1. `contract_ref` present → bind mode;
2. a planning workload will actually run;
3. `plan_authoring_contributors` empty or absent.

**Condition 2 is load-bearing, not defensive.** An implementation-only bind-mode cycle is legitimate and *used* — the SIP-0101 replay demonstration (`cyc_cfe6962e8fc8`) skipped framing via a `workload_sequence` override to resume from a checkpoint. Blocking on `contract_ref` alone would have rejected it. `_authors_a_plan()` mirrors `_required_roles_by_workload()`'s reading of the same surface, legacy `plan_tasks` fallback included, so the two can't disagree about whether planning happens.

| Configuration | Result |
|---|---|
| bind + framing + no contributors — **the shk-6 rolls 1–3 shape** | **block** |
| bind + framing + contributors — the shk-7 green shape | allow |
| bind + implementation-only — **the replay shape** | allow |
| author mode, sole-author — the common case | allow |
| legacy: bind, no `workload_sequence`, `plan_tasks` default | **block** |
| legacy: bind + `plan_tasks: false` | allow |
| bind + framing + explicitly empty list | **block** |

**Blocks rather than warns**, per the module's own rule: every input is knowable in the resolved config at create time, and the downstream rejection is deterministic — a warning would spend a framing workload to reach a certain failure.

## The message names the remedy

Why it's bind mode, what's missing, every valid role, and both exits (add contributors, or drop `contract_ref` for author mode). A message that says only "invalid configuration" rebuilds the diagnosis cost this check removes — pinned by a test.

## Rode along: one vocabulary, not two literals

`VALID_PLAN_AUTHORING_CONTRIBUTORS` moves from a private constant in `task_plan` to `cycles/models.py`, beside `WORKLOAD_REQUIRED_ROLES` — whose comment already states exactly this pattern: one source of truth shared by dispatch-time step resolution and the create-time preflight, "so the two never drift." A duplicated literal is how a role accepted at create becomes a `CycleError` at dispatch. An identity test pins it.

## Deliberately *not* changed

**`validated-fullstack.yaml`.** The shakedown write-up reads as "the CRP lacks `plan_authoring_contributors`," which invites adding it there. That would be wrong: bind mode comes from `execution_overrides.contract_ref` at launch, **not** from the CRP, so `validated-fullstack` is a perfectly valid *author-mode* profile whose users would silently inherit three extra proposer tasks and their cost.

**Option 2** (thread the criteria index into the sole-author merge path) is foreclosed with a named trigger — `merge_plan`'s exclusion is a documented choice, so reversing it is a design change, not a bug fix. No configuration wants sole-author bind mode today.

## Tests

Seven-case truth table (including both false-positive shapes that must **not** block and both legacy paths), remedy-content assertions, the single-sourcing identity, and a composition test — because an uncomposed check is one that never runs.

**Regression: 6497 passed, 1 skipped.**

### DIFF
--- a/src/squadops/api/routes/cycles/cycles.py
+++ b/src/squadops/api/routes/cycles/cycles.py
@@ -35,6 +35,7 @@ from squadops.cycles.models import (
 from squadops.cycles.preflight import (
     Finding,
     PreflightDecision,
+    bind_mode_authoring_decision,
     combine,
     model_availability_decision,
     required_check_tooling_decision,
@@ -125,6 +126,10 @@ async def _run_create_preflight(profile: SquadProfile, config: dict) -> tuple[Fi
     """
     decision = combine(
         required_roles_decision(profile, config),
+        # #762: bind mode with no plan_authoring_contributors is unwinnable by
+        # construction — the sole-author path never receives the criteria index,
+        # so every framing attempt is rejected. Fail in seconds, not per re-roll.
+        bind_mode_authoring_decision(config),
         model_availability_decision(profile, await _pulled_model_names()),
         # SIP-0096 §6.5: a required check whose tooling is knowably absent is a
         # create-time reject, never a mid-run blocked_unverified surprise.
--- a/src/squadops/cycles/models.py
+++ b/src/squadops/cycles/models.py
@@ -210,6 +210,13 @@ WORKLOAD_REQUIRED_ROLES: dict[str, frozenset[str]] = {
     WorkloadType.IMPLEMENTATION: frozenset(),
 }
 
+# Roles that may appear in ``plan_authoring_contributors`` (SIP-0093 Rev 1). Homed
+# here for the same reason as WORKLOAD_REQUIRED_ROLES above: two readers key off it —
+# dispatch-time sequence construction (task_plan.build_planning_steps) and the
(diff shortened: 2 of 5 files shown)


jasonLadd (commented):
reviewed

PR #774, opened 2026-08-07
fix(telemetry): only attempt Langfuse prompt linkage when assets come from Langfuse (#766)

jasonLadd:
Second item of the v1.6 queue front. Design gate (D1–D6) posted to the issue before code.

Closes #766

## What was happening

`_handle_generation` passed the **filesystem** asset's front-matter version into **Langfuse's** prompt registry — two registries with independent version namespaces. Under `SQUADOPS__PROMPTS__ASSET_SOURCE_PROVIDER=filesystem` (the schema default *and* the deployed value) the lookup cannot succeed for any asset at any version, so it produced a vendor-logged `ERROR` per agent per cycle while the linkage it exists to create was never established.

## Two defects, kept separate

**The noise** is fixed by not making a call that cannot succeed. The `ERROR` originates in the Langfuse SDK's own logger *before* our `except` sees anything — catching harder or downgrading our own log would not have touched it. **Muting the vendor logger is explicitly not the fix**: that hides the inert-capability defect behind the noise defect.

**The inert capability is not built here — it is declared.** One `INFO` line at construction stating linkage is disabled and why. This is the `DECLARED_UNBUILT_CHECKS` pattern (#730/D4) one level down: a gap that is named is governable, a gap that is silent is a false capability. The trigger for actually building linkage (publishing filesystem assets into Langfuse's registry) is recorded on the issue — the concrete candidate is v1.8's scorecard wanting per-prompt-version attribution.

## Ownership

The adapter is **told** the provider at the composition root; it does not read config itself and does not carry a copy of the setting. `PromptsConfig.asset_source_provider` (SIP-0084) owns that fact and keeps owning it. Both construction sites (`api/runtime/main.py`, `agents/entrypoint.py`) already held the full `AppConfig`.

**Rejected:** a field on `LangFuseConfig` — that's observability config, not prompt-source config, and the same fact living in two config surfaces is exactly the drift class.

## The gate is "can this ever succeed", not "did it error"

A 404 under the `langfuse` provider is legitimate and informative — the prompt genuinely isn't registered at that version — so the existing best-effort catch is untouched.

## Tests

- filesystem path makes **zero** `get_prompt` calls with `prompt_name` *and* `prompt_version` both present — the exact shape that produced `Prompt not found: 'request.planning_task_base' with version 3` on every authoring agent during shk-7;
- the generation is **still recorded** without a `prompt` kwarg, so skipping linkage never costs an observation;
- the disclosure fires **once at construction**, not per generation — otherwise we'd trade an ERROR per cycle for an INFO per cycle and fix nothing;
- no disclosure when linkage is enabled.

**Two existing tests now declare `prompt_asset_provider="langfuse"`.** That is not test-weakening — both assert behavior that only applies where resolution is *attempted* (does linkage fire when it should; does a resolution failure still record the generation), and their assertions are unchanged. The harness default is `filesystem`, matching `PromptsConfig` — the deployed path, not the convenient one.

**Regression: 6501 passed, 1 skipped.**

### DIFF
--- a/adapters/telemetry/factory.py
+++ b/adapters/telemetry/factory.py
@@ -135,6 +135,7 @@ def create_llm_observability_provider(
     provider: str = "langfuse",
     config: LangFuseConfig | None = None,
     secret_manager: SecretManager | None = None,
+    prompt_asset_provider: str = "filesystem",
 ) -> LLMObservabilityPort:
     """Create LLM observability provider (SIP-0061).
 
@@ -146,6 +147,10 @@ def create_llm_observability_provider(
         provider: Provider name (currently only "langfuse")
         config: LangFuseConfig from AppConfig
         secret_manager: For resolving secret:// references in config
+        prompt_asset_provider: ``PromptsConfig.asset_source_provider`` (SIP-0084).
+            Threaded through so the adapter is told where prompt assets come from
+            instead of reading config itself or carrying a duplicate setting (#766);
+            prompt→generation linkage is only attempted when it is ``langfuse``.
 
     Returns:
         LLMObservabilityPort implementation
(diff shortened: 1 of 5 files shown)


jasonLadd (commented):
reviewed

Issue #772, opened 2026-08-07
contract/scaffold disagree on the default success status — a POST-to-collection with no declared success_status yields an unwinnable contract


jasonLadd:
## The defect

When an interface manifest omits `success_status` on an endpoint, the contract deriver and the scaffold expander **default differently**:

| Component | Behavior when `success_status` is undeclared |
|---|---|
| deriver — `capabilities/scaffold_contract.py:267` | `"expect": {"status": ep.success_status or 201}` for POST-to-collection |
| deriver — `capabilities/scaffold_contract.py:327` | `"expect": {"status": child.success_status or 200}` for child routes |
| scaffold — `capabilities/scaffold.py:901` | `if ep.success_status is not None:` → the `status_code=` kwarg is **omitted entirely**, so FastAPI's own default (200) applies |

For a **POST-to-collection with no declared `success_status`**, the derived contract asserts **201** while the scaffold emits a route that returns **200**. The probe fails on a correctly-implemented app: the contract is unwinnable, and it was authored innocently — nothing in the manifest looks wrong.

## Why it is not firing today

Measured on the live pair (2026-08-07), manifest v4 `art_8becd104e9fc`:

```
GET   /runs                   declared=None
POST  /runs                   declared=201     <-- the only endpoint reaching the `or 201` branch
GET   /runs/{run_id}          declared=None
POST  /runs/{run_id}/join     declared=None    <-- child route, both sides default 200, agree
POST  /runs/{run_id}/leave    declared=None    <-- child route, both sides default 200, agree
```

`success_status` is declared on **1 of 5** endpoints — and that one endpoint happens to be the single case that reaches the `or 201` branch. The disagreement is masked entirely by that coincidence. Nothing enforces it: a new manifest that adds a second collection-level POST, or drops the explicit 201 from this one, hits the trap immediately.

## Why this is filed rather than left in the plan

It is a **seeded-mode defect**, not only an authored-mode hazard. Any hand-authored manifest — including the ones written for future FAY windows or a second stack — can hit it. The v1.6 plan records it as context, but a plan is where work is planned, not where defects are tracked.

## The fix, and the judgment it removes

There is a real semantics question buried here — *is the deriver presumptuous (201 is invented) or is the scaffold under-specified (201 is correct REST and the route should declare it)?* Those are opposite fixes, and picking one bakes a convention into every future stack blueprint.

[...]

PR #775, opened 2026-08-07
docs(1.6): verification cadence — V1–V8, and the two corrections it exposed

jasonLadd:
The plan named gates but left *when a cycle is actually worth running* to be improvised. Asking the question surfaced two things the plan had **wrong**, so this is a correction as much as an addition.

## The cadence

Most of 1.6 is offline and deterministic — M0a is a fixture test, M0b is seed-time wiring, M3/M2/M6 are gates provable against hand-made manifests. Running cycles through that stretch burns budget to learn nothing; the first work that *requires* a live cycle is M1. Cycles need a deployed image, so this is also the deploy cadence.

| # | Trigger | What runs | Kind |
|---|---|---|---|
| **V1** | queue front deployed | two create probes (#762 must 422; the replay shape must still pass) + a boot check for #766's disclosure | diagnostic |
| **V2** | **after M0b, before M1** | Guard 1b + **a full seeded-mode control cycle** | diagnostic — **banks the control** |
| **V3** | M3/M2/M6 landing | **no cycles** — adversarial hand-made manifests, #772's trap as the designed-failure probe | — |
| **V4** | M1 lands | first authored cycles, unscored, several expected | diagnostic |
| **V5** | M4/M5 land | integration shakedown — HITL gate, provenance, B1 emitting | diagnostic |
| **V6** | before pre-registration | authored-mode viability run | diagnostic; **gates V7** |
| **V7** | frozen deploy | **the FAY window** — pre-registered, unfiltered | **evidentiary** |
| **V8** | at the cut | confirmation shakedown + **the seeded control re-run**, compared against V2 | diagnostic |

## Correction 1 — the seeded control moves earlier

The plan put a *single* seeded shakedown at the cut. That is a **claim, not a control**: with nothing to compare against, a degraded number is indistinguishable from a number that was always that.

The seeded control cycle now runs at **V2**, immediately after M0b — the last moment the pipeline is fully deterministic and a red is unambiguously *ours* rather than ambiguous between our change and model variance — and is **re-run at V8**. The comparison is the control; either observation alone is not.

This matters specifically because **M0b changes which contract a cycle runs against, on the path seeded mode uses too.**

## Correction 2 — a pre-registration viability run (V6)

The plan went from integration straight to the pre-registered window. But pre-registering a window for a capability that has **never once succeeded** wastes it — the 1.4 FAY bar was set at ≥4/6 because there was reason to believe it was achievable. V6 answers *do we have reason to believe?*, is explicitly **not part of the window**, and its rolls are **not counted** — recorded so a later reader can't mistake it for a filtered first attempt.

## Two things stated because they collapse under pressure

> **A shakedown is diagnostic and you fix what it finds. A window is evidence and you fix nothing until it closes.**

V4–V6 are where reacting to what you see is the point. V7 is where reacting invalidates the measurement.

And **V4's failures are the product, not a setback.** The first authored cycles failing at the winnability gate is the system working — the gate doing in seconds what would otherwise cost a full implementation. What to watch at V4 is not the pass rate but whether **M6's taxonomy can name which subsystem to blame**. If every V4 failure lands in one undifferentiated bucket, M6 isn't finished and V7 will produce a number nobody can act on.

Each gate now names the verification that closes it.

### DIFF
--- a/docs/plans/1-6-0-authorship-plan.md
+++ b/docs/plans/1-6-0-authorship-plan.md
@@ -30,6 +30,13 @@ to a guard plus wiring; Track M gains an explicit **gates-before-author** build 
 the `success_status` rider is upgraded from a schema-tidiness argument to a live latent
 defect with a reproduction.
 
+**Revised 2026-08-07 — verification cadence added.** The plan named gates but left *when a
+cycle is worth running* to be improvised. V1–V8 below make it explicit, and two of them are
+corrections rather than additions: the seeded-mode control observation **moves earlier**
+(one shakedown at the cut is a claim, not a control — it needs a clean baseline to compare
+against), and a **pre-registration viability run** is added, because pre-registering a
+window for a capability that has never once succeeded wastes the window.
+
 ## Character
 
 An **even minor**: a feature release led by headline SIPs, which gate the version.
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #776, opened 2026-08-07
fix(maintainer): derive SIP frontmatter from the body instead of failing (#770)

jasonLadd:
Last item of the v1.6 queue front. Design gate (D1–D6) posted to the issue before code.

Closes #770

## Two findings from re-verifying the premise

1. **`status` is the only genuinely required field.** Downstream, the promotion path uses `title`/`sip_uid`/`author`/`created_at` — and all four already fall back (`"Untitled"`, `None`, `"Unknown"`, `now()`). Only the missing *block* was fatal, not the missing fields.
2. **The duplicate scan silently skipped frontmatter-less files** (`if not metadata: continue`), so those same 12 drafts were invisible to duplicate detection — a quieter second consequence the issue didn't name. It now warns: a file we cannot read is a file we cannot check, and a silent skip makes the net look complete when it has a blind spot exactly the size of the un-frontmattered drafts.

## The fix reads what the author already wrote

Title from the H1 (`SIP-0XXX:` / `SIP:` / `IDEA —` prefixes stripped), status from either declaration shape with the parenthetical dropped (`Proposed (stub)` → `proposed`, `Draft` → `proposed`), author and `created_at` from the body when stated and **omitted when not** — a guessed value is worse than an absent one because it looks authoritative.

**Rejected as the primary fix:** a better error message that still requires a four-line paste. It would fail 11 more times by construction.

## Also closes the #253 class, one form down

`_BODY_STATUS_RE` matched `**Status:** X` but not a `## Status` heading with the value on a following line — the shape **SIP-0103 had**, so its body would have kept saying `Draft (proposed)` while frontmatter and registry said `accepted`. I hand-fixed that during the acceptance; nothing stopped it recurring.

The new pattern is deliberately conservative: the value line must **begin** with a status keyword, optionally emphasised. The existing prose-after-heading no-op test still returns `False` and stands unchanged as the guard.

## The backfill is the test corpus

The same deriver ran once over the 11 `SIP-*.md` drafts. That reuse is the point, and it earned its keep immediately: the emphasised-value shape `**Proposed** (draft, 2026-08-07)` **failed the first implementation** and would never have been caught by a fixture written to match my own regex.

The `IDEA-` file is excluded (D6) — it isn't a SIP, and an `IDEA-` file in `sips/proposed/` is a naming/placement question rather than something to quietly stamp frontmatter onto. Raised separately, not moved here.

## Tests

Derivation across all three real H1 shapes and both status shapes including both parenthetical forms; author/created omitted rather than invented; underivable status fails with an actionable message; end-to-end `ensure_frontmatter` round-trip through the real extractor; existing frontmatter never overwritten; both heading-form rewrites (plain and emphasis-preserving).

**Regression: 6501 passed, 1 skipped.**

## One note on how this was verified

Smoke-testing the promotion path, I ran the real script against `SIP-Stack-Blueprint-Contract.md` and it promoted it to SIP-0104 — the one SIP we deliberately decided *not* to accept until a second stack exists. Fully reverted (registry back to `last_assigned: 103`, no 0104 entry, file back in `sips/proposed/` with its backfilled frontmatter, verified clean). Re-smoked against a throwaway fixture in a temp directory instead. Worth knowing that this script does the whole promotion in one shot with no dry-run flag — arguably a follow-up.

### DIFF
--- a/scripts/maintainer/update_sip_status.py
+++ b/scripts/maintainer/update_sip_status.py
@@ -149,6 +149,23 @@ _BODY_STATUS_RE = re.compile(
     re.IGNORECASE | re.MULTILINE,
 )
 
+# The OTHER declaration shape drafts actually use: a "## Status" heading with the
+# value on a following line ("## Status\nDraft (proposed)"). #253 closed the
+# inline form; this is the same bug class one form down — SIP-0103 had exactly
+# this shape, so its body would have kept saying "Draft" while the frontmatter and
+# registry said "accepted" (#770).
+#
+# Deliberately conservative: the value line must BEGIN with a status keyword,
+# optionally emphasised ("**Proposed** (draft, ...)" is how one real draft writes
+# it). A heading followed by prose ("## Status\n\nSome explanation") still matches
+# nothing, so the caller still warns rather than rewriting arbitrary text. The
+# emphasis is captured into `heading` so a rewrite preserves it.
+_BODY_STATUS_HEADING_RE = re.compile(
+    r"^(?P<heading>#{1,6}[ \t]+[^\n]*Status[^\n]*\n(?:[ \t]*\n)*\*{0,2})"
+    r"(?P<word>" + "|".join(_STATUS_WORDS) + r")",
+    re.IGNORECASE | re.MULTILINE,
+)
+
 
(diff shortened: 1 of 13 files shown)


jasonLadd (commented):
reviewed

PR #778, opened 2026-08-08
feat(contract): pin the deriver to the deployed reference pair (#777, M0a)

jasonLadd:
First code item of the v1.6 M lane.

Closes #777

## What M0a turned out to be

The plan opened believing derivation had to be **built**. A premise pass found the deriver already exists — `scaffold_contract.emit_contract_dict`, SIP-0098 phase 98.2 — and that contract v9 is exactly its output for manifest v4. So M0a isn't construction; it's making that equality impossible to lose quietly, because three things now rest on it: **M0b**'s derive-at-seed, **M3**'s winnability dry-run, and **Guard 1b**'s byte-identical requirement.

## The repo already had half of it

`examples/03_group_run/interface_manifest.yaml` is **byte-identical** to the ingested manifest v4 (`art_8becd104e9fc`), so no duplicate copy was added — the existing reference instance is used. The ingested contract v9 (`art_4f368ea08799`) is committed beside it as **the vault's bytes**, not a regenerated copy: regenerating locally would make the equality test a tautology, the deriver proving it equals itself.

## Both ends pinned, because either alone is defeatable

| Guard | Stops |
|---|---|
| manifest `sha256` | the deriver "matching" a reference that has drifted |
| contract `sha256` | the committed contract being quietly regenerated from the deriver under test |
| byte equality | serialization drift — the emitted YAML *is* the ingested artifact |
| per-section comparison | a failure that says "these files differ" instead of naming the section |

**No regeneration escape hatch, deliberately.** This is not a golden tracking code evolution; it's a reference tied to banked evidence — the 1.4 FAY window (6/6, five consecutive) was measured against this contract. A one-env-var regen invites exactly the casual update the plan's M0 divergence taxonomy exists to prevent.

## Tripwire verified, not assumed

| Mutation | Result |
|---|---|
| `CONTRACT_VERSION` 1 → 2 in the deriver | 2 failed; semantic test names `'contract_version'` |
| tamper the committed contract | 2 failed |
| restored | 5 passed |

## One finding worth your attention

**The first mutation I tried was a no-op by construction.** Changing `ep.success_status or 201` → `or 202` changed nothing, because the only endpoint reaching that branch (`POST /runs`) declares 201 explicitly.

That's **#772 in action**, and it widens that issue: the masked default isn't only a latent unwinnable-contract trap, it's a **test blind spot** — no test using the reference pair can observe a change to that branch. Recorded as evidence on #772, where it strengthens the fix-by-deletion resolution: once `success_status` is required, both defaults are *deleted* rather than left unreachable, and unreachable-but-present code is exactly what looks tested because everything around it is.

**Regression: 6506 passed, 1 skipped.**

---

## Correction after review: the pinned contract moved out of `examples/`

First version of this PR put the contract at `examples/03_group_run/verification_contract.yaml`, beside the manifest it derives from. That was wrong, and checking the objection proved it wrong for a stronger reason than "it's test data":

**`scripts/dev/contract_gate.py` *generates* that file** from the example manifest — it's one of the two CI jobs — and then prints the `artifacts ingest --type verification_contract --file …` command to upload it. The file's own header reads *"Regenerate with the expander; do not hand-edit."*

So the first version committed a **generated artifact into the source tree**, in a directory that reads as user-facing examples. That's the smell CLAUDE.md names, and the clearest evidence it was wrong is that it misled the owner into thinking we'd hard-wired a locked design into the very release meant to eliminate hand-authored designs.

**Now:** `tests/fixtures/reference_contract/contract_v9_art_4f368ea08799.yaml` — named for the artifact it copies, so its provenance is legible from the filename rather than a docstring.

**The manifest stays in `examples/`**, deliberately: it *is* a genuine product input — `contract_gate.py` defaults to it, and it's the file ingested as `art_8becd104e9fc` — so a human should be able to read and copy it. The pair is now split across two directories on purpose, and that costs nothing, because the coupling was never directory adjacency: the manifest's own `sha256` pin is what fails if it drifts from the contract.

Nothing else changed. The committed bytes are still the vault's rather than a regenerated copy, so the guard stays non-tautological. Regression 6506, `examples/` untouched.


jasonLadd (commented):
reviewed

PR #780, opened 2026-08-08
feat(cycles): derive the verification contract when a cycle seeds a manifest but no contract (#779, M0b)

jasonLadd:
Second code item of the v1.6 M lane, and the first that changes runtime behavior.

Closes #779

## The gap

Bind mode is keyed strictly on a seeded `contract_ref` — correct for seeded mode, impossible for authored mode, where a manifest the squad has just written has no pre-existing contract to point at.

Today an operator runs `contract_gate.py`, then `squadops artifacts ingest --type verification_contract`, then passes the returned id. **Three manual steps** to produce something mechanically derivable from the manifest — an equality #777 just pinned as exact.

## What changes

A cycle that seeds a manifest and **no** `contract_ref` now gets the contract derived, stored as an artifact, and `contract_ref` set to the stored id. After that the pipeline is unchanged: bind mode sees an ordinary ref.

**Derived *and persisted***, not derived in memory at run start — a contract with no artifact id cannot be replayed, audited, or hash-frozen. Seeded mode's guarantee is a pinned artifact; authored mode gets the same one rather than a weaker version.

## Three decisions worth the reading

**Never overrides a supplied `contract_ref`.** An operator who passes one is pinning a *specific* contract — possibly deliberately unlike what today's deriver emits, as when replaying an older run — and replay compatibility keys on `contract_ref` (`REPLAY_COMPATIBILITY_ELEMENTS`). Silently replacing it would break replay.

**An unusable manifest rejects at create; it never falls through to author mode.** Seeding a manifest *is* the request for bind mode, and an unbound green carries none of the criteria that were asked for. Same stance `_load_contract_for_run` already takes toward a broken ref.

**Same artifact type as the manual ingest path.** Nothing downstream can tell a derived contract from an ingested one; provenance lives on the artifact record rather than in a divergent type every consumer would have to learn about.

## Rode along: one selection rule, not two

The seeded-manifest predicate (`filename == interface_manifest.yaml` **or** `artifact_type == interface_manifest`) is hoisted out of the executor into the shared module, and the executor delegates to it. Two copies is how the create path and the executor start disagreeing about which ref is the manifest.

## Scope boundary

M0b covers a manifest that exists **at create time**. Authored mode's manifest is produced by framing, so its derivation happens at the framing→implementation boundary — that belongs with **M1**, which reuses this helper rather than growing a second derivation path.

## Caught in regression rather than shipped

The first version called `get_artifact_vault()` unconditionally, which broke eight create tests that wire no vault. Fixed by checking for `plan_artifact_refs` first, so a cycle with nothing seeded never depends on vault wiring at all.

**Regression: 6517 passed, 1 skipped.**

## What follows

Per the plan's verification cadence this is the item **V2** attaches to: Guard 1b plus a full seeded-mode control cycle, because M0b changes what contract a cycle runs against on the path seeded mode uses too. That needs a deploy, so it's the release's first deploy window.

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -3047,17 +3047,16 @@ class DispatchedFlowExecutor(FlowExecutionPort):
         ``_load_interface_manifest_for_run`` reads at the implementation run — so the
         manifest the gate hash-checks against the contract is the same one the
         skeleton will expand from. Unreadable refs are skipped, not fatal: an absent
-        seeded manifest is reported by the caller as the #494/#496 rejection."""
-        for ref_id in cycle.execution_overrides.get("plan_artifact_refs") or []:
-            try:
-                ref, content_bytes = await self._artifact_vault.retrieve(ref_id)
-            except Exception:
-                continue
-            if ref.filename == "interface_manifest.yaml" or (
-                getattr(ref, "artifact_type", None) == "interface_manifest"
-            ):
-                return content_bytes.decode(errors="replace")
-        return None
+        seeded manifest is reported by the caller as the #494/#496 rejection.
+
+        The selection rule itself lives in ``cycles.contract_derivation`` (#779) so the
+        create path, which derives a contract from this same artifact, cannot disagree
+        with the executor about which ref is the manifest."""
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #782, opened 2026-08-08
feat(cycles): winnability gate — prove a manifest can be won before spending on it (#781, M3)

jasonLadd:
Third code item of the v1.6 M lane. Design gate (D1–D5) posted to the issue before code.

Closes #781

## What it does

A manifest that *parses* is not a manifest that can be *satisfied*. Six closed-surface proofs, all deterministic:

| Proof | Catches |
|---|---|
| `parses` | malformed document (returns alone — nothing else can run) |
| `lint` | the SIP-0099 parses-but-unexpandable classes |
| `expands` | expander refusal, or a skeleton with **no fill slots** — every file frozen, so the cycle "succeeds" having produced nothing |
| `contract_derives` | no contract derivable, so nothing could state what done means |
| `checks_live` | a derived check that would skip at every evaluation forever — the slot ships unverified while appearing covered |
| `testid_coverage` | a route with no DOM anchor, leaving qa guessing at markup |
| `status_declared` | #772 — see below |

Whether the design actually *satisfies the PRD* is judgment; that stays with `decisions[].warrant` and the human gate.

## The status proof is narrower than the plan's shorthand — deliberately

The plan says "per-endpoint `success_status`." Implemented literally, **that rejects the reference manifest**, which declares it on 1 of 5 endpoints. A rule that rejects the artifact the 1.4 evidence was measured against is the wrong rule, not a wrong manifest.

Measuring the actual defect gives the exact rule:

- **collection POST** (`POST`, no `{` in path) → deriver expects `success_status or 201`, scaffold omits `status_code=` so FastAPI returns **200** → unwinnable;
- **child POST** → deriver expects `or 200`, scaffold default is also 200 → agree, harmless;
- **GET** → no status probe derived at all.

One rule, aimed at the one real failure. The reference manifest passes it; #772's shape fails it — the pair a gate needs. Both non-failing shapes are tested too, so the narrowness reads as intended rather than as an oversight.

**This delivers #772's protection without waiting on the manifest v5 schema change.** That change still deletes the defaulting entirely (fix-by-deletion); this stops the class reaching a cycle in the meantime.

## The history was already in the code

`success_status`'s own docstring documents this costing a roll at **pf-39**: the expander emitted a bare `@router.post` while the derived probe pinned 201, so *"green then depended on the dev agent volunteering `status_code=201`: pf-38 did, pf-39 did not, and that single token was the whole difference."*

The field was added as the fix. Leaving it **optional** left the trap open. This gate closes it.

## Other decisions

**Pure function, no I/O** — provable against adversarial manifests long before an authoring stage exists to produce one.

**All proofs run; findings accumulate.** Short-circuiting would make an author fix one defect per revision and exhaust `manifest_max_attempts` on a manifest that had three.

**Findings are typed, not free text**, so M6's failure taxonomy can attribute a rejection without parsing prose — the "one vocabulary, not three" intention applied before there's anything to reconcile.

**No call-site wiring (D5).** The gate is built and proven; wiring it into the authoring loop is M1's job, where the thing it gates first exists. Both in one PR would leave neither independently reviewable.

**Regression: 6527 passed, 1 skipped.**

### DIFF
--- a/src/squadops/cycles/manifest_winnability.py
+++ b/src/squadops/cycles/manifest_winnability.py
@@ -1,0 +1,208 @@
+"""Can this interface manifest actually be won? (#781, M3)
+
+A manifest that *parses* is not a manifest that can be *satisfied*. It can name a route
+the expander cannot place, promise a view with no DOM anchor to query, or imply a
+contract holding a check that is dead on arrival — and each of those costs a full
+implementation workload to discover.
+
+This module answers the question deterministically, before anything downstream spends
+on it. **Closed-surface proofs only** (SIP-0103 §5b Q2): every finding is mechanical
+and reproducible. Whether the design actually *satisfies the PRD* is judgment, and it
+stays with the ``decisions[].warrant`` discipline and the human review gate.
+
+Pure by construction — manifest text in, findings out. No vault, no cycle, no deploy —
+so the gate is provable against adversarial manifests long before an authoring stage
+exists to produce one.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+#: Stable finding classes. Typed rather than free text so M6's authoring failure
+#: taxonomy can attribute a rejection without parsing prose — the "one vocabulary, not
+#: three" intention from the post-1.5 reconciliation, applied before there is anything
+#: to reconcile.
+PROOF_PARSES = "parses"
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #784, opened 2026-08-08
feat(cycles): schema gate — provenance and the decisions[] judgment record (#783, M2)

jasonLadd:
Fourth code item of the v1.6 M lane. Design gate posted to the issue before code.

Closes #783

## The premise was worse than the plan assumed

`decisions` is present in the reference manifest — 4 entries, each `id` / `choice` / `warrant` — but was **not parsed into the model at all**. The citation discipline SIP-0103 leans on existed only as YAML nobody read.

Now parsed, with the §5c.10 **`unresolved: true`** form: an author surfaces a design question it declines to answer instead of silently defaulting.

## The constraint that governs the whole item

Decisions stay **out of `_canonical()`**. That projection is *"every field the expander reads to produce the skeleton, and nothing else,"* with provenance excluded precisely so a provenance-only edit cannot move the hash and invalidate the contract bound to it.

Decisions are judgment, not structure. Adding them to the projection would have moved the manifest hash, broken contract v9's binding, invalidated #777's guard, and broken **every bind-mode cycle**.

Pinned two ways: the hash still equals the one contract v9 names (`bb472e26…`), and revising a warrant plus appending a decision leaves it unmoved. **Consequence worth stating:** an author may improve an explanation without re-deriving anything, because the design didn't change.

## Lifecycle rule 3 holds by construction — pinned, not checked

The plan says derivation may not depend on an unresolved decision. Since decisions are absent from both the structural projection *and* the derived contract, **no derived check can depend on one**.

A detector for an impossible condition is the "rejects the guess but never supplies the answer" pattern this repo has ruled against, so the test proves the exclusion rather than re-checking it.

## Schema-gate proofs

| Proof | Rejects |
|---|---|
| `provenance` | empty `source_prd` — a design that doesn't say what it was designed from can't be reviewed |
| `decision_record` | no `id` (can't be referenced across revisions); a `choice` with no `warrant` — **indistinguishable from an invention**; `unresolved` with no `question` — deferring without saying what, which *reads as diligence while communicating nothing* |

Citation stays at **decision granularity** (`source_prd` + `decisions[].warrant`), not per entry — §5b Correction 3 ruled per-entry would bloat authoring for little gate value, and the schema has no field for it.

## Rename

`manifest_winnability.py` → `manifest_gates.py`. It now holds both authoring gates and the name should say so. Done now while the only importer is its own test (no call sites yet, per M3's D5) — the rename gets expensive once the authoring loop wires them.

**Regression: 6537 passed, 1 skipped.**

### DIFF
--- a/src/squadops/capabilities/scaffold.py
+++ b/src/squadops/capabilities/scaffold.py
@@ -127,6 +127,30 @@ class Frontend:
 
 
 @dataclass(frozen=True)
+class Decision:
+    """One design judgment the schema cannot express mechanically (SIP-0103 §5c.4).
+
+    Pagination behavior, authorization boundaries, idempotency, caching — decisions a
+    designer must make and the manifest's structural fields cannot hold. Recording them
+    with a PRD ``warrant`` makes the judgment explicit, reviewable at the human gate,
+    and auditable afterwards.
+
+    ``unresolved`` marks a question the author declines to answer (§5c.10): a PRD that
+    does not determine something should surface as a stated question rather than a
+    silent default. An unresolved entry carries its ``question`` instead of a
+    ``choice``.
+
+    Deliberately NOT part of the structural projection — see ``_canonical``.
+    """
+
+    id: str
+    choice: str = ""
+    warrant: str = ""
+    unresolved: bool = False
+    question: str = ""
+
+
+@dataclass(frozen=True)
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #786, opened 2026-08-08
feat(cycles): authoring failure taxonomy — one classification, three consumers (#785, M6)

jasonLadd:
Fifth code item of the v1.6 M lane, and the last before M1. Design gate posted to the issue before code.

Closes #785

## One record, not three

M5's revision reason, B1's baseline, and the window's attribution were three asks for the same thing. Built separately they become three vocabularies to reconcile later — the problem the post-1.5 reconciliation names as intention 2. Built once, **before the first authored manifest is rejected**, because a rejection recorded without a class is data that cannot be recovered afterwards.

| Class | Who fixes it |
|---|---|
| `authoring_defect` | the author — retry with rejection context |
| `schema_gap` | schema change ⇒ blueprint version |
| `blueprint_limitation` | blueprint governance (S5's ≥2-stack rule) |
| `derivation_defect` | M0's code, never the manifest |
| `prd_insufficiency` | `decisions[]`, `unresolved: true`, the human gate |

Ownership is **borrowed** from the check registry's `OWNERSHIP_*` constants rather than restated — two answers to "whose problem is it?" is the drift this codebase keeps paying for. A test pins that every class names an owner from that shared set.

## Two positions that shaped it

**`prd_insufficiency` is not a gate failure.** An unresolved decision *passes* the M2 schema gate by design — declaring an open question is correct behavior. So the assessment reads the manifest's **declared uncertainty**, not only its rejections; a findings-only classifier would report zero PRD insufficiency on a manifest whose author explicitly said the requirements were ambiguous. Open questions are counted **without rejecting**.

That's the whole point of the class: a PRD that under-specifies must not look identical to a squad that cannot design, because the remedies are opposite.

**Some classes are recurrence-determined, and single-occurrence precision would be faked.** "The expander refused this design" is genuinely ambiguous between *the author asked for something outside the vocabulary* and *the vocabulary should grow*; nothing at the gate separates them. So a single occurrence takes the conservative class and **recurrence promotes it** — measured across cycles by B1, which is exactly why B1 records recurrence rather than only totals. Documented where the classifier lives, so a later reader can't mistake a conservative class for a confident one.

**Derivation failures classify to us, not the author.** A manifest that already passed lint and expansion but produces an unusable contract is the deriver's problem by elimination — and asking an author to work around a broken deriver teaches a superstition.

## Tests worth noting

Every gate proof is **enumerated from the gate module** and must have a class, so a proof added later without one fails the suite rather than silently re-creating the bucket. An unknown proof is filed conservatively rather than lost — losing a finding is worse than filing it under the author's column.

**Regression: 6552 passed, 1 skipped.**

## What's next

**M1** — the authoring stage. Everything since the queue front has been building the things that make M1's failures readable; this was the last of them. Per the verification cadence, M1 is followed by **V4**: the first authored cycles, unscored, where the thing to watch is not the pass rate but whether this taxonomy can name which subsystem to blame.

### DIFF
--- a/src/squadops/cycles/authoring_failure.py
+++ b/src/squadops/cycles/authoring_failure.py
@@ -1,0 +1,191 @@
+"""Why did this manifest fail — and whose problem is it? (#785, M6)
+
+One classification, three consumers: M5's provenance records *why* each revision
+happened, B1's baseline counts *which classes recur*, and the authored-mode measurement
+window needs *which subsystem* a failure is attributable to. Built three times, those
+become three vocabularies that must later be reconciled — the exact problem the post-1.5
+reconciliation names as intention 2.
+
+It exists before the first authored manifest is rejected, because **a rejection recorded
+without a class is data that cannot be recovered later**. Without it the window's output
+collapses into one "manifest rejected" bucket, which cannot say whether authored mode is
+limited by the author, the schema, the blueprint, or the deriver — the entire question
+the release is asking.
+
+Keyed alongside the vocabulary that already exists (``FailureEvidenceCategory``,
+``OWNERSHIP_*``, locus classification) rather than as a parallel set: this adds the
+*authoring* dimension those lack, and borrows their ownership semantics rather than
+inventing a second answer to "whose problem is it?".
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #790, opened 2026-08-08
docs: rotate CHANGELOG for v1.4.0–v1.5.0 and guard the rotation (#789)

jasonLadd:
Steps 1 and 3 of #789. Step 2 (the GitHub Release objects) is out-of-tree and follows this merge, so the releases quote the merged changelog text.

Closes #789

## Premises verified before acting

All confirmed: `gh release list` still advertises **v1.3.1 as Latest**; all six tags v1.4.0–v1.5.0 exist; `CHANGELOG.md` was last touched 2026-07-26, five days before the 1.4.0 cut, and its `[Unreleased]` block still asserted *"Nothing below has shipped in a release"* and *"Lane S — the Ephemeral Application Sandbox — has not started"*, both false since 2026-07-31.

**One correction to the brief, and to my own earlier reporting.** Dates come from the annotated tags, which are authoritative — and v1.5.0's tagger date is **2026-08-07**, not the 2026-08-06 I reported at the cut (I'd converted a UTC timestamp with the wrong offset). That also means `docs/ROADMAP.md` carried the wrong date in two places, fixed here.

That drift is instructive: the existing `roadmap-current-header` guard matches `### v(X.Y.Z) (date) — Current` but only asserts the *version*, so the date was never checked.

## Rotation

The existing `[Unreleased]` body **was** the 1.4 arc, so it becomes `[1.4.0]` with its two false claims corrected and the cut-gate supersession recorded — keeping the honest claim language: *a specified contract, not PRD-to-app*.

Five new sections written **from the release records, not re-derived**: ROADMAP's timeline, `1-4-1-hardening-patch-plan.md`, `1-5-0-stabilization-plan.md`, `post-1-5-roadmap-reconciliation.md`. 1.4.2–1.4.4 are short patch entries because that's what they were. A fresh `[Unreleased]` is opened for the in-flight 1.6 Authorship work.

## Prevention — rule 4 on the owning guard, not a new CI step

The brief suggested extending the CI docs-drift guards; the actual home is `tests/unit/architecture/test_docs_version_sync.py`, which already owns rules 1–3 and runs inside the regression gate. Extending the owning seam beats adding a parallel one.

- the pyproject version must have a dated `## [X.Y.Z] — YYYY-MM-DD` section;
- `[Unreleased]` must not still describe the released version — catching the other half, where someone renames the section and never opens a fresh one.

**Verified by mutation, not assumed.** Bumping pyproject to 1.6.0 with no matching section fails 7 tests with `CHANGELOG.md has no `## [1.6.0] — YYYY-MM-DD` section`; restoring passes 16.

## Checklist

`CLAUDE.md` gains the full six-step cut checklist, with **rotate the changelog** and **cut the GitHub Release** as numbered steps beside rebuild-deploy-verify — so the half that got skipped five times is written down rather than folklore.

**Regression: 6555 passed, 1 skipped.**

## After merge

I'll cut six GitHub Releases from these sections against the existing annotated tags — **no re-tagging** — and mark v1.5.0 Latest.


jasonLadd (commented):
reviewed

PR #792, opened 2026-08-08
feat(cycles): the squad authors its own interface manifest (#791, M1)

jasonLadd:
**M1 of the 1.6 Authorship line** — the squad writes its own interface manifest.

A dedicated `development.author_manifest` framing stage, placed between dev's technical
design and qa's test strategy, that emits `interface_manifest.yaml` and revises it against
the deterministic gates before anything downstream spends on it.

Design gate posted to #791 before any code; three of its four premise findings changed the
shape of the work.

## §6 open question 1 — resolved: single author, and the reviewers are the gates

§5a's recommendation with one amendment, measured against main. The two things §5a assigns
a QA reviewer — *every view declares its anchors, every endpoint a probe-able status
contract* — are **already proven mechanically** by M3's `testid_coverage` and
`status_declared`. An LLM reviewer on top would put judgment where determinism already
holds, which is the §5c.2 / #464 style-lottery class.

So: one author (dev), reviewed by M3's proofs and M4's human gate, with
`qa.define_test_strategy` reading the manifest downstream as a free non-gating lane — which
is also why the stage sits *before* qa rather than after. Multi-role proposers stay the
recorded fallback; that migration direction has one, the reverse does not.

## Three corrections against main

**1. Authoring already existed, in the wrong place and ungated.** SIP-0099 99.2 asked the
dev *proposer* to emit the manifest as a second fenced block beside its plan tasks, and the
merger re-emitted it (dev preferred, qa fallback). This is a **relocation**: the old path is
deleted, not left running beside the new one — two emitters of one artifact, with proposer
ordering deciding which won, is exactly the dual path the repo forbids.

**2. A framing-authored manifest never reached the implementation run.** The framing→
implementation forwarding filtered promoted artifacts to
`{document, control_implementation_plan}`, dropping `interface_manifest`. So the manifest
was stored, gate-validated, promoted — and then silently not carried, leaving the
implementation unscaffolded while framing believed it had designed a skeleton. Invisible
because every cycle since 99.2 has run bind mode, where the manifest rides
`plan_artifact_refs` from cycle creation (#496) and never travels that seam. Authored mode
could not have worked without this fix; it lands here with a test.

**3. The taught schema would have been rejected on every roll.** The authoring asset
predated M2/M3 and showed no `source_prd`, `decisions[]`, `success_status` or `testids` —
all four now required. That is #629's pattern: the system holds the rule and shows the
author only the rejection. Closed the way #686 closed it for plans — a
`manifest_authoring_rules` classification table keyed to the `PROOF_*` constants, bound by
test to the managed asset, so a proof cannot be added to the gates without a decision about
whether authors are told. Derivation-owned proofs are required to be **absent** from the
asset: teaching an author to work around a broken deriver both teaches a superstition and
files our defect under their name.

## Wiring the gates found a defect in M6

M2, M3 and M6 were pure modules with no callers until now. M1 wires all three — the gates as
the revision loop's in-stage verdict and as the framing gate's rejection, the taxonomy as
the outcome record. Doing so surfaced one defect: both gates parse independently, so an
unparseable manifest was reported **twice** and counted as `authoring_defect: 2` in a
baseline measuring one defect. Fixed at the composition, with the assertion added to M6's
existing double-report test.

## Shape

- **Mode is derived, not flagged** — a scaffoldable stack with no pinned `contract_ref`,
  the same condition the removed proposer path keyed on, lifted to step selection. No new
  config key; rollback is a profile pointing at a seeded contract.
- **Budgets are the existing seams (§5b Q4)** — `manifest_max_attempts` buys in-stage
  revisions with the gates' findings as corrective context (all findings per attempt, not
  one per attempt), and an exhausted manifest is **still emitted** so the framing gate can
  reject it into a free `framing_max_rerolls` re-roll rather than failing the run with
  nothing for an operator or the taxonomy to read.
- **The §5c.1 input contract is enforced, not asserted** — a declared allow-list of envelope
  keys with a test that fails if the handler reads anything else, including the shapes
  contamination would arrive in (a reference manifest, a cross-cycle recall).
- **Guard 1a holds** — the emitted artifact carries the same filename and type the seeded
  rail uses, so nothing below framing can tell who wrote it. The seeded framing sequence is
  pinned unchanged by test.

## Verification

Full regression green (6,984 unit + 6,598 regression). 36 new tests across four files:
mode/placement/dispatch, the rules binding, the handler's revision loop and failure
surfaces, and the two executor seams (gate rejection + forwarding).

No cycles run — M1 is the first item that needs one, and per the plan's cadence that is
**V4**, after deploy. The seeded control banked at V2 is untouched.

Closes #791


jasonLadd (commented):
reviewed

Issue #410, opened 2026-07-13
Ollama adapter silently drops qwen3 thinking tokens — ~60% of paid generation time is discarded unread and invisible to LangFuse


jasonLadd:
## Finding (measured during PR #408 validation, cyc_6841d75f167c)

qwen3.6:27b generates thinking by default via Ollama (`message.thinking`). The adapter (`adapters/llm/ollama.py`) never sends a `think` parameter and reads only `message.content` — so every call pays wall-clock for the full `eval_count` (thinking + content) but the thinking is dropped unread and never reaches LangFuse.

**Direct probe:** `"Say exactly: hello"` → `eval_count=143`, content=`"hello"` (1 token), `message.thinking`=556 chars. Thinking is ON by default and invisible.

**Measured across the implementation run's 11 instrumented calls:** 39,408 tokens generated, ~15.5k chars-estimated content survived → **~60% of generation time bought thinking that nobody (human or system) can even inspect.** Per-call overhead ranged 13–81%. At the Spark's 11.4 t/s ceiling this is roughly 35 min of a 77-min run.

## Why it matters
1. **Wall-clock**: dominant, controllable cost on single-GPU deployments.
2. **Observability**: LangFuse shows content only — the paid tokens don't appear in any telemetry, so the cost was undiagnosable until cross-referencing `eval_count` vs stored output length.
3. **Quality signal**: if thinking is *helping* output quality (plausible on 27b, see dense-vs-MoE reasoning-depth rationale), we should know; if not, it should be off. Today it's neither observable nor configurable.

## Proposed direction
- Adapter: support an explicit `think` option per request (Ollama `think: true/false`), threaded from capability/handler config — per-capability knob, not a global.
- When thinking is on: capture `message.thinking` and emit it to LangFuse (redaction rules apply) so the spend is visible.
- Candidate defaults: thinking ON for design/framing/dev capabilities, OFF for mechanical verdict/repair-validation calls (e.g. `qa_validate_repair` spent 57–69% overhead to produce a PASS/FAIL verdict).

Found while analyzing token-utilization during the #407/#408 Spark validation cycle.

jasonLadd:
## Qwen3.8-27B changes this issue's shape: the dial is now documented, and the default moved the wrong way

Confirmed still live on `main` @ `2aaf5f73` — `adapters/llm/ollama.py` sends no `think` parameter anywhere, and `_build_chat_payload:188` builds `options` from `num_predict` + `temperature` only.

[...]

Issue #301, opened 2026-07-01
Composition root (api/runtime/main.py) bypasses existing llm/queue factories — provider selection + config wiring inconsistent


jasonLadd:
## Problem
`src/squadops/api/runtime/main.py` (the runtime composition root) is **inconsistent** about the factory pattern. It routes most adapters through factories (auth `create_auth_provider`, cycles/registry, telemetry `create_llm_observability_provider`, events, audit) but **direct-constructs others even though factories exist**:

- `RabbitMQAdapter(url=RABBITMQ_URL)` (main.py:289) — bypasses `adapters/comms/factory.py::get_queue_adapter`
- `OllamaAdapter(...)` (main.py:379) — bypasses `adapters/llm/factory.py::create_llm_provider`
- `A2AClientAdapter()` (main.py:423) — no factory exists at all

## Why it matters (bigger than any one setting)
1. **Provider selection is lost.** The factory is where config selects the provider — the whole point of ports & adapters. Hardcoding the concrete `OllamaAdapter`/`RabbitMQAdapter` in the composition root means the runtime can't config-swap the LLM/queue provider despite the port + factory supporting it.
2. **Config wiring is bypassed → drift.** The factory is where config flows into the adapter. Direct construction re-implements it and diverges. Concrete instance: #158/#299 externalizes adapter timeouts to `SQUADOPS__*` config + wires the factories, but the *running* llm/queue adapters won't pick it up because main.py bypasses those factories. Every adapter setting has this latent drift.
3. **Double-maintenance.** Construction changes (new dep, new param) must be applied to both the factory and each direct site; easy to miss one.

## Fix
Route the runtime's adapter construction through the factories (`get_queue_adapter`, `create_llm_provider`, a new a2a factory), so provider-selection + config-wiring happen in one place. Then #158/#299's config-driven timeouts (and future per-adapter config) take effect uniformly.

## Ownership
Cross-lane: the composition-root change is in `api/runtime/main.py` (Mac); the factories reading config is `adapters/*/factory.py` (Spark). Structural/arch — fits the 1.3.0 stabilization bucket (with #186/#152/#234) unless it's causing active pain sooner.

Surfaced during #158 (adapter timeout externalization). Related: #299, #300.

jasonLadd:
[...]

PR #794, opened 2026-08-08
SIP (proposed): Atlas Provider Adapter — config-selected inference providers behind a conformance gate

jasonLadd:
Files a new proposed SIP for design review. **Merging this PR only *files* the draft — it is not acceptance.** Promotion to `accepted/` happens separately, on an explicit maintainer ruling (workflow table, CLAUDE.md).

## What it proposes

Adding Atlas — the in-house LLM inference engine hand-tuned for DGX Spark — as a second provider. The SIP's claim is that this is **not "write an adapter"**: it is "write an adapter and discover, one live roll at a time, every place the vendor already leaked past the port."

Four moves: provider selection becomes config (`LLMConfig.provider`, defaulting to `ollama`); capabilities are declared rather than inferred by `isinstance`; model availability moves on-port; and the adapter ships **dark behind a conformance suite** that no adapter may bypass.

## What the code audit found

Verified against the tree, reproducible via the SIP's Appendix A:

- **Provider selection does not exist.** `LLMConfig` has no `provider` field, and `create_llm_provider(provider="ollama")` is never passed the argument by any caller. `SQUADOPS__LLM__PROVIDER=atlas` would do nothing, silently.
- **Seven code sites name `OllamaAdapter` across 3 files in `src/`** — including two `isinstance` checks in API routes and a composition root that constructs the vendor adapter directly, bypassing the factory.
- **The quietest leak is the worst**: `MODEL_SPECS` is keyed on Ollama tag syntax, and `_resolve_model_budget` treats an unrecognized name as *no clamp*, not an error. The registry's own comment records what one missing key already cost. A migration renames every key at once.

## Scope boundary

**1.7 ships the neutrality groundwork and the dark adapter. Cutover — any change to a shipped default — is a separate 1.8+ decision.** Every phase is inert on merge: no compose file, `.env.example` value, profile, or agent image changes the model runner. This keeps it invisible to the Spark lane's 1.6 validation sprints.

## Issue linkage — deliberately partial

- **`#313`** — absorbed as P1 (doctor's `ollama list` hardcode + off-port `list_pulled_models`). P1's PR will carry `Closes`.
- **`#301`** — **referenced, not closed.** This SIP fixes only the LLM half; the queue half stays in the 1.7 composition-root cluster behind its own design gate.
- **`#410`** — related, explicitly *not* absorbed; ruled diagnostic rather than gating (§10.3, with the reasoning for the reversal recorded).
- **`#793`** — filed *from* this work: `tokens_per_second` never reaches LangFuse because the redaction copy drops 1 of 11 fields. Independent bug; the SIP makes fixing it a cutover precondition that gates none of the adapter work.

This PR resolves none of the above. It files a design.

## Needs a ruling before acceptance

Two open decisions are marked ⓞ in §1.1 and cannot be settled by an implementer:

1. **§10.4 — release placement.** The argument for landing a second provider adapter in a feature-free odd minor is a *reading* of the convention. §4.3 makes the case (inert by construction, rails-before-mechanism precedent ×3, and 1.7's stated identity is literally "every port is actually a port"). A clean fallback exists if rejected.
2. **§C.4 — decision thresholds.** Proposed margins for "materially faster" and acceptable parity loss. The numbers are contestable; what the SIP argues is non-negotiable is that *some* threshold is fixed before the benchmark runs, or it gets fitted to the result.

## Execution note

Churn analysis on the target files (in-conversation, not in the SIP) splits cleanly: `ports/llm/provider.py`, `adapters/llm/factory.py`, `ollama.py`, and `routes/cycles/models.py` are cold (0–1 commits in 60 days). `api/runtime/main.py` (18), `routes/cycles/cycles.py` (15), and `agents/entrypoint.py` (8) are hot with 1.6.

Suggested sequencing: P1, P3, and most of P0 land in cold files and are safe now; **P2 and P0's preflight site should wait for the post-1.6 window**, which independently matches the 1.5 plan's ruling that the #301 composition-root cluster needs a design gate before code.

## Review focus

The design questions worth arguing are §3.2 (capability declaration replacing `isinstance`), §3.6 (the conformance suite as a factory gate), §4 (whether the dark-ship rule is actually sufficient protection for the Spark lane), and §C.4's thresholds. Appendices B and C are non-normative.

### DIFF
--- a/sips/proposed/SIP-Atlas-Provider-Adapter.md
+++ b/sips/proposed/SIP-Atlas-Provider-Adapter.md
@@ -384,8 +384,20 @@ queue port.
 - **A/B tier** — same prompts, same models, same box, through both adapters. **This is the
   migration decision instrument** and the reason the SIP exists (§2.5). It reports two
   numbers, and both gate the cutover:
-  - **Throughput** — tokens/sec and wall-clock per generation. The expected win, and the
-    only reason to switch engines.
+  - **Wall-clock to a validated artifact** — time to an accepted output, the honest
+    headline. **Not tokens/sec.** t/s counts tokens *generated*, and `eval_count`
+    includes thinking tokens the system discards unread (#410), so t/s misreads the
+    comparison in both directions: an engine that thinks *less* delivers the same
+    artifact faster at equal-or-lower t/s (a win scored neutral), and one that is faster
+    per-token but thinks *more* shows higher t/s and worse wall-clock (a regression
+    scored a win). t/s stays as a diagnostic underneath, never the verdict.
+
+    Worked example from a live qa run — `qa_test_handler LLM throughput: 10.6 t/s
+    (4866 tokens)` = ~7.6 minutes for one generation, of which #410's measured 13–81%
+    thinking range means somewhere between ~1 and ~6 minutes is unreadable reasoning.
+    Nothing today can narrow that interval for a given call, which is exactly the
+    problem: **the A/B is uninterpretable until thinking and content tokens are counted
+    separately** (§10.3).
   - **Quality parity** — extraction health (clean / recovered / failed rates per
     SIP-LLM-Emission-Contracts §3.4) and usage-accounting consistency. **A throughput win
     with a quality regression is a loss**, and without this half the FAY and 98.5 lineages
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reivewed

jasonLadd (commented):
reviewed

PR #797, opened 2026-08-08
fix(telemetry): tokens_per_second reaches LangFuse — replace the hand-copied redaction rebuild

jasonLadd:
## The defect

`LangFuseAdapter.record_generation` rebuilt `GenerationRecord` field-by-field to apply redaction, and copied **ten of eleven** fields — omitting `tokens_per_second`. It defaults to `None`, so the omission raised nothing. **LangFuse has reported null throughput for every generation since SIP-0061.**

```python
# before — ten of eleven
redacted_record = GenerationRecord(
    generation_id=record.generation_id,
    ...
    latency_ms=record.latency_ms,
    #  tokens_per_second=record.tokens_per_second,   <-- never written
    prompt_name=record.prompt_name,
    prompt_version=record.prompt_version,
)

# after
redacted_record = replace(
    record,
    prompt_text=self._redaction.redact(record.prompt_text),
    response_text=self._redaction.redact(record.response_text),
)
```

## Why it stayed invisible

1. The field is optional — had it been required this would have raised on the first call.
2. **`latency_ms` sits directly above it in the same constructor and *is* copied**, so LangFuse timing data looked healthy, which made the whole telemetry path look healthy.
3. The handler logs throughput on a **separate code path** (`capabilities/handlers/cycle/base.py:702`), so the number stayed visible in the Prefect UI and the gap never presented as missing data.

## Why `replace()` rather than adding the keyword

`GenerationRecord` is `@dataclass(frozen=True)`, and `dataclasses.replace()` is the pattern CLAUDE.md already names for this repo. *Copy-everything-redact-two* states the actual intent, and a field added tomorrow is carried without touching this adapter.

Adding `tokens_per_second=` would have fixed today's symptom and left the trap armed for field twelve. The same change is applied to `record_event`/`StructuredEvent` — complete today only because no field has been added to it since it was written.

## Tests — verified against the unfixed adapter

Five added. Stashing the fix and re-running gives **3 failures with the right messages** (`assert None == 10.6`, `tokens_per_second was dropped or altered by the redaction copy`). The 2 that pass guard regressions the *fix itself* could introduce, so passing both before and after is correct.

| Test | Bug it catches |
|---|---|
| `test_tokens_per_second_survives_redaction` | the exact defect |
| `test_every_generation_field_but_the_redacted_pair_is_carried_through` | field twelve added and dropped — each field gets a distinct non-default value so a drop reads as `None` rather than coinciding with a default |
| `test_every_event_field_but_the_message_is_carried_through` | the same latent shape in `record_event` |
| `test_redaction_still_happens` | trading a dropped field for a leaked secret |
| `test_tokens_per_second_reaches_the_langfuse_generation_call` | the second hop — surviving redaction but not landing in the payload handed to the SDK |

## Out of scope, per the issue

Sending real `start_time`/`end_time` so LangFuse can derive latency natively (it currently receives a single point-in-time observation created on the flush thread), and a durable per-generation throughput record — `tokens_per_second` is persisted nowhere. Both are worth doing; neither is needed to restore the metadata field.

## Verification

- Regression suite: **6602 passed, 2 skipped**
- `ruff check` / `ruff format`: clean
- Behavior change is confined to telemetry metadata — no domain, contract, or wire-format surface is touched.

Closes #793

### DIFF
--- a/adapters/telemetry/langfuse/adapter.py
+++ b/adapters/telemetry/langfuse/adapter.py
@@ -15,7 +15,7 @@ import queue
 import random
 import threading
 import time
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from enum import StrEnum
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #798, opened 2026-08-08
fix(cycles): an authored manifest binds its own run (#796)

jasonLadd:
Derive and pin the verification contract the moment the authoring stage lands a manifest,
mid-framing, so every task dispatched after that point binds to the design the squad just
wrote. Found by **V4 roll 1** — the first cycle ever run in authored mode.

## What V4 measured

The manifest was good: both gates clean, contract derives, 19-file skeleton, 4 fill slots.
The plan authored beside it, in the same framing run:

```
FILL SLOTS : backend/routes.py, frontend/src/views/{CreateRunView,RunDetailView,RunListView}.jsx

task 0  FROZEN     backend/models.py          task 5  INVENTED  frontend/src/components/RunList.jsx
task 1  FROZEN     backend/main.py            task 6  INVENTED  frontend/src/components/CreateRun.jsx
task 2  INVENTED   backend/routes/runs.py     task 7  INVENTED  frontend/src/components/RunDetail.jsx
task 3  INVENTED   backend/routes/runs.py     task 8  INVENTED  backend/tests/test_runs.py
task 4  FROZEN     frontend/src/main.jsx
```

**Zero of nine on a fill slot.** The recorded rejection was #673's dual-claim rule catching
tasks 2 and 3 — the one contract-independent net that happened to fire. Everything else was
invisible, because every contract-gated net keys on `_is_bind_mode`, which an authored cycle
(no seeded `contract_ref`) never satisfies. Authored mode was a second pipeline: one with
frozen-artifact ownership, qa ownership, module existence and criteria binding all switched
off.

## The fix

The manifest is a design artifact; the instant it exists it is fixed. So the cycle stops
being author mode right there and becomes what a seeded cycle already is — a manifest, and
the contract derived from it. Four seams, each on rails that already existed:

1. **Derive at acceptance** (`_bind_authored_manifest`) and append the contract to the run's
   own artifact refs, so it is listed in `artifacts list`, seen by the gate, promoted at
   approval, and forwarded.
2. **Inject at dispatch-time enrichment** — `_enrich_envelope`, the declared single
   composition point (#663), whose "dispatch keys shadow plan-time keys" rule was built for
   exactly this. A seeded contract exists at plan generation and is injected there; an
   authored one cannot exist that early. Same injector, same registry, two moments —
   mutually exclusive by construction.
3. **The gate recognises a run-derived contract**, so its bind-mode nets engage.
4. **Forwarding sets `contract_ref`** for the implementation run; an operator-supplied ref
   always wins, matching every other forwarded key.

This is SIP-0103 §3.5 made true rather than asserted — *"authored mode is a mode of manifest
provenance, not a second pipeline"* — and §5a's third containment bound: at freeze the
manifest hash-freezes and the contract derives, after which determinism is identical to
seeded mode by construction.

**Seeded mode is untouched by construction**, which matters because it is the control: the
derivation no-ops when a `contract_ref` exists, and enrichment injects only when handed a
contract derived during *this* run. Pinned by test.

## Verification

The V4 manifest and its plan are committed as a matched fixture pair (`tests/fixtures/authored_v4/`,
with a README recording their provenance) and the suite **replays the observed failure**:
with the contract derived, the gate rejects that exact plan on three frozen claims plus the
impossible `backend.routes.runs` import — ten errors where author mode raised one.

16 new tests covering derivation (once per run, never in seeded mode, latest manifest wins,
underivable manifests left to the gate), injection (right task types, byte-identical without
a derived contract), the gate replay, and forwarding. Full regression green: 7,000 unit,
6,614 regression.

## Two names corrected in passing

- `is_seeded_manifest` → `is_interface_manifest`. It never checked provenance, and after #791
  half its callers ask about an authored manifest.
- `_inject_contract_inputs` → `inject_contract_inputs`, having gained a second legitimate
  caller; a cross-module import of a private name is the smell that says promote it.

Closes #796

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,10 @@ seeds a manifest but no contract (#779), M3 winnability gate (#781), M2 schema g
 the `decisions[]` judgment record (#783), M6 authoring failure taxonomy (#785), and **M1
 the authoring stage itself (#791)** — a dedicated `development.author_manifest` framing
 step that revises against the gates, replacing the ungated proposer-side emission it
-relocates. M4 (the human manifest gate) and M5 (provenance) follow.
+relocates, and **#796** — the fix V4 roll 1 exposed: an authored manifest now derives and
+pins its contract mid-framing, so the plan authors bind to the design their own squad just
+wrote instead of inventing paths, and every contract-gated net engages. M4 (the human
+manifest gate) and M5 (provenance) follow.
 Plan: `docs/plans/1-6-0-authorship-plan.md`.
 
 ## [1.5.0] — 2026-08-07
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -46,7 +46,7 @@ from squadops.cycles.acceptance_evaluation import resolve_check_stack
 from squadops.cycles.agent_config import build_agent_resolver
 from squadops.cycles.build_completeness import compute_missing_required_files
 from squadops.cycles.checkpoint import RunCheckpoint
-from squadops.cycles.contract_derivation import SEEDED_MANIFEST_FILENAME
+from squadops.cycles.contract_derivation import CONTRACT_ARTIFACT_TYPE, SEEDED_MANIFEST_FILENAME
 from squadops.cycles.frozen_check_validation import frozen_check_violations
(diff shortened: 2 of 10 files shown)


jasonLadd (commented):
reviewed

PR #799, opened 2026-08-08
test(llm): LLM port conformance suite — one contract, adapter-parameterized (SIP Atlas P3)

jasonLadd:
Implements **P3** of `sips/proposed/SIP-Atlas-Provider-Adapter.md` (filed in #794).

Test-only. No production file is touched, so this is inert against the 1.6 lane by construction.

## What this is

One behavioral suite every `LLMPort` adapter must pass. With a single live provider it **characterizes** the current contract; it becomes a conformance suite the moment a second adapter registers. The 1.5 stabilization plan asked for exactly this, and was explicit about the trap:

> it must declare semantic capabilities … as required/optional/extension per provider — **not** enshrine Ollama transport behavior as the contract.

So **no assertion names a route, a payload key, or a streaming format.** Everything provider-shaped is confined to an `AdapterCase`'s dialect handler. Atlas (P4) and vLLM (P6) each land as one registry entry — no shared assertion changes.

## Why the wire is mocked at the transport

The neighbouring Ollama tests patch `_get_client`. That's fine for testing *that* adapter and useless for a shared suite: it presumes a private method every adapter must happen to have. `httpx.MockTransport` intercepts a layer lower, so the adapter's real URL construction, payload shaping, response parsing, and error mapping all execute — which is the code conformance is actually about.

## Coverage — 27 tests, 6 dimensions

Generation (incl. stream reassembly) · usage accounting · model listing · error translation · health · port surface.

Assertions nothing covered before:

| Assertion | Bug it catches |
|---|---|
| `list_models()` performs no network I/O | the port documents this as a MUST NOT and nothing tested it — a sync method reaching the network blocks the event loop for every caller on that thread |
| usage absent from the wire is `None`, never `0` | zero is indistinguishable from a free call and silently drags every average toward zero |
| `tokens_per_second` is a rate, not a count | a ns/s unit error yields a value off by 1e9 that still looks like a plausible float |
| refresh failure preserves the last known list | a transient blip would otherwise read downstream as "no models present" and block a valid cycle |
| streaming errors translate to typed exceptions | locus classification (#568) reads these types; a raw httpx error escaping reclassifies an outage as a squad mistake and burns a correction attempt |

## Validated by mutation, not by passing

A new suite that goes green first try is unverified, not good. Eight deliberate defects were injected into `OllamaAdapter`; **each fails the suite**:

`ns→s unit error in t/s` · `404 no longer → ModelNotFound` · `list_models leaks the internal list` · `refresh failure empties the cache` · `zero-filled usage` · `dropped final stream frame` · `ignored request model override` · `untranslated stream errors`

**That pass caught two vacuous tests of my own**, both now fixed:

1. **Adapters cache their httpx client**, so entering a second `patch` block had no effect — "now the backend fails" assertions kept talking to the healthy transport. A cache-preservation test passed against an adapter that emptied its cache on failure. Replaced with one transport dispatching to a swappable handler (`Wire`), immune to how any adapter manages connections.
2. **The streaming fixture's terminal `done` frame carried empty content**, so an adapter ignoring done-frame content passed regardless. The final fragment now rides that frame — which is also how Ollama actually behaves.

## Deliberately absent

`capabilities()` honesty and listing-vs-absence need P0/P1's port surface. They join this file when it lands, rather than being stubbed now.

## Verification

- Regression suite: **6624 passed, 2 skipped**
- `ruff check` / `ruff format`: clean · `lint_test_quality.py`: clean
- `tests/unit/llm/` is already in `REGRESSION_DIRS` — no allowlist change

Refs #313, #301 — neither is closed by this PR; P3 adds no production behavior.

### DIFF
--- a/tests/unit/llm/conformance.py
+++ b/tests/unit/llm/conformance.py
@@ -1,0 +1,223 @@
+"""Shared machinery for the LLM port conformance suite (SIP-Atlas-Provider-Adapter P3).
+
+Not a test module — imported by ``test_llm_port_conformance.py``.
+
+**What this is.** One behavioral suite that every ``LLMPort`` adapter must pass.
+With a single live provider it *characterizes* the current contract; it becomes a
+conformance suite the moment a second adapter is registered. The 1.5 stabilization
+plan asked for exactly this distinction and was explicit about the trap:
+
+    it must declare semantic capabilities ... as required/optional/extension per
+    provider — **not** enshrine Ollama transport behavior as the contract.
+
+So the assertions live in the shared suite and speak only the *port's* vocabulary.
+Everything provider-shaped — routes, payload keys, streaming framing — is confined
+to an :class:`AdapterCase`'s dialect handler. Adding an adapter is one registry
+entry, never a new test.
+
+**Why the wire is mocked at the transport, not at the adapter.** The neighbouring
+Ollama tests patch ``_get_client``, which is fine for testing *that* adapter and
+useless for a shared suite: it presumes a private method every adapter must happen
+to have. ``httpx.MockTransport`` intercepts one layer lower, so the adapter's real
+URL construction, payload shaping, response parsing, and error mapping all execute.
+That is the code conformance is actually about.
+"""
+
+from __future__ import annotations
+
+import json
+from collections.abc import Callable, Iterator
+from contextlib import contextmanager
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #800, opened 2026-08-08
feat(llm): declare capabilities on the port; model listing speaks ModelInfo (SIP Atlas P0 + P1 listing)

jasonLadd:
Implements **P0** and **P1's on-port listing half** of `sips/proposed/SIP-Atlas-Provider-Adapter.md`.

## Scope was chosen by churn, not by phase boundary

Both phases add to the same two files and converge on the same caller, so splitting them means touching those files twice. Everything here is **cold against the 1.6 line**:

| File | commits/60d | last touched |
|---|---|---|
| `ports/llm/provider.py` | 0 | 4 months |
| `llm/models.py` | 0 | 4 months |
| `adapters/llm/ollama.py` | 1 | 5 weeks |
| `routes/cycles/models.py` | 1 | 6 weeks |

Nothing config-shaped and nothing in the cycle-create path is touched.

## The port says what it can do

- **`LLMCapability` + `capabilities()` / `supports()`.** Defaults to **all-False**, so an adapter declaring nothing supports nothing — a silent omission fails closed instead of presenting as a working feature. Carries #572's doctrine verbatim from the queue port: *a flag that overstates the implementation is worse than a missing feature.*
- **`list_available_models() -> list[ModelInfo]`**, plus `pull_model`/`delete_model` on the port. All default to **raising `NotImplementedError`, never returning `[]`** — "cannot enumerate" must stay distinct from "enumerated nothing", because an empty list reads downstream as *no models present* and would block work that should proceed.
- **`ModelInfo` replaces raw `/api/tags` dicts.** The dict *was* the coupling: callers had to know Ollama's payload keys, so reaching for metadata made them Ollama-specific by construction.
- **`pull_model`/`delete_model` return `None`.** A provider response body handed back through the port is that same leak in another shape; no caller used it.

**`thinking_tokens: False` is honest, not incidental.** qwen3-family models emit `message.thinking` through Ollama by default, but this adapter never sends a `think` parameter and reads only `message.content` — so reasoning tokens are paid for and inseparable here (#410). Declaring True would tell a caller it could split them.

## Routes ask the port, not the class

`/api/v1/models/*` drop `isinstance(port, OllamaAdapter)` for `port.supports(capability)`. **The 503 stays** — it's the honest answer for a provider that genuinely cannot manage models — but a provider that *can* is no longer refused for having the wrong class.

## What is deliberately NOT migrated

`cycles.py`'s model-availability preflight. It has **15 commits in 60 days, last touched 19 hours ago**, and sits in the cycle-create path every 1.6 validation run hits.

It still calls `list_pulled_models`, so that method stays as an **explicitly deprecated shim** delegating to `list_available_models`. Removing it outright would have raised `AttributeError` *inside that preflight's broad `except Exception`*, degrading silently to warn-and-allow — a false green rather than a visible break. The shim's deletion is a named line item of the preflight migration, and it keeps its test coverage until then.

## Verification

**Conformance suite gains two dimensions** — capability honesty and `ModelInfo` shape (34 tests, up from 27). Mutation-verified, since a suite that passes first try is unverified:

| Mutation | Result |
|---|---|
| flip a declared capability to False | **fails** — declared-False but works |
| misspell a capability key | **fails** — key set no longer matches |
| surface blank-named models | **fails** — unnamed entries must be dropped |

**Live against real local Ollama, no mocks:** 7 models translated to `ModelInfo`, deprecated shim agrees with its source, `refresh_models` agrees with `list_available_models`, health up.

**Existing tests updated, not deleted.** `pull`/`delete` now assert the port contract (`is None`) while keeping their exact-wire assertions; the models API fixture drops its `__class__` isinstance fixup for a capability-declaring `LLMPort`. New coverage for the surface this creates: every management route 503s **naming the missing capability** when undeclared, and the code-defined registry route survives an incapable backend.

Regression suite: **6656 passed, 2 skipped**. ruff + `lint_test_quality.py` clean.

## Not deployed

The local stack is running v1.3.1 and I did not rebuild `runtime-api` — that restarts a service you've had up 11 days, and it's your call. The adapter half is live-verified above; a rebuild is the remaining step if you want the route half exercised end to end before merge.

Refs #313, #301 — neither closed. #313 needs the doctor half; #301 keeps its queue half.

### DIFF
--- a/adapters/llm/ollama.py
+++ b/adapters/llm/ollama.py
@@ -18,8 +18,8 @@ from squadops.llm.exceptions import (
     LLMModelNotFoundError,
     LLMTimeoutError,
 )
-from squadops.llm.models import ChatMessage, LLMRequest, LLMResponse
-from squadops.ports.llm.provider import LLMPort
+from squadops.llm.models import ChatMessage, LLMRequest, LLMResponse, ModelInfo
+from squadops.ports.llm.provider import LLMCapability, LLMPort
 
 logger = logging.getLogger(__name__)
 
@@ -74,6 +74,23 @@ class OllamaAdapter(LLMPort):
     def default_model(self) -> str:
         """Return the default model name."""
         return self._default_model
+
+    def capabilities(self) -> dict[str, bool]:
+        """Declare what this adapter actually does (#572's rule, see the port).
+
+        ``thinking_tokens`` is False deliberately, and it is not a statement
+        about the models: qwen3-family models emit ``message.thinking`` through
+        Ollama by default, but this adapter never sends a ``think`` parameter and
+        reads only ``message.content``. Reasoning tokens are paid for and cannot
+        be separated from content here (#410). Declaring True would tell a caller
+        it could split them.
+        """
+        return {
+            LLMCapability.MODEL_LISTING: True,
+            LLMCapability.MODEL_MANAGEMENT: True,
+            LLMCapability.STREAMING_USAGE: True,
+            LLMCapability.THINKING_TOKENS: False,
+        }
 
     async def _get_client(self) -> httpx.AsyncClient:
         """Get or create async HTTP client."""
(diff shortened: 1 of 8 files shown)


jasonLadd (commented):
reviewed

PR #801, opened 2026-08-08
test(llm): live-tier conformance — the same contract against a real server (SIP Atlas P3)

jasonLadd:
Completes **P3**, which specified "unit + live tiers" and shipped only unit.

Test-only plus a SIP doc update. No production file touched.

## What it is

The same contract, pointed at a real server. **This is the harness P4/P5 run against Atlas on the Spark, unchanged** — the adapter comes from `create_llm_provider`, so switching providers is an env change rather than an edit:

```
SQUADOPS_CONFORMANCE_PROVIDER=ollama \
SQUADOPS_CONFORMANCE_BASE_URL=http://localhost:11434 \
SQUADOPS_CONFORMANCE_MODEL=qwen2.5:3b-instruct \
  pytest tests/integration/llm -v
```

Skipped entirely when unset, so it never blocks a normal run. Not in `REGRESSION_DIRS` (unit-only), so the regression suite is untouched.

## Property-shaped, not value-shaped

A live server's model list, token counts and latencies are not knowable in advance. This tier asserts what must hold for **any** conforming provider — which is exactly what makes it reusable against Atlas instead of becoming a second Ollama test.

Covers what a fixture structurally cannot:

| Check | Why a mock can't do it |
|---|---|
| real streaming framing | the unit tier proves an adapter parses a *canned* stream; only a live server proves it parses **that server's** |
| `t/s` within physical plausibility | a duration-unit error yields a well-formed float ~1e9 out — it looks fine until you bound it |
| the configured model is present in the live listing | passing generation while the model is absent from the listing means the two surfaces disagree, and the **preflight** is the one that blocks work |
| `refresh_models` agrees with `list_available_models` | a caller's answer must not depend on which method it happened to call |
| real 404 bodies, real refused connection (discard port 9) | fixtures approximate these; the translation contract is about the real ones |

**Deliberately not exercised: `pull_model` / `delete_model`.** A conformance run must never delete an operator's weights or spend minutes pulling gigabytes. Their honesty is covered at the unit tier, where it costs nothing.

## A real bug in my own fixture, kept as a comment

A module-scoped adapter fixture caches an `httpx.AsyncClient` bound to the **first test's event loop**. With `asyncio_default_fixture_loop_scope=function`, every later test got `RuntimeError: Event loop is closed`.

Worse — and the reason the comment stays: **`refresh_models` swallows that in its blanket `except` and returns an empty cache**, so the failure *presented* as "the backend listed no models" rather than as a loop error. A function-scoped fixture removes the whole class for the price of one connection per test.

## Verification

- **Green against real local Ollama: 14 checks, ~4.5s.** Generations capped at 16 tokens — this is a contract check, not the SIP's §C.2 benchmark, and the two must not be confused.
- Verified it skips cleanly (14 skipped) with the env unset.
- Regression suite: 6656 passed, 2 skipped. ruff clean.

## SIP update

§6.1 claimed "the Mac can run the full live-tier conformance suite against local Ollama." That was an assertion; it is now a recorded result with the command that produces it.

Refs #313, #301 — neither closed.

### DIFF
--- a/sips/proposed/SIP-Atlas-Provider-Adapter.md
+++ b/sips/proposed/SIP-Atlas-Provider-Adapter.md
@@ -753,6 +753,23 @@ $ curl -s http://localhost:11434/v1/models
 {"object":"list","data":[{"id":"qwen2.5:14b","object":"model","created":...,"owned_by":"library"}, ...]}
 ```
 
+**The live tier exists and this claim is no longer hypothetical** — P3 shipped
+`tests/integration/llm/test_llm_port_conformance_live.py`, run green against real
+local Ollama (14 checks, ~4s). It names no vendor: the adapter comes from
+`create_llm_provider`, so pointing it at Atlas on the Spark is an env change, not
+an edit.
+
+```
+SQUADOPS_CONFORMANCE_PROVIDER=ollama \
+SQUADOPS_CONFORMANCE_BASE_URL=http://localhost:11434 \
+SQUADOPS_CONFORMANCE_MODEL=qwen2.5:3b-instruct \
+  pytest tests/integration/llm -v
+```
+
+Skipped entirely when those are unset, so it never blocks a normal run.
+Generations are capped at 16 tokens — this is a contract check, not the §C.2
+benchmark, and the two must not be confused.
+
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #802, opened 2026-08-08
docs(plan): M4 becomes question-gated review, not mandatory manifest review

jasonLadd:
Narrows M4 from "the operator approves every authored manifest" to "the pipeline stops only when the design asks a question."

## The evidence

V4 roll 2's framing bundle went through `progress_plan_review` and was approved. The approval note discussed fill slots and criteria counts — facts M2 and M3 had already proven deterministically. Meanwhile the manifest carried one genuinely unresolved question:

> **expansion-gating** — *PRD §4.1 specifies expansion only after core stability, but does not define the exact checkpoint or time threshold for triggering Tier 1 items.*

**Nothing surfaced it to the reviewer.** The design then implemented with 15/15 criteria verified, zero corrections, and a delivered app that installs, builds, boots and answers all five probes (`audit_delivered_app.py` PASS).

So the mandatory review contributed nothing, and the one thing a human uniquely held went unasked. A gate that gets rubber-stamped is worse than no gate — it manufactures the appearance of review, and a later reader cannot tell a considered approval from a reflex.

## The inversion

| | shape |
|---|---|
| Default | **No gate.** M2 + M3 approve the design; the cycle proceeds. |
| Blocks | **Only an unresolved-critical decision.** The note *is* the question; answering it is the whole interaction. |
| Everything else | Readable after the fact, never blocking. |

A narrowing of §3.4, not a contradiction: §5c.10 already had unresolved-critical entries landing in the gate note as questions. This promotes that from passenger to trigger. `RETURNED_FOR_REVISION` and revise-don't-re-roll are unchanged for when it does fire, and machinery stays nil per §5b Correction 4 — a policy entry with a conditional firing rule.

## What replaces the review

Removing a gate removes an observation. **FAY yield does not cover design quality** — roll 2 flattened the reference's typed `Participant` entity into an untyped list and still went green.

So: **sampling, not gating**, carried by the §5c.7 diagnostics already required (structural diff vs the reference, revision counts, M6's rejection taxonomy, surface counts). They move when quality moves. A design gets read when a number looks wrong.

## Two corrections recorded rather than left to drift

- **The divergence from accepted SIP-0103 §3.4**, written into the plan the way M0's premise correction was. The SIP stays accepted; it is a design commitment on main, not a living document.
- **#796 moved the freeze.** §5a says the manifest hash-freezes and the contract derives *"the instant the HITL gate approves."* It now derives at authoring acceptance, mid-framing, because the plan authors need it to bind — without that, roll 1's plan hit zero of nine fill slots. Approval gates whether implementation proceeds, not whether the contract exists.

## Gate consequence

Gate 2's exit criterion now requires the question path to have **fired**, not merely to exist — a release where it never triggered has not shown it works. V5 carries the manifest that makes it fire.

### DIFF
--- a/docs/plans/1-6-0-authorship-plan.md
+++ b/docs/plans/1-6-0-authorship-plan.md
@@ -159,7 +159,7 @@ Slice 1, SIP-0096's inert core, and M0a here):
 - Ordering M1 first inverts that: every early gate defect would surface *as* an authoring
   failure, and the window's attribution (M6) would inherit the confusion.
 
-M4 (the HITL gate) follows M1 because there is nothing to review until something is
+M4 (the review path) follows M1 because there is nothing to ask about until something is
 authored; M5 (provenance) is last because it stamps a pipeline whose shape is settled by
 then.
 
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #804, opened 2026-08-08
feat(scaffold): authoring provenance on the manifest (#803, M5)

jasonLadd:
A system-owned `provenance` block recording **how a manifest came to be**: authored mode, the cycle and task that wrote it, the attempt count, and the **classified reason for each revision** drawn from M6's taxonomy.

```yaml
provenance:
  mode: authored
  cycle_id: cyc_77cbb5aab7ca
  task_id: task-run_516f0fed-003-development.author_manifest
  attempts: 2
  revisions:
    - attempt: 1
      classes: {authoring_defect: 1}
      proofs: [source_prd]
```

The reasons **cost nothing to produce** — `DevelopmentAuthorManifestHandler` already accumulates an `AuthoringOutcome` per attempt for its corrective feedback, and M6 already classifies every finding. M5 records what the loop had computed. That's the payoff for having built M6 before M1.

Why classes rather than a count (§5c.5): *"took three tries"* cannot say whether those were schema failures, winnability rejections, or the author's own refinement — three causes with opposite remedies. One vocabulary, already shared with the gate rejection note and B1's baseline.

## Four properties, each pinned

- **Excluded from `_canonical()`**, so stamping cannot move `content_hash()` — the value the verification contract binds, M0a's guard pins, and every bind-mode cycle checks. This is the M2 `decisions` lesson: a provenance-only edit that moved the hash would break all three at once while looking like bookkeeping. The reference's `bb472e267e53` is asserted **literally**, not just self-consistently, so the manifest and its test can't drift together. A parametrized test also proves the *contents* are ignored — two runs that authored the same design in different attempt counts must hash identically, or the golden benchmark starts comparing luck instead of designs.
- **Appended, never re-emitted.** Round-tripping through the parser would discard the comments and grouping a design document exists to carry. Asserted by `stamped.startswith(reference)`.
- **Optional, because absence is the signal.** Present ⇒ a squad authored it and the block says how; absent ⇒ it didn't come from this system. The reference instance has to keep parsing, and does.
- **Author-supplied blocks are discarded, not rejected.**

## One correction to the issue's own design

The issue proposed a schema-gate rule rejecting an author-written `provenance` block. **That would have false-rejected at net-b**, where the same gate re-runs on the *stamped* manifest and a provenance block is entirely legitimate. Discarding is also the proportionate response: the field is system-owned, so overwriting costs nothing, while a rejection would spend a revision on bookkeeping the system replaces anyway. Logged at INFO when it happens.

## Name collision resolved

`PROOF_PROVENANCE` → **`PROOF_SOURCE_PRD`**. It meant *"`source_prd` is declared"* and its rule id was already `name-the-source-prd`, so introducing a real `provenance` block would have given one word two meanings in one module. It now names the field it checks, like `PROOF_STATUS_DECLARED` and `PROOF_TESTID_COVERAGE` around it. The rename was caught by the existing binding tests in `manifest_authoring_rules` and M6 — which is what those tables are for.

## Scope cut, declared

§5c.5 also asks for *"after any operator edit at the gate — the edit's own record."* **Out**, with the trigger named: M4 was narrowed to question-gated review (#802) and no operator-edit path exists to record. Building the record first would be machinery for a hypothesis.

## Verification

12 new tests. Full regression green: 7,055 unit / 6,669 regression.

Closes #803


jasonLadd (commented):
reviewwed

PR #805, opened 2026-08-08
feat(llm): VLLMAdapter — second provider, OpenAI-compatible dialect (SIP Atlas P6)

jasonLadd:
Implements **P6** of `sips/proposed/SIP-Atlas-Provider-Adapter.md`, plus the two SIP amendments this work produced.

All files cold against the 1.6 line: a new adapter, `factory.py` (0 commits in 5 months), and test files added by this arc.

## The design changed — §3.5a amended

An earlier revision proposed a shared `OpenAICompatibleAdapter` for vLLM and Atlas. **Withdrawn**, for two reasons that only appeared once the dialect was examined rather than assumed:

1. **Atlas is already expected to diverge.** §10.2 asks it to emit duration fields the OpenAI shape lacks — so a shared class would need provider conditionals on day one, which is the identity-branching #559 bans. Separate types *are* the polymorphic answer.
2. **It inverts the no-gold-plating rule.** Extracting a base on the expectation that Atlas matches vLLM builds shared structure on the very §10.2 assumption this SIP otherwise refuses to rely on. Extraction happens *after* two implementations exist and overlap is demonstrated.

So: `OllamaAdapter`, `VLLMAdapter`, `AtlasAdapter` — independent, each declaring its own capabilities. **The conformance suite is what makes that affordable**: one defect, one failing test, every adapter.

## §10.2 upgraded from question to ask — with measurements

Probed the same server serving both dialects:

| Field | Ollama native `/api/chat` | that server's `/v1` shim |
|---|---|---|
| token counts | yes | yes |
| `eval_duration` (decode) | yes | **absent** |
| `prompt_eval_duration` (prefill) | yes | **absent** |
| `load_duration` (residency) | yes | **absent** |
| `total_duration` | yes | **absent** |

Those four are exactly §3.6.1's `prefill_ms` / `load_ms` / `total_ms`. Two consequences now recorded in the SIP: **Ollama keeps its native adapter** (the OpenAI shape would delete the attribution half of the A/B *on the baseline arm* and drop `model_management`), and **if Atlas ships plain OpenAI, the A/B is asymmetric**. Atlas is in-house — so: emit the durations, settled before P4 opens.

## The adapter

Built against a live OpenAI-compatible server, not documentation: SSE `data: {json}` frames terminated by `data: [DONE]`, `delta.content` streaming vs `message.content` not, `/v1/models` keyed on `id`, 404 + `{"error":{...}}` for unknown model. `generate()` is a single-turn chat because `/v1/completions` is deprecated upstream and not universally served.

Declares `model_management: False` (no pull/delete in the shape — callers get an honest 503) and computes tokens/sec from wall-clock, since there is no timing field. **That rate includes prefill where Ollama's is decode-only** — not interchangeable, which is why §3.6.1 makes wall-clock the decision field and rate a diagnostic.

## Adding a second adapter did its job immediately

**Four shared assertions had `qwen2.5:7b` baked in.** The "provider-neutral" suite was secretly Ollama-shaped — precisely the trap the 1.5 plan named. Expectations moved off parallel name-keyed dicts onto `AdapterCase`, so one entry fully describes a provider and Atlas becomes one object rather than five dict edits.

## Mutation-verified

Eight deliberate defects against the new adapter: SSE prefix mishandling · zero-filled usage · dropped 404 translation · wrong `delta` key · over-declared `model_management` · wrong listing field · absent `[DONE]` handling · missing `total` derivation.

**The last two initially survived** — the shared fixture never exercised them. Now covered in `test_vllm_adapter.py`, which holds dialect specifics *deliberately outside* the shared suite: forcing Ollama to fake a `[DONE]` frame would enshrine vLLM's transport as the contract, the mirror image of the mistake just fixed.

## A correction carried into the code

An earlier probe suggested Ollama's `/v1` ignores `stream_options.include_usage`. That was `head -3` SIGPIPE-ing curl and truncating the stream before the usage frame. **It honors it** — the frame arrives with empty `choices`, exactly as the fixture models. Docstrings carrying the wrong claim are corrected.

## Verification

- **Live against a real OpenAI-compatible server:** `prompt=32, completion=6, total=38, t/s=20.66` — internally consistent.
- **Live conformance tier green for both providers**, 14 checks each.
- Unit suite: 130 in `tests/unit/llm/`. Regression: **6696 passed, 2 skipped**. ruff + `lint_test_quality.py` clean.

Nothing selects this adapter: `LLMConfig` still has no `provider` field (that's P2), so the default path is untouched.

Refs #313, #301 — neither closed.

### DIFF
--- a/adapters/llm/__init__.py
+++ b/adapters/llm/__init__.py
@@ -2,14 +2,17 @@
 
 Provides implementations of LLM ports:
 - OllamaAdapter: Local Ollama LLM server adapter
+- VLLMAdapter: vLLM OpenAI-compatible server adapter
 
 Part of SIP-0.8.7 Infrastructure Ports Migration.
 """
 
 from adapters.llm.factory import create_llm_provider
 from adapters.llm.ollama import OllamaAdapter
+from adapters.llm.vllm import VLLMAdapter
 
 __all__ = [
     "OllamaAdapter",
+    "VLLMAdapter",
     "create_llm_provider",
 ]
--- a/adapters/llm/factory.py
+++ b/adapters/llm/factory.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from adapters.llm.ollama import OllamaAdapter
+from adapters.llm.vllm import VLLMAdapter
 from squadops.ports.llm.provider import LLMPort
 
 if TYPE_CHECKING:
@@ -26,7 +27,7 @@ def create_llm_provider(
     """Create an LLM provider.
 
     Args:
-        provider: Provider name ("ollama")
+        provider: Provider name ("ollama" or "vllm")
         secret_manager: Optional secret manager for resolving secret:// refs
         base_url: Base URL for the LLM server (may be secret:// ref)
         default_model: Default model to use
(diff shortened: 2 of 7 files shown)


jasonLadd (commented):
reviewed

PR #808, opened 2026-08-09
feat(cycles): the manifest gate stops only when the design asks a question (#807, M4)

jasonLadd:
The build item behind the plan section rewritten in #802 — that PR was docs-only.

A manifest declaring no unresolved decision has already been approved by the deterministic gates, so the cycle proceeds. One that declares a question stops, and **the question is what the operator is shown** rather than a bare "approve?".

## The evidence is also the fixture

V4 roll 2's gate was approved with a note repeating what M2 and M3 had already proven, while the manifest's one real question reached nobody:

> **expansion-gating** — *PRD §4.1 specifies expansion only after core stability, but does not define the exact checkpoint or time threshold for triggering Tier 1 items.*

That manifest is now committed as `interface_manifest_roll2.yaml` and is the fixture that must stop a gate. A guard test asserts it declares a question **and** that roll 1's manifest (already in the fixture directory, all decisions resolved) declares none — otherwise pointing the suite at the wrong file would make it pass while proving the opposite.

That mistake actually happened during the build: I first wired the tests to roll 1's manifest and they failed `assert 0 == 1`. The guard exists because of it.

## Shape

- **Keyed on the design, never on who wrote it.** Both rails are read — a manifest the run authored, and an operator-seeded one on `plan_artifact_refs`. Reading only the first would leave seeded cycles with a mandatory review and authored ones without, which is behavior correlated with authoring mode and exactly what **Guard 1a** forbids.
- **Three-valued on purpose.** No manifest at all returns `None` and keeps today's behavior, because a plan-only cycle's gate is a *plan* review and M4 never claimed to touch it.
- **The pass-through is recorded**, with its own `decided_by`. The argument for question-gating is that a rubber stamp can't be told from a considered approval; a machine pass-through leaving no row would reproduce that defect from the other side.
- **The synthesized decision runs through the same exhaustive dispatch** a human's answer does, so a pass-through cannot reach a path an approval would not. (My first attempt used `continue`, which in that loop means *re-run the same workload* — an infinite loop, caught by the suite timing out.)

## The "unresolved-critical" reading

§5c.10 and the plan both say *unresolved-critical*, but `Decision` is `id / choice / warrant / unresolved / question` — there is no criticality field.

Reading taken: **`unresolved: true` is itself the assertion that it matters.** An author that declines to guess and states the question has already judged it worth surfacing. A criticality field would ask them to grade their own question — the self-assertion problem M5 just removed from provenance — and would be a schema change that should ride the #772/#795 boundary. If authors declare unresolved liberally once it stops the pipeline, that is measurable and criticality can be added with evidence.

## Two guard families pointed back at their intent

- The #494/#496/#509 inverse guards asserted *"no gate decision at all"* when they meant *"plan validation did not auto-reject"*. Rewritten to assert the rejection they were written for, rather than flipped to match new behavior.
- The emit-site count moved 46 → 47.

## Disclosed behavior change

A **seeded** cycle whose manifest declares no open question now auto-approves where it previously waited for a human. The deliverable is unaffected — approval doesn't influence what gets built — and it removes human-latency variance from V8's seeded control re-run. But it is a change to the control configuration, so it is stated here rather than absorbed.

## Verification

12 new tests. Full regression green: 7,107 unit / 6,721 regression. Live at V5, which also needs B1 emitting.

Closes #807

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,9 @@ step that revises against the gates, replacing the ungated proposer-side emissio
 relocates, and **#796** — the fix V4 roll 1 exposed: an authored manifest now derives and
 pins its contract mid-framing, so the plan authors bind to the design their own squad just
 wrote instead of inventing paths, and every contract-gated net engages. M4 (the human
-manifest gate, narrowed to question-gated review) and **M5 authoring provenance (#803)** —
+manifest gate, narrowed to question-gated review — #807: the gate stops only when the
+design declares an unresolved decision, and the question itself is what the operator is
+shown; a design that asks nothing records a distinguishable system approval and proceeds) and **M5 authoring provenance (#803)** —
 a system-owned block recording mode, cycle, task, attempt count and the *classified* reason
 for each revision, excluded from the manifest's canonical projection so recording how a
 design was written can never move the hash its contract binds.
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -46,9 +46,16 @@ from squadops.cycles.acceptance_evaluation import resolve_check_stack
 from squadops.cycles.agent_config import build_agent_resolver
 from squadops.cycles.build_completeness import compute_missing_required_files
 from squadops.cycles.checkpoint import RunCheckpoint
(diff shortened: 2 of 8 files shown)


jasonLadd (commented):
reviewed

PR #810, opened 2026-08-09
feat(cycles): the pre-memory rejection baseline (#809, B1)

jasonLadd:
The one item in 1.6 whose omission would be **permanent**. Cross-Cycle Memory's value claim is that recurrence of the same mistake falls; testing that needs a picture of how often each class occurs *before* memory exists, and the moment memory is live that picture is unrecoverable.

## Measuring first made this much smaller

Three of B1's four inputs were already durable — two of them because M5 and M6 landed ahead of it:

| input | durable before this PR? |
|---|---|
| authoring failure classes | ✅ M5's `provenance.revisions[].classes` |
| attempts consumed | ✅ `provenance.attempts` |
| framing re-rolls consumed | ✅ derivable — one framing run per re-roll |
| **plan-validation classes** | ❌ **nothing recorded them** |

So the work is the fourth. A rejected plan recorded *what* went wrong as a joined error string and never *which class* — and re-parsing that prose at window time is the failure mode the plan warns against ("one vocabulary, not a free-text field that each reader parses differently"), not the fix.

## Shape

- **Classified at the point of rejection.** The gate calls its validators one at a time, so the producing validator is known exactly where the errors are collected — no validator signature changes, no error tagging, no prose to parse back.
- **The instrument cannot alter the experiment.** `collect()` is a pass-through around the existing `errors.extend(...)`, returning its input unchanged, so a counting bug cannot change which plans get rejected. Asserted directly.
- **The vocabulary is the one authors are taught** — `ImplementationPlan.validate_*`, which `plan_authoring_rules` keys on and the authoring-rules asset is bound to. A baseline speaking different names could not answer *"did teaching this rule reduce it?"*, which is the only question it exists to support.
- **A `rejection_record` artifact, not a schema change.** Durable, typed, attached to the run that was rejected, no migration. A jsonb column on `cycle_gate_decisions` would be a second answer to where run facts live.
- **Never raises.** A baseline is a record, not a gate: losing one cycle's bookkeeping is a gap in a dataset; failing the rejection path would turn it into a lost cycle.
- **Both dimensions**, per the plan — recurrence, and time-to-resolution from attempts + re-rolls. Memory that halves recovery cost without changing recurrence is a real win a recurrence-only baseline would score as zero.

## The aggregation is a script, deliberately

What is impossible afterwards is the **data**, not the report. With the inputs structured and durable, the baseline can be produced at any time — including retrospectively over cycles already run, which means V4 rolls 1 and 2 are already partly covered. `scripts/dev/emit_rejection_baseline.py` reads only stored facts.

**Read by nothing in 1.6**, exactly as the plan specifies.

## Verification

15 new tests, including a **replay over the V4 fixtures**: roll 1's real plan through the real validator reproduces its known dual-claim class end to end, so the shape is proven on observed data rather than a hand-made record.

Full regression green: 7,122 unit / 6,736 regression / 22 architecture.

Closes #809

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ shown; a design that asks nothing records a distinguishable system approval and 
 a system-owned block recording mode, cycle, task, attempt count and the *classified* reason
 for each revision, excluded from the manifest's canonical projection so recording how a
 design was written can never move the hash its contract binds.
+**B1 (#809)** closes the one item whose omission would have been permanent: rejections are
+now classified at the moment they happen — plan-validation by producing validator, authoring
+by M6 class — so the pre-memory recurrence baseline Cross-Cycle Memory will be measured
+against is capturable. Read by nothing in 1.6, deliberately.
 Plan: `docs/plans/1-6-0-authorship-plan.md`.
 
 ## [1.5.0] — 2026-08-07
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 
 import asyncio
 import dataclasses
+import json
 import logging
 import os
 import time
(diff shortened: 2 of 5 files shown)


jasonLadd (commented):
Reviewed

PR #813, opened 2026-08-09
docs(plan): the seeded control becomes conditional; add Track S's missing verification point

jasonLadd:
Three amendments, all decided after V5.

## 1. The seeded control is no longer a cut item

**Its unique coverage is a strict subset of an authored cycle's.** Seeded and authored runs differ only in the framing half — below framing they are one code path *by construction* (Guard 1a), and #796 made that literal by having an authored cycle derive its contract and become bind mode. Expansion, contract binding, plan validation, dispatch, checks and corrections are exercised identically by both.

So a seeded run adds no coverage. It adds **lower variance on the shared part** — precision-per-cycle, not reach — and a multi-roll authored window already has N observations of the same machinery.

Two facts decided it:
- **The control has never fired.** V2 banked one observation (43/43) and nothing has been compared against it. A detector that has never fired is either useless or means nothing regressed, and one observation can't distinguish those.
- **It measures a configuration the release doesn't ship.** 1.6's claim is authored mode; a scheduled detector for the fill-in-a-supplied-design path spends the cut's budget proving the 1.4 capability still works.

**What replaces it** is what most of its value already was — Guard 1b, replay zero-diff, and M0a's equivalence guard, all offline.

**V8's seeded re-run becomes conditional.** *"Is it the design step or the machinery?"* is the one question a seeded run uniquely answers, and it only arises when a window disappoints. Retire the schedule, keep the capability — deleting it would remove the instrument for precisely the moment you'd most want an answer and least want to be building one.

## 2. The frozen pair is a fixture, not an anchor

Its entire dependency surface: two sha256 constants in M0a's test, one committed fixture, two vault artifact ids. Re-basing is a fixture update, not a measurement event.

The prohibition on that hash moving was always about it moving **incidentally** — a provenance edit silently invalidating a binding, the #494 class — never about a deliberate versioned re-base. Recorded because the constraint had been overstated, including by me in this thread, and because it unblocks #795 and #772, whose fixes move the hash.

## 3. Track S had no verification point at all

All eight V-points were M-track. Building a second stack — new expander, contract derivation, test-runner and sandbox integration — had none, which would have surfaced when S3 was half-written and someone asked what its schema had been validated against.

Added **VS**, gating S3: does the Node/TS stack expand, derive a satisfiable contract, and carry a cycle end to end? The blueprint schema is written against two stacks; one of them being unproven makes it a schema against one stack and a hope.

## Also recorded

The remaining sequence — **#811 + Gate 2 wording → #812 → S1 → Node stack → VS → S3 + S5 → riders → V6 → pre-registration → freeze → V7 → cut** — and the two decisions still open before pre-registration: whether the window records contract size per roll, and whether the worked example should be made unlike the benchmark domain.

### DIFF
--- a/docs/plans/1-6-0-authorship-plan.md
+++ b/docs/plans/1-6-0-authorship-plan.md
@@ -139,6 +139,24 @@ while 1.6's first deploy window is still far out, or any one of them turning liv
 
 ## Track M — Squad-Authored Manifest (SIP-0103)
 
+### Remaining sequence *(recorded 2026-08-09, after V5)*
+
+Track M and B1 are complete. What is left, in order:
+
+> **#811 (close the revision loop) + Gate 2 wording → #812 (gate attribution) → S1 →
+> build the Node/TS stack → VS → S3 + S5 → riders → V6 → pre-registration → freeze → V7 → cut**
+
+Why M's leftovers come first: #811 is a behavioural gap in the gate that just shipped — the
+system can ask a question and cannot act on the answer — and deferring it past an entire
+second stack is how it becomes 1.6.1, then 1.7, then a thing someone rediscovers from a log
+line. #812 is small and blocks V7 from being able to state its own configuration.
+
+**Two decisions still open before pre-registration**, both from V4/V5 and both cheap now:
+whether the window records **contract size per roll** (verification depth varies with an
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #814, opened 2026-08-09
fix(cycles): an answered design question reaches the author as a revision (#811)

jasonLadd:
The gate could ask and could not be answered. `RETURNED_FOR_REVISION` stopped the workload sequence and revision required a manual retry run — so at V5 a human could be asked *"the PRD does not define the expansion checkpoint — which is it?"* and their answer would go nowhere. A system that asks a question it cannot act on is the rubber stamp M4 replaced, wearing a better costume.

## The revision replays the prefix it does not invalidate

The first cut of this re-executed the **whole** framing workload — 58 minutes to change one document — and I wrongly recorded that as an architectural limitation. It wasn't. SIP-0101 Slice 3 already restores a source run's boundary checkpoint into a new run's namespace and suppresses those dispatches.

**One thing genuinely blocked it:** `translate_checkpoint_for_replay` rebinds a deterministic id's run prefix and **raises** on anything else — and framing task ids were `uuid4().hex`, so no framing checkpoint could ever be translated onto a new run. Framing now joins implementation in the deterministic namespace (`task-{run}-{index}-{type}`). Ids stay unique within a run and now carry position and type, which is also what lets a boundary be selected by task type rather than by guesswork. **Zero test churn** from the change.

## Where a revision restarts, and why it isn't the manifest

`development.design_plan`, not `development.author_manifest`.

A reviewer's note is a change to the **design**. Restarting at the manifest alone would leave `technical_design.md` describing an interface the revised manifest no longer has, and no later reader could tell which one the squad meant. So the design stage now receives the revision context too and answers the note; research and the objective frame are upstream of the design, unaffected by a note about it, and restore from the checkpoint.

## Degrades, never fails closed

Replay fails closed because a replay that silently ran the full plan would claim a saved prefix it never earned. **A revision claims nothing** — the full re-execution is simply the slower correct answer, and it is exactly what this did before the optimisation. An unresolvable boundary (pruned checkpoints, or a source run predating deterministic ids) logs and re-runs everything.

## Budget: `manifest_max_attempts`, counted separately

Not `framing_max_rerolls` — it **defaults to 0** and the validated profiles don't set it, so sharing it would have shipped the feature switched off, and it conflates a stochastic fault the system retries with a deliberate human instruction. `manifest_max_attempts` is what §5c.6 specifies and is already CRP-allow-listed.

## Revise, don't re-roll

The context carries the notes **and the prior manifest**. Given only a note the author re-derives everything, so decisions that were accepted come back changed — the fay-6 new-dice failure one level subtler.

## Two defects the guards caught mid-build

- **The golden context guard** flagged that `development.design_plan` now receives rejection context, and told me to regenerate in the same PR so it reads as a behaviour change. Done, in the same commit.
- Regenerating exposed a wording defect I'd introduced: **both revision appendices also render on *system* re-rolls**, where nobody human returned anything. Neither may say "a reviewer asked". Both now name what came back without asserting who sent it.
- A test also caught me mis-writing the deterministic id format (`run_id[:12]`) in a fixture.

## Two guard families pointed back at their intent

- `test_returned_for_revision_stops_orchestration` guarded #466's false-approve as *"no run may be created"*. Post-#811 a run **is** created — a framing revision. Rewritten to assert what it always meant: framing, never implementation. Flipping it to match new behaviour would have quietly retired the #466 guard.
- Emit-site count 47 → 48.

## Verification

14 tests in the new file plus a rewritten guard and a budget-bound test. Full regression green: 7,147 unit / 6,761 regression / 22 architecture.

Closes #811

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,9 +27,12 @@ a system-owned block recording mode, cycle, task, attempt count and the *classif
 for each revision, excluded from the manifest's canonical projection so recording how a
 design was written can never move the hash its contract binds.
 **#811** closes the revision loop: a design question the gate asks can now be *answered* —
-`RETURNED_FOR_REVISION` re-executes framing with the reviewer's notes and the prior manifest
-as authoring context, bounded by `manifest_max_attempts`, instead of stopping the sequence
-for a manual retry run.
+`RETURNED_FOR_REVISION` re-executes framing with the notes and the prior manifest as
+authoring context, bounded by `manifest_max_attempts`, instead of stopping the sequence for a
+manual retry run. It **restores the prefix the note does not invalidate** via SIP-0101's
+checkpoint translation — enabled by framing task ids becoming deterministic in the same
+change — and re-runs from the technical design, so the design answers the note rather than
+describing an interface the revised manifest no longer has.
 **B1 (#809)** closes the one item whose omission would have been permanent: rejections are
 now classified at the moment they happen — plan-validation by producing validator, authoring
 by M6 class — so the pre-memory recurrence baseline Cross-Cycle Memory will be measured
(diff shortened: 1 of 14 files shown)


jasonLadd (commented):
reviewed

PR #815, opened 2026-08-09
feat(api): a gate decision records who made it (#812)

jasonLadd:
## The issue's premise was wrong, and the reality was worse

I filed this saying `decided_by` "is derived from the authenticated principal." It wasn't derived from anything — `runs.py:248` was the hardcoded literal `"system"` with a `TODO: extract from auth context` beside it. Measured:

```
system                 | approved              | 140
system:plan_validation | rejected              |  34
system                 | rejected              |   4
system                 | returned_for_revision |   1
```

**All 140 human approvals in this project's history carry the word that means "no human was involved"** — in the same namespace the machine paths use for `system:plan_validation` and `system:no_open_questions`. Not indistinguishable by convention: colliding in vocabulary.

## The fix

`decided_by` now composes `{actor}:{principal}` from the request identity, using the `IdentityType` vocabulary that already existed (`human` / `service`) rather than inventing one. Machine paths keep `system:*` unchanged.

**An agent can declare itself** — `decided_by_kind` on the API, `--as-agent` on the CLI. An agent authenticating as a person is invisible to the token, so declaring is the only mechanism available. **Declared, never inferred:** a guess from timing or note style would be usually right, which is exactly what would make the field unreliable in the cases anyone would actually check it.

Only `human` and `agent` are declarable. `service` comes from the token and `system` belongs to the executor — a caller who could claim either could dress an agent's decision as a machine's, which is the same laundering from the other direction.

No identity at all records `unattributed` rather than being promoted to a person. Auth can be off in a local deployment, and naming what is known beats inventing what is not.

## The legacy rows are left alone, deliberately

`is_machine_decision()` reads the `system:` prefix so a future mechanism is covered the day it ships — and deliberately does **not** match the bare legacy `"system"`. Those 140 rows are decisions people actually made; the field simply couldn't say so. Matching them as machine decisions would launder retroactively the precise confusion this ends.

## Why it's load-bearing, not cosmetic

- **The authored-mode window (V7)** must be able to state whether its design questions were adjudicated by a human or an agent. A banked number describing "squad-authored design with human adjudication" means something different from one describing "…with an LLM adjudicating its own squad's questions".
- **1.8's Cross-Cycle Memory** seeds itself from gate and revision history. A corpus filing agent answers under a human's name teaches the wrong thing about who resolves what.

## Honest limit

An **undeclared** agent still records as the token's type. The system cannot detect an agent, only offer a way to declare one and record it faithfully. That's a policy problem, not a mechanism problem — what changes here is that being honest becomes *possible*, which it was not. Recorded on the issue.

## Verification

17 new tests. Full regression green: 7,164 unit / 6,778 regression / 449 API+CLI.

Closes #812

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,11 @@ manual retry run. It **restores the prefix the note does not invalidate** via SI
 checkpoint translation — enabled by framing task ids becoming deterministic in the same
 change — and re-runs from the technical design, so the design answers the note rather than
 describing an interface the revised manifest no longer has.
+**#812** makes a gate decision say who made it. `decided_by` was a hardcoded `"system"` with
+a TODO beside it, so all 140 human approvals in the project's history carried the word that
+means *no human was involved* — in the same namespace the machine paths use. It now composes
+`{actor}:{principal}` from the request identity, and an agent can declare itself
+(`--as-agent`) since a token cannot tell.
 **B1 (#809)** closes the one item whose omission would have been permanent: rejections are
 now classified at the moment they happen — plan-validation by producing validator, authoring
 by M6 class — so the pre-memory recurrence baseline Cross-Cycle Memory will be measured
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #816, opened 2026-08-09
refactor(scaffold): one registration per stack, not five scattered facts (S1)

jasonLadd:
Opens **Track S**. "A stack" was an identifier indexing four module-level dicts plus one function with the answer written inline, so a new stack had to remember to appear in five places — and forgetting produced a **plausible wrong answer** rather than an error.

## The trap, concretely

```python
def fill_slot_paths(manifest):
    if manifest.stack not in _EXPANDERS:      # is it registered?
        raise ValueError(...)
    views = tuple(f"frontend/src/views/{r.view}.jsx" for r in manifest.frontend.routes)
    return ("backend/routes.py", *dict.fromkeys(views))   # ...FastAPI's answer, always
```

A registered Node/TS stack would have been told to fill `backend/routes.py` and `.jsx` views — and every downstream consumer would have agreed with it: the contract's covered files, frozen-artifact ownership, `fill_slot_signature`. **The guard it passed asks a different question than the one being answered.**

## What changed

One `ScaffoldStack` per stack now carries all five facts — `expand`, `fill_slots`, `qa_test_namespace`, `harness_entry_modules`, `check_stack`. `acceptance_evaluation`'s `_BUILD_PROFILE_CHECK_STACKS` is sourced from it rather than declaring the known-stack set a second time, so *"which stacks does this system know?"* has one answer.

**Named `ScaffoldStack`, not `StackBlueprint`, carrying today's fields only.** The Stack Blueprint SIP is deliberately unaccepted until two real stacks exist — its own acceptance gate says generalising from one instance "produces the FastAPI contract with generic field names, worse than no contract". A consolidation that quietly minted that vocabulary would prejudge the question the SIP exists to answer against evidence.

## The plan's exact test, verified

| | result |
|---|---|
| `expand()` output | unchanged, 19 files |
| manifest `content_hash` | **`bb472e267e53d5ad`** — unmoved |
| derived contract sha256 | **`7622f570c949fe95`** — unmoved |
| M0a reference guard | green |
| contract lint gate | green |
| regression | 6,789 |

## The headline test

`test_a_second_stack_does_not_inherit_the_first_stacks_fill_slots` registers a second stack and asserts it gets **its own** slots. That's the bug this removes — and nothing could have caught it while only one stack existed, which is precisely why S1 comes before building the Node stack rather than after.

Also guarded: a half-declared stack (registered as expandable while silently answering `()` elsewhere), and the refusal drifting between accessors so an unknown stack raises in one place and returns `()` in another.

## Verification

11 new tests. Full regression green: 7,175 unit / 6,789 regression.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,12 @@ means *no human was involved* — in the same namespace the machine paths use. I
 now classified at the moment they happen — plan-validation by producing validator, authoring
 by M6 class — so the pre-memory recurrence baseline Cross-Cycle Memory will be measured
 against is capturable. Read by nothing in 1.6, deliberately.
+**Track S opens with S1**: the five per-stack facts — expander, fill slots, QA namespace,
+harness boundary, check-stack vocabulary — collapse from four module-level dicts plus one
+inline answer into a single `ScaffoldStack` registration. Pure refactor; the reference
+manifest and contract hashes are unmoved. It removes a silent-omission trap: `fill_slot_paths`
+guarded on whether a stack was *registered* and then returned FastAPI's slot map to whoever
+asked, so a second stack would have inherited `backend/routes.py` with nothing objecting.
 Plan: `docs/plans/1-6-0-authorship-plan.md`.
 
 ## [1.5.0] — 2026-08-07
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #819, opened 2026-08-09
docs(plan): correct nine unverified claims found by a source sweep

jasonLadd:
Docs-only. Every load-bearing code claim in `docs/plans/1-6-0-authorship-plan.md` (953 lines) checked against source. **Nine were wrong. Seven overstated what exists.**

The sweep was prompted by a defect in this plan's own S4 section: it described what would happen to a second stack's typed checks without ever opening the contract emitter, and got the failure mode wrong in the reassuring direction. That became #818. The obvious next question was how many other claims in the document had the same origin.

## What was wrong

| Line | Claim | Source |
|---|---|---|
| §M0 scope table | `endpoint_defined` **per fill slot**; `field_present` derivable from `entities` | emitted **once**, routes slot only (measured: 1 across 4 `fill_files`). `field_present` is **never emitted** — the deriver's whole vocabulary is 7 checks |
| §M3 proof table | status completeness = **per-endpoint** `success_status` | `manifest_gates.py::_status_findings` is *"deliberately narrow"* — collection POST only. Its docstring: requiring it everywhere "would reject the reference manifest" (1 of 5 endpoints declare it) |
| §S4 | typed checks would **silently skip** stack #2 | `is_check_applicable(*, "*.ts")` is **False** for all four named checks → `manifest_gates.py:252` rejects, `task_plan.py:462` strips with a WARNING. Not silent — and the real failure (#818) is upstream and worse |
| §seeded control | replay zero-diff "(#734 method)" | #734 is *"workspace-revision provenance — stamp revision identity on every acceptance verdict"* |

## What was presented as existing and is not

| Line | Claim | Reality |
|---|---|---|
| Guard 1a | "**pinned** by an architecture test" | No test covers authoring mode. The **invariant holds** — `authors_interface_manifest` has exactly one production call site, `task_plan.py:394`, inside framing — but nothing pins it. `test_plan_gate_seams.py` was cited as the *precedent for the shape* and reads as the thing itself |
| Guard 1b | listed at "offline, milliseconds" among mechanisms replacing the seeded control | **No test exists.** M0a's own docstring names Guard 1b as a *dependent*: "Guard 1b **requires** the reference manifest to produce byte-identical downstream artifacts" |
| replay zero-diff | same table | No script, no test, wrong issue cited |
| §M5 | provenance records "any **operator edit's** own record" | `Provenance` is `mode, cycle_id, task_id, attempts, revisions`. No operator-edit field — and **no operator-edit code path exists** anywhere to record |

## The seeded control — retirement stands

Two of the three mechanisms justifying the retirement are unbuilt, which reopened the question. **Owner ruling: the retirement holds, and the control stays a comparison instrument available on demand rather than a cut checkbox.**

The correction genuinely does not weaken the retirement, because that table was never the load-bearing argument. The load-bearing argument is structural — *the control's unique coverage is a strict subset of an authored cycle's*, since below framing the two are one code path by construction and #796 made it literal by having an authored cycle derive its contract and become bind mode. That argument survives the sweep intact, and an unbuilt mechanism cannot turn a subset into a superset.

What it does change is honesty in the meantime: until Guard 1b exists, nothing offline isolates a provenance change from a transformation defect.

## Scope changes this makes

- **Guard 1a's architecture test and Guard 1b become explicit cut-gate build items** rather than assumed fixtures. Both are small.
- **Replay zero-diff is dropped** — it never existed, and inventing it at the tail of a release is new scope, not a guard.
- **#818 moves ahead of the Node stack** on the S track. S4 stops being a disclosure obligation and becomes the residual coverage question after #818 lands.

## The standing rule this adopts

A load-bearing claim about code behavior in this document carries a `file:line`, or it is marked unverified. **Bounding claims carry it first** — *narrower*, *only*, *already*, *merely*, *does not affect* — because those are the ones that stop further investigation. An alarming claim gets checked by whoever has to act on it; a reassuring one closes the question, and if it is wrong nobody finds out until the code runs.

All nine defects shared one origin: a claim about code sourced from a document *about* the code — this plan's earlier sections, the SIP's prose, an issue number from memory, a test file cited as precedent. **Not one was a misreading of source.** Having a secondary source that describes the code turned out to be the highest-risk condition, not the lowest, because it makes opening the file feel redundant.

Refs #818

### DIFF
--- a/docs/plans/1-6-0-authorship-plan.md
+++ b/docs/plans/1-6-0-authorship-plan.md
@@ -37,18 +37,42 @@ corrections rather than additions: the seeded-mode control observation **moves e
 against), and a **pre-registration viability run** is added, because pre-registering a
 window for a capability that has never once succeeded wastes the window.
 
-**Claim sweep 2026-08-09 — every load-bearing code claim in this document was checked against
-source.** Prompted by a defect in this plan's own S4 section, which described a failure mode it
-had never verified. **Nine claims were wrong; seven overstated what exists.** All nine shared
-one origin: a claim about code sourced from a *document about the code* — this plan's earlier
-sections, the SIP's prose, an issue number from memory, a test file cited as precedent. Not one
-was a misreading of source. Corrections are inline below, each marked `Correction 2026-08-09`
-and carrying a `file:line`.
+**Claim sweep 2026-08-09 — this plan, SIP-0103, and the issues filed this release were all
+checked against source.** Prompted by a defect in this plan's own S4 section, which described a
+failure mode it had never verified (now #818). Results:
 
-**Standing rule adopted from it:** a load-bearing claim about code behavior in this document
-carries a `file:line`, or it is marked unverified. Bounding claims — *narrower*, *only*,
-*already*, *merely*, *does not affect* — carry it first, because those are the ones that stop
-further investigation, and a wrong one is not discovered until the code runs.
+| Artifact | Load-bearing code claims | Wrong | Direction |
+|---|---|---|---|
+| this plan | swept in full | **9** | 7 overstated what exists |
+| SIP-0103 | swept in full | **4 wrong + 1 normative bullet silently dropped** | all 4 overstated |
+| issues #762–#818 | premises spot-checked | **0 in bodies** (1 title imprecise; #812's premise was already corrected on the issue) | — |
+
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #821, opened 2026-08-09
feat(scaffold): the contract emitter refuses a stack it has no criteria for (#818)

jasonLadd:
Implements the design gate posted on #818, all three calls as ruled.

## The defect, demonstrated

The pre-change dispatch (`_routes_criteria if path == _ROUTES_PATH else _view_criteria`, with `_ROUTES_PATH = "backend/routes.py"` hardcoded) applied to a Node/TS-shaped slot:

```
src/routes.ts     -> ['frontend_compiles']
backend/routes.py -> ['command_exit_zero', 'endpoint_defined', 'import_present', 'module_imports']
```

A second stack's slots are `.ts`, so nothing matched `_ROUTES_PATH` and **every slot — including the API routes file — fell to the view branch.** The API routes file was handed to the frontend bundler.

**Why no guard caught it.** `endpoint_defined` was never *emitted*, so the two mechanisms that refuse to credit a check which cannot run had nothing to work with: `manifest_gates.py` `PROOF_CHECKS_LIVE` saw no inapplicable check, and `task_plan.py`'s dispatch-time strip had nothing to strip. The contract was not incomplete, it was **wrong, and every gate accepted it** — which is worse than the "narrower enforced surface" the plan's S4 section predicted before anyone opened this file.

## What changed

`CriteriaPack` owns **the partition as well as the criteria**. "What kinds of fill slot does this stack have, and how do I tell them apart" was the stack-specific claim that inline conditional was making silently, and it is the part that breaks first for a second stack.

`ScaffoldStack` gains `criteria_pack: str` — a *name*, not a callable, so the check vocabulary stays in the emitter layer. Same indirection S1 established with `check_stack`, and no new stack vocabulary: today a pack is named for the stack that needs it, and the indirection exists so a later FastAPI+Vue stack can share these backend criteria without minting one.

**The default is deliberately asymmetric with `check_stack`, and that asymmetry is the point.** An unset `check_stack` means the typed checks *skip* — safe, because `CheckOutcome.skipped` is not executed-and-passed under SIP-0096, so it surfaces as unverified. An unset `criteria_pack` **refuses**, because the failure it prevents is not a missing check but a wrong contract. Visible-and-unverified has a safe default; silently-wrong does not.

**Untouched, which is why the seam is small:** probes, coverage expectations, frozen-file hashing and `skeleton.expander` are manifest-derived already. That is S2's HTTP-constant containment argument paying off exactly where it was meant to.

## Two sub-findings, both fixed

**`_required_capabilities` had a closed universe.** It returned `[c for c in (CAP_PYTHON, CAP_NODE) if c in found]`, silently dropping a third — in a field documented as "declared from what the criteria actually require, so the two can't drift." A Go or Rust pack would have under-declared what the sandbox must provide. Known capabilities keep their historical order (python before node, which is serialized into the pinned reference contract), then anything else sorted.

**`CheckSpec.supported_stacks` is deleted.** Declared on 11 specs, read by nothing repo-wide, and a *third* stack vocabulary that was internally inconsistent — `{"fastapi"}` a framework, `{"python"}` a language, `{"python","javascript","typescript"}` languages. What it appeared to do is done by `applicable_extensions`, which is read at two live call sites, plus each evaluator's own stack branch.

Deleted rather than wired, deliberately: it is the seam someone would reach for to fix this issue, and wiring an unvalidated vocabulary would mint the stack-blueprint schema by accident — which is exactly what S1 declined to do when it named its type `ScaffoldStack`. This is S3's falsification pass applied early, and it takes the vocabulary count from three to two.

## Proof

| Exit criterion | Result |
|---|---|
| Reference contract byte-identical | ✅ verified against a captured baseline |
| Contract hash `7622f570c949fe95` | ✅ unmoved |
| Manifest hash `bb472e267e53d5ad` | ✅ unmoved |
| M0a equivalence guard | ✅ green |
| A stack with no pack raises | ✅ `test_a_stack_with_no_criteria_pack_is_refused_not_given_the_first_stacks_criteria` |
| `supported_stacks` deleted with the argument recorded | ✅ |
| Full regression | ✅ **6798 passed, 1 skipped** |

Nine new tests in `test_contract_emitter_stack_seam.py`, each written against a bug that was reachable: the routes file taking view criteria, a pack name nothing answers to, a refusal that does not say where to register the pack, a second stack inheriting the first's toolchain (`tests_pass` demanded `CAP_PYTHON` unconditionally), a second registration perturbing the reference, the dropped third capability, and a binding test so that two registries cannot drift apart — which is the failure mode S1 removed one layer down.

## Not in scope

The Node/TypeScript stack itself, which is the next item and the first thing that will actually exercise this seam. Kept separate deliberately: landing them together would make a moved contract hash ambiguous between "the refactor changed something" and "the new stack revealed something," and the byte-equivalence proof above is only meaningful because nothing else moved.

Also unresolved by design, for S3: whether a pack should be **data rather than callables**. Callables are what exists and prove no behavior change here; data would make S3's falsification pass mechanical instead of manual. Recorded on the issue as an owner fork.

Closes #818

### DIFF
--- a/src/squadops/capabilities/scaffold.py
+++ b/src/squadops/capabilities/scaffold.py
@@ -1463,6 +1463,20 @@ class ScaffoldStack:
     #: implementation's vocabulary rather than the scaffold's — "fastapi", not the profile
     #: name. Empty means the checks skip, which is the conservative default #503 chose.
     check_stack: str = ""
+    #: #818: the verification-criteria pack ``scaffold_contract`` emits for this stack —
+    #: which kinds of fill slot it has, what evidence proves each kind, and what proves the
+    #: deliverable builds and its suite ran. A *name*, not a callable, so the check
+    #: vocabulary stays in the emitter layer; same indirection as ``check_stack``.
+    #:
+    #: **The default is deliberately asymmetric with ``check_stack``.** An unset
+    #: ``check_stack`` means the typed checks *skip*, and a skip is safe because
+    #: ``CheckOutcome.skipped`` is not executed-and-passed under SIP-0096 — it surfaces as
+    #: unverified. An unset ``criteria_pack`` means the emitter **refuses**, because the
+    #: failure it prevents is not a missing check but a *wrong* contract: before #818 a
+    #: stack with no pack fell through to the FastAPI view branch and derived a contract
+    #: that every gate accepted. Visible-and-unverified has a safe default; silently-wrong
+    #: does not.
+    criteria_pack: str = ""
 
 
 _STACKS: dict[str, ScaffoldStack] = {
@@ -1473,6 +1487,10 @@ _STACKS: dict[str, ScaffoldStack] = {
         qa_test_namespace=("backend/tests/", "frontend/src/tests/"),
         harness_entry_modules=("backend.main", "app.main", "main"),
         check_stack="fastapi",
+        # Today a pack is named for the stack that needs it; the indirection exists so a
+        # later stack can share one (a FastAPI+Vue stack wants these same backend criteria)
+        # without minting a fourth stack vocabulary to express it.
+        criteria_pack="fullstack_fastapi_react",
     ),
 }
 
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

Issue #822, opened 2026-08-09
S2: build the Next.js + TypeScript stack — the second real stack the Blueprint SIP's acceptance gate requires


jasonLadd:
> **Rewritten 2026-08-09.** The original scope specified an Express API plus a typed React SPA (S2's 2026-08-07 ruling). **Superseded: stack #2 is Next.js + TypeScript** — owner ruling 2026-08-09, amended into the plan by #826. The original is wrong in both halves and is not preserved here; the plan's S2 section carries the full reasoning and the rejected alternatives.

The second stack. Its purpose is not "support another language" — it is **the acceptance gate for the Stack Blueprint SIP**, which declines its own acceptance until a second real stack exists, because *"generalising from one instance produces the FastAPI contract with generic field names — worse than no contract, because it looks authoritative and the second stack will quietly bend itself to fit rather than reveal the mismatch."* Every bend this stack is forced into is S3's evidence.

## What it breaks — verified against source 2026-08-09

| Hardwired assumption | Broken by |
|---|---|
| server code is Python, so the 9 `ast.parse` checks apply | TypeScript — `tsc --noEmit` becomes the hygiene tier |
| boot is `uvicorn backend.main:app` | `next start` |
| **the app boots from source** | `next build` must run first; **no build exists anywhere in today's probe path** |
| tests under `backend/tests/`; Python import boundary | co-located; no equivalent boundary |
| build runs in `frontend/` via `npm run build` | one build at the project root |
| backend and frontend are separate trees | one tree |
| the manifest has an `api` half and a `frontend` half | both are routes — some return JSON, some markup |

**What it does *not* break**, recorded so S3 does not imply it: *"endpoint discovery is static and declared up front"* stays on the unvalidated list. Next discovers routes from the filesystem, but the expander still works from a manifest that **declares** paths and places files at derived locations (`POST /runs` → `app/api/runs/route.ts`). Declared-then-placed is what an expander does. HTTP is likewise held constant.

## The surface map

**Already a per-stack seam — registration, not refactor:**

| Layer | Landed as |
|---|---|
| `capabilities/scaffold.py::_STACKS` | S1 (#816) — raises for an unregistered stack |
| `capabilities/scaffold_contract.py::_CRITERIA_PACKS` | #818 (#821) — refuses a stack with no pack |
| `capabilities/handlers/probe_runner.py::_PROFILES` | #823 — refuses to boot a stack that declares no profile |
[...]

PR #823, opened 2026-08-09
feat(probes): the probe runner boots the stack the cycle actually builds (#822)

jasonLadd:
First of two PRs on #822. The boot seam, landing alone so that a probe failure when the Node stack arrives is unambiguous between "the parameterization broke FastAPI" and "the new stack doesn't boot."

## The defect

`run_probes` took `profile: ExecutionProfile = DEFAULT_PROFILE` as a **default argument**, and no caller overrode it. Every stack was booted with `python -m uvicorn backend.main:app`.

This is loud rather than silent — a Node app doesn't start under uvicorn, boot fails, and `ProbeOutcome.skipped` means not-executed, which SIP-0096 declines to credit. That's why it's scope and #818 was the emergency. But *"boot the wrong thing and read the failure"* is a diagnosis someone has to reconstruct; a declared refusal is one they're handed.

## This corrects the design gate

The gate on #822 said to derive the profile from the sandbox `EnvironmentContract`'s `START_APPLICATION`. **Verified before building, and it doesn't work:**

```
EnvironmentContract.START_APPLICATION:
    (".sandbox-venv/bin/python", "-m", "uvicorn", "backend.main:app", "--host", "0.0.0.0",   "--port", "8000")
probe_runner.DEFAULT_PROFILE:
    (sys.executable,             "-m", "uvicorn", "backend.main:app", "--host", "127.0.0.1", "--port", "{port}")
```

That argv runs **inside the sandbox container**, against the `.sandbox-venv` its `INSTALL_DEPENDENCIES` step builds. `run_probes` is called from `handlers/cycle/qa_test.py` **in the qa container**, against a fresh `TemporaryDirectory` where no venv exists — hence `sys.executable`. Two execution contexts, not one fact duplicated: the **interpreter is context-specific**, only the launcher and entry point are stack-specific.

The gate flagged host/port and missed the interpreter, which is the disqualifying half. Correction was posted on #822 before this was written.

## The shape

`ScaffoldStack.probe_profile` names a profile registered in `probe_runner` — a name, not a callable, so the boot vocabulary stays in the runner layer. Exactly `criteria_pack` from #818.

| Case | Behavior |
|---|---|
| explicit `profile` | use it (preserves the existing test seam) |
| stack declares one | use it |
| **registered stack declaring none** | report not-executed, logged |
| no stack resolvable | **`DEFAULT_PROFILE`** — today's behavior, byte-identical |

That last row is the important one: a config typo must fall back, never silently delete probe evidence.

**The refusal does not raise**, unlike #818's. `qa_test` states probes are *"additive evidence only … a failed probe surfaces at the run verdict/rollup, not as a task failure here"* — raising would convert missing evidence into a dead QA task. Forgetting is caught at **build time** instead, by a completeness test across all three per-stack registries. S1's real lesson was that forgetting must be an error, not that there must be one registry.

`scaffold_stack_for()` resolves `build_profile` to the **scaffold** stack name, deliberately not reusing `resolve_check_stack`, which maps to the **evaluator** vocabulary (`"fastapi"`). Two names, and S3 still owes the reconciliation — collapsing them here would key a lookup on a name the registry doesn't hold.

## Proof

| | |
|---|---|
| Reference contract `7622f570c949fe95` | ✅ unmoved |
| Manifest `bb472e267e53d5ad` | ✅ unmoved |
| Full regression | ✅ **6812 passed, 1 skipped** |

Four existing test doubles in `test_build_handlers.py` didn't accept the new kwarg. **Tightened rather than made permissive** — they capture the stack instead of swallowing `**kwargs`, and a new test proves the handler threads it end to end. A double that absorbs unknown arguments would have hidden exactly the wiring regression this PR exists to prevent.

Eleven new tests in `test_probe_boot_seam.py`, including the inertness guard for every caller that exists today — if that path stopped booting, the behavioral half of every contract would quietly stop being evidence.

## Next

The Node/TypeScript stack itself: expander, criteria pack, environment contract, bend register. Then **VS**, which gates S3.

Refs #822

### DIFF
--- a/src/squadops/capabilities/handlers/cycle/qa_test.py
+++ b/src/squadops/capabilities/handlers/cycle/qa_test.py
@@ -265,6 +265,7 @@ class QATestHandler(_CycleTaskHandler):
             return
 
         from squadops.capabilities.handlers.probe_runner import probe_check_rows, run_probes
+        from squadops.capabilities.scaffold import scaffold_stack_for
         from squadops.cycles.patch_verification import materialize_artifacts
         from squadops.cycles.verification_contract import Probe
 
@@ -295,7 +296,11 @@ class QATestHandler(_CycleTaskHandler):
         with tempfile.TemporaryDirectory(prefix="qa_probes_") as tmp:
             workspace = Path(tmp)
             materialize_artifacts(artifacts, workspace)
-            outcomes = await asyncio.to_thread(run_probes, workspace, probes)
+            # #822: boot the stack this cycle actually builds. Before this, `run_probes`
+            # took the FastAPI profile as a default argument and nothing overrode it, so
+            # every stack was booted with `uvicorn backend.main:app`.
+            stack = scaffold_stack_for(inputs.get("resolved_config"))
+            outcomes = await asyncio.to_thread(run_probes, workspace, probes, stack=stack)
 
         vr = outputs.setdefault("validation_result", {})
         vr.setdefault("checks", []).extend(probe_check_rows(outcomes))
--- a/src/squadops/capabilities/handlers/probe_runner.py
+++ b/src/squadops/capabilities/handlers/probe_runner.py
@@ -26,6 +26,7 @@ from __future__ import annotations
 
 import contextlib
 import dataclasses
+import logging
 import os
 import socket
 import subprocess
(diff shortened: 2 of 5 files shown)


jasonLadd (commented):
reviewed

Issue #820, opened 2026-08-09
DEFERRED to 1.7: SIP-0103 §3.3's interface self-consistency proof — endpoints/params/testids mutual coherence


jasonLadd:
## What is deferred

SIP-0103 **§3.3's third bullet**, verbatim:

> *"interface self-consistency: endpoints/params/testids the manifest promises are mutually coherent and cover the PRD's promised surface (the #565 `{id}`/`{run_id}` naming-prior class becomes an authoring-time check instead of a live-roll loss)."*

It is normative in an accepted SIP and **it does not ship in 1.6** (owner ruling 2026-08-09). This issue exists so the commitment has a home outside a release plan that gets superseded at the cut.

## Why it needs an issue rather than a plan line

Found by the 2026-08-09 claim sweep (PR #819): the bullet was never built, and the v1.6 plan's M3 proof table had **dropped it without recording the drop** — it was lost between the two documents rather than decided. The same sweep established that things recorded only in plan prose do not propagate (six premise corrections found at build; one propagated when it happened). Recording this deferral only in the plan would repeat the failure that surfaced it.

## What M3 actually proves today

`cycles/manifest_gates.py` declares `parses`, `lint`, `expands`, `contract_derives`, `checks_live`, `testid_coverage`, `status_declared`, plus M2's `source_prd` and `decision_record`.

The relevant one is `_testid_findings` → `PROOF_TESTID_COVERAGE`: **every frontend route declares ≥ 1 testid.** That is *coverage*, not *coherence*.

**Unproven, and the content of this issue:**

- a path parameter for the same logical entity is named consistently across endpoints;
- declared testids correspond to anything the endpoints or views promise;
- endpoint/param mutual coherence generally.

It is a **closed-surface proof over the manifest's own declarations** — deterministic, no PRD semantics, no LLM. That places it squarely in the Phase-1 class §5b Q2 admits, which is why it cannot be folded into the "semantic PRD coverage" the plan deliberately defers. Different thing, deferred for different reasons.

## Why deferring is defensible

The motivating instance was **PR #565 — *"feat(manifest): rename group_run path param `{id}` → `{run_id}`"*** — a fix to the **hand-authored** reference manifest. The class fires when a *human* declares a parameter name that fights the model's prior, and the implementation then writes the prior instead of the declaration.

[...]

PR #824, opened 2026-08-09
docs(sip-0103): record post-acceptance amendments in the SIP, not only the plan

jasonLadd:
Governance fix. Three divergences and four unimplemented dispositions were recorded **only** in `docs/plans/1-6-0-authorship-plan.md` — superseded at the cut, after which SIP-0103 becomes the sole surviving description of a design it no longer matches.

That's not tidiness. §5c.9 hands the cross-cycle learning channel to the Cross-Cycle Memory SIP explicitly. Someone opening SIP-0103 in 1.8 reads §3.4, sees a mandatory HITL manifest gate, and builds against a gate that does not exist — and reads §5c.7's four diagnostics as available when two were never built.

## The mechanism was never missing

**§5a is literally titled "Amendments from the 1.5 boundary."** §5a, §5b and §5c are all amendment rounds. The convention exists, is demonstrated three times in this document, and stopped being used at the exact moment acceptance made it matter — because the divergences were instead justified with a principle I invented and reused verbatim: *"the SIP stays as accepted — it is a design commitment on main, not a living document — and the correction lives here."*

That has a real basis (CLAUDE.md does say acceptance is a commitment, not an implementation artifact) but it stretches a claim about **sequencing** into a licence to leave the spec stale. It sounds like governance and functions as a bypass: divergence costs one paragraph in a document with a shelf life.

## What §5d records

**A — premises falsified after acceptance.** §5b Correction 1 is wrong in both halves (the deriver exists; v9 is its exact output for v4, so the SIP's headline scope addition was satisfied before work started). Its derivability split is wrong in detail (`endpoint_defined` emitted once, not per fill slot; `field_present` never emitted). And §5a's *"zero new plumbing"* — **authored mode was dead on arrival**; forwarding dropped `interface_manifest`, and V4 roll 1 hit 0 of 9 fill slots before #796.

**B — design narrowed by owner ruling.** §3.4's mandatory manifest review → question-gated (#807), ruled 2026-08-08 on V4 evidence: a human approved a bundle by discussing facts the deterministic gates had already proven, while the one genuinely unresolved question in the manifest was never surfaced. Recorded as consistent with §5c.10 rather than as a contradiction, with the code sites named.

**C — dispositions not implemented**, because silence reads as shipped. §5c.7 (2 of 4 diagnostics), §5c.5 (operator-edit record — and no operator-edit path exists at all), §5c.3 (blueprint-owned manifest → 1.7), §3.3 bullet 3 (→ #820).

**The one flagged as more than disclosure:** B1 removed a control citing four diagnostics, two of which were never built. That is the only place the delivered system is weaker than the ruling authorized. Second PR, ahead of the pre-registered window.

## CLAUDE.md gains step 5a

The contributor workflow had eight steps and no row for *"implementation showed the design was wrong."* Without one, the next implementer hits the same silence and fills it the same way. Step 5a says: amend the SIP **in the PR that diverges** — plus three things worth stating because each was violated here:

- **a release plan is not an amendment** (it is superseded);
- **a code comment is not either** (`open_questions()` cites *"§3.4 as narrowed"* — correct, and invisible to a reader of the SIP);
- **a disposition deliberately not built is an amendment too.**

## One convention question for you

I did **not** touch `updated_at` in the SIP frontmatter or `sips/registry.yaml`. Both are stamped by `update_sip_status.py` and currently differ by microseconds, so they look script-owned, and hand-editing one would create the drift this PR exists to remove. But a reader checking "has this SIP changed since acceptance?" would look exactly there and see 2026-08-07.

If amendments should re-stamp it, that's a small change to `update_sip_status.py` (an `--amend` mode) rather than a hand edit. Your call — flagging rather than deciding, since it's maintainer-owned.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -182,6 +182,7 @@ Key principle: **acceptance is a design commitment on main, not an implementatio
 - **A code comment is not an amendment either.** Necessary, not sufficient: a reader of the SIP never sees it.
 - Each amendment names **what changed, the evidence, and who ruled it.** "We decided otherwise" without evidence is how a spec becomes advisory.
 - **A disposition that is deliberately not built is an amendment too** — silence reads as "shipped."
+- **Do not touch `updated_at`.** It means *last status transition*, not last content change: `update_sip_status.py` is the only writer (frontmatter and registry are stamped separately, hence the millisecond skew between them), so a hand edit invents drift. **The amendment section's own date is the content record** — and it is the better one, since it says *what* changed rather than only when.
 
 ### Key Implemented SIPs
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -168,11 +168,20 @@ SIPs govern architectural decisions. Located in `sips/` with lifecycle:
 | 3. Accept | Merge to main | Maintainer | Runs `update_sip_status.py ... accepted`, merges |
 | 4. Branch | Feature branch off main | Implementer | `git checkout -b feature/sip-NNNN-...` from main (which now has the accepted SIP) |
 | 5. Implement | Feature branch | Implementer | Incremental commits per phase |
+| **5a. Amend** | **The diverging PR** | **Implementer** | **Implementation showed the accepted design was wrong, or a decision narrowed it → add an amendment section to the SIP *in the PR that diverges*** |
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #825, opened 2026-08-09
feat(diagnostics): build the two SIP-0103 §5c.7 signals M4 spent before they existed

jasonLadd:
Second of the two PRs from the SIP-0103 audit. #824 records the divergences; this closes the one gap where the delivered system was **weaker than any decision authorized**.

## The gap

M4 removed the mandatory human manifest review. The justification — mine — was that design quality moves to *sampling, not gating*, carried by the four diagnostics §5c.7 requires:

| diagnostic | status when M4 shipped |
|---|---|
| revision/attempt counts | ✅ M5 (#803) |
| gate-rejection taxonomy | ✅ M6 (#785) |
| **structural diff vs the human reference** | ❌ never built |
| **manifest size/surface counts** | ❌ never built |

So the argument for dropping a human gate was half-funded from the moment it was made. That's the **second instance of one pattern** — the seeded control was retired citing three offline mechanisms, two of which were unbuilt. Both arguments removed a control by naming replacements and not checking that the replacements existed.

## Why these two, and not just "for completeness"

**FAY is structurally blind to what they measure.** V4 roll 2 flattened the reference's typed `Participant` entity into an untyped list and still went green — installed, built, booted, answered all five probes. A yield number cannot see a design regression the app happens to survive.

**The sharper risk is that the squad authors the exam it sits.** The manifest decides the contract and the contract decides the checks, so a roll can win by *declaring less* — 29 executed checks on V4 roll 2 against 57 on V5, same PRD. `surface_counts` is what makes that visible. A floor would be the wrong instrument; it invites padding in the other direction.

## It works on the first run

Against the real V4 roll-2 fixture, unprompted:

```
| entities | 1 | 2 | -1 |
### entities
- only in reference: `Participant`
```

That's the motivating case, caught mechanically. It also surfaces `/runs/:run_id` against the reference's `/runs/:id` — **#820's deferred naming-prior class**, now showing up as a visible non-gating signal rather than a silence. That materially strengthens the deferral: the class isn't unobserved, it's unblocked.

## Three properties enforced structurally, not promised

**Operator-run, never a pipeline stage.** The reference manifest is excluded from squad inputs (§4, §5c.1), so a diff computed inside a cycle would make it an input and contaminate authored mode. An architecture test asserts nothing under `src/squadops/` or `adapters/` can reach the module — the same shape as Guard 1a.

**Non-gating by construction.** The diff returns a report and has no failure mode, pinned by a test. A diagnostic that can raise becomes a gate the first time it does, and it would do so inside the window where nothing may be fixed.

**Not a score.** §5c.7 declines maintainability/elegance metrics because no deterministic representation exists and an LLM-graded score is the evidence laundering A6 forbids. This reports *what* differs and leaves *whether that is worse* to a reader — distance from the human design is not error, and roll 2's 409-for-conflict was better than roll 1's 400.

## Also

Corrects the plan's M4 section, which read *"diagnostics §5c.7 **already requires**"* — which a reader takes as "already has."

Eleven tests, each against a reachable bug: the flattened entity, a shrinking declared surface, differing detail on a shared key being dropped by a naive set diff, zero-vs-absent in the counts, noise on an unchanged design, the contamination reach, and the non-gating property.

**Regression 6823 passed, 1 skipped.**

Sequencing: owed **before V6 and pre-registration**, since this is the instrument the window relies on and V7 is the one place nothing may be fixed mid-flight. Does not block the stack decision.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -168,11 +168,20 @@ SIPs govern architectural decisions. Located in `sips/` with lifecycle:
 | 3. Accept | Merge to main | Maintainer | Runs `update_sip_status.py ... accepted`, merges |
 | 4. Branch | Feature branch off main | Implementer | `git checkout -b feature/sip-NNNN-...` from main (which now has the accepted SIP) |
 | 5. Implement | Feature branch | Implementer | Incremental commits per phase |
+| **5a. Amend** | **The diverging PR** | **Implementer** | **Implementation showed the accepted design was wrong, or a decision narrowed it → add an amendment section to the SIP *in the PR that diverges*** |
 | 6. Code review | PR to main | Maintainer + team | Review the implementation |
 | 7. Merge | main | Maintainer | Merge the feature PR |
 | 8. Promote | main | Maintainer | Runs `update_sip_status.py ... implemented` after verification |
 
 Key principle: **acceptance is a design commitment on main, not an implementation artifact.** The feature branch starts from a main that already has the accepted SIP, so the implementer works from an approved spec. This separates "we agree this is the right design" from "the implementation is correct."
+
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #826, opened 2026-08-09
docs(plan): amend S2 — stack #2 is Next.js + TypeScript

jasonLadd:
Owner ruling 2026-08-09, superseding the 2026-08-07 selection. The criterion and the rejected alternatives are retained unedited; only the selected instance changed.

## Why the original pick was too narrow

Express-or-Fastify + typed React **barely changes the frontend**. It holds React, the component model, `frontend/`, npm, vite, and bundler-as-only-check all constant — so it exercises the backend axes and leaves `frontend_dir = workspace_root / "frontend"` and the `npm run build`-in-a-subdirectory assumption untouched. This plan's own risk row warns that a near-twin validates nothing; the original was a near-twin on one half.

## What Next.js adds over it

Everything the original broke, plus: one build at the project root, one tree instead of two, the manifest's `api`/`frontend` split — and one nobody had noticed, **"the app boots from source."** `next build` must run before `next start`, and there is no build step anywhere in today's probe path; `frontend_build` is a separate check and probes boot the backend directly from the materialized workspace.

## Two corrections recorded so they aren't inherited

I asserted twice during selection that **file-system routing breaks "endpoint discovery is static and declared up front."** It doesn't. Next discovers routes from the filesystem, but the expander still works from a manifest that *declares* paths and places files at derived locations — `POST /runs` → `app/api/runs/route.ts`. Declared-then-placed is what an expander does. That assumption stays on the unvalidated list, and **the manifest maps onto Next.js better than the discussion implied**, which lowers the release risk I kept citing against it.

Also: `[run_id]` is a **third** path-parameter convention after `{run_id}` and `:run_id`. #820's trigger 3 fires regardless of which candidate had been picked.

## What made it possible

Two owner rulings, both recorded in the section: the backend/frontend split may break so long as the squad can compose a viable app, and *"I don't want to force decisions to make the manifest comfortable; if the manifest pattern needs to break down then so be it."* Together they convert the manifest's two-section shape from a **constraint on selection** into **a thing under test**.

## Risk

**Attribution, not mapping.** Next.js moves the most variables at once, so a poor VS makes "the schema doesn't generalize" hard to separate from "the squad can't write server components." M6's taxonomy is the instrument — at VS, watch attribution rather than pass rate, the same discipline V4 was given.

**NestJS + React-TSX is recorded as the fallback** if Next.js fails to land; it would still give S3 a second instance, and its decorator routing is structurally analogous to FastAPI's, which would make a TypeScript `endpoint_defined` tractable for S4.

## The pack-combo idea, kept but reordered

The owner raised decomposing a stack into independently-variable packs, so `(fastapi, react_ts)` would be recombination rather than authoring. Good idea, wrong order — deriving a composition model now generalizes from **one and a half** instances, weaker footing than the two the Blueprint SIP insists on. Once stack #2 lands, the cross-combo is nearly free and *is* the composability test, with real packs instead of an abstraction designed for it.

**Build consequence:** record backend/frontend entanglement in the bend register while writing the pack. Entanglement that can't be separated later is cheap to notice now and expensive to reconstruct.

Docs only. #822's scope and design gate are rewritten separately, next.

### DIFF
--- a/docs/plans/1-6-0-authorship-plan.md
+++ b/docs/plans/1-6-0-authorship-plan.md
@@ -662,7 +662,86 @@ Consolidate into one object carrying **today's fields**. Pure refactor, exact te
 unmoved, both emission gates 6/6, regression unchanged. This removes the silent-omission
 failure mode immediately and prejudges nothing.
 
-### S2. Stack #2 — **decided: a Node/TypeScript HTTP stack** *(owner ruling 2026-08-07)*
+### S2. Stack #2 — **AMENDED: Next.js + TypeScript** *(owner ruling 2026-08-09, superseding 2026-08-07)*
+
+> **The 2026-08-07 ruling below — an Express-or-Fastify API with a typed React frontend — is
+> superseded.** Its reasoning is retained unedited because the *criterion* still governs and
+> the rejected alternatives still stand; only the selected instance changed. The trigger was
+> the owner's observation that the original choice **barely changes the frontend at all**.
+
+**Why the original selection was too narrow.** Express + typed React holds React, the component
+model, `frontend/`, npm, vite, and the bundler-as-only-check all constant. Measured against the
+hardwiring this repo actually carries, it exercises the backend axes and leaves the frontend
+ones — `frontend_dir = workspace_root / "frontend"` (`cycles/acceptance_checks.py`, the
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

Issue #788, opened 2026-08-08
failure analyzer recommended a provably-impossible relative import, and the repair followed it into a worse failure


jasonLadd:
## What happened

V2 control cycle `cyc_dd3855f353c0`, run `run_e40065a2e83a`, round 1 analysis of a `tests_pass` collection failure:

> *"The emitted test file incorrectly used `from .store import reset` … Fixing the relative import path to **`from ..store import reset`** or using an absolute import will resolve the collection failure."*

The repair took the first suggestion. Round 2 failed with:

```
ImportError: attempted relative import with no known parent package
```

`from ..store` cannot work from `backend/tests/` — there is no parent package to ascend to. The suggestion was not merely unhelpful; it was **provably impossible**, and it consumed the second of two correction rounds. A4 then terminated the chain as `plan_defect`.

The analysis did mention "or using an absolute import" — the form that works — as a trailing alternative. The repair took the first, specific, wrong one.

## Why this is a diagnosis-quality defect and not bad luck

The analyzer had everything needed to be right. `conftest.py` is in the workspace and states the answer in its header (workspace root on `sys.path`, so `import backend` resolves). The failing traceback names the exact Python rule being violated. Nothing about the correct answer required inference — only reading a file that was present.

This lands in the #431/#687 family (diagnosis quality feeding the repair chain), with a sharper edge than either: those addressed evidence that was **missing or lossy**. Here the evidence was **present and ignored**, and the analyzer's confidence in a wrong specific fix beat its own correct general one.

## Two candidate directions (not a decision)

1. **Rank the alternatives, or emit one.** An analysis offering "X or Y" hands the repair an unranked choice, and the repair took the wrong branch. Either the analyzer commits to one recommendation, or the ordering has to mean something.
2. **Make impossible advice unrepresentable for the class we can check cheaply.** A relative-import recommendation against a path outside a package is statically decidable. This is the deterministic-net-beside-a-judgment-call pattern the repo already uses (#671 module-existence, #689 undefined-names): the model may still be wrong, but not in a way a two-line check could have caught.

Not proposed: more prompt instruction. The analyzer was already free to say "use an absolute import" and did, second.

## Relationship to #787

They are the same failure from two ends and both should land. #787 stops the author writing the bad import; this stops the analyzer *recommending* one when it happens anyway. Fixing only #787 leaves the repair path able to reintroduce the same class from a different starting point.

## Home

[...]

PR #827, opened 2026-08-09
feat(probes): a stack that cannot run from source builds before it boots (#822)

jasonLadd:
First of the seam PRs for stack #2, landing alone so that a probe failure when Next.js arrives is unambiguous between "the parameterization broke FastAPI" and "the new stack doesn't build."

## The gap

Every stack today is **interpreted**. `uvicorn backend.main:app` runs the materialized workspace directly, and there is **no build step anywhere in the probe path** — `frontend_build` is a separate acceptance check, and probes boot the backend on their own. Next.js requires `next build` before `next start`.

## Why this isn't just a longer boot timeout

`startup_timeout_s = 25.0`, and a cold `npm ci` plus a production build is minutes. Raising the boot timeout is the obvious fix and it's wrong: it conflates two failures with two different diagnoses. *"The app is slow to start"* and *"the app does not build"* would both arrive as `subject did not boot`.

That matters concretely — **#687's failure analyzer keys on these reason strings**, and pointing it at a subject that never started is how **#788** happened: a repair chasing a cause that was never there.

So preparation is its own stage with its own bound and its own reason prefix.

## Shape

`ExecutionProfile` gains `prepare_argv` and `prepare_timeout_s`, run once in the workspace before boot.

**Inert until a stack declares it** — empty is the default and that's every stack today. Pinned two ways: a test asserting **no subprocess runs at all** for a profile without it, and a test naming `DEFAULT_PROFILE` directly rather than a look-alike fixture. That property is the one that matters most here; a regression would silently delete the behavioral half of every contract's evidence.

**Failure is not-executed, never raised.** A subject that cannot be built is `skipped` exactly as one that cannot boot is, because `qa_test` states probes are *"additive evidence only … not a task failure here."* Three modes reported distinctly:

| mode | reason |
|---|---|
| non-zero exit | `subject preparation failed (exited N): <output tail>` |
| never finished | `subject preparation failed (no exit within Ns)` |
| tool absent | `subject preparation failed (could not launch)` — the #462 missing-tooling class |

**Output is captured and tailed** for the reason boot stderr is (#512), and the case is stronger here: a boot failure leaves a process to interrogate; a build failure leaves nothing. If the compiler's message isn't captured at that moment it exists nowhere. Both streams are read — `tsc` writes diagnostics to stdout while most tools use stderr, and reading only stderr would drop the message for exactly the toolchain this exists to support.

**A failed build does not proceed to boot.** Otherwise probes would run against a stale or absent artifact and report whatever they found as truth.

## Proof

Ten tests, each against a reachable bug: the inertness guard, a build failure misreported as a boot failure, output reaching the reason from either stream, the unbounded hang, the missing tool, the happy path actually continuing to boot (a preparation step that swallowed control would make every compiled stack silently unprobeable — the #818 shape one stage over), and the build running in the workspace rather than wherever the runner happens to be.

**Regression 6833 passed, 1 skipped.**

## Next

`frontend_build`'s hardcoded `workspace_root / "frontend"` (design gate decision 6), then the stack itself, then VS.

Refs #822

### DIFF
--- a/src/squadops/capabilities/handlers/probe_runner.py
+++ b/src/squadops/capabilities/handlers/probe_runner.py
@@ -68,6 +68,19 @@ class ExecutionProfile:
     startup_timeout_s: float = 25.0
     request_timeout_s: float = 10.0
     poll_interval_s: float = 0.1
+    #: #822: a one-time command run in the workspace *before* boot, for stacks whose subject
+    #: cannot run from source. Empty means none — which is every stack that exists today, so
+    #: this is inert until one declares it.
+    #:
+    #: **Why this is not just a longer boot timeout.** Compiling and starting are different
+    #: failures with different diagnoses: "the app is slow to start" and "the app does not
+    #: build" would otherwise arrive as the same `subject did not boot` reason, and the
+    #: build's own output — the only thing that explains it — is on a process that already
+    #: exited. Separating them keeps `_boot_failure_reason` meaning what it says.
+    prepare_argv: tuple[str, ...] = ()
+    #: Generous on purpose: a cold `npm ci` plus a production build is minutes, and this
+    #: bound exists to stop a hang, not to police build speed. Boot keeps its 25s.
+    prepare_timeout_s: float = 600.0
 
 
 # The default profile: boot the FastAPI backend with uvicorn on an allocated port, using the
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #828, opened 2026-08-09
feat(checks): the buildable project directory is a stack fact, not "frontend/" (#822)

jasonLadd:
Second seam PR for stack #2, landing alone. Reference contract `7622f570c949fe95` and manifest `bb472e267e53d5ad` **unmoved**; regression **6841 passed, 1 skipped**.

## Premise correction — the design gate named the wrong evaluator

Gate decision 6 said *"`frontend_build` hardcodes `workspace_root / "frontend"`."* **It doesn't.** Verified before implementing:

- `run_frontend_build` takes `target_dir` as a parameter, the vitest dispatch already passes `None` (whole tree), and `_find_package_json_dir` *discovers* the package.json location rather than assuming it (#303).
- The hardcoded path is in **`FrontendCompilesCheck`** — the per-view bundler check — at `acceptance_checks.py`.

So the behavioral build check was already stack-tolerant and the per-view one was not. This PR fixes the one that's actually broken.

## Why it matters more than a skip usually would

For a second stack, every view's `frontend_compiles` reported `no_frontend_tree`. SIP-0096 declines to credit a skip, so this is visible rather than a false green — but the consequence is still severe:

`frontend_compiles` is the **only** bundler-level coverage a view has. #648 built it because fay-4 and fay-8 each shipped a view with a rollup bind-time error *invisible to every static check and to `node --check`*, first surfacing at final verification where no correction budget could reach it. And stack #2 has **no AST tier at all** — the nine `ast.parse` checks are Python-only.

So stack #2's views would have shipped with an entirely empty static surface, every check reporting not-executed.

## The change

`frontend_compiles` gains an optional `project_dir`, defaulting to `"frontend"`. Stack #1's criteria pack emits nothing new, so its contract is byte-identical — which is the release's banked evidence and the reason the parameter is optional rather than required.

Declared in **`path_params`** as well as `optional_params`/`param_types`, so it inherits the authoring-time traversal rejection in `implementation_plan` and the pre-eval `_safe_resolve`. That's not decoration: the value now arrives from a contract rather than being a literal, and a bare `workspace_root / value` join would happily resolve outside the workspace and run `npm install` there. Verified `"."` resolves to the root and passes the lint, while `../outside`, `/etc` and `frontend/../../..` are rejected at both layers.

## What is deliberately *not* parameterized

The build **command**. S5's admission rule wants a field demonstrated on two stacks before admission: `frontend/` vs `.` is demonstrated; both stacks run `npm run build`. A `build_argv` or `package_manager` parameter would be a one-stack guess wearing a general name — the exact decay that rule exists to prevent. Pinned by a test so a future PR has to argue rather than drift.

## Tests

Eight, each against a reachable bug: the root-built project found when declared, the default still resolving stack #1's layout, **the defect reproduced** (root-built project without the param → `no_frontend_tree`), three traversal escapes, the declaration wiring that makes both validation layers cover it, and the not-parameterized guard.

## Next

The stack itself — expander, criteria pack, probe profile, environment contract, completeness test, bend register. Then **VS**, which gates S3.

Refs #822

### DIFF
--- a/src/squadops/cycles/acceptance_check_spec.py
+++ b/src/squadops/cycles/acceptance_check_spec.py
@@ -574,10 +574,10 @@ CHECK_SPECS: dict[str, CheckSpec] = {
         name="frontend_compiles",
         applicable_extensions=frozenset({".js", ".jsx", ".ts", ".tsx"}),
         required_params=frozenset({"file"}),
-        optional_params=frozenset({"timeout_s"}),
-        param_types={"file": str, "timeout_s": int},
+        optional_params=frozenset({"timeout_s", "project_dir"}),
+        param_types={"file": str, "timeout_s": int, "project_dir": str},
         requires_stack_context=False,
-        path_params=frozenset({"file"}),
+        path_params=frozenset({"file", "project_dir"}),
         example={"file": "frontend/src/views/RunsListView.jsx"},
         # #648: fay-4 and fay-8 both shipped a view with a rollup bind-time
         # error (an undefined identifier) — invisible to every static check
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #830, opened 2026-08-10
feat(correction): view source for repair targeting comes from the contract (#829)

jasonLadd:
Third seam PR for stack #2, and the one with the highest cost per occurrence. Contract `7622f570c949fe95` and manifest `bb472e267e53d5ad` unmoved; regression **6852 passed, 1 skipped**.

## The defect

`_widen_target_for_frontend_build` identified view source as `p.startswith("frontend/")` — stack #1's directory layout stated as a property of views. **A root-built stack unions nothing.**

That function exists because of fay-8, recorded in its own docstring: *"five correction rounds, four identical `frontend_build` failures, every repair emitted backend/test files only … the loop polished a passing backend while the build-breaking view sat outside every target."* For a second stack the trap reopens intact — **in the correction path, where it costs a full round each time** rather than a single check.

Third instance of the same assumption: #818 (contract emitter), #828 (`frontend_compiles` project dir), this one.

## The fix, and why not a wider prefix

The precedent is twenty lines up in the same file. `_probe_owned_slots` (#688) resolves repair targets from **contract data**, deliberately: *"This relation is authoring-independent: it comes from the contract."* It was written because prefix and package scoping made reachability depend on where the squad happened to author its suite.

Same relation, one criterion over: a fill slot is a **view** when the contract attaches `frontend_compiles` to it. `VerificationContract.view_slots()` reads it, `inject_contract_inputs` threads it as `contract_view_slots` beside `contract_endpoint_owners`, the widening consumes it.

**Measured against the reference contract, `view_slots()` returns exactly the three files the prefix matched** — so this is a strict generalization, not a behavior change on the configuration the release's evidence was measured on.

**No fallback, and none is needed.** `_frontend_build_failed` requires a `frontend_build` row, which only a contract emits, so a contract-less cycle can never reach the widening and the absent-input path is byte-identical to today. That avoids the legacy/new dual path the repo rules out.

## Two things checked rather than assumed

**Does the input reach the right task?** `qa.test` carries `bind_behavioral_surface` and `frontend_build` runs inside it, so the failing task whose inputs the correction reads does receive `contract_view_slots`. Verified before relying on it.

**Is the value carried, not just computable?** A test asserts it reaches task inputs through `inject_contract_inputs`, because #796 was exactly the other thing — an artifact stored, validated, promoted, and silently not forwarded, so the capability was dead while looking built.

## On the two modified tests

The two #650 tests failed, and that's the right kind of failure — their fixtures predate this input. **Their setup gained `contract_view_slots`; every assertion is unchanged.** The assertions are about behavior (the views reach the target), which is still required; what changed is how the runner learns which files are views. The fixture is now closer to a real `qa.test` envelope, which carries the key.

Flagging it explicitly because "tests changed to make the code pass" deserves scrutiny by default.

## Also

`CHECK_FRONTEND_COMPILES` joins the named-constant block beside `CHECK_ENDPOINT_DEFINED`, for the reason those comments already give: a literal that stops matching after a rename resolves to *"this stack has no views"* rather than erroring.

Closes #829 · Refs #822

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -180,12 +180,13 @@ def _widen_target_for_frontend_build(
     """
     if not _frontend_build_failed(failure_evidence):
         return target
-    frontend_source = [
-        p
-        for p in (failed_inputs.get("implementation_artifacts") or [])
-        if isinstance(p, str) and p.startswith("frontend/")
-    ]
-    return list(dict.fromkeys([*target, *frontend_source]))
+    # #822: which files are views is the CONTRACT's answer, not a directory prefix. This read
+    # `p.startswith("frontend/")` — stack #1's layout stated as a property of views. A stack
+    # that builds at the project root would union nothing, and the fay-8 trap this function
+    # exists to close would reopen intact for it. Same authoring-independent relation
+    # `_probe_owned_slots` uses (#688), one criterion over.
+    view_slots = [p for p in (failed_inputs.get("contract_view_slots") or []) if isinstance(p, str)]
+    return list(dict.fromkeys([*target, *view_slots]))
 
 
 def _failed_probe_ids(failure_evidence: Any) -> list[str]:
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #831, opened 2026-08-10
docs(plan): stage the remaining 1.6.0 sequence

jasonLadd:
Replaces the stale `Remaining sequence` section — it still described Track M's leftovers, all of which landed.

**Scope confirmed unchanged:** 1.6.0 cuts on both headlines, S1–S3 + S5 release-defining. The considered-and-declined 1.6.0/1.6.2 split is recorded so it is not relitigated, along with why the version arithmetic did not work (a patch cannot carry a headline feature SIP).

**Stages, which progress will be reported against:**

- **1a–1e** — finish the S lane: hardwiring sweep → parameterization → the Next.js stack → bend register → VS
- **2** — S3 promotion + S5 governance
- **3a–3e** — release-wide, parallel: Guard 1a, Guard 1b, #376, #194, capacity roll
- **4a–4g** — worked-example decision → V6 → contract-size decision → pre-registration → freeze → V7 → V8 → cut

**Why 1a leads.** It is the S lane's missing M0a — the fourth application of the rails-before-mechanism rule this plan already names three times. S1's framing (*"an identifier indexing four module-level dicts plus one function"*) was accurate about `scaffold.py` and was used as though it described the system. Four further per-stack surfaces (#818, #823, #828, #829) were then found one at a time, by accident, each while fixing the last. 1a closes the set instead of discovering it.

Also records the hard constraint that everything touching code lands before pre-registration, because V7 requires a frozen deploy — which is what makes Stage 3 parallel and Stage 4 strictly last.

### DIFF
--- a/docs/plans/1-6-0-authorship-plan.md
+++ b/docs/plans/1-6-0-authorship-plan.md
@@ -183,23 +183,72 @@ while 1.6's first deploy window is still far out, or any one of them turning liv
 
 ## Track M — Squad-Authored Manifest (SIP-0103)
 
-### Remaining sequence *(recorded 2026-08-09, after V5)*
+### Remaining sequence — **staged, recorded 2026-08-09 after the S-lane seam run**
 
-Track M and B1 are complete. What is left, in order:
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #834, opened 2026-08-10
feat(cycles): a cycle declares its stack once, not twice (#832)

jasonLadd:
**Stage 1b, step b1** — the first class-(b) item from the 1a sweep, and the only one carrying a design fork. Regression **6868 passed, 1 skipped**.

## The defect

A cycle named its stack in two config keys, indexing two registries, set to the same literal on adjacent CRP lines and bound by convention alone:

```yaml
  dev_capability: fullstack_fastapi_react   # -> DEV_CAPABILITIES
  build_profile:  fullstack_fastapi_react   # -> _STACKS
```

Nothing verified they agreed. The failure is not a crash: **the skeleton expands for one stack while the dev agent is instructed to write another's files**, so every emission lands outside the fill slots and the cycle surfaces as *"the plan claims no slots"* — a symptom several layers from its cause, a full framing workload after the mistake.

Fifth per-stack registry, after `_STACKS`, `_CRITERIA_PACKS`, `_PROFILES` and the sandbox's `_CONTRACTS`.

## Not a collapse — one rule instead

`python_cli`, `python_api` and `react_app` are free-form capabilities for cycles with **no** scaffold stack, so the two registries are not the same set. Not every dev capability is a stack; every stack needs one.

**The stack declares it; the config may restate it but not contradict it.**

| Config | Behavior |
|---|---|
| scaffoldable `build_profile`, `dev_capability` absent | **derived** |
| …agreeing | proceed |
| …**contradicting** | **blocked at create** |
| no scaffoldable `build_profile` | untouched |

**Blocked rather than derived-over**, per #762's precedent: both inputs are knowable at create, the downstream failure is deterministic rather than probable, and silently overriding an explicit config value hides the drift instead of ending it. `resolve_dev_capability` returns `None` for contradiction — deliberately distinct from `""` — so the caller must decide rather than quietly pick a side.

## A second defect the sweep exposed

**Five call sites defaulted an absent `dev_capability` to `"python_cli"`.** A scaffolded fullstack cycle that merely *omitted* the key was handed CLI prompts — an omission rather than a contradiction, so no equality check would ever have caught it. All five now resolve through the stack.

That one was not in the issue; it surfaced while wiring the resolver.

## Tests

Sixteen. The resolution truth table, the create-time block, the message naming both values and the fix, free-form cycles left untouched (a net that false-positives on a working configuration is worse than the gap it closes — #762), and the registry binding that keeps the fifth registry from drifting like the other four.

Closes #832 · Refs #822

### DIFF
--- a/src/squadops/api/routes/cycles/cycles.py
+++ b/src/squadops/api/routes/cycles/cycles.py
@@ -40,6 +40,7 @@ from squadops.cycles.preflight import (
     model_availability_decision,
     required_check_tooling_decision,
     required_roles_decision,
+    stack_dev_capability_decision,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #835, opened 2026-08-10
refactor(task_plan): injection asks the check what it can parse (#833)

jasonLadd:
**Stage 1b, steps b2–b4** — delivered far smaller than filed, because the issue's premise was wrong. Regression **6876 passed, 1 skipped**; behavior-identical for stack #1.

## The correction

**Filed as:** stack #1 layout hardwiring — a stack-aware seam ANDed with a literal that undoes it, needing a `source_suffixes` field on `ScaffoldStack`.

**Actually:** the Python-*analysis* boundary, and correct at all six sites.

| Site | What `.py` gates | Verdict |
|---|---|---|
| `task_plan.py:629, 655, 684` | injecting three checks that each declare `applicable_extensions={".py"}` | correct, wrongly expressed |
| `interface_conformance.py:121, 201` | `_model_classes` / `_route_decorators`, both `ast.parse` | correct by construction |
| `scaffold_enforcement.py:236` | restoring `status_code` into a Python decorator; docstring already says *"a non-Python or unparseable artifact passes through untouched"* | correct and documented |

`is_qa_test_path_for_stack` answers *"is this a QA test path for this stack"*; the suffix answered *"can the check parse it"*. Both are needed and neither undoes the other. **Stack #2 loses nothing that was not already disclosed** — the AST tier does not cover TypeScript, which is the S4 disclosure already in the plan.

**So no `source_suffixes` field.** It would have been invented to parameterize something that does not vary by stack — exactly what S5's admission rule exists to prevent.

## What was actually owed

The three `task_plan` sites restated the boundary by hand, in a place that cannot notice when the check's own declaration moves. They now call **`is_check_applicable`** — the same seam `manifest_gates` `PROOF_CHECKS_LIVE` and the dispatch-time strip already use — so the manifest gate and the injection filter read one source, and a future TypeScript implementation of any of those checks is picked up without touching these sites.

`CHECK_HARNESS_BOUNDARY` joins the named-constant block by the criterion those comments already state.

## A method defect in the 1a sweep, recorded on the issue

The sweep classified by **grep hit with one line of context** rather than by reading the enclosing function — the same proximity-for-scope failure as the #822 design gate naming the wrong evaluator.

**Consequence: the class (a)/(c)/(d) calls from Stage 1a carry the same risk and are provisional.** One of the four class-(b) items dissolved on contact; the others were confirmed by implementation. Stage 1c should re-read rather than trust the classification.

Closes #833 · Refs #822

### DIFF
--- a/src/squadops/cycles/acceptance_check_spec.py
+++ b/src/squadops/cycles/acceptance_check_spec.py
@@ -320,6 +320,12 @@ CHECK_CONTRACT_ASSERTIONS = "contract_assertions_match"
 # moves both together.
 CHECK_FILL_SLOT_SIGNATURE = "fill_slot_signature"
 
+# SIP-0100 / #833: the QA harness-boundary check. Named here because `task_plan` filters its
+# injection on the check's own applicability — the same single-source rule as the constants
+# above, and it earns one for the same reason they do: a literal in an injection filter
+# silently injects nothing after a rename.
+CHECK_HARNESS_BOUNDARY = "harness_boundary"
+
 # #822: the per-view bundler check. Named here because `VerificationContract.view_slots`
 # filters on it to identify a stack's view files for repair targeting — the same
 # single-source rule as the constants above, and the same failure mode if it drifts: a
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #836, opened 2026-08-10
feat(scaffold): the Next.js + TypeScript stack — stack #2 (#822, Stage 1c)

jasonLadd:
**Stage 1c.** Expander, criteria pack, probe profile, sandbox environment contract and dev capability — all five per-stack registries. Stack #1 byte-unmoved (`7622f570c949fe95` / `bb472e267e53d5ad`); regression **6890 passed, 1 skipped**.

17 files, 7 fill slots, against stack #1's 19 and 4. **Parity of ambition deliberately** — a better skeleton would make VS-vs-V5 uninterpretable.

## The bend that matters

**The API is many fill slots, not one.** FastAPI holds every endpoint in a single `backend/routes.py`. Next derives the URL from the directory, so the reference manifest's five endpoints become **four route files**:

```
app/api/runs/route.ts                  <- GET /runs + POST /runs
app/api/runs/[run_id]/route.ts
app/api/runs/[run_id]/join/route.ts
app/api/runs/[run_id]/leave/route.ts
```

The blueprint never expressed this — `_ROUTES_PATH` was singular. **Argued as a genuine cross-stack convention rather than a concession:** an API is a set of addressable operations, and how many files hold them is the stack's business. Supporting evidence — `endpoint_owners()` already returned a *dict* and survives unchanged, so the relation was modelled correctly even while the partition was not.

Five more bends are in the module docstring, including two disclosures: the per-slot bundler check costs seven whole-tree builds (kept at parity because #641/#648 anchor blame to the failing file), and **the route slots have no static check beyond the build** — the nine AST checks parse Python by construction, and `tsc --noEmit` is safelisted but *not provisionable* (`tsc` lives in `node_modules/.bin`, never on PATH — #707's "passes the allowlist but cannot run" class). `next build` runs tsc itself, so the bundler check is the type check.

## The two earlier seams pay off

**#827's `prepare_argv`** gets its first consumer: `next start` serves `.next/`, which only `next build` produces. A profile stopping at install would boot into *"no production build found"* and report it as a boot failure — exactly the conflation #827 separated.

**#828's `project_dir`** likewise: the pack emits `"."` because Next builds at the root.

## Three stack-#1-pinned tests updated — flagging for scrutiny

**The 1a sweep covered `src/` and `adapters/` but not `tests/`.** Tests encode stack assumptions too, and three did:

| Test | Change |
|---|---|
| `..._single_answer_to_which_stacks_exist` | asserted the set was exactly one stack. The invariant was never the *count* — it is that membership has one source — so it now asserts both, plus that the accessor still refuses a near-miss |
| `test_frontend_capabilities_declare_build_support` | required `index.html` of every frontend capability. That is **Vite's** entry point, not a property of frontends; Next generates its HTML and ships none, so the assertion would have forced a capability to declare a file its stack does not have. `package.json` — the real #290 invariant — is kept |
| `test_registry_has_four_capabilities` | **removed**, not bumped to 5. A count names no bug class, breaks on every legitimate addition, and is the shape `docs/TEST_QUALITY_STANDARD.md` rules out. What it appeared to protect is covered by its neighbours; the binding that matters is pinned by #832's test |

The removal is the one worth arguing about, so it is called out rather than buried.

## Fifteen new tests

Written against where the two stacks *disagree*, not that this one works — a suite that only proved it works would hide the evidence S3 needs. Includes the four-route-file partition, both parameter translations, the stubs failing their own probes (SIP-0098 §7), the pack requiring node and never python, and all five registries answering for the stack.

## Next

**1c-iv/v** are done as part of this. Remaining in Stage 1: **1d** (bend register — currently in the module docstring; S3 may want it as its own document) and **1e/VS** — a Next.js cycle end to end, which needs a deploy.

Refs #822

### DIFF
--- a/src/squadops/capabilities/dev_capabilities.py
+++ b/src/squadops/capabilities/dev_capabilities.py
@@ -334,6 +334,87 @@ DEV_CAPABILITIES: dict[str, DevelopmentCapability] = {
         max_completion_tokens=12000,
         test_timeout_seconds=180,
     ),
+    # #822 stack #2. Bound to the scaffold stack of the same name by
+    # ``ScaffoldStack.dev_capability`` (#832), so the two registries can no longer disagree
+    # about which stack a cycle is building.
+    "nextjs_ts": DevelopmentCapability(
+        name="nextjs_ts",
+        system_prompt_supplement=(
+            "You are generating source code for a Next.js (App Router) application in "
+            "TypeScript. One project at the repository root — there is no separate backend "
+            "or frontend tree.\n"
+            "Emit each file as a fenced code block with the language and path separated by "
+            "a colon. Examples:\n"
+            "```typescript:app/api/runs/route.ts\n"
+            "```tsx:app/runs/page.tsx\n\n"
+            "Paths must be clean relative paths with no colons in the path or spaces."
+        ),
+        file_structure_guidance=(
+            "\n\nGenerate complete, runnable source for a Next.js App Router application in "
+            "TypeScript.\n\n"
+            "### Routing\n"
+            "- Server endpoints are ROUTE HANDLERS at app/api/<path>/route.ts, exporting one "
+            "async function per HTTP method (`export async function GET(request: Request)`).\n"
+            "- Do NOT use server actions for endpoints the design declares: they have no "
+            "stable URL, so nothing can address them over HTTP.\n"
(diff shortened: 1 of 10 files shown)


jasonLadd (commented):
reviewed

PR #837, opened 2026-08-10
docs(stack-2): the bend register as its own document (#822, Stage 1d)

jasonLadd:
**Stage 1d.** Moves the bends out of `stack_nextjs_ts.py`'s docstring into `docs/plans/stack-2-bend-register.md`, leaving a **pointer rather than a copy** — a claim maintained in two places is this release's own recurring failure.

## Applying S3's definition strictly split six bends into four kinds

S3's definition is precise: a bend is *"a convention the stack had to adopt that it would not otherwise use, purely to satisfy the schema."* Held to that, the six became:

| # | Finding | Kind | Direction |
|---|---|---|---|
| 1 | the API is many fill slots, not one | **schema defect** | the blueprint was wrong |
| 2 | path parameters translate `{x}` → `[x]` | **cleared** | neither side bent |
| 3 | `frontend.routes[].view` has no home | **field with no home** | the schema does not map |
| 4 | route handlers, not server actions | **the bend** | the stack conceded |
| 5 | seven whole-tree builds | disclosure | a cost |
| 6 | no static check on route slots | disclosure | already known |

**The split is the useful part** — *"the schema is wrong here"* and *"the stack conceded here"* demand opposite responses, and lumping them together would have hidden both.

## One true bend

Entry 4 is the only one, and the register argues it **in both directions**: *"a design is declared as a set of addressable HTTP operations"* holds as a cross-stack convention and the probe tier depends on it — but it removes a genuinely idiomatic option from the stack, and it is silent for any future non-HTTP transport, which is narrower than "cross-stack" claims to be. **Recorded as argued, not settled.** S3 decides.

## Entry 1 was filed as a bend and is not one

The stack adopted nothing unnatural — one file per route segment is Next's own convention. It is the **blueprint** that could not express it. Calling it a bend would have put the concession on the wrong side and left the schema unexamined.

Its supporting evidence is unusually good: `endpoint_owners()` already returned a **dict** and survives this stack unchanged. The relation was modelled correctly all along; only the partition that produced it was singular. That is the difference between a schema that generalizes and one that happened to fit.

## Entry 2 is recorded despite being clean

Because a register that lists only problems cannot show that anything was *checked*. The path-parameter translation was examined — #820's third trigger firing exactly as predicted — and the schema came out clean.

Docs only. Plan step 1d updated with the outcome.

Refs #822

### DIFF
--- a/docs/plans/1-6-0-authorship-plan.md
+++ b/docs/plans/1-6-0-authorship-plan.md
@@ -201,7 +201,7 @@ against these stage/step labels.
 | **1a** | **Hardwiring sweep.** Enumerate every stack-#1 assumption in deliverable-facing modules and classify each: already-a-seam · needs-parameterizing · belongs-in-the-FastAPI-pack · not-on-stack-2's-path | every hit classified; the "needs" class filed |
 | **1b** | Parameterization PRs for the "needs" class | each: FastAPI byte-unmoved, regression green |
 | **1c** | **The Next.js stack** — expander, criteria pack, probe profile, environment contract, completeness test across all four per-stack registries | `expand()` yields a tree; `emit_contract_dict` yields a satisfiable contract |
-| **1d** | **Bend register**, written *during* 1c | zero unexplained bends (S3's exit criterion) |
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #840, opened 2026-08-10
docs(plan): decompose 1e and 4g

jasonLadd:
Intended for #839 but it merged first, so this is the same audit's remaining half. Docs only.

## 1e — VS acquired a prerequisite after the plan was written

It ran 2026-08-10 (`cyc_afa934886acd`), framed **75m33s**, and was rejected: the cycle built **stack #1 while configured for stack #2**, with every component behaving correctly. The chain is **#838**.

So 1e is now **1e-i** (three fixes) → **1e-ii** (re-roll) → **1e-iii** (read it). As one line it understated Stage 1's remaining work by a wide margin — it read as "run a cycle" and is "land three fixes, then run a cycle."

Records that **`PROOF_STACK_MATCHES_CONFIG` goes first**: cheapest of the three, independently valuable, and it converts a 75-minute discovery into a seconds-long one for the re-rolls the other two fixes will themselves need.

Also records what 1e-iii must show — the seven fill slots, whether the three required checks *execute*, and whether M6 can name the subsystem — and that **none of it is the pass rate**.

## 4g — criteria without a procedure

The plan carried cut **criteria** (the six-item gate) and no cut **procedure**. That is not academic:

**#789 records six consecutive releases tagged but never advertised** — v1.4.0 through v1.5.0, with `gh release list` still showing v1.3.1 as Latest and `CHANGELOG.md` last rotated at 1.3.1.

The backfill landed (v1.5.0 is Latest today), but **#789's item 3 — the CI guard meant to prevent recurrence — never did.** Verified 2026-08-10: **no workflow or dev script references `CHANGELOG` at all.** So the class is fixed historically and **unprevented going forward**, and a step represented by the single word "Cut" will be skipped a seventh time.

Now **4g-i…vi**: bump · marker sync · **CHANGELOG rotation** · ROADMAP entry · SIP-0103 status decision · tag **and** GitHub Release marked Latest. The tag alone is what six cuts did.

**4g-v records a recommendation rather than an action:** SIP-0103 **stays `accepted`**. Main implements a corrected, narrowed version of it — three falsified premises, one owner-ruled narrowing, four unimplemented dispositions — and promoting it would assert otherwise. Precedent: the 2026-08-03 audit kept four SIPs at `accepted` with gaps named.

## Two steps deliberately not decomposed

**4a** conflates a decision with a small prompt edit — a clarification, not a decomposition. **3b (Guard 1b)** has an unanswered *how*: feeding a hand-authored manifest through a path that authors manifests. Better resolved by whoever builds it than invented now.

Refs #822, #838, #789

### DIFF
--- a/docs/plans/1-6-0-authorship-plan.md
+++ b/docs/plans/1-6-0-authorship-plan.md
@@ -204,6 +204,31 @@ against these stage/step labels.
 | **1d** | **Bend register** — `docs/plans/stack-2-bend-register.md`, written *during* 1c | zero unexplained bends (S3's exit criterion). **Landed**: six findings, of which **one is a true bend** — the rest split into a schema defect, a cleared check, a field with no home, and two disclosures |
 | **1e** | **VS** — a Next.js cycle end to end | gates S3 |
 
+**1e is not one step — it acquired a prerequisite after this plan was written.** VS ran on
+2026-08-10 (`cyc_afa934886acd`), framed for 75m33s, and was rejected: the cycle built **stack
+#1** while configured for stack #2, and every component behaved correctly while doing so. The
+causal chain is **#838**.
+
+| Step | Work | Gate |
+|---|---|---|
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #841, opened 2026-08-10
feat(gates): a manifest must be for the stack the cycle is building (#838)

jasonLadd:
**Stage 1e-i(a)** — the first of #838's three fixes. Taken first because it is the cheapest, independently valuable, and it turns a 75-minute discovery into a seconds-long one for every re-roll the other two fixes will themselves need. Regression **6902 passed, 1 skipped**.

## The defect

VS (`cyc_afa934886acd`) was configured `build_profile=nextjs_ts`, the authored manifest declared `fullstack_fastapi_react`, and **every component behaved correctly while building the wrong application for 75 minutes.**

`lint()` accepts any *registered* stack — *is it known*, never *is it the one asked for* — so nothing downstream had a reason to object. The rejection that eventually fired was incidental (a frozen-file claim), not the real problem.

## The proof

Compares the declared stack to the cycle's `build_profile`, threaded from the handler where it is already resolved for the prompt. **Empty skips the comparison** — seeded cycles and the gate's own adversarial fixtures have no cycle config, and inventing a mismatch for them would reject correct manifests (#762's lesson that a net which false-positives is worse than the gap it closes).

**It returns alone**, like a parse failure and for the same reason: a wrong stack invalidates the frame every other proof reasons in. The remaining proofs run happily against the wrong application and produce real-looking findings about testids and statuses — for a design about to be replaced wholesale. Telling an author to fix a testid in a manifest whose stack is wrong is worse than telling them nothing.

## Two drift guards required feeding, which is the #686 pattern working

**M6's `_PROOF_CLASS`** → `AUTHORING_DEFECT`, with a condition recorded in the code: **if this class recurs once the design-stage plumbing lands, it is a `blueprint_limitation` instead** — and B1's recurrence data is exactly what tells the two apart. That is the taxonomy's own "single occurrence takes the conservative class, recurrence promotes it" rule applied at the moment it matters.

**`manifest_authoring_rules`** → `AUTHOR_FACING`, with a new taught rule `use-the-stack-you-were-given` in the managed asset (#448).

**Filed `AUTHOR_FACING` rather than `COVERED_ELSEWHERE` deliberately.** The authoring request's schema block *does* render `stack: {{stack}}`, so "already taught" was available and would have been less work. **VS falsified it** — the author saw that line and wrote another stack's name anyway, having inherited it from a technical design naming the other one eight times. Classifying it as covered would assert a teaching that has been measured failing.

The rule's last sentence is the one that matters: *if the technical design you were given describes a different architecture than the stack you were given, the stack wins and the design is the thing that was wrong.*

## What this does not fix

**The cause is still upstream.** The design stage is never told the stack — that is 1e-i(b) — and `BUILD_PROFILES` has no `nextjs_ts` entry, with an `except Exception` that would swallow its absence — 1e-i(c). This proof is the deterministic backstop that holds regardless of what any model writes; it is not the plumbing.

Refs #838, #822

### DIFF
--- a/src/squadops/capabilities/handlers/planning/manifest.py
+++ b/src/squadops/capabilities/handlers/planning/manifest.py
@@ -100,7 +100,9 @@ class DevelopmentAuthorManifestHandler(_PlanningTaskHandler):
                     f"No `{SEEDED_MANIFEST_FILENAME}` block was found in your response. "
                     f"Emit exactly one fenced block whose header carries that filename."
                 )
-            outcome = assess_authoring_outcome(yaml_or_none)
+            # #838: the gate compares the declared stack to the one this cycle is
+            # configured for. `stack` is already resolved above for the prompt.
+            outcome = assess_authoring_outcome(yaml_or_none, stack)
             outcomes.append(outcome)
             if not outcome.rejected:
                 return yaml_or_none, None
--- a/src/squadops/cycles/authoring_failure.py
+++ b/src/squadops/cycles/authoring_failure.py
@@ -35,6 +35,7 @@ from squadops.cycles.manifest_gates import (
     PROOF_LINT,
     PROOF_PARSES,
     PROOF_SOURCE_PRD,
+    PROOF_STACK_MATCHES_CONFIG,
     PROOF_STATUS_DECLARED,
     PROOF_TESTID_COVERAGE,
     WinnabilityFinding,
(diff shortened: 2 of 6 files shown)


jasonLadd (commented):
reviewed

jasonLadd (commented):
reviewed

PR #842, opened 2026-08-10
feat(framing): the configured stack outranks whatever the PRD says (#838)

jasonLadd:
**Stage 1e-i(b)** — and the premise changed on investigation. Regression **6920 passed, 1 skipped**.

## What #838 said, and what is actually true

The issue said *"the design stage is never told the stack, so it designed FastAPI from the only architecture prose that exists."* Reading further:

**The group_run PRD names its stack in prose, twice.**

```
line 36: "a coherent, runnable full-stack vertical slice (FastAPI + React)"
line 42: "Deliver a runnable full-stack MVP (FastAPI backend + React frontend)"
```

Research summarised it (fastapi 3, react 4, **next 0**), the design stage designed it, and the manifest author inherited it through `prior_outputs` — a *declared* input. **The design stage was not uninformed. Every stage was obediently following the requirements**, which disagreed with the cycle's configuration and won for being the louder input.

That is the same defect class yet again: another unbound declaration of the stack, this time in the benchmark's own requirements.

## So the fix is a precedence rule, not plumbing

**The stack the cycle configures outranks anything the PRD says about architecture** — stated on the shared planning base so **all four** framing stages inherit it.

Telling only `development.design_plan` would have left the contamination path open one step upstream, in research, which is where the FastAPI mention was first summarised. That is the difference between fixing the stage that failed and fixing the stage that fed it.

**Rendered after the PRD** in the template: a precedence rule read *before* the thing it overrides is preamble.

Prose lives in managed assets (#448) — `request.target_stack_section` carries the rule, `profile_narratives/nextjs_ts.md` the architecture.

## One deliberate asymmetry

`narrative_for_stack()` is **tolerant** where `_narrative()` is **loud**. The existing one raises at import for a `BUILD_PROFILES` entry missing its file — a registration error, correctly fatal. This answers a different question, at prompt-assembly time, about a stack that may legitimately not have a narrative yet — and a framing stage must not die because one is absent. It returns the empty string, and the section renders nothing rather than an authoritative-looking empty heading.

## Tests

Thirteen, including the one that matters most: **every framing stage** gets the section, not just the design one — parametrized across `DataResearchContextHandler`, `StrategyFrameObjectiveHandler` and `DevelopmentDesignPlanHandler`. Plus silence for free-form cycles, silence for a registered stack with no narrative, and the template ordering assertion.

## What this does not do — and the decision it raises

**It does not touch the PRD.** Removing the stack from the requirements would be the *cleaner* fix — architecture is the squad's job, which is SIP-0103's whole claim — but the PRD is the benchmark's fixed input, and **changing it would invalidate V4/V5 comparability for V7.** Raised separately rather than decided here.

Refs #838, #822

### DIFF
--- a/src/squadops/capabilities/handlers/build_profiles.py
+++ b/src/squadops/capabilities/handlers/build_profiles.py
@@ -43,6 +43,21 @@ def _narrative(profile_name: str) -> str:
     return text[:-1] if text.endswith("\n") else text
 
 
+def narrative_for_stack(stack: str) -> str:
+    """The architecture narrative for a scaffold stack, or ``""`` if it has none (#838).
+
+    Deliberately tolerant where :func:`_narrative` is loud. That one raises at import for a
+    BUILD_PROFILES entry missing its file — a registration error. This answers a different
+    question, asked at prompt-assembly time about a stack that may legitimately not have one
+    yet, and a framing stage must not die because a narrative is absent.
+    """
+    path = _NARRATIVES_DIR / f"{stack}.md"
+    if not path.is_file():
+        return ""
+    text = path.read_text(encoding="utf-8")
+    return text[:-1] if text.endswith("\n") else text
+
+
 # ---------------------------------------------------------------------------
 # QA handoff section name constants (D12 — single source of truth)
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #845, opened 2026-08-10
feat(build): register nextjs_ts, and stop swallowing a missing build profile (#838)

jasonLadd:
**Stage 1e-i(c)** — the last of #838's three fixes, **plus a live defect I introduced in #842** that this work surfaced. Regression **6933 passed, 1 skipped**.

## The registry gap

`nextjs_ts` was registered in five per-stack registries and absent from `BUILD_PROFILES`, the sixth. `_seed_skeleton_artifacts` resolves `get_profile(manifest.stack)`, which raises for an unregistered name — and the call site wrapped it in one broad `except Exception` returning `[]`.

**So a *correct* `nextjs_ts` manifest would have been silently unscaffolded, leaving the *wrong* manifest as the only one that worked.** The failing path was the correct one.

Registration and expansion are now separate concerns: a registered-but-unprofiled stack **raises**, naming the disagreement; an expansion failure on a properly registered stack **stays tolerant**, because that fallback predates SIP-0099 and making it loud would fail cycles that work today. Same separation #827 made for the probe path, for the same reason — two defects, two diagnoses.

An unscaffolded cycle has no contract, no fill slots and no frozen files. It is not a degraded run; it is a run whose entire verification story is absent, and it would report `completed`.

## A live defect in #842, already merged

**#842 pointed framing's target-stack section at `prompts/profile_narratives/` — which is the *builder's* asset.** Those prompts open *"You are assembling…"* and instruct *"do not regenerate application code — focus on packaging, configuration, and operational readiness."*

So since #842 merged, every framing stage on a `fullstack_fastapi_react` cycle has been handed the packaging prompt under a heading reading **"TARGET STACK — decided, not proposed"** — telling the stage that authors the technical design not to write application code.

Two audiences reading one asset. That is the ownership smell CLAUDE.md's edit-time rule exists to catch, arriving one release after that rule was written, from me.

```
prompts/profile_narratives/  ->  how is this stack PACKAGED   (builder)
prompts/stack_narratives/    ->  how is this stack WRITTEN    (framing)
```

`capabilities/stack_narratives.py` owns the second; `build_profiles` keeps its own loader unchanged. This also required writing the FastAPI **architecture** narrative that never existed — only its packaging one did.

**The #92 invariant makes the split concrete:** a builder narrative must not enumerate filenames, while an architecture narrative is largely *about* where things live. One asset could not have satisfied both, which is why the test caught it.

## Two test changes, flagged rather than buried

**`test_registry_has_four_profiles` removed**, not bumped to 5. A count names no bug class, breaks on every legitimate registration, and is the shape `TEST_QUALITY_STANDARD.md` rules out — same removal as the `DEV_CAPABILITIES` count in #836. The binding that matters is pinned in the new `test_build_profile_registration.py`: every scaffoldable stack has a profile, and its absence raises.

**The #92 filename heuristic now exempts framework product names containing a dot.** It could not tell `Next.js` from `main.js`, and the existing narratives escaped it only by naming frameworks that happen to have no dot (FastAPI, React, Vite). Naming the framework is not what #92 forbids — enumerating the files the validator grades is. This is the change most worth a second opinion, so it is called out.

`nextjs_ts`'s builder narrative joins the #452 byte-pin regime. It has no pre-move literal, so its pin establishes a baseline rather than verifying a move — the same guarantee going forward.

## #838 is now complete

All three fixes landed: the deterministic backstop (#841), the precedence rule (#842), and this. **VS can re-roll** — 1e-ii.

Closes #838 · Refs #822, #842

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -1764,9 +1764,32 @@ class DispatchedFlowExecutor(FlowExecutionPort):
 
         if not is_scaffoldable_stack(manifest.stack):
             return []
+
+        # #838: a REGISTRATION failure and an EXPANSION failure are different defects with
+        # different diagnoses, and one broad `except` conflated them. `is_scaffoldable_stack`
+        # has already passed, so the stack is in `_STACKS`; if it is absent from
+        # `BUILD_PROFILES` those two registries disagree, and swallowing that produced the
+        # inversion VS found — a *correct* `nextjs_ts` manifest would have been silently
+        # unscaffolded, leaving the wrong manifest as the only one that worked. An
+        # unscaffolded cycle has no contract, no fill slots and no frozen files: it is not a
+        # degraded run, it is a run whose entire verification story is absent. Loud.
         try:
-            files = get_profile(manifest.stack).expand(manifest)
+            profile = get_profile(manifest.stack)
+        except ValueError:
+            logger.error(
+                "stack %s is registered as scaffoldable but has no build profile — the two "
+                "registries disagree, and proceeding would produce an unscaffolded cycle with "
+                "no contract, no fill slots and no frozen files",
+                manifest.stack,
+            )
+            raise
+
+        try:
+            files = profile.expand(manifest)
         except Exception:
+            # An expansion failure on a properly registered stack stays tolerant: the run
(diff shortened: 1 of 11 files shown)


jasonLadd (commented):
reviewed

PR #847, opened 2026-08-10
fix(checks): one command allowlist, measured against the images (#707)

jasonLadd:
## What this fixes

Two allowlists gated `command_exit_zero` and **disagreed in both directions** — #707's own measurement, re-confirmed on `main`. #846's Family A is the bill coming due: VS's Next.js re-roll (`cyc_0edb55919384`) lost a 75-minute framing run because the squad reached for `tsc --noEmit` **exactly as the safelist advertised it**, and plan validation refused it.

## Measured, not inferred

I ran the tools in both agent containers (`squadops-eve`, `squadops-neo`, 2026-08-10) rather than reading requirements files:

| form | eve (qa) | neo (dev) | old safelist | old env gate | verdict |
|---|---|---|---|---|---|
| `python -m py_compile` | ✓ | ✓ | ✓ | ✓ | **kept** |
| `node --check` | ✓ | ✓ | ✓ | ✓ | **kept** |
| `pyflakes <file>` | ✓ | ✓ | ✓ | ✗ | **kept** — the one Python linter installed was the one refused |
| `ruff check` | MISSING | MISSING | ✓ | ✗ | dropped |
| `tsc --noEmit` | MISSING | MISSING | ✓ | ✗ | dropped |
| `python -m mypy` | MISSING | MISSING | ✓ | **✓** | dropped |
| `eslint <args>` | v6.4.0 | v6.4.0 | ✓ | ✗ | dropped — present but exits 2 with no config before reading a line (#645's fails-on-any-content class) |

`python -m mypy` is the instructive one: it **cleared both gates** and then failed at evaluation with `No module named mypy`. `argv[0]` is `python`, so an executable-name gate is structurally blind to it — which is why forms now declare the tool they *need*, not the binary they start with.

## The fix is structural, not a resync

- `CommandPattern` gains `tool` — what the form actually needs.
- `CHECK_ENV_TOOLS` declares what the images provide, **beside** the safelist. Two modules cannot hold one fact without eventually disagreeing about it.
- `_assert_safelist_is_runnable()` **refuses to import** a self-contradictory vocabulary. A test catches drift; raising means the contradictory state cannot exist (the #845 precedent). Static module content only — it can fire on an edit to that file, never on runtime data. Verified by mutation.
- `_CHECK_ENV_EXECUTABLES` **deleted** rather than resynchronised. Asking the safelist is strictly stronger: the old test passed `["python", "-c", "1"]` on `argv[0]` alone and left the refusal to evaluation, an hour of correction budget later.

This satisfies #707's acceptance: *one structure answers "may a plan author this command", with no second list to keep in sync*, and a parity test derived from the structures.

## A second hole, found by the same measurement

`node --check` refuses `.ts` with `ERR_UNKNOWN_FILE_EXTENSION` exactly as it does `.tsx` (node v20.19.2), but `_NODE_UNPARSEABLE_SUFFIXES` listed only `.jsx`/`.tsx`. Invisible while the tree held one Python stack. After the prune it was **the one form a TypeScript author could still reach**, and it fails on correct code.

## So stack #2 has no command form — and that is declared, not omitted

The Stack Blueprint SIP argues against its own `analysable_suffix` sketch on exactly this ground: the frontend *"was not modeled as a second language without checkers; it was silently absent, and absence demands no handling… A partial model is worse than none: it implies the rest does not exist."*

A shorter safelist with nothing said about TypeScript repeats that. So `LANGUAGES_WITHOUT_COMMAND_FORM` declares the gap with its reason, and `verified_by` is the load-bearing field — an empty command surface is only acceptable while something else carries the claim, and `next build` does. It renders into the generated `docs/architecture/typed-check-menu.md`, replacing the standing caveat that called this surface *"untrustworthy until #707's allowlist inventory"*.

This does **not** mint per-stack vocabulary: it lives in `acceptance_check_spec` beside `DECLARED_UNBUILT_CHECKS`, not on `ScaffoldStack`. Stage 2a/2b owns the schema question and this deliberately does not prejudge it.

## Three prompt surfaces restated the safelist by hand, and all three had drifted

The dev fragment also told authors that dev containers carry `ruff` and that node exists only in QA — **both false** (neither image has ruff; both have node/npm).

- The Python-hosted block now renders from `command_safelist_names()`.
- The two fragments are prose and cannot render, so a test binds them **in both directions**: naming a retired tool is the trap that fired; failing to name a live form is the quieter half, since an unadvertised check is an unused one.

That test caught this PR's own first draft, which named `tsc` in order to *warn* about it. Naming a forbidden command in a prompt is how it gets emitted, so the fragments now state only the positive.

## Bend register

Entry 6 amended in this PR per its own written-during rule. Its stated fact ("`tsc --noEmit` **is** on the safelist but is not provisionable") stops being true here. The disclosure comes out **stronger**, not weaker.

## Verification

- Regression **6930 passed**, 1 skipped.
- Reference contract `7622f570c949fe95` and manifest `bb472e267e53d5ad` **unmoved** (M0a's pins).
- Import tripwire verified by mutation — adding a `tsc` form raises at import.
- Fragment manifest hashes regenerated via `scripts/dev/regen_fragment_manifest.py --write`.

## Scope note

Deliberately **not** here: #846's Family A2 (plan validation judging a vitest suite by pytest discovery) and Family B. Those follow in a stacked PR off this branch. #846 filed both families together; they split cleanly because this half was already specified in #707 with its own acceptance criteria, and burying it would have hidden that the reconciliation was diagnosed long before it was paid for.

Closes #707
Refs #846, #822, #645

### DIFF
--- a/docs/architecture/typed-check-menu.md
+++ b/docs/architecture/typed-check-menu.md
@@ -23,9 +23,31 @@ Regenerate: `UPDATE_CHECK_MENU=1 pytest tests/unit/cycles/test_check_governance.
 | `regex_match` | authored | product | yes | yes | yes | yes | error |
 | `undefined_names` | injected | product | yes | yes | yes | yes | error |
 
-Caveat — `command_exit_zero` ownership is per-command in truth and
-untrustworthy until #707's allowlist inventory + precedence ruling
-(recorded in the registry beside the entry).
+`command_exit_zero` ownership is per-command in truth. The forms it may take
+are inventoried below — this replaces the standing caveat that called the
+surface untrustworthy pending #707's allowlist inventory.
+
+## Authorable `command_exit_zero` forms
+
+One list, not two (#707): a form is authorable exactly when the tool it needs is
+provisioned, and `acceptance_check_spec` refuses to import if that stops being true.
+Entries are measured in the agent images, never assumed.
+
+| form | tool needed |
+|---|---|
+| `python -m py_compile <file>` | `python` |
+| `node --check <file>` | `node` |
+| `pyflakes <file>` | `pyflakes` |
+
+### Languages no form reaches
+
+Declared, not omitted — a list that simply fails to mention a language reads as
+complete. What carries the claim instead is named, because an empty command
+surface is only acceptable while something else verifies the code.
+
+| language | why no form | verified instead by |
+|---|---|---|
+| TypeScript (.ts/.tsx) | no TypeScript checker is provisionable — tsc lives in the app's own node_modules/.bin and never on PATH, and `node --check` refuses both extensions before parsing (ERR_UNKNOWN_FILE_EXTENSION, node v20.19.2, measured 2026-08-10) | frontend_compiles / frontend_build — `next build` runs tsc itself and next.config.mjs declines to ignore type errors, so the bundler check IS the type check (#822 bend register entry 6) |
 
 ## Declared-unbuilt (visible, not evaluable, not authorable)
 
--- a/docs/plans/stack-2-bend-register.md
+++ b/docs/plans/stack-2-bend-register.md
@@ -125,13 +125,33 @@ targeting for wall-clock, and repair targeting is what #650 and #688 were both w
 
 ## 6 — No static check on the API route slots beyond the build · *disclosure*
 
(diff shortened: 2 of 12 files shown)


jasonLadd (commented):
reviewed

PR #848, opened 2026-08-10
fix(plan): plan validation knows the stack, and the sole author gets the contract (#846)

jasonLadd:
**Stacked on #847** (`feature/707-one-command-allowlist`) — both touch `implementation_plan.py`. Merge #847 first; this PR's diff against `main` will then be just its own two commits.

Covers #846's **A2** and **Family B**. Family A1 is #847, because it was already specified in #707 with its own acceptance criteria.

---

## A2 — plan validation judged a vitest suite by pytest discovery

`validate_check_applicability` asserted `test_*.py` / `*_test.py` unconditionally. On a vitest stack that rejects **every correct qa task**: VS's re-roll was refused for declaring `__tests__/store.test.ts`, and the remedy the message offered — "include a test_*.py" — would have been wrong.

The stack has declared this since SIP-0072 (`DevelopmentCapability.test_file_patterns`; `nextjs_ts` names the four vitest forms). Nothing asked it. Now reached through `ScaffoldStack.dev_capability` (#832) — an **explicit declared pointer**, not a naming convention, which is the distinction Stage 2a exists to close.

**This was my class-(d) miss in the Stage 1a sweep.** I verified `run_node_tests` discovers `package.json` and concluded the vitest path was fine. That was about *execution*. Plan validation carried its own hardcoded assumption and I never looked at it.

One deliberate non-reuse: the basename matcher is now shared with `handlers.cycle.validation._is_test_file`, but its `/__tests__/` clause is **not**. That generosity is right for *excluding* files from a source set and wrong for "will the runner collect this?" — `__tests__/helpers.py` is collected by neither runner, and sharing the clause would have quietly widened #715's guard for every stack. Pinned by test.

Also collapses `resolve_dev_capability(cfg) or "python_cli"`, written out at six call sites across two handlers. A seventh copy was the wrong way to add the first non-Python stack.

---

## Family B — the answer is neither option the issue offered

#846 framed B as *"a wiring gap (the #796 class, partial) versus an authoring defect"* and correctly marked it unverified. It is neither.

**The proposers never ran.** `merge_decisions.yaml` for `cyc_0edb55919384`:

```yaml
authoring_mode: sole_author
sole_author_reason: no_contributors_configured
proposal_ids: []
canonical_tasks:
- merge_action: gap_filled
  reason: Sole-author fallback via PlanAuthoringService — no proposer contribution available.
  # ...for all 8 tasks
```

`bind_criteria_index` is declared only on `*.propose_plan_tasks`. `governance.merge_plan` had **no entry in `CONTEXT_CONTRACTS` at all**, so `inject_contract_inputs` injected nothing and `produce_plan` built its prompt from the PRD and framing documents alone.

So the one authoring path that reaches implementation **on every CRP except `validation-multirole`** was the only one never given the contract.

### The evidence in the emitted plan is unambiguous

| | |
|---|---|
| `criteria_refs` in `implementation_plan.yaml` | **0** — it could not bind an index it never saw |
| frozen files claimed as deliverables | 3 |
| invented paths | 8 |
| fill slots claimed by no task | 3 |

The invented paths being *plausible variants* — `app/api/runs/[id]/route.ts` for the slot's `[run_id]`, `app/create/page.tsx` for `app/runs/new/page.tsx` — is precisely what an author holding the manifest's endpoint list and none of the contract's file list produces. #846 suspected this and could not confirm it; the zero `criteria_refs` count settles it.

This is **#796's class at a seam #796 did not touch**: #796 fixed *when* the contract exists in authored mode. This fixes *who is told*.

### The fix

Split along the #663 S3 line — the registry declares who receives an input class, the handler owns the mechanics:

- `governance.merge_plan` gains `ContextAssemblyContract(bind_criteria_index=True)`.
- `contract_surface_sections` renders the two **existing** managed assets and is shared with the proposer's two methods. No new prose (#448).

### The test that pinned the defect

`test_injection_follows_the_registry_not_the_task_name` asserted `"contract_criteria_index" not in inputs` for merge_plan, docstringed *"the merger consumes proposer outputs and is deliberately excluded."* True of the merge path, and silent about the sole-author path that actually runs. Inverted here with the reason recorded, plus `test_strategy_proposer_is_still_excluded` as the polarity guard — without it, "everyone gets everything" would pass.

A test pinning an exclusion is a claim about **every** path through that task type, not just the one its author had in mind.

### A premise correction on my own earlier work

At #762 I deliberately did not add contributors to `validated-fullstack.yaml`, reasoning it "would tax author-mode users." The consequence is that author-mode users got the path with **no contract at all** — and #762's preflight block requires `contract_ref`, so authored mode falls through its middle condition by design. Configuring contributors would have masked this rather than fixed it.

---

## Verification

- Regression **6938 passed**, 1 skipped.
- Both fixes verified by **mutation** — removing either makes its test fail on the intended assertion.
- Plan-context golden regenerated **in the same commit** (exactly two keys added to the merge_plan envelope), so the behavior change reads as one rather than as a silent refactor side effect.
- Author-mode polarity guard: with no contract, the sole-author prompt is byte-identical.

## Still open on #846

**B1's remaining question.** With the author now told which files are frozen, is a frozen claim still possible? `validate_frozen_artifact_ownership` exists and is author-facing (`no-frozen-claims`), so the gate should catch what teaching misses — but VS's plan reached the gate *with* frozen claims, and whether that rule fired is worth confirming on the next roll rather than asserting here.

Refs #846, #796, #762, #832, #715, #663

### DIFF
--- a/src/squadops/capabilities/context_assembly.py
+++ b/src/squadops/capabilities/context_assembly.py
@@ -157,15 +157,25 @@ CONTEXT_CONTRACTS: dict[str, ContextAssemblyContract] = {
     # --- planning chain (#657): upstream documents on an envelope-local
     # prior_outputs copy; all four authoring types receive a re-roll's
     # rejection context (#669 — fresh dice previously meant zero teaching,
-    # fay-10 re-tripped the same class three times). governance.merge_plan is
-    # DELIBERATELY absent from BOTH: by merge time both proposals collide on
-    # the proposed_plan_tasks.yaml filename, so artifact threading would
+    # fay-10 re-tripped the same class three times). governance.merge_plan takes
+    # NEITHER artifact threading nor rejection context: by merge time both proposals
+    # collide on the proposed_plan_tasks.yaml filename, so artifact threading would
     # silently drop one proposal (the merger consumes brief_outcome/
     # proposal_outcome output keys instead), and its normal merge path is
     # deterministic — no prompt for rejection context to teach.
     # bind_criteria_index (SIP-0098 98.3): dev/qa propose BUILD tasks and so
     # bind covered-file criteria; strategy proposes guidance and the brief
     # author frames — neither is indexed.
+    #
+    # #846: "its normal merge path is deterministic" was true of the merge and hid the
+    # case that matters. When no plan_authoring_contributors are configured the proposers
+    # never run, and merge_plan authors the whole plan itself through an LLM — the
+    # sole-author fallback, which is the DEFAULT for every CRP but validation-multirole.
+    # Reading merge_plan's absence here as "this task type has no prompt" left the one
+    # author that actually reaches implementation as the only one never given the
+    # contract. Measured on cyc_0edb55919384: 0 criteria_refs, 3 frozen files claimed,
+    # 8 invented paths. So it is present now, for the index alone.
+    "governance.merge_plan": ContextAssemblyContract(bind_criteria_index=True),
     "governance.prepare_plan_authoring_brief": ContextAssemblyContract(
         artifact_filter=ArtifactFilter(
             by_producing_task=(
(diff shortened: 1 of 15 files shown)


jasonLadd (commented):
reviewed

PR #850, opened 2026-08-10
fix(contract): criterion ids identify a slot, and a malformed contract is refused (#849)

jasonLadd:
Closes #849 — all three findings.

## A · Criterion ids now identify a slot, not a filename

The `nextjs_ts` pack slugged the **basename**. App Router fixes the filename and puts identity in the directory, so `cyc_7899ed1feae5`'s contract came out as **nine criteria under four ids**:

```
vc-compiles-route   x4   (app/api/runs/route.ts, .../[run_id]/route.ts, .../join, .../leave)
vc-compiles-page    x3   (app/page.tsx, app/runs/new/page.tsx, app/runs/[run_id]/page.tsx)
```

`criterion_index()` is keyed on id and documents itself as last-writer-wins, so **five of seven fill slots resolved to no criterion at all** — while the plan, which bound all seven refs correctly, read as fully covered and passed every gate.

That is #818's class verbatim (*"an incorrect contract that is believed"*), one layer down, and the same FastAPI-shaped assumption as `_ROUTES_PATH`: that a filename identifies a file. Mine, from #822.

**`_view_criteria` is deliberately untouched.** Stack #1's views land in one flat directory, so its basenames are unique *by construction of its own expander* rather than by luck — and changing it would move `7622f570c949fe95`, the hash #777's guard and every bind-mode cycle are pinned to.

## B · The linter that catches this had no production caller

`lint()._lint_ids` detects duplicate ids precisely and has existed throughout. Its only callers, repo-wide, were `scripts/dev/contract_gate.py` and unit tests — **both running against the reference FastAPI manifest, whose ids cannot collide.** The guard was exercised solely by the one input incapable of failing it, and was absent from every path that derives a contract at cycle time.

`_contract_findings` now lints before reading the contract's contents, and returns alone on failure: a contract whose own linter rejects it is not something the per-check findings below it can reason about — duplicate ids in particular make `criterion_index()` lossy, so those findings would be computed over criteria no ref can resolve to.

Linting at the winnability gate rather than at storage is deliberate. This is the gate that already answers *"is this manifest winnable"*, a malformed contract is exactly a way for it not to be, and a rejection here re-rolls framing for free (#522) instead of failing mid-run.

## C · A frozen file with no declarations may not be checked

**I mis-diagnosed this initially and corrected it before building.** The appendix already says *"do not write typed checks against a frozen file unless the line below proves it passes"* — the rule was never missing.

The real gap is one layer down: `frozen_surface_index_lines` renders declarations for `.py` only, so on stack #2 all ten frozen files are bare names and the rule's escape hatch has **no line to reason from**. Roll 3's single rejection was an author guessing into that vacuum.

The appendix now states that an entry with no declarations may not be checked at all. Teaching the index to read TypeScript is S4's territory (typed checks off hardcoded `.py`) and is recorded there rather than built here under deadline.

## Test shape

Both new suites are **parameterized over `_STACKS`**, not written against `nextjs_ts`. The assumption "a filename identifies a file" is FastAPI-shaped and the next stack gets to break it too — and stack #1 passes by construction, which is exactly why a stack-specific test would have proved nothing.

The gate test **monkeypatches the pack to re-create the defect** rather than reverting the emitter fix. Otherwise it would silently degrade into a duplicate of the emitter's uniqueness test and stop covering the wiring — the half that was actually missing.

There is also a third assertion beyond uniqueness: ids must name the *directory that distinguishes the slot*. Uniqueness alone is satisfiable by a counter (`vc-compiles-1`), which is stable only until a slot is inserted in the middle, and would make a criterion id in evidence say nothing about which file it judged.

## Verification

- Regression **6946 passed**, 1 skipped.
- Stack #1 contract `7622f570c949fe95` and manifest `bb472e267e53d5ad` **unmoved**.
- Both stacks now emit **14 criteria / 14 distinct ids** and lint clean.
- **Mutation-verified both ways**: reverting the emitter fix fails 3 tests; unwiring the lint fails the 2 gate tests.
- Stack #2's contract hash moves, which is the point.

## Note on how this was found

Roll 3 found this by failing on an unrelated criterion. Had that regex not been authored, the cycle would have passed into implementation with **most of its verification contract silently unenforced** — a green run over an unverified surface. The rejection that looked like the problem was the only thing that stopped a worse one.

Every defect these three rolls surfaced was statically decidable; none needed a cycle to discover. That is the argument for the mechanical stack inventory recorded on #849 rather than another re-roll, and it is what comes next.

Closes #849
Refs #822, #818, #777

### DIFF
--- a/src/squadops/capabilities/scaffold_contract.py
+++ b/src/squadops/capabilities/scaffold_contract.py
@@ -236,13 +236,31 @@ def _nextjs_ts_slot_criteria(manifest: InterfaceManifest, path: str) -> dict[str
     Consequence stated plainly: **no criterion here proves the declared endpoints exist in
     the source.** ``endpoint_defined`` is Python-only, so that claim is carried by the
     behavioral probes alone — late and functionally, rather than early and structurally.
+
+    **The criterion id is derived from the whole slot path, not the basename (#849).** It was
+    the basename, copied from ``_view_criteria``, and on this stack that produced *colliding
+    ids*: App Router fixes the filename and puts identity in the directory, so all four route
+    files slugged to ``route`` and all three pages to ``page`` — nine criteria, four distinct
+    ids. ``VerificationContract.criterion_index()`` is keyed on id and documents itself as
+    last-writer-wins, so five of seven fill slots resolved to no criterion while a plan that
+    bound all seven refs read as fully covered and every gate passed. Measured on
+    ``cyc_7899ed1feae5``. Not a coverage gap — #818's "incorrect contract that is believed",
+    one layer down, and the same FastAPI-shaped assumption as ``_ROUTES_PATH``: that a
+    filename identifies a file.
+
+    ``_view_criteria`` still slugs the basename and is deliberately left alone. Stack #1's
+    views are emitted into one flat directory, so its basenames are unique *by construction of
+    its own expander* rather than by luck — and changing it would move the contract hash
+    ``7622f570c949fe95`` that #777's guard and every bind-mode cycle are pinned to. The
+    general protection against a future stack repeating this is the id linter, now wired at
+    derivation, not a second hand-audit of the emitters.
     """
     return {
         "interface": [],
         "implementation": [
             {
                 "check": "frontend_compiles",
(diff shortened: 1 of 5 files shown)


jasonLadd (commented):
reviewed

PR #851, opened 2026-08-10
test(stacks): 2a — the stack inventory as an executable enumeration

jasonLadd:
**#850 is merged, so this targets `main` directly.** Stage 2a from the 1.6 plan: *"one place answering 'which stacks exist and what does each declare', with today's six registries as members."*

## Why a test rather than a document

1a was the document version of this step, and it **overclaimed three times**: a premise checked by grepping with one line of context instead of reading the enclosing function; a sweep scoped to `src/`+`adapters/` that missed `tests/`; and a module never looked at, which became #838 and cost a 75-minute framing run to rediscover.

A read-and-classify pass fails the way reading fails. An enumeration derived from the registries cannot skip a member, because it doesn't know which members it's looking at.

## What it checks

Four properties, each a defect that actually shipped, generalized so the next stack meets it as a test failure:

| check | the defect it generalizes |
|---|---|
| membership across all six registries | #838 — `BUILD_PROFILES` had no `nextjs_ts` entry, swallowed, silently unscaffolded |
| every fill slot reachable by a `criteria_ref` | #849 — colliding ids dropped 5 of 7 slots while the plan read fully covered |
| every guard has a production caller | #849 — `lint()` called only from CI and tests, both running the one input incapable of failing it |
| declared language coverage on per-stack renderers | the frozen index, below |

Membership records **how** each registry is addressed, not just that it is. Three resolve through an explicit `ScaffoldStack` field; two are convention-bound (`sandbox _CONTRACTS`, `BUILD_PROFILES`) — and **both convention-bound ones are the ones that broke**. That set is pinned so it can only shrink.

The guard sweep deliberately does **not** count `tests/` or `scripts/` as callers. A guard only CI calls does not protect a cycle — that is precisely how `lint()` looked healthy.

A membership matrix alone passes today, since both stacks are complete on all six. That's exactly why it isn't the only check: a test that passes the moment it's written has described the system rather than closed it.

## What it found immediately

**The frozen-surface index describes Python only — and this is not a stack-#2 gap.**

Stack #1 renders **10 of its 15** frozen entries bare, including the whole of `frontend/` (`App.jsx`, `api.js`, `main.jsx`). pf-42's index has only ever described the Python half, on both stacks. Nobody noticed because the checks that would trip on a frontend guess are bundler-level rather than declaration-level.

**Stack #2 was worse: zero declarations for any frozen file.** So the plan appendix's rule — *"do not check a frozen file unless the line proves it passes"* — had **no line anywhere**, and no frozen file could ever be legitimately checked. That vacuum is what #849's live rejection was an author guessing into.

This reframes #850's finding C, which I'd scoped as a stack-#2 disclosure. It's a language gap both stacks have always had.

## The fix

`_typescript_surface`, mirroring `_python_surface` **including field names**:

```
- `lib/models.ts` — defines Participant(id, name), RunEvent(id, title, datetime, location, distance, ...)
- `lib/store.ts` — functions reset, all, insert, find, nextId
- `lib/errors.ts` — exports ERROR_STATUS; functions errorResponse
```

Field names matter specifically: pf-42 died on `meeting_location` for the declared `location`, and pf-43 on omitted module state. **A name the author cannot see is a name it will invent** — listing `RunEvent` without its fields is that same defect one level up, which is why there's an explicit test for it beyond the interface names.

Regex rather than a parser, and safe **here specifically** because these files come from the stack's own templates so the shapes are fixed. The docstring says it is not a general TypeScript reader and must not be reused as one — teaching the typed *checks* to parse TS is S4's work, over code the squad wrote rather than code we emit.

`DESCRIBED_FROZEN_SUFFIXES` now declares the coverage, with `.json` / `.html` / `.mjs` / `.js` recorded as still undescribed on both stacks.

## Two bugs in the test itself, found on first run

Worth recording because they're the failure modes this kind of test invites:

1. The convention-bound detector **inferred** "addressed by its own name" from the resolved value — wrong, because stack #1's `criteria_pack` *is* `"fullstack_fastapi_react"`. Now the field name is declared, not inferred. That same confusion is plausibly how two registries stayed convention-bound without anyone deciding they should be.
2. Asserting that every described-suffix file renders declarations is **false**: `backend/__init__.py` is empty and correctly declares nothing. The assertion is one-directional now (an *undescribed* suffix rendering declarations means the tuple lies), with a separate check that the reader produces something for at least one file — otherwise "everything bare" would satisfy the rule while meaning the parser silently returns nothing.

## Verification

- Regression **6966 passed**, 1 skipped.
- Stack #1 contract and manifest pins **unmoved** — M0a's byte-equality test green.
- The guard-wiring sweep is clean, which makes it a regression guard for #849's class rather than a discovery today.

## What this is for

Every defect the three VS rolls surfaced was statically decidable; none needed a cycle to discover. This is the instrument that finds them in a batch, offline, instead of one per 54–75 minute framing run — and its output feeds 2b's schema directly.

Refs #822 (Stage 2a), #838, #846, #707, #849


jasonLadd (commented):
reviewed

jasonLadd (commented):
reviewed

PR #852, opened 2026-08-10
docs(2b): blueprint schema drafted against both stacks

jasonLadd:
Stage 2b. Extracted **mechanically** — reflecting over the six per-stack declarations and comparing values across both stacks, not reading the source. That discipline is the point: 1a was the read-and-classify version and overclaimed three times.

## The headline

The plan's 2c section records two single-stack fields and calls 2c *"the step most likely to bite"* — having measured **one of six declarations**.

| | count |
|---|---|
| fields total | **47** |
| differ across both stacks | 29 |
| identical on both | **15** |
| populated by one stack only | 3 |

2c's subject is **18 fields, not two**.

## Three findings that shape the schema

**`ExecutionProfile` is 6/8 identical.** Only `boot_argv` and `prepare_argv` vary — it is a defaults object with two stack-specific fields, not a per-stack object. A blueprint carrying all of it asserts variability two stacks don't support: the *"FastAPI contract with generic field names"* the SIP declines acceptance over, reached from the opposite direction.

**10 of the 29 varying fields are `DevelopmentCapability`, and 8 of those are prompt prose.** The largest genuine difference between two stacks is *what the agents are told* — not paths, not commands. A schema modelling only structure is silent about the majority of what actually differs.

**`ExecutionProfile.prepare_argv` is the only field where stack #2 is the populated side** (`next build` must precede `next start`). Every other single-stack field is stack #1's, so this is the one piece of evidence that a field can be introduced by the *second* stack rather than inherited from the first.

## Verified, not assumed

`EnvironmentContract.image` is identical across both stacks — by deliberate #822 decision, since the sandbox already carries Node 20/npm 10 and no new runtime enters the pipeline. **Unfalsified, not decorative.** That distinction is what 2c has to make fifteen times, and getting it wrong in either direction is how a schema either bloats or loses a real field.

## Proposed shape

Three tiers rather than a flat field list — declared data (schema-governed, falsifiable), pack-supplied callables (`expand`/`fill_slots`/`slot_criteria`, since generation as data is a template language), core-owned everything else. Plus the rule from #818 generalized: every declaration is required-or-refuse or optional-with-a-visible-consequence, **never silently defaulted** — every defect this release found was a silent default.

## Open questions carried to 2g

Two now have Next.js evidence: **composition** (Next collapses the api/frontend split entirely — evidence against composition as the primary shape) and **`expand` as callable vs data** (its file placement is a computation over the manifest, not a table).

Docs only, no code change. Regenerate the matrix with 2a's extractor, which enumerates from the registries so a third stack enters automatically.

Refs #822 (Stage 2b)

### DIFF
--- a/docs/plans/stack-blueprint-schema-draft.md
+++ b/docs/plans/stack-blueprint-schema-draft.md
@@ -1,0 +1,102 @@
+# Stack Blueprint schema — draft against two stacks (Stage 2b)
+
+**What this is.** The 1.6 plan's step 2b: *"draft the blueprint schema against **both** stacks | gate: every field traced to a real declaration in each."* It is the input to 2c (falsification), 2e (disclosures) and 2g (promoting the Stack Blueprint SIP).
+
+**Method: mechanical, not read-and-classify.** Every field below was extracted by reflecting over the six per-stack declarations and comparing values across `fullstack_fastapi_react` and `nextjs_ts`. Nothing here is a reading of the source. That discipline is deliberate — 1a was the read-and-classify version of a sweep and overclaimed three times, and the classifications below are exactly the kind that look obvious and are not.
+
+Regenerate with the extractor in `tests/unit/capabilities/test_stack_inventory.py` (2a), which enumerates from the registries so a third stack enters automatically.
+
+---
+
+## The headline: the falsification surface is nine times larger than the plan estimated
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #853, opened 2026-08-10
docs(2b): the blueprint schema itself, not just the field inventory

jasonLadd:
**#852 was mislabelled.** It was titled *"blueprint schema drafted against both stacks"* and contained no schema — the traceability matrix, the 29/15/3 split, and a proposed shape in prose. The filename said `-draft` and there was no draft in it. This writes the thing, into the same file, which makes the name accurate and the rename I proposed unnecessary.

## The schema

**Tier 1 — declared data** (12 fields, blueprint-owned, falsifiable at 2c). Each row carries its value in *both* stacks, which is 2b's actual gate: `id`, `qa_test_namespace`, `source_extensions`, `authored_extensions`, `test_file_patterns`, `test_framework`, `required_tools`, `operation_commands`, `boot_argv`, `packaging.optional_files`, `packaging.validation_rules`, `criteria_families`.

**Tier 2 — pack-supplied callables.** `expand`, `fill_slots`, `slot_criteria`, `build_criteria`, `suite_criteria`. Callables rather than data because Next.js derives file *location* from route path (`POST /runs/{run_id}/join` → `app/api/runs/[run_id]/join/route.ts`) — a computation over the manifest, not a table. Expressing it as data makes a template language: code with worse tooling and no type checking.

**Tier 3 — core-owned.** The 15 fields identical on both stacks, hoisted out of the per-stack registries. Promoting them into the blueprint would assert variability two stacks don't support — the *"FastAPI contract with generic field names"* failure reached from the opposite direction.

**Optional capabilities** are declared with what an unset value *does*, not defaulted: `harness_entry_modules` and `check_stack` (S1 only), and `prepare_argv` — the only field in the entire inventory introduced by the *second* stack.

## A spec, not a dataclass — reversing what I proposed a turn earlier

An unwired dataclass is a declaration nothing reads. That's the exact class #818 **deleted** rather than wired (`CheckSpec.supported_stacks`: 11 declarations, 0 readers) precisely because wiring it would have minted the blueprint vocabulary by accident. Promotion is 2g's job.

2c can still falsify a spec mechanically — delete a field, check whether either stack's real declaration becomes inexpressible. That is a stronger test than corrupting a dataclass nothing consumes.

## Two dispositions recorded rather than glossed

**Prompt prose has no home in the schema, and it's the largest varying surface** — 8 of the 29 demonstrated-varying fields are LLM instruction text. The blueprint declares *which asset* carries it, not the text, because prompt content belongs in fragments via PromptService (#448). The consequence is stated rather than hidden: a blueprint alone is then not sufficient to define a stack, and a pack ships prose alongside it. A boundary, not a gap.

**`ready_path` is one fact in two declarations.** Both stacks say `/health` because both scaffolds *emit* a `/health` route and the probe profile restates it. That's the drift shape this release has paid for three times; the honest fix is for the probe to read the scaffold's convention rather than re-declare it.

## Verification

Every Tier-1 value was checked against the live declarations by script before commit — twelve claims, all confirmed. The last three things I wrote described the intended state rather than the achieved one, so the values here are verified rather than transcribed.

Docs only, no code change.

Refs #822 (Stage 2b)

### DIFF
--- a/docs/plans/stack-blueprint-schema-draft.md
+++ b/docs/plans/stack-blueprint-schema-draft.md
@@ -1,102 +1,112 @@
 # Stack Blueprint schema — draft against two stacks (Stage 2b)
 
-**What this is.** The 1.6 plan's step 2b: *"draft the blueprint schema against **both** stacks | gate: every field traced to a real declaration in each."* It is the input to 2c (falsification), 2e (disclosures) and 2g (promoting the Stack Blueprint SIP).
+**What this is.** The 1.6 plan's step 2b: *"draft the blueprint schema against **both** stacks | gate: every field traced to a real declaration in each."* Input to 2c (falsification), 2e (disclosures) and 2g (promoting the Stack Blueprint SIP).
 
-**Method: mechanical, not read-and-classify.** Every field below was extracted by reflecting over the six per-stack declarations and comparing values across `fullstack_fastapi_react` and `nextjs_ts`. Nothing here is a reading of the source. That discipline is deliberate — 1a was the read-and-classify version of a sweep and overclaimed three times, and the classifications below are exactly the kind that look obvious and are not.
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #855, opened 2026-08-10
fix(gates): an approved gate promotes, however it was approved (#854)

jasonLadd:
Closes #854. Roll 4's framing passed the gate — first time in four rolls — and implementation refused to start reporting the plan absent. **The plan was not absent.** It was authored, validated, stored, and never handed forward.

## The chain, measured on `cyc_c82a401a21f8`

| step | what happened |
|---|---|
| gate approved by `system:no_open_questions` | M4's question gate (#807) approves **inside the executor** |
| `_promote_run_artifacts` | one call site — the HTTP gate route — which that path never reaches |
| framing artifacts | all eleven still `promotion_status="working"` on disk |
| `_build_next_workload_overrides` | reads **promoted only**, got an empty list, never wrote `plan_artifact_refs` |
| `_load_plan_for_run` | reads that key exclusively → `None` |
| #424 | refused to build with its instrumentation contract silently absent |

**Every link behaved correctly. Only the first never fired.** #424's refusal is the system working — it's why this surfaced as a clean failure instead of a build running with no verification contract.

## Why it took four rolls to see

`system:no_open_questions` has **one row in the entire database**, recorded today, against 37 `system:plan_validation` rejections. Every gate before this was decided over HTTP — a human running `squadops runs gate --approve`, or an agent answering. V5's question gate fired *with* questions, so it took the HTTP path too. M4 shipped the auto-approval on 2026-08-09 without the promotion, and roll 4 is the first cycle to take it.

## Fix

Promotion becomes a consequence of *the gate being approved*, not of *how it was approved* — extracted to `cycles/gate_promotion.py`, called by both the route and the executor.

Failures stay logged-not-raised. The decision is already recorded by the time promotion runs, so letting a transient vault error propagate would trade a loud downstream refusal for a lost approval — the wrong direction, given the loud refusal is exactly what made this findable.

## Test shape

Asserts **on the vault**, not on a call count. "Did we promote" is the question the failure turned on, and a mock assertion would pass against a call that promoted nothing. Plus a polarity guard that a vault failure doesn't lose the approval.

## The pattern, fourth time in this line

- `governance.merge_plan` absent from `CONTEXT_CONTRACTS` → default authoring path never got the contract (#846)
- two command allowlists each gating a question the other also answered (#707)
- `VerificationContract.lint()` with no production caller (#849)
- promotion reachable only through the HTTP route (this)

None was a missing feature. Each was correct code whose wiring was never checked against every path through it. 2a's guard-wiring sweep catches the *call-site-count* version of this; it would not have caught this one, because promotion had a caller — just not on both paths.

Regression **6969 passed**. Fix verified by mutation.

Refs #807, #424

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -1954,6 +1954,16 @@ class DispatchedFlowExecutor(FlowExecutionPort):
             ),
         )
         await self._cycle_registry.record_gate_decision(run_id, decision)
+        # #854: an approval promotes the run's artifacts, and that is a consequence of the
+        # gate being APPROVED rather than of how it was approved. It lived only in the HTTP
+        # gate route, so this path — the one M4 added — recorded an approval and forwarded
+        # nothing. VS roll 4 is the first cycle to take it: framing passed, the plan
+        # validated, the gate approved, and the implementation run then refused because
+        # `plan_artifact_refs` was never written (#424 correctly declining to build with
+        # its instrumentation contract absent).
+        from squadops.cycles.gate_promotion import promote_run_artifacts
+
+        await promote_run_artifacts(self._artifact_vault, run_id)
         self._cycle_event_bus.emit(
             EventType.GATE_DECIDED,
             entity_type="run",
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #857, opened 2026-08-11
fix(plan): the sole author is shown the rules its plan is validated against (#856)

jasonLadd:
Closes #856.

Roll 5 was rejected at the framing gate on **one** defect: a `qa.test` task declaring only `__tests__/qa_execution_report.md`. The intent was legitimate — run an earlier task's suite, emit a coverage report — and the correct form is `expected_artifacts: []`. The rule that says so, `qa-tests-must-be-discoverable`, states it in its own second sentence.

**The author never saw it.** `_authoring_rules_section` renders that appendix and had exactly two callers — `planning/propose.py` and `planning/brief.py`. `_plan_authoring_service.produce_plan` rendered it nowhere, and that is the path that writes the plan on every CRP except `validation-multirole`.

Verbatim the failure the asset was written for, quoted from its own docstring:

> shk-1's first framing authored the #673 dual-claim shape with the contract, the manifest and the typed-acceptance vocabulary all present, because the rules themselves appeared in no prompt: **the validator knew, the author never did.**

## My miss at #848

#846's Family B found this same path missing the *contract* surfaces, and I fixed those. I went looking at `bind_criteria_index` and never asked what **else** the proposers receive that this path doesn't. So I closed #846 declaring B resolved with half the forwarding gap still open.

## Fifth time, same shape

| | |
|---|---|
| #846 | `governance.merge_plan` absent from `CONTEXT_CONTRACTS` |
| #707 | two command allowlists, each gating what the other also gated |
| #849 | `VerificationContract.lint()` with no production caller |
| #854 | promotion reachable only through the HTTP gate route |
| **this** | authoring rules rendered by two of three authoring paths |

And a limit worth stating plainly: **2a's guard-wiring sweep catches neither #854 nor this one.** Both had production callers — just not on every path. The sweep asks *"does this have a caller"*; the recurring defect is *"does this have all its callers."* That's a narrower check than I claimed when I built it.

## Shape of the fix

`authoring_rules_section` extracted to `handlers/_plan_authoring.py` beside `contract_surface_sections`, called by both the base handler and `produce_plan` — rather than adding a second copy, which is how this class starts. Unconditional in both, since the rules hold for every plan on every roll. Ordered **before** the cycle-specific contract surfaces (general frame, then specifics), pinned by test so a later append can't invert it.

## Tests

Four existing tests updated, **two of them mine from #848**. They asserted the sole-author prompt equals the base string exactly, which an unconditional section correctly breaks. Rewritten to assert what they actually guard — that no *contract* appendix appears when there is no contract — rather than loosened to keep passing.

Regression **6971 passed**. Fix verified by mutation.

Refs #686, #846, #629

### DIFF
--- a/src/squadops/capabilities/handlers/_plan_authoring.py
+++ b/src/squadops/capabilities/handlers/_plan_authoring.py
@@ -31,6 +31,34 @@ from squadops.llm.models import ChatMessage
 logger = logging.getLogger(__name__)
 
 
+async def authoring_rules_section(renderer: Any) -> str:
+    """The plan-shape rules every deterministic validator enforces (#686), or ``""``.
+
+    Unconditional — unlike the contract surfaces below there is no input to key on, because
+    these rules hold for every plan on every roll.
+
+    **Shared because it had three authors and reached two (#856.)** The proposers render it
+    (`planning/propose.py`) and the brief author renders it (`planning/brief.py`); the
+    sole-author path in `_plan_authoring_service.produce_plan` did not, and that is the path
+    that authors the plan on every CRP except ``validation-multirole``.
+
+    VS roll 5 (`cyc_d430f25fd01d`) is the measurement: a `qa.test` task declared
+    `expected_artifacts: ['__tests__/qa_execution_report.md']` — a report, no test file — and
+    `validate_check_applicability` rejected it. The rule that covers it,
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #860, opened 2026-08-11
fix(nextjs): the route file is derived from the URL it serves (#859)

jasonLadd:
Closes #859. Roll 6 reached implementation on stack #2 for the first time and burned its full 7200s budget across 16 tasks and three correction rounds, on a defect **the contract itself created**.

## The defect

The manifest declared `/api/runs` — the URL Next actually serves, what an App Router author writes, what the frontend fetches. `_route_groups` prefixed `app/api/` unconditionally:

| | |
|---|---|
| derived fill slot | `app/api/api/runs/route.ts` |
| that file serves | `/api/api/runs` |
| contract probe requests | `/api/runs` |

The app could never answer its own verification. The dev emitted exactly what it was told, three times. Every gate passed it — M3's proofs, the #849 lint, plan validation.

## The fix, in two parts

**1.** `path` means the same thing on every stack — the URL served — and the file derives from it. `/api/runs` → `app/api/runs/route.ts`.

**2.** `_assert_routing_tree_is_coherent`. Dropping the prefix puts handlers and pages in one `app/` tree, where Next refuses a directory holding both `route.ts` and `page.tsx`, and refuses two slug names at the same dynamic position. Without this, part 1 trades a *silent double-prefix* for a *silent collision* — the same defect class facing the other way.

Raising rather than relocating is deliberate: silently moving a file would ship an app answering different URLs than its own contract, which is the failure being fixed. M3's `PROOF_EXPANDS` turns the raise into a framing rejection costing a free re-roll rather than an implementation run.

## The reference manifest fails the new rule

That's the sharpest evidence for it. Its API is `/runs`, its page is `/runs/:id` — under one tree, two slug names at the same position. So declaring the API under a distinct prefix is a **requirement of stack #2**, not decoration.

**17 stack-#2 tests failed on exactly this**, every one of them building its manifest by re-stacking the reference. They now route through a single shared fixture that expresses the reference *per stack*, so a third stack adds its clause in one place rather than each test growing a private workaround.

## Recorded, not acted on

*"Do API and page routes share a routing tree?"* — stack #1 no, stack #2 yes. Demonstrated with opposite values on two stacks, which is exactly what S5's admission rule asks of a candidate blueprint field. This fix expresses it as **pack behavior**, not declared data. Defensible (the rule is Next's and needs the stack's own knowledge), but a schema reader cannot tell that stack #2 constrains its author's URL space and stack #1 doesn't. **2c owes it a disposition** — noted in the 2b draft rather than decided here.

## Verification

- Regression **6975 passed**, 1 skipped.
- Stack #1's contract and manifest pins **unmoved** — M0a byte-equality green. Stack #2's hash moves, which is the point.
- **Both halves mutation-verified**: restoring the prefix fails 1 test; removing the collision guard fails 2.
- Bend register **entry 7** added in this PR per its written-during rule — the first bend found by *running* the stack rather than building it.

Refs #822 (Stage 1e)

### DIFF
--- a/docs/plans/stack-2-bend-register.md
+++ b/docs/plans/stack-2-bend-register.md
@@ -28,8 +28,11 @@ here" and "the stack conceded here" demand opposite responses.**
 | 4 | route handlers, not server actions | **bend** | the stack conceded |
 | 5 | seven whole-tree builds per acceptance pass | disclosure | a cost, not a concession |
 | 6 | no static check on route slots | disclosure | a coverage gap, already known |
+| 7 | the API prefix is the manifest's to declare | **bend** | the stack conceded |
 
-**One true bend.** That is the number S3 has to argue, and #4 is argued below.
+**Two true bends.** #4 was the only one until VS roll 6 produced #7 — which is the first bend found
+by *running* the stack rather than by building it, and the only one whose evidence is a manifest
+that cannot be built at all.
 
 ---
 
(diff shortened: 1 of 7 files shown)


jasonLadd (commented):
reviewed

PR #862, opened 2026-08-11
fix(dev): the developer is told what the frozen files declare (#861, #858)

jasonLadd:
Closes #861 and #858. Both were measured by VS roll 7 (`cyc_0e301961f099`), and both are the same class: **supply the authoritative fact to the agent that needs it.**

## #861 — the index reached the plan author, not the developer

Roll 7 cleared framing with a clean plan (7/7 fill slots, 7 distinct `criteria_refs`) and failed implementation after 38 minutes as `plan_defect`:

| round | emitted | outcome |
|---|---|---|
| 0 | `import … from '../../lib/store'` | unresolvable relative path |
| 1 | `import { runStore } from '@/lib/store'` | **`runStore` is not exported** — same signature, A4 terminated |

`lib/store.ts` exports `reset, all, insert, find, nextId`.

`development.develop` declared `manifest_surfaces=(error_contract, model_surface, testid_surface)` and no frozen surface. That index rides on `bind_criteria_index`, which it does not declare; its `artifact_filter` threads planning documents and its own prior output, not the scaffold's files; `acceptance_workspace` is explicitly separate from prompt context (#643).

So the line naming those exports reached the plan author and stopped. The repair path was blind identically, which is why round 1 reproduced round 0 instead of converging — **the correction terminated on a defect no repair could see.**

`SURFACE_FROZEN` now rides the same transport #588/pf-45/#659 used to give the developer the error-contract convention, the model field surface and the DOM testid inventory. Each of those exists for this exact reason.

#851 was the prerequisite and is already merged: before it, this index rendered `lib/store.ts` as a bare name with no exports.

## #858 — two of four authoring attempts lost to unquoted `list[X]`

An unquoted `[` inside a `{ … }` flow mapping opens a flow sequence and the document fails to parse before any gate reads it. The example taught flow style with three scalar fields and **no collection type**, so the first list an author needed hit the trap — while the reference manifest, kept out of squad inputs by SIP-0103 §4, has carried the quoted form all along.

The example now shows `type: "list[Tag]"`, and the rule is stated in prose because a trailing comment is dropped when an author copies the block.

## Two weaknesses in my own tests, caught by mutation

- The #858 assertion checked the whole file, and `type: "list[Tag]"` also appears in the prose I added — so **deleting the example still passed**. Now scoped to the fenced block, and it fails under mutation.
- #861's assertion is on the **derived fragment**, not the registry flag. A flag can be set while the deriver returns nothing — precisely what `.py`-only rendering did to this stack before #851.

## Recorded, not fixed here

Threading this index to the developer also hands it **#787**'s defect: the index describes each file's *own* imports, which models the relative form — and round 0's failure was exactly a relative import. On stack #2 that's harmless and helpful (the scaffold uses `@/`, so the index shows the correct reachable form). On stack #1 it would teach `./views/CreateRunView.jsx`. **This raises #787 from latent to live for stack #1**, noted on that issue.

## Verification

- Regression **6984 passed**, 1 skipped.
- Both fixes **mutation-verified** (the #858 one only after its assertion was fixed).
- Context-assembly golden regenerated in the same commit; the delta is exactly the `frozen_surface` key on the two develop scenarios.

Refs #787, #822 (Stage 1e), #851

### DIFF
--- a/src/squadops/capabilities/context_assembly.py
+++ b/src/squadops/capabilities/context_assembly.py
@@ -45,6 +45,12 @@ SURFACE_TESTID = "testid_surface"
 #: The qa-keyed landing of the SAME testid inventory (#667): identical
 #: deriver, distinct envelope key — each handler reads its own variant.
 SURFACE_DOM_TESTID = "dom_testid_surface"
+#: #861: what each scaffold-frozen file declares. The plan author has had this since pf-42;
+#: the developer that must IMPORT from those files had not, and roll 7 died of it — it wrote
+#: `import { runStore } from '@/lib/store'` against a module exporting
+#: `reset, all, insert, find, nextId`, then repaired to the same invented name because the
+#: repair path was blind identically. A name the author cannot see is a name it will invent.
+SURFACE_FROZEN = "frozen_surface"
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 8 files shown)


jasonLadd (commented):
reviewed

PR #865, opened 2026-08-11
fix(scaffold): the frozen index publishes call signatures, not just names (#863)

jasonLadd:
Closes #863.

Roll 8 imported the right symbol from the right module and wrote `all()` against `all(table)`. tsc rejected it, `frontend_compiles` failed, the correction terminated.

| | roll 7 | roll 8 |
|---|---|---|
| import path | `../../lib/store` → `@/lib/store` | correct first try |
| symbol | **invented `runStore`** | **correct `all`** |
| failure | symbol does not exist | wrong arity |

#862 threaded the index to the developer and roll 8 shows that working. What remained is that the index publishes **names without shapes**:

```
index:  functions reset, all, insert, find, nextId
truth:  reset()   all(table)   insert(table, row)   find(table, id)   nextId()
```

## The asymmetry is mine, in both readers

`_ecmascript_surface` (#851) renders `defines Participant(id, name)` — fields included — and `functions reset, all, …` — parameters omitted. `_python_surface` has had the same shape since pf-42: `_class_field_names` gives `User(id, email)`, `functions` is a bare name list.

pf-42's lesson was applied to types and not to functions, in both languages. Its own docstring states the principle the omission breaks — *a name the author cannot see is a name it will invent*. Roll 8 is that sentence one level down.

Parameter **names**, not annotations: the index is prose for an author, and `insert(table, row)` carries arity and meaning where the full type costs readability for something the caller never restates. Optional markers survive (`api(path, init?)`) because they change how the function may be called.

## Three things this PR caught in itself

- The first TS pattern required the argument list to follow the name directly, so `export async function api<T>(…)` stopped matching and **`lib/api.ts` went from described to bare** — a file made worse by a fix meant to improve it. Now generic-aware, pinned by a test naming that declaration.
- `_ts_param_names` splits on **top-level** commas only. Naive splitting would read `Record<string, unknown>`'s comma as a separator and publish a third parameter that doesn't exist — teaching an author to pass three arguments to a two-argument function, which is the same defect this fix removes.
- My own new test split the rendered list on `", "`, cutting `insert(table, row)` in half — **it reported the code broken when the test was.** It now strips whole `name(params)` matches and requires only separators to remain.

## Verification

- Regression **6989 passed**, 1 skipped.
- Both readers **mutation-verified** (reverting Python params fails 1; dropping the generic clause fails 1).
- Goldens regenerated in the same commit; the delta is signatures only — `functions health()`, `register_error_handlers(app)`, `client()`.
- #851's assertion updated, which expected bare `functions reset, load`.

## Filed alongside, not fixed here

**#864** — the repair path diagnosed both roll 7's and roll 8's failures accurately (*"invokes `all()` with zero arguments, but the type signature requires exactly one"*), chose `patch`, and re-emitted the same call. This fix will likely stop those repairs from being needed, which **masks that defect rather than resolving it**.

Refs #851, #862, #822 (Stage 1e)


jasonLadd (commented):
reviewed

PR #866, opened 2026-08-11
fix(context): every judged agent is shown the facts its checks depend on

jasonLadd:
## The defect class, closed as a surface

Seven closed defects this arc share one shape: **an agent is judged against a fact it cannot see** (#846, #707, #849, #854, #856, #861, #863). Each was fixed one cell at a time, each cost a roll to find. This PR closes every remaining cell of the matrix at once and adds the instrument that makes the next omission a test failure instead of a two-hour roll.

## The cells, each with its roll evidence

| agent | judged by | was blind to | evidence |
|---|---|---|---|
| `qa.test` | `tests_pass` | the tree it imports from + the installed package set | roll 9 (`cyc_a92eaa4f4052`): `import request from 'supertest'` — not in package.json; V2 (`cyc_dd3855f353c0`): `from .store import reset` — reachable only as `backend.store` (#787) |
| `builder.assemble` | plan-authored `regex_match` checks | its own task description and every authored check | roll 9: plan authored 5 regex checks on `qa_handoff.md`, 2 on sections the profile calls optional; the prompt carried none — `system_prompt_for_files` documents the same class from 2026-05-03, closed one-sidedly (validator's list surfaced, plan's checks never) |
| repair envelopes | the same criteria as the initial author | the frozen surface | roll 7 (`cyc_0e301961f099`): #861 fixed the initial author; the repair "invented the same name again because it was blind identically" → plan_defect |
| every reader of the frozen index | — | the reachable module path; the dependency surface | the index modeled each file's *own* imports (relative form) as its only import guidance (#787), and rendered `package.json` bare — my #851 call that `.json` "declares nothing", falsified by roll 9 |

**One matrix cell was wrong when re-measured**: `development.develop` *does* see the criteria it is judged by — `criteria_refs` resolve into `acceptance_criteria` at plan time (`task_plan.py`) and the handler renders them as the Contract Expectations block. Pinned by test instead of "fixed".

## What changed

- **`scaffold.py`** — Python lines open with `` import as `backend.store` `` (path-derived; `conftest.py` excluded, `__init__.py` collapses to the package); both readers relabel interior imports as `its own imports`; `package.json` renders its `dependencies`/`devDependencies` names (`DESCRIBED_FROZEN_BASENAMES`, names not versions).
[...]

PR #868, opened 2026-08-12
fix(stack2): the harness can execute the idiom the stack teaches

jasonLadd:
## Roll 10's two blockers (cyc_43a216d43e1e), both #818's class in the stack's own wiring

An incorrect contract that is believed — this time in the harness: the stack teaches the `@/` import form and requires `npm ci`-style preparation, and neither could ever succeed on the tree it emits.

### 1. vitest cannot resolve the alias the stack teaches
tsconfig `paths` covers `tsc`/`next build`; vitest resolves through vite, which does not read tsconfig paths. The emitted `vitest.config.ts` declared no `resolve.alias`, so **every** suite importing app code the taught way failed collection — including the scaffold's own frozen harness test. `tests_pass` was unwinnable by construction; no gate executes vitest, so nine rolls never saw it (roll 9's phantom-package import failed resolution first and masked it).

**Reproduced before fixing** (exact scaffold shapes, vitest 1.6.0): `Failed to load url @/lib/store`. Re-run with the alias: 1 passed.

### 2. Probe preparation used the lockfile-only installer against a lockfile-less tree
The expansion emits no `package-lock.json` (offline-deterministic — it cannot), and `npm ci` EUSAGE-refuses without one. Every probe reported "subject preparation failed" with npm's usage text; 4 of 11 contract criteria unverified. `npm install` matches the frontend build check and the vitest runner, so preparation now cannot succeed or fail differently from the checks beside it.

### Pins
- `test_vitest_resolves_the_alias_the_stack_teaches` — tsconfig, vitest config, and the harness test's import form must agree, asserted from the **expanded** files.
- `test_the_probe_prepare_installs_without_a_lockfile` — a lockfile-less expansion forbids `npm ci` in `prepare_argv`.

Both mutation-verified. Regression **7014 passed / 0 failed**.

Roll 10 context: develop recovered from its one type-error correction, builder passed all five plan-authored checks first try (#866's surfaces working), qa was the only failing surface and both causes are these. Full post-mortem in the morning report.

### DIFF
--- a/src/squadops/capabilities/handlers/probe_runner.py
+++ b/src/squadops/capabilities/handlers/probe_runner.py
@@ -114,7 +114,14 @@ _PROFILES: dict[str, ExecutionProfile] = {
         # so a prepare step that stopped at install would boot into "no production build
         # found" and report as a boot failure — the exact conflation #827 separated. `sh -c`
         # for the two-step, matching the sandbox contract's INSTALL_DEPENDENCIES precedent.
-        prepare_argv=("sh", "-c", "npm ci --no-audit --no-fund && npx next build"),
+        #
+        # `npm install`, NOT `npm ci` (roll 10, cyc_43a216d43e1e): the scaffold emits no
+        # `package-lock.json` — an offline-deterministic expansion cannot produce one — and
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #869, opened 2026-08-12
fix(crp): validated-fullstack budget 7200 -> 10800s

jasonLadd:
One line + the reason in place. Roll 11 executed 33/33 checks green and died of time mid-qa with an 8-develop-task authored design — the budget was the binding constraint, and design size is a per-roll lottery (5 tasks roll 10, 8 tasks roll 11, same PRD). Roll 12 validated 10800s: full correction arc to a natural terminal state at 2h05m. V7's window pre-registers its own budget, so this changes roll configuration, not measurement methodology. Also removes the need for per-launch typed API-body overrides (`--set` is string-only and the executor compares numerically).


jasonLadd (commented):
reviewed

PR #872, opened 2026-08-12
fix(scaffold): TS frozen surface publishes types, returns, and classes (#871)

jasonLadd:
The ECMAScript surface reader concealed exactly the two facts roll 12's repair invented against (#871, found by the #864 investigation):

- **`export class` had no clause at all** — frozen `lib/errors.ts` rendered as `exports ERROR_STATUS; functions errorResponse(err)`, so `ApiError`, the one seam the error contract expects every route to use, was never named. Roll 12's original defect is the invention this predicts: the route passed a bare string to `errorResponse()` (TypeScript-legal, the param is `unknown`), `instanceof ApiError` failed at runtime, and every error path returned 500 where the contract pins 422/409.
- **The #863 names-only ruling hid the types**: `find(table, id)` concealed `id: string` and `nextId()` concealed its `string` return, so the repair wrote `find('runs', Number(run_id))` / `nextId('runs')` and the tree stopped compiling.

## What changed

- Functions render full param text + return type; each signature is backtick-wrapped so commas inside `Record<string, unknown>` never read as list separators.
- New `classes` clause: class name + constructor call form (parameter-property modifiers stripped). It surfaced **stack #1's own invisible `ApiError`** in `frontend/src/api.js` the moment it landed — the same gap existed there too.
- Interface fields now carry their types (`Run(id: string, ...)` — a fourth independent contradiction of roll 12's numeric-id invention).
- `_python_surface` deliberately unchanged: names-only stands until a Python roll fails on a concealed type (its frozen sources annotate optionally, so rendered types would be intermittent anyway). The divergence is recorded on both readers.

## Rendered result for roll 12's frozen files

```
classes `ApiError(code: string, detail: string = '')`; exports ERROR_STATUS; functions `errorResponse(err: unknown): Response`
functions `reset(): void`, `all(table: string): ...`, `find(table: string, id: string): Record<string, unknown> | undefined`, `nextId(): string`
```

## Verification

- 60 surface/inventory/completeness tests pass; full `tests/unit/capabilities/` 1656 passed; ruff clean.
- Four mutations each killed by ≥1 test: drop class clause; drop return types; regress params to names-only; remove the next-export bound on constructor search (which would attribute a constructor to the wrong class).

Closes #871


jasonLadd (commented):
reviewed

PR #873, opened 2026-08-12
fix(correction): file-owned patch gate + rejected-repair evidence carry (#870)

jasonLadd:
Both halves come from the #864 investigation of roll 12 (`cyc_e4b2444fa300`). **One claim in the issue as filed was wrong and is corrected below.**

## 1. File-owned patch gate

Patch verification judged a repair only against the *failed task's* typed criteria. A qa.test failure has none, so roll 12's dev repair — which rewrote four routes that no longer compiled — was judged against nothing structural and sailed into the behavioral retest.

`verify_patched_artifacts` now accepts `file_owned_criteria`: the contract criteria owned by the files the repair actually rewrote, selected by the new `resolve_criteria_for_files` (criterion index → `resolve_contract_refs`, the same seams bind-mode dispatch uses). An **executed blocking failure rejects the patch before the retest**, with named checks.

**The gate is monotone by design**: evaluator errors and skips are recorded and change nothing. This matters because the executor may lack the stack's toolchain — runtime-api has no node, so stack #2's compile checks cannot execute there today, and a non-monotone gate would have severed the retest path (the pf-47/pf-49 deadlock) for exactly the repairs that need it. On stack #1 the gate bites today (verified: `import_present` rejects a garbage `backend/routes.py` through the full executor wiring); on stack #2 it lights up the day the executor gains node. The behavioral retest remains the compile gate in the qa environment — it is what correctly caught roll 12's repair, in 25 seconds.

## 2. Rejected-repair evidence carry

A rejected repair left no trace but `patch_retest status=FAILED passed=False`. The next correction round re-analyzed the task blind to WHY the previous repair was rejected. Now every rejection path (patch verification's named failed checks; the retest's verdict text + failing rows) appends to a run-lived carry (`_record_repair_rejection`, bounded 3×500 chars), which `_inject_deterministic_evidence` renders as a **PRIOR REPAIR REJECTED** authoritative block — the exact transport `scaffold_enforcement` already uses.

Drive-by fix in the same evidence surface: the pf-33 failed-checks log comprehension matched dict rows against what are actually `PatchCheckRecord` dataclasses, so every rejection logged `failed=-`. It now names the checks as intended.

## Correction to the issue

The issue claimed a failed repair "poisons the re-dispatched attempt via latest-wins." **That was wrong**: `_enrich_envelope` already excludes repair-candidate artifacts on fresh dispatches (pf-31 Fix E), and roll 12's attempt 2 demonstrably saw the accepted pre-repair tree (its "16 source file(s)" count matches exactly). No rollback mechanism is needed; none is built here. Attempt 2 failed honestly against the still-defective accepted tree.

## Verification

- 2843 cycles+repair-handler tests pass; regression 7043 passed (+31 new) with the same 2 pre-existing `TestCommandExitZero` environmental failures main shows under the regression script.
- Five mutations each killed by ≥1 test: executor wiring deleted (gate derivation), monotone violated (skips treated as failures), retest rejection not recorded, evidence injection dropped, prompt block dropped.
- Live probe: reference-manifest contract resolves 4 file-owned criteria for `backend/routes.py`; garbage content → `PATCH_FAILED reason=file_owned_criteria` end-to-end through `_try_accept_patch`.

Closes #870. Refs #864, #456, #389.

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -368,6 +368,7 @@ def _inject_deterministic_evidence(
     artifact_contents: dict[str, str] | None,
     scaffold_enforcement_carry: list[str] | None,
     bound_record: Any = None,
+    repair_rejections: list[str] | None = None,
 ) -> None:
     """Deterministic authoritative-evidence injection for the correction chain.
 
@@ -380,6 +381,8 @@ def _inject_deterministic_evidence(
       cannot drift, and reporting that it does aims repairs at bytes the producer
       may not write; shk-2 lost three attempts to exactly that).
     - ``scaffold_enforcement`` (3.4b): prior attempts' frozen-emission instructions.
+    - ``prior_repair_rejections`` (#870): what happened to this task's previous
+      repair — patch-verification's named failed checks or the retest verdict.
     - ``contract_expectations`` (pf-31 Fix A): the failed task's typed criteria
       as exact expectation lines.
     - ``error_contract`` (pf-34): the ApiError raise convention + code→status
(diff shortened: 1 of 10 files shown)


jasonLadd (commented):
reviewed

Issue #875, opened 2026-08-12
TS frozen surface renders consts as bare names — ERROR_STATUS's type and values are invisible, and roll 13's qa author invented both


jasonLadd:
Third instance of the #871 class, one clause over: the TS frozen-surface reader renders `export const` as a **bare name**, so a const's shape and values are facts no author is shown — and roll 13's qa author invented both.

## The invention

Frozen `lib/errors.ts` declares:

```ts
export const ERROR_STATUS: Record<string, number> = {
  validation_error: 400,
  ...
}
```

Its surface line renders as `exports ERROR_STATUS`. Roll 13's qa suite (`cyc_732b773cf323/run_90efd238c88d`, first attempt) then:

- called `ERROR_STATUS.get('validation_error')` — assuming a `Map`; on a `Record` this is a runtime `TypeError` inside the tests (correction-00's failure analysis names it);
- asserted 422 where the map says 400 — the *values* are as invisible as the type (the primary confusion is #874 — the derived probe also says 422 — but a surface line showing `validation_error: 400` would have put the truth in front of the author regardless).

"A name the author cannot see is a name it will invent" has now run the full ladder on this one file: fields (pf-42), module state (pf-43), signatures (#863), classes and types (#871, roll 12), and now const shape/values (roll 13). Each clause got its fix only after a roll died on it.

## Fix direction

Render the declared type beside the const name, the same way #871 did for functions and classes: `exports \`ERROR_STATUS: Record<string, number>\``. That kills the Map assumption mechanically.

Whether to render the *entries* (the code→status map itself) is a judgment call: for `ERROR_STATUS` specifically the values are the load-bearing fact, but a general "render initializer contents" rule would bloat lines for large consts. A bounded form (`ERROR_STATUS: Record<string, number> = { validation_error: 400, run_not_found: 404, ... }` for object literals under some size) covers the real case without a general dump. Note the error-code→status facts also reach dev tasks via `error_seam_instructions` (pf-34) — the check worth doing first is whether qa.test's context contract carries that surface at all; if not, that is the cheaper, already-built fix for the values half.

Refs #871 (the class), roll 13 evidence: `run_90efd238c88d/art_5a47564d2e05/failure_analysis.md`.

PR #876, opened 2026-08-12
fix(contract): the rejects-blank expectation is the pack's call (#874)

jasonLadd:
Roll 13 (`cyc_732b773cf323`) authored `validation_error: { http: 400 }` — a legitimate choice — and the deriver emitted a contract that contradicts itself in one YAML: the frozen seam and coverage expectations say 400, the `rejects-blank` probe says 422 (**hardcoded**, on the stack-1 assumption that pydantic rejects blank fields before app code runs). The probe was unwinnable by construction, and the run burned a correction round on it before triage. Every prior authored roll masked this by happening to declare 422.

## The fix

pf-39's lesson applied to the rejection side (the comment block directly above the hardcode records the *success* status being un-hardcoded the same way): the expected status is a **per-stack fact on the CriteriaPack** (#818's seam):

- `fullstack_fastapi_react`: `blank_rejection_status=422` — framework-fixed; pydantic preempts the seam, the authored mapping is unenforceable for this probe (deriving here would create the mirrored defect).
- `nextjs_ts`: `blank_rejection_status=None` — derive from `error_contract.codes[validation_error].http`; an unmapped code **omits the probe** (the frozen handler resolves unmapped codes to 500, so any declared expectation loses).

## Verification

- Reference (stack-1) contract byte-identical — the M0a guard's subject unmoved; stack-2 fixture (which maps 422) also derives identically, so no hash pins move anywhere.
- New tests pin all four faces: seam-owned derives the authored 400 (the roll-13 shape), framework-fixed ignores a declared 400, unmapped omits, and **probe-vs-coverage-expectation self-consistency** (the exact disagreement roll 13 shipped — an #820-flavored assertion, live early).
- Three mutations each killed: regress to hardcoded 422; make stack #1 derive; fall back to 422 when unmapped.
- 4489 capabilities+cycles tests pass; regression 7050 passed with the same 2 pre-existing `TestCommandExitZero` environmental failures main shows in this shell.

Closes #874. Refs #820, #593, #651, #875.

### DIFF
--- a/src/squadops/capabilities/scaffold_contract.py
+++ b/src/squadops/capabilities/scaffold_contract.py
@@ -63,7 +63,8 @@ class CriteriaPack:
     Deliberately NOT covering the manifest-derived tiers. Probes, coverage expectations,
     the frozen-file hashing, and ``skeleton.expander`` are language-agnostic already, which
     is why this seam is small — and is S2's HTTP-constant containment argument paying off
-    exactly where it was meant to.
+    exactly where it was meant to. (#874 carved out one probe fact: the blank-rejection
+    status is framework-dependent, so it lives here — see ``blank_rejection_status``.)
     """
 
     name: str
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

Issue #878, opened 2026-08-12
tests_pass signature carries no intra-suite identity — #435 reads any two failing suite rounds as an exact repeat


jasonLadd:
The #435 failure signature is `(check_id, subject_file, reason_token)` per failing check row (`src/squadops/cycles/correction_signature.py:31`). For `tests_pass`, the reason token is the aggregate suite summary from `test_runner.py:71-75` — `"tests failed (exit code 1, N test file(s), M source file(s))"` — which carries **no failing-test identity**. Consequences:

- **`MOVEMENT_PROGRESS` is unreachable inside a suite.** A repair that fixes 15 of 16 failing tests still exits 1 with the same file counts → exact signature repeat. The progress-aware rule (partial reduction resets the repeat condition, `correction_signature.py:97`) can never see intra-suite progress because the suite is one row.
- Combined with candidate presence in ~89% of deltas (the module's own design note), #435 becomes a **de-facto 2-round cap for qa.test failures**: any two adjacent failing-suite rounds with candidates terminate as `plan_defect` at round 1.
- Roll 14 is the live instance: round 0 (`fetch failed`) and round 1 (`Failed to parse URL from /api/runs`) are genuinely different crash modes — the repair changed behavior — yet both collapsed to `tests_pass||failed` and terminated. The outcome happened to be right (the strategy was doomed; see the companion qa-guidance issue), but by coincidence of coarseness, not discrimination.

The design comment is explicit that "evidence text never alters the signature, so a re-described identical failure still repeats" — that property must be kept. The enrichment therefore has to be **stable identity, not prose**: e.g. the sorted failing-test identifiers (`file::test_name`) surfaced as a structured field on the tests_pass check row that `failure_signature` folds into the subject/reason. Same failures re-described still repeat; different failures (or fewer failures) stop repeating.

## Cosmetic rider: the termination message misattributes the candidate

`should_terminate_plan_defect` requires only that both rounds carry *some* non-`none` candidate (`correction_signature.py:104-127`) — they need not match. But the `_ExecutionError` text (`adapters/cycles/correction_runner.py:868-870`) renders "with structural plan-change candidate '{candidate}' on both decisions", naming only the terminal round's candidate. Roll 14's round 0 candidate was `add_task`, round 1 `tighten_acceptance`; the run report asserts `tighten_acceptance` "on both decisions", which is false. The typed `CorrectionTermination` record has the same single-candidate field. Reword the message (and consider recording both candidates in the termination record).

[...]

PR #879, opened 2026-08-12
fix(qa): state the suite execution model on both guidance surfaces (#877)

jasonLadd:
Closes #877.

Roll 14 (`cyc_25b4a9b0b637/run_02cc78b7acbe`) failed on the only check its app didn't own: eve authored the qa suite as live-HTTP tests against `localhost:3000`, but `tests_pass` executes `vitest run` in a plain Node process — no server exists, nothing boots one. Both correction rounds repaired the crash (`fetch failed` → relative paths → `Failed to parse URL`) instead of the strategy, and #435 terminated the run as `plan_defect`. The context-completeness class again: the author was judged against an execution model it was never shown.

## The fix — same fact, both guidance surfaces

The qa author sees two prose surfaces (`qa_test.py:733-739` system, `qa_test.py:235` user); each already owns its half of the concern, and each was silent or wrong:

**`task_type.qa.test.md`** (0.9.23 → 0.9.24, manifest regenerated):
- New **Execution Environment (hard rule)**, placed ahead of the Discovery Contract: the suite runs in the runner's own process; **no application server is running during test execution, on any stack**; a test that fetches localhost fails unconditionally. In-process alternatives named per stack — `TestClient`/`httpx.AsyncClient` for Python workspaces, route-handler import for Next.js.
- The pytest Discovery Contract rule was stated as universal ("Backend tests MUST be Python pytest files") — nonsense on a single-tree TypeScript workspace, and contradicting the stack-2 supplement in the same prompt. Now scoped to workspaces declaring Python dependency manifests, with an explicit TypeScript-workspace branch (the suite is the declared JS runner's; scaffolded Next.js workspaces collect `__tests__/**/*.test.ts`).

**`nextjs_ts.test_prompt_supplement`** (`dev_capabilities.py`): states the serverless execution model and shows the working pattern — `import { POST } from '@/app/api/runs/route'`, invoke with `new Request(...)`, assert on the returned `Response`, with the `{ params }` form for dynamic routes (verified against the scaffold's pinned `next` 14.2.5, where params is a plain object).

## Tests

- `test_stack_nextjs_ts.py`: the supplement states the serverless model (`NO server`, `localhost`, `vitest run`) and shows in-process handler invocation (import example, `new Request(`, params form).
- `tests/unit/prompts/test_qa_test_fragment.py` (new): the execution rule exists, covers both stacks, and precedes the Discovery Contract; the pytest rule is conditioned and the TypeScript branch exists.

These are content pins on load-bearing guidance — the bug each catches is a later rewrite silently dropping the fact that killed roll 14.

Full regression: 7057 passed, 1 skipped.

## Not in scope

#878 (tests_pass signature granularity — the reason this terminated at round 1 rather than converging on a later repair) is filed separately and not built here.


jasonLadd (commented):
reviewed

PR #882, opened 2026-08-13
fix(executor): never re-seed the walking skeleton on resume (#881)

jasonLadd:
Closes #881.

Resuming a failed run re-ran walking-skeleton seeding, and the fresh stub set shadowed the dev-completed fills in every subsequently materialized workspace — the resumed run tested the skeleton, not the app. Live instance: roll 14's resume (`cyc_25b4a9b0b637/run_02cc78b7acbe`), where probes that passed 38/39 in the original run returned 500 from the moment it resumed, and two correction rounds (~66 min of generation) burned diagnosing an environmental defect as a work-product one.

## Mechanism

Three cooperating pieces, none individually wrong:

1. `execute_run` seeded the skeleton unconditionally — on resume this stored a second full stub set under **new** artifact ids (`scaffold_seeded: True`).
2. `_execute_sequential` restores checkpoint artifacts (old seeds → dev fills → builder) first, then `_seed_prior_artifacts` appends the fresh stubs **after**.
3. `_resolve_artifact_contents` resolves per-filename by last-writer-wins over insertion order — so the fresh stub beat the dev fill for every fill slot. Stubs throw by design, so every route 500s.

## Fix

- **`execute_run`**: skeleton seeding is guarded on *no self-checkpoint* — the restored checkpoint's `artifact_refs` already carry the original seed set. A checkpoint-less resume (failed before any task completed) still seeds; there is nothing to shadow. Replay (SIP-0101) and revision (#811) paths are untouched: they have no self-checkpoint, and their translated/boundary checkpoints carry no implementation fills.
- **`_seed_prior_artifacts`**: ids already restored from the checkpoint are skipped — plan refs are id-stable across resume and were being re-appended after the restored state.
- **`runs assemble`** (second consumer of the same defect): one artifact per filename — produced code beats a scaffold-seeded version regardless of recency, then latest `created_at` wins. Previously it wrote `artifact_refs` in list order with filename overwrite, which on a resumed run ships the stubs.

With the guard in place, tonight's poison artifacts are never referenced again: the next resume restores checkpoint 13 (which predates them) and materializes the dev fills. No vault surgery.

## Tests (each mutation-verified — the fix reverted, the test fails)

- `TestResumeDoesNotReseedSkeleton` — full `execute_run` through the existing harness: a resumed run stores zero `scaffold_seeded` artifacts and still completes; paired with a fresh-run test so the resume assertion can't pass vacuously.
- `TestSeedPriorArtifactsDedupe` — already-restored ids are skipped; fresh ids still seed.
- `TestAssembleArtifactPreference` — the exact roll-14 shape (newer re-seeded stub loses to older produced fill), latest-wins among equals both ways, and legacy artifacts without metadata never mistaken for stubs.

Full regression: 7065 passed, 1 skipped.

## Operational context

Roll 14's resumed run is still burning down its correction chain against the poisoned workspace (bounded by budget, lands FAILED — deliberately not cancelled, since cancel would forfeit resume eligibility and `runs retry` is #880-broken). After this merges + runtime-api rebuild, resuming it again gives the #877 experiment a clean workspace.

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -420,8 +420,13 @@ class DispatchedFlowExecutor(FlowExecutionPort):
             # Data-driven: no manifest -> seed_artifact_refs unchanged = byte-identical to
             # today (the manifest itself is already among plan_artifact_refs; this ADDS the
             # expanded skeleton). Logic lives in helpers (#290 god-file rule).
+            # #881: never on resume. The checkpoint's artifact_refs already carry the
+            # original seed set, and a fresh set stores NEW artifact ids that
+            # _seed_prior_artifacts appends AFTER the restored state — last-writer-wins
+            # per filename then hands every fill slot back to a stub that throws by
+            # design, so the resumed run tests the skeleton instead of the app.
             interface_manifest = await self._load_interface_manifest_for_run(cycle, run)
-            if interface_manifest is not None:
+            if interface_manifest is not None and existing_checkpoint is None:
                 seed_artifact_refs.extend(
                     await self._seed_skeleton_artifacts(interface_manifest, cycle, run_id)
                 )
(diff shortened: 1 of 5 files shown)


jasonLadd (commented):
reviewed

PR #883, opened 2026-08-13
fix(executor): scaffold-seeded artifacts never shadow produced content (#881)

jasonLadd:
Closes #881 (second half — reopened after PR #882 proved incomplete).

The seeding guard in #882 prevents *new* stub sets, but the poison from the first resume is already baked into the run's checkpoints: `run_02cc78b7acbe` checkpoint 21 carries the dev fill at index 26 and the re-seeded stub at index 68 of 92. `_restore_checkpoint_state` appends in checkpoint order and `_resolve_artifact_contents` was last-writer-wins per filename — so the first post-#882 resume (01:54Z) re-poisoned the workspace with zero new seeding. Verified live before the in-flight attempt even finished.

## Fix

`_resolve_artifact_contents` now applies the same rule `assemble` got in #882 — the invariant lives at the consumer, where #882's own PR body said it belonged ("artifact_refs are a log, not a manifest"):

- a **scaffold-seeded artifact never replaces produced content** for the same filename;
- among produced versions, later-in-order still wins (RC3 latest-correction-attempt semantics preserved);
- among seeded versions, later still wins (frozen files exist only as seeded artifacts, so a re-seeded frozen fix stays visible).

One `continue` plus a holder set; byte-budget accounting unchanged.

## Tests (mutation-verified — rule removed, the checkpoint-21 test fails)

In `TestArtifactContentsPreResolution`:
- `test_reseeded_stub_after_fill_loses` — the exact checkpoint-21 shape (stub after fill → fill wins);
- `test_produced_after_stub_wins` — the normal fresh-run shape unchanged;
- `test_latest_produced_version_still_wins` — RC3 semantics not frozen by the new rule;
- `test_frozen_file_latest_seeded_wins` — seeded-only files still take the latest seed.

Full regression: 7069 passed, 1 skipped.

## Operational context

The 01:54Z resume attempt is burning to FAILED against the still-poisoned workspace (bounded by #435/budget; cancel would forfeit resume eligibility). After this merges + runtime-api rebuild, the next resume restores checkpoint order as before but resolution now picks the fills — the #877 experiment finally gets its clean workspace.

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -1246,8 +1246,16 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                 if art_id not in selected_ids:
                     selected_ids.append(art_id)
 
-        # Resolve content (D3: 512KB limit)
+        # Resolve content (D3: 512KB limit). Per filename, later-in-order wins
+        # (RC3: the latest correction attempt supersedes earlier ones) with ONE
+        # exception (#881): a scaffold-seeded artifact never shadows produced
+        # content. The run's artifact list is a log, not a manifest — a resumed
+        # run's checkpoint can carry a re-seeded stub set AFTER the dev fills,
+        # and stubs throw by design, so order alone would hand every fill slot
+        # back to the skeleton. Frozen files exist only as seeded artifacts and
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #886, opened 2026-08-13
fix(correction): repair routing honors ownership facts before heuristics (#884)

jasonLadd:
Closes #884.

Roll 14's resumes #3/#4 both hit the same misroute, and resume #4 showed its full cost: a qa-suite *placement* failure ("No test files found" — the suite emitted at a path vitest's include glob doesn't match) fell through to the #688 same-language fallback, the dev chain was handed eve's suite plus 7 app files, and neo's repair — with zero qa guidance — rewrote the suite live-fetch on #3 and on #4 introduced `router.push(\`/runs/${run.id}\`)` with `run` undefined into `app/runs/new/page.tsx`. The suite passed the vitest-only retest (pages are never imported by the suite), the run **completed**… and the verdict came back `blocked_unverified` because the probes' `next build` died on that exact line. The misroute converted a trivial placement defect into a shipped compile break.

Two facts existed that would each have prevented it; neither was consulted. The fix makes routing honor them in the precedence order ruled in review — facts first, heuristics last:

## 1. The fact source: vitest's second no-suite message (`test_runner.py`)

`_VITEST_SUITE_BROKEN_MARKERS` knew `"No test suite found"` (a file that collects but contains no tests — resume #1's empty stub, routed correctly to qa) but not `"No test files found"` (the include glob matched zero files). One marker line: `suite_broken=True` now fires, `classify_failure_locus` returns OWN_ARTIFACT, and the placement defect routes to `qa.test_repair` → eve, per the existing #568 own-artifact table.

## 2. The ownership veto (`correction_runner.py` + `task_plan.py`)

`_resolve_repair_target` unions the failed task's own artifacts into every target (the pf-21 lesson — kept), including targets dispatched to a **foreign role**. New `_apply_ownership_veto` at the dispatch loop: a repair step whose role differs from the failed task's producing role has that task's own artifacts stripped from its target. Ownership derives from the own-artifact repair table (`own_artifact_role` — never re-typed, the #559 rule); task types without an entry are already repaired by their own role and pass through untouched. If the veto empties a target, that's logged as a locus-classifier miss — the boundary holds rather than handing artifacts across it.

Net precedence: `suite_broken` fact → own-artifact routing; provenance veto at the role boundary; the same-language fallback still exists for its legitimate case (suite ran, app violates contract) but can no longer reach another role's artifacts.

## Tests (all mutation-verified)

- `test_no_test_files_found_is_broken` — the exact resume-#4 stderr; killing the marker fails it.
- `TestOwnershipVeto` (5 tests) — the resume-#3/#4 strip shape, own-role pass-through, no-entry pass-through, veto-empties boundary, and table derivation of the owner role.
- `TestOwnershipVetoWiring` — full correction-protocol run asserting the dispatched dev repair envelope excludes the qa-owned suite while keeping app files; added specifically because the pure-function tests missed an unwired-veto mutation (135 passed with the veto computed but unused — this test kills it).

Full regression: 7090 passed, 1 skipped.

## Not in scope

The second gap resume #4 exposed — the vitest-only retest cannot see page-level type errors, so a compile-broken patch can be *accepted* and only caught at probe prepare (verdict `blocked_unverified`) — is the known stack-2 compile-gate residual (#870's "runtime-api has no node"; the durable fix is SIP-0102's app-container territory). Observed, not filed, pending owner call.

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -476,6 +476,55 @@ def _locus_and_repair_target(
     return (failure_locus, expected, focus, description)
 
 
+def _apply_ownership_veto(
+    target: list[str],
+    failed_task_type: str,
+    step_role: str,
+    failed_own_artifacts: list[str],
+) -> list[str]:
+    """#884: a repair step may not receive another role's artifacts.
+
+    ``_resolve_repair_target`` unions the failed task's own artifacts into
+    every target (the pf-21 lesson: the failing artifact can carry its own
+    bug) — but when the repair chain runs under a DIFFERENT role than the one
+    that produced those artifacts, handing them over invites a guidance-less
+    cross-role rewrite: roll 14's resume #3/#4 (#884) had the dev chain
+    rewrite the qa suite into a live-fetch version, and its page rewrites
+    shipped a compile break that blocked the verdict. Ownership comes from
+    the own-artifact repair table (``own_artifact_role``); task types without
+    an entry are already repaired by their own role and pass through
+    untouched. If the veto empties the target, the locus classifier missed an
+    own-artifact case — logged as such, and the repair proceeds empty rather
+    than handing the artifacts across the boundary.
+    """
+    from squadops.cycles.task_plan import own_artifact_role
+
+    owner = own_artifact_role(failed_task_type)
+    if owner is None or step_role == owner:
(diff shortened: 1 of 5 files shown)


jasonLadd (commented):
reviewed

PR #887, opened 2026-08-13
fix(correction): suite-health verdict joins the tests_pass signature (#878 minimum)

jasonLadd:
Partial fix for #878 — the minimum discriminator; failing-test identity extraction remains open on the issue.

Roll 15 (`run_783f50a2d564`) was killed by a false exact-repeat: round 0 = suite ran, assertions failed (a real app gap — empty-body POST 500s before validation); round 1 = "No test files found" (a placement defect that #886 now routes to eve's own-artifact repair). Both rounds collapsed into the single coarse `tests_pass||failed` signature element, and #435 terminated a **39/40-green, all-probes-passing** run one round before the fixed routing would have repaired the second failure.

The runner's structured `suite_broken` verdict now joins the reason token (`;suite_health=ran|broken`) when present:

- mechanical-vs-behavioral rounds no longer read as repeats (the roll-15 kill, mutation-verified);
- two same-health rounds with the same aggregate reason still repeat — the shk-4 true positive is preserved;
- rows without the verdict (legacy, non-suite checks) produce byte-identical elements, so pre-fix and post-fix signatures compare correctly.

It's a machine fact, not prose — the module's "evidence text never alters the signature" rule is intact.

Rider from the issue: the termination message claimed the terminal candidate appeared "on both decisions"; the rule requires candidate *presence* on both, not identity (roll 14: `add_task` then `tighten_acceptance`). Reworded to name the terminal candidate explicitly.

Full regression: 7093 passed, 1 skipped.

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -913,10 +913,14 @@ class CorrectionRunner:
                 candidate,
                 "; ".join(termination.repeated_signature),
             )
+            # #878 rider: the rule needs a structural candidate PRESENT on both
+            # decisions, not the same one — naming only the terminal candidate
+            # "on both" misstated roll 14's add_task/tighten_acceptance pair.
             raise _ExecutionError(
                 f"plan_defect: correction terminated at round {correction_attempts} — "
-                f"failure signature repeated from round {termination.first_seen_round} "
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #889, opened 2026-08-13
fix(plan): build-profile required_files are a floor the plan must cover (#888)

jasonLadd:
Closes #888.

Roll 15's dead end: the plan's builder task listed `qa_handoff.md` but not `Dockerfile`, bob validated only his task list (#107 — framing may decompose builder work), the suite **converged** (first stack-#2 convergence) — and the run died at the #291 completion gate, which raises a run-level error with no correction chain, and leaves nothing to resume (qa checkpointed complete, builder task can never re-run).

The root problem: the profile's `required_files` are a run-level **floor**, but the per-task list could silently subtract from it — the 1.4.3 "expected_artifacts = floor" lesson, one layer up.

## Fix — shift-left to plan validation

`ImplementationPlan.validate_builder_floor(resolved_config)`: when the plan carries a builder task and a build profile resolves, every profile required file must appear (basename match, same normalization as the builder validator and the #291 gate) in **some** plan task's `expected_artifacts`. Missing → gate rejection → free framing re-roll, before an implementation run is admitted. Unknown/missing profile defers to #426's existing nets; builderless plans owe no floor; coverage may live on any task, not only the builder (if framing assigns a required file to dev, someone owns it — that's what the floor asks).

Wired at **both** plan-gate seams per the #718/#719 rule — the promotion seam (recorded rejection, free re-roll) and the dispatch-admission seam (raises) — and pinned by an AST test that fails if either seam drops the call.

The #291 completion gate stays as the backstop for files no plan could foresee.

## Tests

- `TestValidateBuilderFloor` (7): the exact roll-15 shape (mutation-verified — removing the missing-files computation fails it), full coverage, subdir basename match, any-task ownership, no-profile/unknown-profile deferral, builderless pass.
- `test_builder_floor_rule_is_wired_at_both_seams` — the #718/#719 pin.

Full regression: 7101 passed, 1 skipped.

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -3575,6 +3575,17 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                             parsed_plan.validate_build_config(cycle.resolved_config()),
                         )
                     )
+                    # Roll 15: a plan whose builder task under-covers the build
+                    # profile's required_files converges its whole suite and then
+                    # dies at the #291 completion gate — which has no repair path
+                    # and no incomplete builder task left on resume. Provable
+                    # here; rejection re-rolls framing for free.
+                    errors.extend(
+                        classifier.collect(
+                            "validate_builder_floor",
+                            parsed_plan.validate_builder_floor(cycle.resolved_config()),
+                        )
+                    )
             elif (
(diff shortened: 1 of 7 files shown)


jasonLadd (commented):
reviewed

PR #891, opened 2026-08-14
fix(authoring): builder example shows the profile's required-files floor (#890)

jasonLadd:
Closes #890.

Rolls 15 and 16 both authored builder tasks reproducing the sole-author example's `expected_artifacts` verbatim — `qa_handoff.md` only, `Dockerfile` hedged in description prose ("Dockerfile if applicable"). The v3 authoring-rules appendix states `builder-floor-coverage` abstractly; the example demonstrated violating it, and the example won 2-for-2 — the measured pattern-beats-prose lesson applied to plan authoring.

## Fix

The strongest teaching surface now tells the truth as data (the pf-34 pattern):

- The builder guideline renders the profile's actual `required_files` — the same list `validate_builder_floor` (#888/#889) rejects on — with the floor semantics stated inline.
- The example's `expected_artifacts` **is** that floor, so pattern-matching authors copy compliance instead of violation.
- Unknown/unresolvable profile renders no floor line and keeps the generic example — the offer gate and #426's nets own misconfiguration.

## Tests

- `test_builder_guideline_and_example_carry_the_profile_floor` — nextjs_ts renders `Dockerfile, qa_handoff.md` in the guideline and both files in the example (mutation-verified: reverting the example to the generic shape fails it).
- `test_unknown_profile_keeps_generic_example` — fallback boundary.

Full regression: 7103 passed, 1 skipped (pipefail-verified).

With #889's gate as the net and this as the teaching, roll 17's plan author copies a compliant example and the gate only fires if the author actively diverges.

### DIFF
--- a/src/squadops/capabilities/handlers/_plan_authoring_service.py
+++ b/src/squadops/capabilities/handlers/_plan_authoring_service.py
@@ -90,11 +90,32 @@ async def produce_plan(
     )
 
     if offer_builder:
+        # #890: the example is the strongest teaching surface (rolls 15/16 both
+        # reproduced its expected_artifacts verbatim), so it must SHOW the
+        # profile's required-files floor, not hedge it in description prose —
+        # the floor rendered here is the same list validate_builder_floor
+        # rejects on (#888), stated as data per the pf-34 pattern.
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #894, opened 2026-08-14
feat(sip-0104): Phase 1 — the deterministic test scaffold (contract, generator, seed lifecycle, Gate 1 pins)

jasonLadd:
Phase 1 of the SIP-0104 plan (`docs/plans/deterministic-verification-scaffolding-plan.md`): **Gate 1, the deterministic artifact.** P1 owns generation correctness — deterministic output, authoritative derivation, manifest production.

## Commits, in the plan's ordering (contract first, generator second)

1. **The scaffold contract** (`capabilities/verification_scaffold.py`) — scaffold-manifest schema with the attribution facts a hash mismatch needs (generator version, source-manifest hash, expanded-tree hash, per-file spine hashes, slot region bounds); slot-id grammar and marker parsing with a near-miss guard (a typo'd marker fails instead of freezing into the spine); the normative slot-elided spine-hash canonicalization; derivation-source precedence; the three failure types. Placed in capabilities so P4's region enforcement (cycles domain) imports the same canonicalization the generator uses — one definition, two consumers.
2. **The generator** (`stack_nextjs_ts_tests.py` + `verification_scaffold_emission.py`) — behavior shells derive from the verification contract's own behavioral probes (in-process twins: create, blank-rejection, chained child actions, duplicate-conflict) plus minted read shells where derivable (collection GET, read-by-id via the capture chain, unknown-id 404). Underivable behaviors are demoted to the qa author's additive surface, never guessed. Chained state replays the probe sequence's own semantics per isolated test. Opt-in is an explicit `ScaffoldStack.verification_scaffold` field — nextjs_ts opts in, stack #1 deliberately does not — and the stack-inventory test gains the registry row with a pinned opt-in carve-out.
3. **Seed-time lifecycle** — expand → emit → **validate** → persist → expose: emission validates against its own manifest before persistence, rides the same seed act as the skeleton (never on resume, #881 by construction), and a failure on an opted-in stack fails run setup loudly (the #845 doctrine) instead of burning an LLM round. The scaffold-manifest record is stored as evidence (run id + `verification_scaffold_manifest` type), not materialized into the app tree.
4. **Gate 1 pins** — the committed reference fixture (8 shells + manifest for the reference nextjs_ts manifest at GENERATOR_VERSION 1) with both ends held: input manifest hash and fixture aggregate hashes pinned in the test, no regeneration hatch.

## Gate 1 exit criteria

- **Byte-equivalence**: reference manifest + generator version → byte-identical scaffold and manifest ✅ (standing pin, M0a shape)
- **Derivability mutations**: tampered tree (missing route file / missing export / missing store seam) → refusal naming the disagreement; unmapped `not_found` → shell deterministically omitted; unopted stack → refusal; declared-but-unregistered emitter → refusal ✅
- **Stack #1 unmoved**: existing byte-identity pins (`bb472e26…`, `7622f570…`) green; no stack-1 scaffold; opt-in explicit ✅

Validation: full regression suite **7162 passed, 1 skipped**, ruff clean.

## Deliberate interim state (P1→P4 window)

The emitted shells are seeded workspace artifacts but are **not yet in the bound record's frozen set** — region-level enforcement of the scaffold is exactly P4's phase, so until P4 lands a producer could still rewrite a shell wholesale. Covered by the plan's roll policy (no experimental roll before P1–P4 deployed and proven). Also per plan, `qa.test` does not yet know about the scaffold (fill mode is P3): shells simply ride the workspace and execute with the suite.

Next after merge: **Phase 2 — Gate 2, the executable artifact** (skeleton-execution gate + the negative corpus, including the consistent-inputs/broken-generator case).

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -427,9 +427,20 @@ class DispatchedFlowExecutor(FlowExecutionPort):
             # design, so the resumed run tests the skeleton instead of the app.
             interface_manifest = await self._load_interface_manifest_for_run(cycle, run)
             if interface_manifest is not None and existing_checkpoint is None:
-                seed_artifact_refs.extend(
-                    await self._seed_skeleton_artifacts(interface_manifest, cycle, run_id)
+                skeleton_refs = await self._seed_skeleton_artifacts(
+                    interface_manifest, cycle, run_id
                 )
+                seed_artifact_refs.extend(skeleton_refs)
+                # SIP-0104: the deterministic test scaffold rides the same seed act, and
+                # only on top of an actually-seeded skeleton — the tree its shells'
+                # imports resolve against is the tree this run carries. Same #881
+                # no-resume rule by construction (this whole branch is seed-time only).
+                if skeleton_refs:
+                    seed_artifact_refs.extend(
+                        await self._seed_verification_scaffold_artifacts(
+                            interface_manifest, cycle, run_id
+                        )
+                    )
 
             # LangFuse + Prefect observability setup
             obs_ctx, flow_run_id = await self._init_run_observability(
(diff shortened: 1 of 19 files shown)


jasonLadd (commented):
reviewed

PR #895, opened 2026-08-14
feat(sip-0104): Phase 2 — the execution-readiness gates (static bytes-level + skeleton execution)

jasonLadd:
Phase 2 of the SIP-0104 plan: **Gate 2, the executable artifact** — proving the generated scaffold can execute before any author sees it, and that the gate detects *generator bugs*, not merely missing inputs.

## The two halves

**Static execution-readiness gate** (`capabilities/verification_scaffold_gate.py`, pure) — judged from the **emitted bytes**, never the generator's intent, which is what makes the decisive case catchable: internally consistent inputs + a wrong emitted import path. Checks: every `@/…` import resolves against the expanded tree; named imports and `alias.MEMBER` invocations are exported by the resolved file; bare imports are declared dependencies (the roll-9 `supertest` class); asserted statuses exist in the contract; slot/probe identity survives into the bytes; placement is collectable (the #884 class). Findings accumulate — one report, not one-per-revision.

Wired twice:
- **`PROOF_SCAFFOLD_READY` at the winnability boundary** — a scaffold that cannot be validly emitted costs a free framing re-roll (#522) instead of a failed implementation run. Classified in the M6 taxonomy as DERIVATION_DEFECT (generator-owned by elimination; never taught to the author).
- **The executor seed path** — same tree feeds emission and gate; failure fails run setup loudly (the P1 seam, extended).

Both callers pinned by the guard-wiring inventory test.

**Dynamic skeleton-execution gate** (`capabilities/handlers/scaffold_execution.py`) — materialize stubs + scaffold, `npm install`, `vitest run --reporter=json`, classify. The stub assertion wall is expected and counted (SIP-0098 §7: the bare skeleton must fail); mechanical failures — a suite that dies pre-test, a shell never collected, a runtime crash — are `scaffold-invalid`. Runner-level failures (no npm, timeout, no report) are **infrastructure**, never a verdict. The classifier is pure and its corpus runs everywhere; only the runner needs the node toolchain.

## Real-node validation (node:20-alpine, vitest 1.6, on this branch)

- **Clean reference scaffold**: 9 tests collected (8 shells + harness), harness passed, 8 shells failed as the expected stub wall → verdict **valid, 8 expected assertion failures**.
- **Injected runtime defect** (`new Requests(` — statically well-formed, only execution can see it): verdict **scaffold-invalid**, exactly one mechanical failure naming the shell; the other 7 shells stayed expected.
- The measurement **caught and fixed a real classifier bug**: vitest's JSON reporter emits chai mismatches as bare messages (`expected 500 to be 201 // Object.is equality`) and strips error-name prefixes from thrown errors too (`Requests is not defined`). Classification is therefore an **allowlist of assertion shapes** — unknown shapes stay mechanical, the safe direction — and the corpus pins the measured forms. A blocklist of error-name prefixes would have shipped broken shells.
- A node-gated integration pair (clean + injected defect) is committed; it skips where npm is absent and runs in the agent-container toolchain.

## Gate 2 exit criteria

- imports resolve / handlers exist / statuses in contract / slot boundaries valid / criterion identity survives ✅ (static corpus)
- empty-fill scaffold collects and every shell executes against the skeleton without mechanical crash ✅ (real-node runs + committed e2e pair)
- **the decisive negative test** — consistent inputs + injected generator defect caught as scaffold-invalid ✅ (static: wrong emitted import path; dynamic: injected runtime crash)
- validity failure fails run setup loudly, never an LLM round ✅ (seed-path raise + winnability finding, both pre-LLM)

Validation: full regression **7192 passed, 1 skipped**, ruff clean. No emission bytes changed — GENERATOR_VERSION stays 1 and the P1 byte-equivalence pin holds untouched.

## Sequencing note (named, not silent)

Per-run wiring of the **dynamic** gate into the qa lane is deliberately deferred to P3: it needs the scaffold surfaces in the qa.test envelope, which is exactly the envelope P3 rebuilds for fill mode. Until then the dynamic gate is standing-corpus + integration evidence; the static gate runs per-run at both boundaries. Roll policy (no rolls before P1–P4 proven) covers the window.

Next after merge: **Phase 3 — Gate 3, the bounded agent surface** (merge contract before implementation: slot-id addressing, fill validation, the missing-slot rule, the semantic brief as coverage inventory).

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -1908,18 +1908,27 @@ class DispatchedFlowExecutor(FlowExecutionPort):
         (P4), not a file the app tree should carry. It is reachable by run id +
         artifact type.
         """
-        from squadops.capabilities.scaffold import verification_scaffold_for
+        from squadops.capabilities.scaffold import expand, verification_scaffold_for
         from squadops.capabilities.verification_scaffold_emission import (
             VERIFICATION_SCAFFOLD_MANIFEST_ARTIFACT_TYPE,
             VERIFICATION_SCAFFOLD_MANIFEST_FILENAME,
             emit_verification_scaffold,
+        )
+        from squadops.capabilities.verification_scaffold_gate import (
+            validate_execution_readiness,
         )
 
         if not verification_scaffold_for(manifest.stack):
             return []
 
         try:
-            emission = emit_verification_scaffold(manifest)
+            # One tree for emission AND the readiness gate, so the two cannot check
+            # against different expansions. The gate re-reads the EMITTED BYTES (P2):
+            # emission's own validation proves record/output agreement; this proves the
+            # output's mechanical claims hold against the tree and contract.
+            tree = expand(manifest)
+            emission = emit_verification_scaffold(manifest, expanded=tree)
+            validate_execution_readiness(emission, tree, manifest)
         except Exception:
             logger.error(
                 "test-scaffold emission failed for opted-in stack %s (run %s) — failing "
--- a/src/squadops/capabilities/handlers/scaffold_execution.py
+++ b/src/squadops/capabilities/handlers/scaffold_execution.py
@@ -25,6 +25,7 @@ import asyncio
 import json
 import logging
 import os
+import re
 import shutil
 import tempfile
 from dataclasses import dataclass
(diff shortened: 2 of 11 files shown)


jasonLadd (commented):
reviewed

PR #896, opened 2026-08-14
feat(sip-0104): Phase 3 — the fill protocol and qa.test fill mode (bounded agent surface)

jasonLadd:
Phase 3 of the SIP-0104 plan: **Gate 3, the bounded agent surface** — the merge contract defined before its consumers, and qa.test switched to fill mode on scaffolded stacks.

## The fill protocol (`capabilities/verification_scaffold_fill.py`)

The merge contract, stated normatively in the module and enforced by its corpus:

- **Slot id is the only addressing mechanism** — `fill:slot-<id>` fences, plus an explicit `not_applicable: <reason>` directive. No paths, no line numbers, no diffs.
- **One fill per slot** — a slot filled twice rejects the slot outright rather than picking a winner.
- **Containment** — a fill body may carry assertions only: no slot-marker vocabulary (region forgery), no `import`/`require()` (the spine owns dependencies), no `fetch()`/XHR/WebSocket (the #877 class), bounded size, and no empty-body silent non-coverage.
- **Deterministic merge with verified containment** — slot-body replacement in scaffold order; the merged spine is re-hashed on every merge and must equal the scaffold record's, or the merge refuses its own output.
- **The missing-slot rule** — every declared slot ends in exactly one disposition (`filled` / `not_applicable` / `rejected` / `missing`); missing and rejected slots render a deterministic `expect().toBe` failing state naming the slot, which the P2 classifier reads as an assertion failure (fill layer), never a mechanical crash (generator).
- Misaddressed fills are recorded, never silently dropped.

## qa.test fill mode

- **Envelope**: `bind_verification_scaffold` on the context contract; `inject_contract_inputs` re-derives the scaffold from the manifest (Gate 1 byte-equivalence makes that exact — same derivation style as every other surface in that composer). Presence-keyed: unopted stacks and manifest-less calls stay byte-identical. Retest-forwarded.
- **Semantic brief = coverage inventory only** (SIP §4.5): one line per slot derived from the slot table; all instruction prose lives in the managed asset `request.qa_test_fill_mode_appendix` (#448). No generated coaching — richer briefs stay SIP §12 follow-on.
- **Handler**: fill fences are parsed and stripped *before* file extraction (they'd otherwise extract as files named `slot-…`); a fills-only emission is authorship, not a #566 zero-extraction failure; merged shells join both the stored artifacts and the executed suite; wholesale shell rewrites (initial and self-eval passes) are dropped and recorded; per-slot dispositions land in evidence. Additive test files ride the existing pipeline unchanged.

## Gate 3 exit criteria — all measured

- **Round-trip determinism** ✅ (corpus)
- **Negative corpus** ✅ — oversized/duplicate/misaddressed/marker-smuggling/import/server-access/empty fills each rejected with the finding named
- **Garbage fill degrades one slot, suite still collects** ✅ — proven structurally in the corpus AND **executed against real node** (node:20-alpine): the merged garbage fill's file dies at `Transform failed`, exactly one mechanical failure, the other 7 shells still execute their expected stub assertions
- **Slot containment** ✅ — merged spine hash equals the scaffold's, verified on every merge and refused on violation

Validation: full regression **7231 passed, 1 skipped**, ruff clean. No emission changes — GENERATOR_VERSION stays 1, the P1 byte pin holds.

## Notes for review

- **Expected-artifacts semantics in fill mode**: the plan's expected suite paths remain satisfiable — the asset instructs the author to emit any expected path as a normal additive fence; shells are always produced by the merge. If plans start expecting scaffold paths, that's a plan-authoring surface to revisit (flagged, not silently absorbed).
- **Fill-mode repair locus is deliberately P4**: bounding what a *repair* may do to shells is exactly P4's restore-and-signal subject ("no repair path may modify frozen regions"); P3 bounds initial authorship and self-eval. Until P4, the interim state from P1 stands (shells not yet frozen-enforced; roll policy covers).
- **P4 can now begin**: Gate 3 passes, which was P4's explicit precondition ("region enforcement must not debug an unstable fill protocol").

Next after merge: **Phase 4 — Gate 4, enforcement against adversarial producers** (frozen-spine + slot-boundary verification on every stored emission, the adversarial corpus, restore-and-signal on the repair path — the checkpoint before any expensive roll).

### DIFF
--- a/src/squadops/capabilities/context_assembly.py
+++ b/src/squadops/capabilities/context_assembly.py
@@ -108,6 +108,11 @@ class ContextAssemblyContract:
     plan_rejection_context: bool = False
     bind_criteria_index: bool = False
     bind_behavioral_surface: bool = False
+    #: SIP-0104 P3: this task receives the deterministic test scaffold (slot table +
+    #: shell files) and authors in fill mode when the stack opts in. Derived from the
+    #: interface manifest at injection (Gate 1 byte-equivalence makes re-derivation
+    #: exact); presence-keyed like every surface here.
+    bind_verification_scaffold: bool = False
 
 
 _EMPTY_CONTRACT = ContextAssemblyContract()
@@ -169,6 +174,7 @@ CONTEXT_CONTRACTS: dict[str, ContextAssemblyContract] = {
         acceptance_workspace=True,
         bind_behavioral_surface=True,
         manifest_surfaces=(SURFACE_FROZEN,),
+        bind_verification_scaffold=True,
     ),
     # --- planning chain (#657): upstream documents on an envelope-local
     # prior_outputs copy; all four authoring types receive a re-roll's
@@ -345,6 +351,9 @@ RETEST_PRESENCE_KEYS: tuple[str, ...] = (
     # the forwarded-vs-reassembled distinction, visible for the first time.
     "workspace_revision_id",
     SURFACE_DOM_TESTID,
+    # SIP-0104 P3: a retest that re-authors from scratch must still be in fill
+    # mode — losing the key would flip it back to whole-suite authorship.
+    "verification_scaffold",
 )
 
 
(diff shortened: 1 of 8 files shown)


jasonLadd (commented):
reviewed

PR #897, opened 2026-08-14
feat(sip-0104): Phase 4 — region-level enforcement against adversarial producers (Gate 4)

jasonLadd:
Phase 4 of the SIP-0104 plan: **Gate 4 — enforcement against adversarial producers.** With this, Gates 1–4 together establish the core invariant: **the LLM cannot mechanically rewrite the scaffold.** This is the plan's checkpoint before any expensive roll.

## What changed

- **The bound record binds the scaffold** (`bound_scaffold_record.py`): `verification_scaffold` carries the per-shell spine hashes and slot tables for an opted-in stack, derived at bind time. Fail-soft: a derivation failure degrades to file-level enforcement, never disables the bound record; pre-P4 records deserialize with region enforcement off; a tampered stored record refuses to load (the manifest's own aggregate verification).
- **The region lane in `enforce_frozen_ownership`** — shared by BOTH landing paths (executor storage and the correction runner's repair emissions), so the repair path inherited enforcement the moment it landed. Content-based and role-independent (§4.3's "any role, any locus"): an emission to a shell must parse its slot structure, keep the bound slot set, reproduce the bound spine hash, and keep every slot body containment-clean. The only legal shape is exactly a fill-merge product — body edits inside intact markers, which pass for qa *and* for a dev-locus repair.
- **Two taxonomy codes**, each with a distinct corrective action: `scaffold_region_violation` (spine/structure mutated — the adversarial class) and `prohibited_fill_emission` (spine intact, body smuggles imports/network — the correctable prohibited-fill class). Evidence carries a `detail` field and **spine-vs-spine attribution**: a containment violation shows the spine hashes *equal* (the body was the sin); a region violation shows them diverging.
- **Restore-and-signal on the repair path**: drops retain rather than restore (#691 — the prior valid version stays the workspace copy), and one `enforcement_instruction` selector feeds the carry transport for both the frozen-file and region codes, so the next repair attempt is told the spine is frozen instead of fighting enforcement blind (the #884 class, closed structurally).

## Gate 4 exit criteria — all in the corpus (25 tests, real bound record from the reference manifest)

- **The adversarial-producer end-to-end fixture** (the plan's single strongest check): changing an import / changing the invocation strategy / modifying a status assertion / moving a slot boundary → `scaffold_region_violation`; pointing a slot body at a live server / smuggling an import into a body → `prohibited_fill_emission`; adding a dependency (package.json) / rewriting the harness test → the existing `frozen_path_emission` lane. Each lands in its expected class deterministically; a legal body edit passes for any role; additive files pass.
- **Slot-boundary manipulation, each explicit**: deleted marker pair, deleted single marker, duplicated slot, nested forged slot, renamed slot id, statement injected adjacent to a slot — all caught, each with the failing check named in `detail`.
- **Hash-sufficiency demonstrated**: every mutation class the plan names is caught by the slot-elided canonicalization + structure parse + slot-set check + body containment. AST-level verification remains the named escalation on evidence of a miss (none observed).
- **Restore-and-signal proven on the repair path**: driven through the correction runner's own `_enforce_step_emissions` — the shell rewrite drops in place and the next attempt's carry receives the instruction.

Validation: full regression **7256 passed, 1 skipped**, ruff clean. Two taxonomy pins updated (code count 4→6, evidence field set +`detail`) — the pins fired exactly as designed.

## The roll boundary

After this merges, **P1–P4 are complete in code**. Per the plan, the boundary before any experimental roll is P1–P4 **deployed and proven**: rebuild agent + runtime-api images, restart, and verify the loaded modules in-container (the standing rebuild-before-SIP-test rule). P5 (evidence/classification/correlation) can proceed in parallel with that — its schema work has no runtime prerequisite on deployment.

Honest status of one earlier deferral: the P2 **dynamic** skeleton-execution gate remains standing-corpus + CI evidence (the `scaffold skeleton gate` CI job runs the real vitest pair on every PR); per-run wiring stayed out — the P6 window protocol is the named empirical net for the residue, and any mechanical failure it catches names its uncovered surface and joins the enforced set.

Next after merge: **Phase 5 — Gate 5, evidence and convergence** (four-class failure routing, correlation-not-collapse, run-report fields), and the deploy+prove step for the roll boundary.

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -677,9 +677,8 @@ class CorrectionRunner:
         if bound_record is not None and step_artifacts:
             from squadops.cycles.scaffold_enforcement import (
                 enforce_frozen_ownership,
-                frozen_emission_instruction,
+                enforcement_instruction,
             )
-            from squadops.cycles.task_outcome import ContractComplianceViolation
 
             enforced, integrity_evidence = enforce_frozen_ownership(
                 step_artifacts, bound_record, step_envelope
@@ -689,14 +688,17 @@ class CorrectionRunner:
                 step_artifacts = enforced
                 for record in integrity_evidence:
                     self._emit_scaffold_integrity_evidence(record, step_envelope)
+                    # SIP-0104 P4: one selector covers the frozen-file AND the
+                    # region-level codes — the next attempt is TOLD what enforcement
+                    # rejected instead of fighting it blind (#691; the #884 repair
+                    # class §4.3 prohibits).
+                    instruction = enforcement_instruction(record)
                     if (
-                        enforcement_carry is not None
-                        and record.violation_code
-                        == ContractComplianceViolation.FROZEN_PATH_EMISSION
+                        instruction is not None
+                        and enforcement_carry is not None
+                        and instruction not in enforcement_carry
                     ):
-                        instruction = frozen_emission_instruction(record)
-                        if instruction not in enforcement_carry:
-                            enforcement_carry.append(instruction)
+                        enforcement_carry.append(instruction)
 
         # pf-31 Fix D: drop syntactically invalid .py emissions (truncation
         # guard) — the prior stored version (last known parseable) stays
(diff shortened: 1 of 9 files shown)


jasonLadd (commented):
reviewed

PR #899, opened 2026-08-15
fix(sip-0104): four defects measured on window roll 1 (sandbox install, audit stack, silent non-collection, rollup pollution)

jasonLadd:
Four defects measured while auditing **SIP-0104 window roll 1** (`cyc_04d36309d793`). Two blocked the audit itself; two are evidence-quality gaps the roll exposed. None changes a verdict, and none affects roll 1's standing as a clean window roll.

## 1. The stack-#2 sandbox could never install (blocked the audit)

`NEXTJS_TS`'s environment contract ran `npm ci`, which refuses outright without a `package-lock.json` — reproduced in a clean container: `EUSAGE: The npm ci command can only install with an existing package-lock.json`. The scaffold emits `package.json` only (an offline-deterministic expansion cannot produce a lockfile), so **this operation could never succeed on this stack** — the sandbox path was unreachable-green by construction, and roll 1's deliverable failed install there while passing every probe elsewhere.

This exact correction already existed twice in the codebase and hadn't been propagated: the probe runner's `nextjs_next_start` profile carries it with a comment from roll 10 (`cyc_43a216d43e1e`), and stack #1's own contract states the rule three declarations above (*"no lockfile → npm install"*). Now all three surfaces plus the vitest runner agree.

One consequence, named: `contract_id()` hashes the full declaration, so this changes the contract id and invalidates cached sandbox workspaces keyed to it — a cache miss, not a correctness risk. No test pinned the old command or hash.

## 2. The audit tool was stack-#1-only (blocked the audit)

`audit_delivered_app.py` hardcoded `FULLSTACK_FASTAPI_REACT` and a `backend/routes.py` assembly sentinel, so auditing a nextjs_ts deliverable would stand a Next.js tree up with uvicorn and report **the auditor's** defect as the app's. It now takes the stack from the contract's own `skeleton.expander` and resolves the environment through the existing `get_environment_contract()` accessor (which refuses unknown stacks — no fallback environment), with a per-stack assembly sentinel table.

## 3. Authored suites the runner silently never collected

Roll 1's qa author wrote ~9KB of additive tests at `tests/api/runs.test.ts` — outside this stack's `**/__tests__/**/*.test.ts` include. Nine test files went to the runner; vitest collected eight. The additive half of the semantic layer verified nothing, and every signal read green because the collected suites passed. This is the #884 class one step down: not "no test files found" but "this one was ignored".

Now detected from the JSON report (restricted to files that declare themselves suites — a `*.test.ts`, not a helper module), logged as a warning, named in the test report a human reads, and banked in the evidence summary — §6's "is the authored layer earning its inference spend" question is unanswerable if part of that layer silently never ran.

**Deliberately reported, not verdict-changing.** Making it blocking is a real option — the author's work isn't running and the fix is trivial — but that changes verdict semantics mid-window, so it should be a decision, not a side effect of detection. Flagging for a call.

## 4. The P5 classification polluted the verification rollup

The classification rode a `validation_result.checks` row, and `normalize_task_checks` records **any** row carrying a `status` key — so a diagnostic aggregated as a verification check and surfaced in roll 1's cycle outcome as `unverified: [{check_id: scaffold_failure_classification, reason: unspecified}]`. Non-blocking, but a diagnostic is not evidence (SIP-0096 §6.1).

It now rides the failure-evidence payload beside `emission_failure` and `app_tracebacks` — the transport built for exactly this. The locus classifier reads the classes from there, and as a bonus the analyzer's prompt (which renders the whole evidence dict) now sees the per-failure owner/route rows.

## Validation

Full regression **7282 passed, 1 skipped**, ruff clean. No emission changes — GENERATOR_VERSION stays 1 and the Gate 1 byte pins hold.

After merge: rebuild, re-run the boot audit to close roll 1 on a green audit rather than probe evidence alone, then launch roll 2.

### DIFF
--- a/scripts/dev/audit_delivered_app.py
+++ b/scripts/dev/audit_delivered_app.py
@@ -49,9 +49,17 @@ from squadops.cycles.verification_contract import (
     capture_probe_values,
     resolve_probe_path,
 )
-from squadops.sandbox.environment import FULLSTACK_FASTAPI_REACT
+from squadops.sandbox.environment import get_environment_contract
 from squadops.sandbox.models import RevisionOrigin
 from squadops.sandbox.workspace import WorkspaceStore
+
+#: One file per stack whose absence means assembly itself failed — the deliverable's
+#: irreducible entry surface. Keyed by stack so a third stack declares its own rather
+#: than inheriting a check that silently cannot hold.
+_ASSEMBLY_SENTINELS: dict[str, str] = {
+    "fullstack_fastapi_react": "backend/routes.py",
+    "nextjs_ts": "package.json",
+}
 
 _VAULT_CONTAINER = "squadops-runtime-api"
 _WORKSPACE_ARTIFACT_TYPES = frozenset({"source", "test", "config"})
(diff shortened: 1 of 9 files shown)


jasonLadd (commented):
reviewed

PR #900, opened 2026-08-15
fix(sip-0102): pin verification before build, not after — plus §11a stack-awareness amendment

jasonLadd:
**Roll 1's deliverable now audits PASS** — installs, builds, boots, and answers all 5 contract probes over real HTTP. Getting there surfaced two more defects of the class #899 fixed, this time inside the sandbox rather than its callers, and the second one was a design fork rather than a missing list entry.

## The two defects

**`.next` was not derived state.** `_EXCLUDED_SEGMENTS` listed `dist` — Vite's build output — but not `.next`. Since `current_files()` reads every unexcluded file as UTF-8, pin verification walked into webpack caches and raised `UnicodeDecodeError`. Not over-pinning: a crash, before the app could boot. The new test writes genuinely binary content, which is why the pre-existing exclusion test (all text) never caught it.

**`next build` rewrites its own source.** It generates `next-env.d.ts` and rewrites the scaffold's frozen `tsconfig.json`, injecting `allowJs` and `skipLibCheck` — documented Next behavior. So no tree on this stack can match its pin once built, and §4.6's verification before `start_application` was asserting an invariant that was never true here.

## The ruling (option 2)

**Pin verification happens before the tree is built, not after.** `EnvironmentContract` now declares `build_mutates_source` — a fact about the toolchain, declared per stack — and where it is true, only the operation that precedes any build (`install_dependencies`) verifies.

- The canonical stack is **byte-identical** to today (`False` default): post-build drift still refuses, proven by a test.
- The guarantee kept is the one that carries meaning: *the source we declared is the source we built.*
- Alternatives weighed and rejected in the amendment: excluding `tsconfig.json` (silently drops tamper detection on a frozen scaffold file), re-pinning post-build (faithful but larger, and nothing consumes a post-build revision yet), and changing the scaffold's emitted tsconfig (fixes the root cause but moves generator bytes, which would **reset SIP-0104's measurement window** — deferred until it closes).

## SIP-0102 §11a — why this needed a SIP change, not just code

This is the answer to "was the sandbox built without stack awareness?" — and the document is more honest than that framing suggests. §3's non-goals say *"Not multi-stack generalization (that is v1.6)"* and §4.2 defers the `StackBlueprint` schema until a second real stack exists. **It scoped itself to one stack deliberately.**

What went unrecorded: #822 later added a `nextjs_ts` environment contract — extending the subsystem to a second stack — while the generalization stayed deferred. The contract seam took the entry cleanly; everything around it stayed canonical-stack-shaped. Four defects in one audit is the measured cost.

The amendment records the ruling with its evidence, and deliberately states **what is still ungeneralized**: `_EXCLUDED_SEGMENTS` remains one global set, so a third stack whose build output has another name repeats that defect. The honest scope is "stack #2 works and the lifecycle claim is now true", not "the sandbox is stack-general". `SIP-Stack-Blueprint-Contract` still owns generalization — and its own gate (*"deferred until a second real stack exists"*) is now satisfied, so it is unblocked by its own criterion. That is a separate call for the owner.

**Consequence noted:** `contract_id()` covers the full declaration by design (§7 items 4/15), so the new field changes every contract id and invalidates workspace caches keyed to them — a cache miss, not a correctness risk.

## Validation

Full regression **7286 passed, 1 skipped**, ruff clean. Audit exit 0 on `cyc_04d36309d793` / `run_65f385f58656`.

### DIFF
--- a/adapters/sandbox/container_backend.py
+++ b/adapters/sandbox/container_backend.py
@@ -78,6 +78,7 @@ class ContainerBackend(NoOpExecutionSandbox):
         http_client_factory: Callable[[], httpx.AsyncClient] | None = None,
         environment_contract_id: str | None = None,
         cache_root: Path | None = None,
+        build_mutates_source: bool = False,
     ) -> None:
         self._container = container
         self._store = store
@@ -91,6 +92,7 @@ class ContainerBackend(NoOpExecutionSandbox):
         self._app_port = app_port
         self._ready_path = ready_path
         self._environment_contract_id = environment_contract_id
+        self._build_mutates_source = build_mutates_source
         self._cache_root = cache_root
         self._readiness_timeout = readiness_timeout_seconds
         self._poll_interval = poll_interval_seconds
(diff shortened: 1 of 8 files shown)


jasonLadd (commented):
reviewed

PR #902, opened 2026-08-15
fix(dev): per-stack fill-only guidance — roll 1's app shipped a dead UI because the author was told stack #1's seam

jasonLadd:
Found by opening roll 1's delivered app in a browser. It passed **36/36 checks, all 5 contract probes, `tests_pass`, `frontend_build`, and the boot audit** — and its UI is completely dead. All five page data calls 404:

```
app/page.tsx:39            api<Run[]>('/runs')
app/page.tsx:70            api('/runs', {...})              ← create
app/runs/[run_id]/page.tsx api<Run>(`/runs/${runId}`)       ← detail
                           api(`/runs/${runId}/join`)
                           api(`/runs/${runId}/leave`)
```

The routes are mounted at `/api/runs`; `GET /runs` is a 404 and there are no rewrites.

## The author didn't guess — it was told

One fill-only appendix served both stacks. A `nextjs_ts` author received, verbatim:

- *"Backend route functions in `backend/routes.py`"* — does not exist on this stack
- *"Frontend view components (`frontend/src/views/*.jsx`)"* — does not exist
- *"`apiFetch` is the wired data-access seam; call it (**it prefixes `/api`** and unwraps the error envelope)"* — that is stack #1's helper and stack #1's semantics
- *"Do NOT edit … `frontend/src/main.jsx`, `frontend/src/App.jsx`, `vite.config.js`, `index.html`"* — none of these exist

Stack #1's `api.js` really does `fetch(\`/api${path}\`)`. Stack #2's `lib/api.ts` fetches the path **verbatim**. The author followed the instruction it was given, and every UI call 404'd. This is #866's context-completeness class inverted: not a fact the agent could not see, but a fact it was told wrong.

## The fix

`DevelopmentCapability` gains `fill_only_template`, resolved per stack through the capability that `ScaffoldStack.dev_capability` already binds — reusing an existing registry rather than minting one. `nextjs_ts` gets its own asset describing **its** layout (`app/**/route.ts`, `app/**/page.tsx`, the `lib/` seams) and stating the client contract exactly, with the correct and the wrong call side by side:

```ts
const runs = await api<Run[]>('/api/runs')   // correct — the declared path
await api('/runs')                            // WRONG — 404, no such route
```

**An unset template renders no appendix at all.** Wrong guidance is worse than none (#818's asymmetric-default rule) — this defect is the proof, so a third stack gets silence until someone writes its asset, not stack #1's conventions.

**`lib/api.ts` is deliberately left fetching verbatim.** Hardcoding an `/api` prefix would silently break any manifest that declares its API under a different prefix — #859 makes a distinct prefix a *requirement* of this stack but does not mandate the literal string. The seam is correct; only the instruction was wrong.

## Window safety

Nothing in the scaffold moves: `GENERATOR_VERSION` stays 1, the Gate 1 byte pins hold, and the expanded tree is untouched. This is a prompt-asset and selection change only, so the SIP-0104 measurement window stays comparable and **roll 2, currently in flight, is unaffected** — it will still reproduce the UI defect, which is the honest baseline. The fix applies from roll 3.

## Validation

Full regression **7294 passed, 1 skipped**, ruff clean. The new tests assert both failure modes roll 1 shipped: that a nextjs_ts author sees this stack's layout and *not* `backend/routes.py`, `frontend/src/views`, or `apiFetch`; and that the prefix semantic is stated with both the correct and the wrong form.

## Still open (not in this PR)

Nothing in the pipeline exercises the UI's data path — probes hit the API directly, shells invoke handlers in-process, `frontend_build` only compiles, and the boot audit probes the API. A future roll can still ship a dead front end and bank green. Worth extending the audit to load a page and assert its data renders.

### DIFF
--- a/src/squadops/capabilities/dev_capabilities.py
+++ b/src/squadops/capabilities/dev_capabilities.py
@@ -51,6 +51,17 @@ class DevelopmentCapability:
     # Materialized into the QA build/test workspace so the frontend build check
     # (#290) and vitest actually run instead of skipping on "no package.json" (#296).
     build_support_files: tuple[str, ...] = ()
+    #: The managed asset carrying this stack's fill-only instruction — which files hold
+    #: the slots, and what the scaffold-owned seams mean. Per stack because the answer
+    #: IS the stack: one asset served both, so a `nextjs_ts` author was told to fill
+    #: `backend/routes.py` and `frontend/src/views/*.jsx` and that `apiFetch` "prefixes
+    #: `/api`" — none of which exist here. It wrote `api('/runs')` against a helper that
+    #: prefixes nothing, and every UI call 404'd in a deliverable that passed every gate
+    #: (SIP-0104 window roll 1, cyc_04d36309d793, measured 2026-08-15).
+    #:
+    #: Empty means NO fill-only appendix rather than another stack's — wrong guidance is
+    #: worse than none (#818's asymmetric-default rule; this defect is the proof).
+    fill_only_template: str = ""
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #903, opened 2026-08-15
feat(audit): UI data-path check — does the UI reach its own API?

jasonLadd:
Closes the gap named in #902: **nothing in the pipeline follows the path the UI itself takes.** Contract probes issue HTTP at *declared* paths, SIP-0104's shells import route handlers and invoke them in-process, `frontend_build` only compiles, and the boot audit probes the contract. Roll 1 shipped the consequence — all five page data calls 404'd in an app that passed 36/36 checks, all five probes, `tests_pass`, `frontend_build`, and the audit.

## What it does

`capabilities/ui_data_path.py` extracts the request paths the UI's own source will issue and resolves each through **that stack's** client seam — stack #1's `apiFetch` prepends `/api`, stack #2's `api` fetches verbatim. Encoding that difference is the point: assuming one stack's convention for the other is precisely the defect. An unknown stack yields nothing rather than guessing a prefix, which would invent failures in a working app.

The audit then puts each distinct path to the running app and asks one unambiguous question: **does a route answer where the UI asks?** Whether the response is semantically right stays the contract probes' job.

No headless browser. The defect is a wiring fact, decidable from source plus one request per path; a browser would add a runtime dependency to answer a question that doesn't need one.

## Two failure shapes — the second was found by running the check against the real app

Because App Router serves pages and API handlers from one routing tree (#859), a wrong API path has two ways to fail:

- **`route_missing`** — a 404 the framework produced. Its 404 page is HTML, while an app answering through the scaffold's frozen `errorResponse` returns JSON, so a legitimate "no such run" (404 **with** the envelope) still passes. That discriminator is what keeps correct apps green.
- **`page_not_api`** — a **200 carrying HTML**. The call landed on a *page* whose URL happens to match, so it never 404s: the seam's `res.json()` fails, its `.catch(() => ({}))` yields an empty object, and the view renders blank with no error anywhere.

The first version of this check only knew the 404 rule. Running it against roll 1's real app caught **4 of 5** calls — `api(\`/runs/${runId}\`)` hit `app/runs/[run_id]/page.tsx` and was classified served. That miss is now its own test class. The JSON requirement applies only to seam calls (`api`/`apiFetch`), which parse JSON by construction; a raw `fetch` promises nothing about the body and is judged on route existence alone.

## Verified against the real deliverable

```
FAIL ui-data-path app/page.tsx:39: api('/runs') requests '/runs', which the app serves no route for — this call 404s for every user
FAIL ui-data-path app/page.tsx:70: api('/runs') ...
FAIL ui-data-path app/runs/[run_id]/page.tsx:42: api('/runs/${runId}') requests '/runs/ui-probe-sample', which serves a PAGE, not the API — the response is HTML, so the parsed body is empty and the view renders blank with no error
FAIL ui-data-path app/runs/[run_id]/page.tsx:67: api('/runs/${runId}/join') ...
FAIL ui-data-path app/runs/[run_id]/page.tsx:91: api('/runs/${runId}/leave') ...
```

All five, at file:line, each with its shape. Roll 1's real source is the regression fixture in the unit corpus.

## Scope and consequences

- **The audit now fails on this** (exit 1). That is the point — "the app works" is the audit's question, and a dead UI means it doesn't.
- **Cycle verdict semantics are untouched.** This lives in the audit, not the pipeline, so it does not change what a roll's verdict means mid-window — consistent with the reporting-only ruling.
- **Roll 2, in flight, will fail this audit**, because it was authored before #902's guidance fix. That is the honest baseline: it is a clean *window* roll (the window counts mechanical suite deaths) and a broken *app*. Worth deciding whether a roll banks on window-cleanliness alone or also needs a passing audit — flagging rather than assuming.

## Validation

Full regression **7315 passed, 1 skipped**, ruff clean.

### DIFF
--- a/scripts/dev/audit_delivered_app.py
+++ b/scripts/dev/audit_delivered_app.py
@@ -43,6 +43,14 @@ import httpx
 
 from adapters.sandbox.container_backend import ContainerBackend
 from adapters.tools.docker import DockerAdapter
+from squadops.capabilities.ui_data_path import (
+    LIVE_SERVER,
+    SERVED,
+    classify_ui_response,
+    describe_failure,
+    expects_json,
+    extract_ui_calls,
+)
 from squadops.cycles.task_plan import REPAIR_TASK_TYPES
 from squadops.cycles.verification_contract import (
     VerificationContract,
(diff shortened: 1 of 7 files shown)


jasonLadd (commented):
reviewed

PR #904, opened 2026-08-15
docs(sip-0104): §13a — an audit pass gates the window from roll 3

jasonLadd:
Records the owner's ruling of 2026-08-15 in the two places that govern the window: SIP-0104 §13a (permanent) and the implementation plan's Gate 6 protocol (operational).

## The rule

**§10.2's mechanical criterion is unchanged and still spans all six rolls.** Added: **from roll 3, a roll counts only if its delivered app also passes `audit_delivered_app.py`**, which now includes the UI data-path check (#903).

**Rolls 1–2 are grandfathered.** They were authored before the guidance defect was fixed (#902), so holding them to a bar their own inputs made unreachable would measure the fix rather than the rolls.

## Why the mechanical criterion alone proved insufficient

Roll 1 (`cyc_04d36309d793`) satisfied **every** existing signal:

- verdict `accepted`, 36/36 checks executed and passed
- all five contract probes green over real HTTP
- `tests_pass` — eight scaffold shells collected and executed
- `frontend_build` green
- **and the boot audit passing**: *"installs, builds, boots, answers 5 contract probe(s)"*

Its UI was entirely non-functional. All five page data calls requested unprefixed paths (`api('/runs')`) against routes mounted at `/api/runs`, so list, create, detail, join, and leave each failed for every user. **A human opening the app in a browser found it; no automated layer did** — every layer exercises the API directly, none followed the path the UI itself takes.

The root cause was build-side, not verification-side: one fill-only appendix served both stacks, so a `nextjs_ts` author was instructed to fill `backend/routes.py` and `frontend/src/views/*.jsx` (neither exists here) and told its client seam *"prefixes `/api`"* (stack #1's helper). It followed its instructions exactly.

## The claim this protects

The two criteria are recorded separately per roll and **must never be collapsed**:

> six rolls with zero mechanical suite failures; rolls 3–6 additionally audited as working applications.

Reading "six clean rolls" as "six working apps" would be exactly the false-green the amendment exists to prevent — and roll 1 is the proof that the two are genuinely different claims.

Docs only; no code changes.

### DIFF
--- a/docs/plans/deterministic-verification-scaffolding-plan.md
+++ b/docs/plans/deterministic-verification-scaffolding-plan.md
@@ -94,11 +94,12 @@ The N=6 zero-mechanical-death window measures whether reality matches that claim
 - only rolls generated with the final Gate 1–4 behavior count;
 - partial or hand-assembled scaffolds do not count; resumed runs carrying pre-SIP artifacts do not count;
 - a roll that never reaches qa (unrelated failure) neither counts nor resets;
-- a new mechanical suite failure resets the window and names the uncovered surface.
+- a new mechanical suite failure resets the window and names the uncovered surface;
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #906, opened 2026-08-15
feat(scaffold): a baseline stylesheet for the Next.js skeleton (post-window)

jasonLadd:
> **DRAFT — do not merge until the SIP-0104 P6 window closes.** This moves expander output, which moves frozen hashes, which is a deploy change. Mid-window that resets the window.

Delivered apps rendered with browser default styles because **the skeleton shipped no stylesheet at all** — nothing chose that look, it was simply absent. `app/layout.tsx` emitted a bare `<html><body>{children}</body></html>` and there was no CSS anywhere in the tree.

This adds a frozen `app/globals.css`, imported by the frozen layout. Stack #1 is untouched.

## Why not the PRD

The PRD is deliberately behavioral prose that names no implementation detail — SIP-0098 §6.7 stripped the endpoints out and the doc forbids putting them back; styling instructions are the same category. More practically, PRD styling means **the squad authors CSS**: more tokens, a new failure surface, visual variance between rolls, and it becomes part of what is measured. That is a design agent bought by accident.

The scaffold is the right owner. A baseline stylesheet is mechanically determined and frozen, exactly like `lib/errors.ts` and `layout.tsx` already are.

## Why element-scoped, not classes

**This is the whole design, and it is what makes it free.** A class-based sheet is a vocabulary the fill author must be taught, remember, and apply — a prompt surface to maintain, a new way for a roll to fail, and per-roll visual variance. Selecting on elements means the sheet styles whatever markup the author happens to write, with no coordination at all. The author never needs to know it exists.

That is the difference between a styling *capability* and a styling *floor*, and only the floor is free.

Three consequences accepted deliberately, all documented at the constant:

- **Best-effort by construction.** Semantic markup gets styled; a pile of `<div>`s gets the container and type treatment and little else. Fixing that means teaching a vocabulary, which is the trade this refuses.
- **Light only, no `prefers-color-scheme`.** A dark scheme would invert the page ground while any author-written inline color stays light — a half-dark page that looks broken. A predictable floor beats a conditional one.
- **No reset.** Larger surface, no benefit at this scope.

## Verified in the container, not just in unit tests

18 files materialize; `npm install` + `next build` complete clean across all 8 routes; the app boots under `next start`; the page serves the compiled stylesheet (`<link rel="stylesheet" href="/_next/static/css/…">`, contents confirmed). A stylesheet that broke the build would cost a whole roll, so this was checked against the real sandbox image rather than assumed.

## What moved, and what deliberately did not

`expanded_tree_hash` moves — the tree gained a file. **The aggregate spine and scaffold hashes do not**, because the shells are byte-identical: they derive from probes, not from CSS.

That asymmetry is why **GENERATOR_VERSION is unchanged**. The manifest records `expanded_tree_hash` as its own field precisely so a mismatch is attributable — generator drift, workspace mutation and producer edit are three different diagnoses. Stamping a new generator version here would tell every future reader the generator changed when it did not, and invalidate stored manifests' version comparisons for nothing. `test_verification_scaffold_reference`'s docstring now states the distinction, with the unmoved aggregates named as the evidence that separates the two cases.

## Pins

Three, each mutation-verified (mutation → test fails → restore → green):

| Pin | Bug it catches |
|---|---|
| stylesheet emitted **and** layout imports it | emit without import = dead bytes in every delivered app, all tests still green; import without emit = `next build` fails on a missing module |
| never a fill slot | an author-editable stylesheet reintroduces the per-roll variance the deterministic scaffold exists to remove — silently, since author CSS still builds |
| no class selectors | the design invariant; a class selector still compiles and still builds, it just quietly stops styling anything the author did not opt into |

Regression 7318 passed, 1 skipped.

## Prompt delta

The fill-only appendix gains `app/globals.css` in its do-not-edit list (consistency with every other frozen file) plus a short note that presentation is already handled, nudging semantic markup and forbidding added CSS frameworks — an author reaching for Tailwind would have to edit frozen `package.json` and burn a correction round.


jasonLadd:
Superseded by #1458, which re-authors this onto current main.

This branch is 851 commits behind and conflicts in three files including the reference scaffold fixture, so it is not landable as it stands. #1458 carries the design, the rationale and all three pins over intact, and **regenerates the tree hash rather than copying it** — the skeleton has moved since August, so the value here (`dcd4e892…`) is no longer correct. On current main it is `f173df1e…`, with both aggregate hashes verified unmoved, which is the evidence that this is still a tree-only change and still needs no GENERATOR_VERSION bump.

The container verification was reproduced on main: 18 files materialize, `npm install` and `next build` clean across all 8 routes, and the prerendered page carries its stylesheet link.

The 1.7.5 plan places this as the prelude item that must be in or out before deploy A's checkpoint pair — for exactly the reason this draft was held. Closing in favour of #1458.

PR #908, opened 2026-08-15
fix(corrections): a suite failure names which tests failed (#878 full)

### LINKED ISSUE #878: tests_pass signature carries no intra-suite identity — #435 reads any two failing suite rounds as an exact repeat

jasonLadd:
Closes #878.

## What it fixes

SIP-0104 window **roll 2** (`run_6cb3563d3291`) was terminated as a `plan_defect` at correction round 1 **while it was converging**:

- **Round 0** — error-envelope assertions failed; the repair patched the route handlers.
- **Round 1** — five verification-scaffold slots were unfilled; the missing-slot rule surfaced them and the analyzer routed to the fill layer correctly.

Two unrelated defects, being fixed in sequence. Both rounds produced the identical `tests_pass||failed;suite_health=ran`, the repeat detector read an exact repeat, and the run was cut. **The app was fine** — 30 checks executed, 29 passed, 8/9 contract criteria.

#887 shipped the minimum fix (`suite_health`), which separates mechanical from behavioral failures. Both rounds here were `ran`, so it could not separate them.

## The fix

A check row carrying `failing_tests` now contributes **one signature element per failing test** instead of one aggregate element.

That makes the movement table reachable inside a suite for the first time. Fixing three of eight failures is now a strict subset — `MOVEMENT_PROGRESS`, which resets the repeat condition. The aggregate form could not express partial progress at all, which is what made #435 a de-facto two-round cap for `qa.test`.

**The true positive is preserved and strengthened.** shk-4's three identical rounds against one unwinnable task still repeat, now on the stricter identity of *the same tests failed* rather than *the suite failed again*.

Identity is `file::title`, sorted and deduped; a suite-level death (the file died before any test ran) is identified by its file alone. **Messages and line numbers are deliberately excluded** — the standing rule is that evidence text never alters a signature, so a re-described identical failure must still read as a repeat. Absent (pytest emits no machine report, and every legacy row) falls back to the aggregate element **byte-identically**.

The data was already there: SIP-0104 P5 added the vitest reporter rows to `RunTestsResult.test_failures`. This plumbs identity into the signature.

## Also: two seams, one fact

The `tests_pass` failure row was hand-built in two places — the first-pass path and the retest path. #626 added `runner`/`suite_broken` to both by hand; a third field added to only one is a silent per-path behavior difference, and this repo has paid for that class before. Both now route through `failed_tests_pass_row`, with an AST test pinning it and rejecting a reintroduced hand-built row.

## Verification

**Not a replay, and I want to be precise about that**: the field being added did not exist at roll 2's runtime, so its stored evidence cannot carry it. The tests encode roll 2's real failure shape instead.

Three mutations killed:

| Mutation | Result |
|---|---|
| revert to the aggregate element (pre-fix behavior) | 3 tests fail |
| let messages leak into the identity | 3 tests fail |
| one seam hand-builds the row again | 1 test fails |

Regression **7326 passed**, 1 skipped.

## Deploy note

Executor-side only, so this needs a `runtime-api` rebuild. It is intended to ride the **roll-3 rebuild alongside #902** — consistent with #907's "exactly one named change" rule rather than an exception to it: that rule exists to keep *gratuitous* variables out of the counting rolls, and a defect that destroys the rolls being measured is part of making the measurement possible. #906 still waits.

### DIFF
--- a/src/squadops/capabilities/handlers/cycle/qa_test.py
+++ b/src/squadops/capabilities/handlers/cycle/qa_test.py
@@ -715,20 +715,9 @@ class QATestHandler(_CycleTaskHandler):
                 reason = test_result.error or "runner_error"
                 detail = f"tests_not_executed:{reason}"
                 summary = f"Repaired suite not executed: {reason}"
-            checks.append(
-                {
-                    "check": "tests_pass",
-                    "executed": test_result.executed,
-                    "exit_code": test_result.exit_code,
-                    "tests_passed": test_result.tests_passed,
-                    "passed": False,
-                    # #626: runner identity + the runner-neutral suite-health
-                    # verdict, so locus routing stops reading pytest exit
-                    # semantics into vitest failures.
-                    "runner": test_result.runner,
-                    "suite_broken": test_result.suite_broken,
-                }
-            )
+            from squadops.capabilities.handlers.test_runner import failed_tests_pass_row
+
+            checks.append(failed_tests_pass_row(test_result))
             missing.append(detail)
         else:
             summary = "Repaired suite passed"
@@ -1101,20 +1090,9 @@ class QATestHandler(_CycleTaskHandler):
                 detail = f"tests_not_executed:{reason}"
                 fail_note = f"Tests not executed: {reason}"
 
-            validation.checks.append(
-                {
-                    "check": "tests_pass",
-                    "executed": test_result.executed,
-                    "exit_code": test_result.exit_code,
-                    "tests_passed": test_result.tests_passed,
-                    "passed": False,
-                    # #626: runner identity + the runner-neutral suite-health
-                    # verdict, so locus routing stops reading pytest exit
-                    # semantics into vitest failures.
-                    "runner": test_result.runner,
-                    "suite_broken": test_result.suite_broken,
-                }
-            )
+            from squadops.capabilities.handlers.test_runner import failed_tests_pass_row
+
+            validation.checks.append(failed_tests_pass_row(test_result))
             validation.passed = False
             validation.missing_components.append(detail)
             validation.summary = (
--- a/src/squadops/capabilities/handlers/test_runner.py
+++ b/src/squadops/capabilities/handlers/test_runner.py
@@ -743,6 +743,53 @@ async def run_backend_import_check(
 _VITEST_REPORT_FILENAME = ".vitest_report.json"
 
 
+def failing_test_identities(test_failures) -> tuple[str, ...]:
+    """Stable identities for the tests that failed — ``file::title``, sorted and deduped.
+
+    **Identity only: no messages, no line numbers.** The correction signature's standing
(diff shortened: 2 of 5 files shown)


jasonLadd (commented):
reviewed

PR #909, opened 2026-08-15
test(guards): Guard 1a and Guard 1b — authored mode is provenance, not a fork

jasonLadd:
Cut-gate items **3 and 5**. Both halves of Guard 1 were verified **unbuilt** on 2026-08-09; the invariant has held only because nothing had broken it yet. Tests only — no production change.

SIP-0103's architectural claim is that authored mode is a mode of manifest **provenance, not a second pipeline**. Guard 1 defends it in two halves, and the plan is explicit that one is not sufficient for the other.

## Guard 1a — structure

No execution path branches on authoring mode below framing.

- the predicate has **exactly one production call site** across `src/` and `adapters/`
- that call sits **inside a FRAMING-guarded branch**
- the `authors_manifest` flag **never leaves** `task_plan.py`
- the executor's authoring-module vocabulary is **pinned to the four names** it legitimately imports — the gate principal, the artifact type, the question-gate helper, the revision restart — none of which is the predicate

The bug this catches is quiet: a single `if authors_manifest:` in a handler falsifies the release's claim while every functional test stays green, because both modes still *work*. They just stop being one pipeline.

**One thing found while building it.** The plan records this call site as `build_task_plan`; it has since moved into `_resolve_workload_steps`. So the test is keyed on the FRAMING guard rather than an enclosing function name — a name-based assertion would already have rotted once, on its first day.

## Guard 1b — output

Same manifest, same downstream bytes, different provenance. The **reference instance** — the one the 1.4 Functional App Yield window was measured against — is fed through both provenances, and the expanded skeleton, the derived verification contract, and the structural hash must all be identical.

1a cannot cover this. A fork can also arrive as a field that only authored manifests carry leaking into a hash or an emitted file: nothing branches, and the outputs diverge anyway.

`Provenance`'s own docstring already states this invariant. Stating it is not holding it — rendering provenance into an emitted file for traceability is a genuinely tempting change, and adding it to `_canonical` would move every bind-mode cycle's manifest hash and break the contract binding, with authored mode as the only mode that shows it. #783 already had to pin `decisions` out of the projection for exactly this reason.

**Two of the six tests exist so the other four cannot be vacuous:** the reference instance must still carry no provenance (otherwise both sides of every comparison are authored), and a real structural edit must still move the hash (otherwise `_canonical` could be gutted and every equivalence assertion would keep passing).

## Verification

| Mutation | Killed |
|---|---|
| a second call site of the authoring predicate | 2 tests |
| provenance enters `_canonical` | 2 tests |
| provenance rendered into an emitted file | 2 tests |
| the structural projection gutted (`stack` + `persistence` dropped) | 2 tests |

Regression **7336 passed**, 1 skipped. Offline and millisecond-scale — no deploy impact, so this is safe to merge during the roll window.

### DIFF
--- a/tests/unit/cycles/test_authored_mode_equivalence.py
+++ b/tests/unit/cycles/test_authored_mode_equivalence.py
@@ -1,0 +1,128 @@
+"""Guard 1b — same manifest, same downstream bytes, different provenance.
+
+Guard 1a proves nothing *branches* on authoring mode. That is not the same as proving
+the transformation is identical, and the distinction is the whole reason Guard 1 has two
+halves: a fork can also be introduced by a field that only authored manifests carry
+leaking into a hash, an expansion, or a derived contract. Then no code branches and the
+outputs still diverge.
+
+So this is the output half. Feed the **reference manifest** — the same instance the 1.4
+Functional App Yield window was measured against — through both provenances and require
+byte-identical downstream artifacts: the expanded skeleton, the derived verification
+contract, and the manifest's own structural hash.
+
+``Provenance`` already states the invariant in its docstring ("Deliberately NOT part of
+the structural projection … a manifest whose only change is *how it was written* must
+expand to the same skeleton and keep the contract bound to it valid"). Stating an
+invariant is not holding it: the block is five fields, and adding a sixth to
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #914, opened 2026-08-15
fix(scaffold): show the qa author the error envelope, and stop the brief contradicting itself

### LINKED ISSUE #875: TS frozen surface renders consts as bare names — ERROR_STATUS's type and values are invisible, and roll 13's qa author invented both

### LINKED ISSUE #910: The fill-mode brief says 'do not re-test these' over the exact list it then says to fill — every window roll has lost a round to unfilled slots
The fill-mode brief tells the qa author two opposite things about the same eight slots, and every SIP-0104 window roll so far has burned a correction round on unfilled slots.

## The contradiction

`request.qa_test_fill_mode_appendix.md` renders the coverage inventory under this heading:

> **Already covered deterministically (do not re-test these behaviors mechanically):**
> ```
> slot-vc-probe-api-runs — POST /api/runs -> 201 (mirrors probe vc-probe-api-runs)
> slot-vc-probe-api-runs-rejects-blank — POST /api/runs blank required fields -> 400
> ... (8 lines)
> ```

Four paragraphs later, about the **same list**:

> Fill EVERY slot listed above

The inventory does double duty — it is both "what the deterministic layer already covers" and "the slots you must fill" — and the framing that reaches the author first says *do not re-test these*.

## Evidence

| Roll | First-pass fills |
|---|---|
| 1 (`cyc_04d36309d793`) | slots left unfilled; recovered in correction, went green |
| 2 (`cyc_62ce2fd83bca`) | five slots unfilled at round 1 |
| 3 (`cyc_b20f58cc7cbc`) | **all eight** unfilled — "the QA agent left 8 scaffold fill slots unmodified" (round-0 failure analysis) |

The missing-slot rule catches it every time and routes correctly to qa — that machinery works. But catching it costs a correction round per roll, and roll 3 shows the trend going the wrong way.

## Why it is the brief and not the author

Read in order, the instructions are consistent with what the author did: it is told the behaviors are already covered and not to re-test them, so it leaves them alone. Nothing in the intervening text retracts the "already covered" framing before "fill EVERY slot" arrives.

## Fix direction

Separate the two jobs. The slot list the author must fill is not the same statement as "the mechanical layer is done" — the latter belongs in the surrounding prose, the former needs a heading that reads as an instruction. Keep the inventory's content (SIP-0104 §"the semantic brief is coverage inventory only" is unchanged); change what the heading claims about it.

Related: this is the context-completeness class (#787/#866) — an agent judged against a fact it cannot reconcile.

### LINKED ISSUE #911: The error envelope's body shape is invisible to the qa fill author — it has invented the field name in two consecutive window rolls
Half the fill slots in a typical build are error behaviors, and **nothing the qa author is shown states the error response's body shape.** It has invented that shape in two consecutive window rolls, and no downstream layer can tell the invention from an application defect.

## What the author is shown

Three surfaces could carry it. None does:

1. **The fill brief** (`request.qa_test_fill_mode_appendix.md`) — its only body example is a success field:
   ```
   expect(body.id).toBeTruthy()
   ```
2. **The frozen surface index** — renders the error seam as a signature whose return type is opaque:
   ```
   - `lib/errors.ts` — classes `ApiError(code: string, detail: string = '')`; exports ERROR_STATUS; functions `errorResponse(err: unknown): Response`
   ```
   `Response` says nothing about the JSON body.
3. **`ERROR_STATUS`** — renders as a bare name, values invisible (that half is #875).

Meanwhile the scaffold emits, and freezes, exactly this:

```ts
return Response.json({ error: { code, detail } }, { status })
```

## The consequence, measured

Roll 3 (`cyc_b20f58cc7cbc`), four of eight slots are error behaviors. The fills assert `body.error_code`, which is `undefined`. Roll 2 (`cyc_62ce2fd83bca`) produced the same read. It is the third-plus occurrence of the envelope-path class (rolls 13, 17 preceded it).

Note the author's guess is not careless — it is the only move available. The information is not present in anything it is given.

## Why nothing downstream can catch it

An assertion failure inside a fill — `expected undefined to be 'validation_error'` — is **indistinguishable from "the application is wrong"**, because that is precisely what an assertion failure means. Roll 3's analyzer classified it `work_product` and dispatched repairs at the route handlers for three consecutive rounds. No better analyzer would do differently; the information to distinguish the two cases does not exist at that point.

So this is not separately fixable downstream. Making the envelope unguessable is the cure.

## Fix direction

The scaffold **generates** the envelope, so it can state it. The fill brief should carry the response shape verbatim, derived — not prose — alongside a worked error-path example next to the existing success-path one. Cheapest correct place: the brief, since it is already per-stack and already rendered from the manifest.

Deeper fix, separate issue: the shells could assert the error body themselves rather than leaving it to the fill. That is a generator change and moves the byte pins.

Related: #875 (the `ERROR_STATUS` half), #795 (`error_contract.shape` declared and read by nothing — the manifest says `message` where the generator writes `detail`), #877/#902 (the same "the author was never shown the fact" class).

### LINKED ISSUE #912: error_seam_instructions teaches stack #1's Python seam to every stack — a nextjs_ts repair is told about errors.py and ApiError(code, message)
`scaffold.error_seam_instructions()` renders one text for every stack, and that text is stack #1's. A `nextjs_ts` cycle's repair prompts are told about a Python file that does not exist in their workspace.

## What it emits for a nextjs_ts manifest

Measured against roll 3's authored manifest (`cyc_b20f58cc7cbc`):

> on failure raise `ApiError(code, message)` (imported from `.errors`): the FIRST argument is the error-code STRING, the second a human message … the frozen **`errors.py`** maps the code to the HTTP status and renders the pinned envelope

Three things wrong for this stack:

- **`errors.py`** — the file is `lib/errors.ts`.
- **`from .errors`** — the import is `from '@/lib/errors'`.
- **`message`** — the constructor parameter is `detail`: `ApiError(code: string, detail: string = '')`.

## Why it matters

This is the **#902 class**, unfixed in a second location. #902 split the fill-only appendix per stack after roll 1 shipped a dead UI, because one asset was teaching stack #1's client seam to a stack #2 author. `error_seam_instructions` is the same shape of defect on the same stack: a shared prompt surface asserting stack-#1 facts as universal.

Lower blast radius than #902 — this feeds correction/repair prompts (pf-34) rather than first emission — but it is actively wrong at exactly the moment a repair is trying to fix an error-envelope defect, which is what rolls 2 and 3 were both doing.

## Fix direction

Per-stack, the way `DevelopmentCapability.fill_only_template` now resolves (#902), or derived from the stack's own emitted seam so it cannot drift from what the expander actually wrote. The values are already available — the function takes the manifest, and the manifest knows its stack.

Related: #911 (the same envelope, invisible on the authoring side), #902, #875.

jasonLadd:
Closes #910. Closes #911. Closes #875. Closes #912.

Three fixes against one failure, plus a fourth that could not be separated from them.

**The qa fill author cannot see the error response's body shape, so it invents one.** Window rolls 2 and 3 both died asserting `body.error_code` against a frozen `{ error: { code, detail } }`. Roll 3 exhausted its correction budget on it.

## #910 — the brief contradicted itself

The fill brief rendered the slot list under:

> **Already covered deterministically (do not re-test these behaviors mechanically):**

and then, four paragraphs later, about the same eight slots: *"Fill EVERY slot listed above."*

| Roll | First-pass fills |
|---|---|
| 1 | some slots unfilled |
| 2 | five unfilled |
| 3 | **all eight** unfilled |

Read in order, the author was obeying the first instruction. The heading is now an instruction; the coverage claim moved into the prose beneath it.

## #911 — the envelope, stated instead of guessed

Absent from all three surfaces the author could consult: the brief's only body example was `expect(body.id)`, the frozen surface rendered `errorResponse(err: unknown): Response`, and `ERROR_STATUS` was a bare name. Four of eight slots are error behaviors.

The brief now carries the envelope verbatim plus a worked error-path example beside the success-path one, and names the invented form explicitly — #902 established that showing the correct form alone is weaker than showing correct-and-wrong together, because the wrong form is what the model already believes.

**Nothing downstream could have caught this.** An assertion failure inside a fill — `expected undefined to be 'validation_error'` — is *indistinguishable* from "the app is wrong", because that is what an assertion failure means. Roll 3's analyzer classified it `work_product` and sent three consecutive repairs at the route handlers. No better analyzer would differ; the information to separate the two cases does not exist at that point. Making the name unguessable is the only cure.

## #875 — a const's type and values

`exports ERROR_STATUS` told an author the name and nothing else. Roll 13's suite called `.get()` on it — a runtime TypeError, since it is a `Record` and not a `Map` — and invented the statuses. Now rendered as `ERROR_STATUS: Record<string, number> = { validation_error: 400, ... }`; long or computed literals degrade to name-and-type.

## #912 — a consequence, not scope creep

Doing #911 alone would have minted a **second** error-seam prompt surface beside `error_seam_instructions`, which already owns this concern — and which was itself stack-#1-shaped, telling a TypeScript author about `errors.py` and `ApiError(code, message)`. Both surfaces now render from one `ErrorSeam` per stack, so they cannot drift from each other or from the template that emits the bytes.

## Two deliberate design calls

**`body_field` is per-stack, not read from the manifest.** The manifest declares `message` for `nextjs_ts`; the emitted `lib/errors.ts` writes `detail`. #795 owns that reconciliation. Until it is settled, describing what the code *actually emits* is the only honest thing to tell an author — quoting the declared shape would be confidently wrong.

**`framework_exception` is per-stack for the opposite reason.** My first pass generalized "never `HTTPException`" to "never a framework exception" and **lost the name that made the prohibition work** on stack #1 (pf-33/pf-34). A test caught it. Next.js has no equivalent attractor, so the clause is omitted there rather than invented.

## Verification

Context goldens regenerated deliberately; stack #1's only delta is a more concrete import form (`from .errors import ApiError` rather than "imported from `.errors`"), with `HTTPException` preserved. Regression **7346 passed**.

Agent-side only — no generator change, no byte pins moved, no `GENERATOR_VERSION` bump.

**Not fixed here**, and filed: **#913**, the uncovered surface the window named — the shells pin the status but not the response body, so response-field paths are still re-invented in every fill. That is a generator change and is blocked on #795 settling which envelope is authoritative.

### DIFF
--- a/src/squadops/capabilities/handlers/cycle/qa_test.py
+++ b/src/squadops/capabilities/handlers/cycle/qa_test.py
@@ -481,15 +481,25 @@ class QATestHandler(_CycleTaskHandler):
         renderer = getattr(context.ports, "request_renderer", None)
         if renderer is None:
             return ""
+        from squadops.capabilities.scaffold import error_envelope_lines
         from squadops.capabilities.verification_scaffold import VerificationScaffoldManifest
         from squadops.capabilities.verification_scaffold_fill import coverage_inventory_lines
 
         record = VerificationScaffoldManifest.from_dict(manifest_dict)
         slot_lines = "\n".join(f"- {line}" for line in coverage_inventory_lines(record))
         shell_parts = [f"**{f['name']}:**\n```typescript\n{f['content']}\n```" for f in files]
+        # #911: half the slots are error behaviors and nothing showed the author the
+        # envelope, so it invented `body.error_code` on two consecutive window rolls.
+        # Keyed on the scaffold record's own stack — the fact is the stack's, not the
+        # run's, so it stays correct when the brief is rendered without a manifest.
+        envelope = "\n".join(f"- {line}" for line in error_envelope_lines(record.stack))
         rendered = await renderer.render(
             "request.qa_test_fill_mode_appendix",
-            {"slot_lines": slot_lines, "shell_files": "\n\n".join(shell_parts)},
+            {
+                "slot_lines": slot_lines,
+                "shell_files": "\n\n".join(shell_parts),
+                "error_envelope": envelope,
+            },
         )
         return rendered.content
 
--- a/src/squadops/capabilities/scaffold.py
+++ b/src/squadops/capabilities/scaffold.py
@@ -27,7 +27,7 @@ import ast
 import hashlib
 import json
 import re
-from collections.abc import Callable, Mapping
+from collections.abc import Callable, Mapping, Sequence
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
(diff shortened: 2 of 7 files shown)


jasonLadd (commented):
reviewed

PR #917, opened 2026-08-15
fix(authoring): the qa example taught the criterion the gate rejects

### LINKED ISSUE #916: The qa plan-authoring example applies regex_match to a source file — the gate rejects it, and it cost window roll 4
`request.qa_propose_plan_tasks.md` ships a worked example that, followed exactly, produces a plan the framing gate rejects — and it sits **four lines above** the prose forbidding it.

## The example

```yaml
    acceptance_criteria:
      - check: regex_match
        file: backend/tests/test_users.py     # ← a source file
        pattern: "def test_"
        count_min: 5
```

Line 71 of the same file: "`regex_match` for DOCUMENT artifacts only (`.md`/`.txt`/`.rst`)". `request.plan_authoring_rules_appendix.md:37` says it again as **regex-only-on-documents**.

## Measured, not inferred

Lifting the example verbatim out of the shipped asset and running it through the gate's own validator:

```
VERDICT — errors from the SHIPPED worked example: 1
 - Task 0 (Backend user-CRUD pytest suite): regex_match on source file
   'backend/tests/test_users.py' — regex criteria are allowed only on document
   artifacts (.md/.txt/.rst) …
```

## It cost a roll

SIP-0104 window roll 4 (`cyc_92c44f8704ab`) was system-rejected at `progress_plan_review` after 67 minutes of framing, class `validate_criteria_scope`, with the **same error on the same shape** — `regex_match` on `__tests__/runs.test.ts` to count test functions. The author reproduced the pattern it was shown.

## Third instance of the class

Roll 16 (`cyc_99efd51aedc5`) was rejected because the sole-author builder example taught a Dockerfile-less `expected_artifacts` while the prose rule said otherwise — fixed in #891, whose finding was recorded as *patterns beat prose*. This is the same defect in a different asset, and the V4 qualifier made the same point a third time.

Prose stating a rule does not survive an example demonstrating its violation. The examples are what get copied.

## Fix

Two parts, and the second is the one that matters:

1. Correct the example. The rejection message already names the right check for this exact intent — `function_defined` (#464) — so the example should use it.
2. **Validate every shipped worked example through the gate's own validators, as a test.** A sweep of all authoring assets found exactly one offender today; nothing stops the next one. An example that would be rejected in production is a defect detectable offline in milliseconds, and it is the only durable fix — correcting this instance leaves the class open.

Related: #891/#890 (same class, builder asset), #464 (`function_defined`), #645 (the named check menu — the structural fix this is a symptom of).

jasonLadd:
Closes #916.

**Window roll 4 was system-rejected because it copied a shipped example.**

`cyc_92c44f8704ab`, rejected at `progress_plan_review` after 67 minutes of framing, class `validate_criteria_scope`: the plan applied `regex_match` to `__tests__/runs.test.ts`, a source file.

`request.qa_propose_plan_tasks.md` ships this:

```yaml
    acceptance_criteria:
      - check: regex_match
        file: backend/tests/test_users.py     # ← a source file
        pattern: "def test_"
```

Four lines below it, the same file says "`regex_match` for DOCUMENT artifacts only". The rules appendix says it again as **regex-only-on-documents**.

**Measured, not inferred.** Lifting the example verbatim from the shipped asset and running the gate's own validator:

```
VERDICT — errors from the SHIPPED worked example: 1
 - Task 0 (Backend user-CRUD pytest suite): regex_match on source file
   'backend/tests/test_users.py' — regex criteria are allowed only on document artifacts …
```

The same error, on the same shape, that killed the roll.

## Third instance of the class

Roll 16 was rejected because the builder example taught a Dockerfile-less `expected_artifacts` while the prose rule said otherwise (#890/#891, whose finding was recorded as *patterns beat prose*). V4 made the point a third time. **Prose stating a rule does not survive an example demonstrating its violation** — the examples are what get copied.

## The fix, and the durable half

The example now uses `function_defined` — what the rejection message itself recommends for this intent (#464), and what that check's own `CheckSpec.example` demonstrates.

**Correcting this instance leaves the class open**, so the real fix is the test: every worked example in every authoring asset is lifted and run through the gate's **real** validators. An example production would reject is a defect decidable offline in milliseconds.

Deliberately calls the validators rather than restating their rules. A test that reimplemented "regex-is-document-only" would drift from the gate and begin certifying examples the gate rejects — the exact failure being prevented.

## Two defects it caught during its own construction

Both kept as test cases, because both are the general shape rather than one-offs:

- **My first replacement used `name`/`count_min`**, where `function_defined` declares `name_prefix`/`min_count`. A plausible-looking example with wrong parameter names teaches an unsatisfiable criterion just as effectively as a forbidden one — so the test now checks every example criterion's parameters against its `CheckSpec`.
- **The test's own first version was partly vacuous.** It built the plan and inspected *its* tasks, with `except ValueError: continue` for blocks that are not full plans — which silently swallowed an example naming a nonexistent check, since that is precisely what makes plan construction raise. Two of three mutations survived. Raw extraction is now separate from plan construction and cannot be defeated that way.

## Window impact

Roll 4 never reached qa, so per the protocol it **neither counts nor resets**. The window still stands at roll 1 banked, rolls 2 and 3 lost.

Also worth noting from this roll: it produced the **first `rejection_record` artifact ever generated in the field**, which closes a gap flagged earlier — that machinery was built and unit-tested but had never fired in a real cycle.

Three mutations killed. Regression **7349**. Agent-side asset plus a test; no generator change, no pins moved.

### DIFF
--- a/src/squadops/prompts/request_templates/request.qa_propose_plan_tasks.md
+++ b/src/squadops/prompts/request_templates/request.qa_propose_plan_tasks.md
@@ -59,10 +59,10 @@ tasks:
     expected_artifacts:
       - "backend/tests/test_users.py"
     acceptance_criteria:
-      - check: regex_match
+      - check: function_defined
         file: backend/tests/test_users.py
-        pattern: "def test_"
-        count_min: 5
+        name_prefix: "test_"
+        min_count: 5
         severity: error
     depends_on_focus:
       - "dev:user crud routes"
--- a/tests/unit/prompts/test_authoring_examples_pass_the_gate.py
+++ b/tests/unit/prompts/test_authoring_examples_pass_the_gate.py
@@ -1,0 +1,166 @@
+"""Every worked example in an authoring asset must survive the gate it teaches against.
+
+Prose stating a rule does not survive an example demonstrating its violation — the
+examples are what get copied. That has now cost two window rolls:
+
+* roll 16 (`cyc_99efd51aedc5`) — the sole-author builder example carried a Dockerfile-less
+  ``expected_artifacts`` while the prose rule said otherwise, and the plan reproduced the
+  example twice in a row (#890/#891, whose finding was recorded as *patterns beat prose*);
+* roll 4 (`cyc_92c44f8704ab`) — the qa proposer example applied ``regex_match`` to
+  ``backend/tests/test_users.py``, a source file, **four lines above** the prose saying
+  document-only. The plan reproduced the shape against ``__tests__/runs.test.ts`` and was
+  system-rejected at ``progress_plan_review`` after 67 minutes of framing (#916).
+
+Correcting each instance leaves the class open. This closes it: an example that production
+would reject is a defect decidable offline in milliseconds, so it is decided here instead
+of in a cycle.
+
+Deliberately runs the **real** validators rather than re-stating their rules — a test that
+reimplemented "regex is document-only" would drift from the gate and start certifying
+examples the gate rejects, which is precisely the failure being prevented.
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+import pytest
+import yaml
+
+from squadops.cycles.acceptance_check_spec import CHECK_SPECS
+from squadops.cycles.implementation_plan import ImplementationPlan
+
+pytestmark = [pytest.mark.domain_contracts]
+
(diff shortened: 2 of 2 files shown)


jasonLadd (commented):
reviewed

PR #919, opened 2026-08-15
fix(authoring): the sole-author path had a private, drifted copy of the check vocabulary

### LINKED ISSUE #918: The sole-author plan path renders a private, drifted copy of the check vocabulary — it taught the criterion that cost roll 4
`_plan_authoring_service` renders the typed-acceptance vocabulary from a **hand-written block inlined in Python**, not from `CHECK_SPECS`. That copy had drifted from the registry the gate reads.

## The drifted block

```
# regex_match — pattern present count_min times in a file (stack-agnostic)
- check: regex_match
  description: "At least three test functions exist"
  file: tests/test_users.py
  pattern: 'def test_\w+'
  count_min: 3
```

Three defects in five lines:

- **"stack-agnostic"** — the gate rejects `regex_match` on source files regardless of stack.
- **`tests/test_users.py`** — a source file, the exact shape `validate_criteria_scope` refuses.
- **the document-only note is missing.** `CHECK_SPECS["regex_match"].notes` says *"Documents only (.md/.txt/.rst) — a regex on a SOURCE file is rejected at plan validation (#464)"*, and `render_typed_acceptance_vocabulary()` emits it. This copy does not.

## It cost window roll 4

`cyc_92c44f8704ab`, system-rejected at `progress_plan_review` after 67 minutes of framing. What it authored:

```yaml
- check: regex_match
  description: At least five test functions exist
  file: __tests__/runs.test.ts
```

*"At least three test functions exist"* → *"At least five test functions exist"*. The description is near-verbatim.

## Why it survived this long

The **proposers** have rendered the derived vocabulary since #182 — `request.development_propose_plan_tasks` and `request.qa_propose_plan_tasks` both interpolate `{{typed_acceptance_vocabulary}}`. The **sole-author path** does not, and it is the path that writes the plan on every CRP except `validation-multirole` — every roll of the SIP-0104 window included.

This is the same class as **#856**, on the same path: *"the proposers and the brief author have rendered these since #686; this path did not, so the author that actually writes the plan on every CRP was the one never shown them."* #856 fixed it for the authoring rules. The vocabulary stayed duplicated.

Note also that **#916/#917 fixed this defect in the proposer asset** — a real fix, but not one these rolls exercise.

## Fix

Render `render_typed_acceptance_vocabulary()`. Keep the surrounding prose that is genuinely additional (the `command_exit_zero` safety rules, the behavioral > structural > textual hierarchy).

Guard it two ways: pin that the path renders rather than restates, and parse every example out of the rendered vocabulary and run it through the gate's own validator — the vocabulary is what a sole author copies from, so an example the gate refuses is a guaranteed rejection shipped to every cycle.

Related: #916/#917 (same defect, proposer asset), #856 (same class, same path, authoring rules), #182 (the derived renderer), #707 (the safelist, already sourced rather than restated for this reason).

jasonLadd:
Closes #918.

**The plan author on every window roll was reading a private, drifted copy of the check vocabulary.**

`_plan_authoring_service` inlined its own hand-written vocabulary block instead of rendering `CHECK_SPECS`. It had drifted from the registry the gate reads:

```
# regex_match — pattern present count_min times in a file (stack-agnostic)
- check: regex_match
  description: "At least three test functions exist"
  file: tests/test_users.py
```

Three defects in five lines: `regex_match` is not stack-agnostic (the gate rejects it on source, on every stack); the example targets a **source file**, the exact shape `validate_criteria_scope` refuses; and it omits the document-only note that `CHECK_SPECS` carries and the derived renderer emits.

**Window roll 4 copied it near-verbatim** — *"At least three test functions exist"* became *"At least five test functions exist"* against `__tests__/runs.test.ts` — and was system-rejected at the framing gate after 67 minutes.

## Why it survived

The **proposers** have rendered the derived vocabulary since #182. The **sole-author path** never did — and it writes the plan on every CRP except `validation-multirole`, which is every roll of this window.

That is the same class as **#856**, on this same file, whose own comment reads: *"the proposers and the brief author have rendered these since #686; this path did not, so the author that actually writes the plan on every CRP was the one never shown them."* #856 fixed it for the authoring **rules**. The **vocabulary** stayed duplicated.

## Correction on #917

#917 fixed this same defect in `request.qa_propose_plan_tasks` — a **proposer** asset. That was a real defect and the fix stands, but these rolls are sole-author and never render it, so **#917 alone would not have changed roll 5.** This is the copy they actually read. I would have launched roll 5 on an unfixed path.

## The fix

Render `render_typed_acceptance_vocabulary()`. The surrounding prose that is genuinely additional is kept — the `command_exit_zero` safety rules and the behavioral > structural > textual hierarchy.

Two tests, both mutation-verified:

| Guard | Mutation that kills it |
|---|---|
| the path renders rather than restates, keyed on the drifted strings | re-introducing the hand-written block |
| every example in the rendered vocabulary survives the gate's own validator | pointing `regex_match`'s own `CheckSpec.example` at a source file |

**The #917 sweep could not have caught this** — it walks `.md` assets and this block is Python. That gap is now closed from the other side.

Regression **7351**. Agent-side plus tests; no generator change, no pins moved.

### DIFF
--- a/src/squadops/capabilities/handlers/_plan_authoring_service.py
+++ b/src/squadops/capabilities/handlers/_plan_authoring_service.py
@@ -30,7 +30,10 @@ from squadops.capabilities.handlers.cycle_tasks import (
     _rewrite_manifest_identifiers,
 )
 from squadops.capabilities.handlers.fenced_parser import extract_fenced_files
-from squadops.cycles.acceptance_check_spec import command_safelist_names
+from squadops.cycles.acceptance_check_spec import (
+    command_safelist_names,
+    render_typed_acceptance_vocabulary,
+)
 from squadops.cycles.implementation_plan import (
     ImplementationPlan,
     planner_build_task_types,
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #920, opened 2026-08-15
docs: state the window's final claim by rule, not by roll number (SIP-0104 §13b)

jasonLadd:
Docs only. Corrects a derived roll count that three lost rolls invalidated.

## The problem

SIP-0104 §13a's closing sentence gives the required form of the final evidence:

> *six rolls with zero mechanical suite failures; rolls 3–6 additionally audited as working applications.*

The **rule** it derives from — stated correctly two paragraphs above, and unchanged by this PR — is *"from window roll 3 onward, a roll counts only if the delivered app also passes the audit."* The phrase "rolls 3–6" was arithmetic on the assumption that rolls 2 through 6 would each count.

They did not:

| Roll | Outcome | Counts? |
|---|---|---|
| 1 | banked clean | yes |
| 2 `cyc_62ce2fd83bca` | killed by the correction-signature defect (#878) | no |
| 3 `cyc_b20f58cc7cbc` | correction budget exhausted | no |
| 4 `cyc_92c44f8704ab` | system-rejected at the framing gate, never reached qa | neither counts nor resets |

The window's **sixth counting roll is not the sixth roll launched**, and any fixed numbering in the claim keeps drifting as rolls are lost.

## The correction

> *six counting rolls with zero mechanical suite failures; every counting roll from roll 3 onward additionally audited as a working application.*

**Nothing about the criterion changes.** §10.2's mechanical claim still spans all six counting rolls, the audit gate still begins at roll 3, and the two are still recorded separately and never collapsed. What changes is that the claim stops naming roll numbers it cannot guarantee.

## Why a new section rather than an edit in place

§13a's sentence is the wording the acceptance evidence will be **quoted from**. A silent correction to a quotable claim is indistinguishable from the claim having always said this, so it lands as **§13b** with what changed, the evidence, and the ruling — the shape CLAUDE.md prescribes. `updated_at` untouched; the amendment's own date is the content record.

## Also

**Stage 1e** moves from "credited at roll 3" to "credited at the first roll that passes the delivered-app audit". Roll 3 was the first roll the gate *applied to* — never a guarantee that it would pass, and it did not. Same substance, stated as a rule rather than a number.

Both plans updated to match. Regression **7351**.

### DIFF
--- a/docs/plans/1-6-0-authorship-plan.md
+++ b/docs/plans/1-6-0-authorship-plan.md
@@ -229,9 +229,12 @@ was never in the 1a sweep) · and can **M6's taxonomy name the subsystem** if it
 **Nothing is yet known about whether the Next.js stack works.** The 2026-08-10 run never
 reached expansion for it.
 
-**1e is credited at SIP-0104 P6 roll 3 (owner ruling 2026-08-15).** The P6 window's rolls
-*are* authored-mode `nextjs_ts` cycles on this PRD, so 1e needs no separate roll — but which
-one closes it is a judgment, and this is the ruling.
+**1e is credited at the first SIP-0104 P6 roll that passes the delivered-app audit
+(owner ruling 2026-08-15).** The P6 window's rolls *are* authored-mode `nextjs_ts` cycles on
+this PRD, so 1e needs no separate roll — but which one closes it is a judgment, and this is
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #921, opened 2026-08-15
docs(sip-cba): duty-shaped steward, versioning question, and the pack configuration lifecycle

jasonLadd:
Docs only. Records an owner design discussion into `SIP-Capability-Backed-Agents`, in the §21 form so the implementation SIPs inherit decisions rather than reconstruct them.

The SIP already split *"Iris applies / Glyph stewards"* (§8, §13) and already stated that capability activation is orthogonal to `RuntimeMode` (§14). **What it never stated was each agent's default posture** — and that is the thing that decides what has to be built.

## §25a — Iris is cycle-bound, Glyph is duty-shaped

Iris is activated by a cycle assignment and needs no machinery this platform lacks; it could be a roster member the way Bob is. Glyph's work is not a cycle and does not decompose into one.

**A correction is recorded rather than quietly dropped.** My first framing in discussion argued that "maintain the design system" was an aspiration rather than a duty, on the grounds that a duty needs an external trigger. That was wrong on the axis. Stewardship is a real discipline with real non-cycle work — industry research, adding capability, incorporating gap feedback, exploring new surfaces, usability testing — and a time-boxed duty window is its natural container.

The correct requirement is not a trigger, it is a **unit of output**:

> A Glyph duty window ends with a published design-system version, or nothing published this window.

That is what makes the duty falsifiable. A window that produces prose churn and no version has produced nothing, and says so.

## §25b — the question this forces, and it blocks the pack's schema

When Glyph publishes v1.4 and three delivered projects were built against v1.3, what happens?

| Model | What the design system becomes |
|---|---|
| Pinned library | a dependency, with an upgrade cost and a migration story |
| Additive-only | a growing vocabulary that can never retire a mistake |
| Advisory style guide | documentation, not a contract |

§13's `canonical` vs `project-local` governance **implies pinned-library** — "project-local" only means something if canonical is a version a project can be behind — but the SIP has never said so. Filed as open question 14, deliberately distinct from Q13: that one is about *capability* versions, this is about the versioned *resource* a capability produces.

Raised, not settled.

## §25c — reconciling "an agent plugin with a default identity"

Taken literally that contradicts Product Decision 2 and §2's warning that a pack owning identities ends reusability. Proposed for review: a pack may publish a suggested **default binding** — name, persona, expected capability set — while the roster remains sole authority to adopt, rename, rebind or ignore it.

**The acceptance test that decides which it actually is**, proposed for §18: install the design pack and bind `design-system-application` to an **existing** agent, with no Iris in the roster. If that works, the default identity is a convenience. If not, the pack owns the identity.

## §25d — usability testing has no substrate

Research, gap incorporation and surface exploration all execute against resources. Usability testing needs users or a credible proxy and the platform has neither. Left undecided it becomes an agent producing plausible findings nobody validated — the exact false-green shape SIP-0096 and SIP-0104 exist to prevent, relocated into design. Decide in scope with a named proxy, delegated to a human, or out of scope.

## §25e — Glyph is the forcing function for SIP-0091

Duty durability has zero code and sits in the capacity pool because nothing has pushed on it. An agent whose entire value is bounded self-directed windows producing versioned output pushes harder than anything currently queued against it — which argues for pulling 0091 forward on that evidence rather than treating Glyph as blocked behind it. Iris has no such dependency and is the cheaper first proof.

## §25f — the near-term rung already exists

PR #906's frozen, element-scoped stylesheet is this SIP's thesis at its smallest: the framework owns the deterministic spine, the model fills judgment, and it needs zero cooperation from any agent. The path is versioned resource → Iris selects → Glyph proposes, each rung testable before the next — and the first rung answers §25b empirically rather than by argument.

Cross-referenced from §13, §20, §22 and §23. Regression **7351**.

---

# §26 — Pack lifecycle, attribution, and configuration

Second commit, same discussion. §21 settled the *substrate*; §26 settles how an installed pack becomes a **configured, attributed** one — derived from §21's "declarations are data the host validates without executing pack code" rather than beside it.

**§26a — three events, not one.** Install / configure / activate. Install materializes a commented configuration **stub** from the declared schema, the way `bootstrap` materializes a profile contract. It never produces a binding: Product Decision 4 already holds that install ≠ authority.

**§26b — one artifact, three editors.** File, CLI and console pane all write the same store. Three independent write paths is the two-seams-one-fact defect this project keeps paying for — #856 and #918 were both a second hand-maintained copy of something already derived, and pack config is exactly the sort of thing that would sprout a console-side copy.

**§26c — attribution is stated, never inferred:** **adopt** (bind to an existing agent — Neo gains `architecture-review`), **mint** (instantiate the suggested default binding under a name the importer chooses), or **ignore**. Roster stays sole authority in all three.

**§26d — packs declare model *requirements*, the host resolves.** A pack naming a concrete model writes a deploy-specific fact into a distribution unit. Advances open question 12 and buys a preflight check: *"this pack requires vision, no pulled model provides it"* at configure time rather than mid-cycle.

**§26e — secrets by name, never by value.** A tool URL is configuration; a key is not. The pack declares `requires_secrets: [figma_token]`, the operator binds it to an existing provider entry, and the pack reaches the tool only through the mediated invoker. Config artifacts holding key material would mint a second secrets path beside the one that exists and degrade §21's permission scoping from structural to advisory.

**§26f — the objection this section exists to answer:** *won't a domain-specific pack need its own verbs?*

"Verb" is covering two things. **Configuration is generic while its parameters are domain-specific** — the host needs types, constraints and secret-ness, not meaning. `CHECK_SPECS` already proves that pattern one layer down: plan validation rejects a malformed criterion with no semantic knowledge of what `name_prefix` does, and the authoring-example guard added today validates parameters across checks it knows nothing about.

**Domain *actions* are capabilities, not CLI verbs.** "Sync tokens from Figma" through a capability gets mediated SPI, typed evidence, permission scoping and a ledger entry; through a CLI verb it gets none of them.

The genuine edge case — interactive one-time setup like pasting a vendor key — gets a **declarative setup form** the host renders from the pack's declaration. The line is not *"may pack code run"* (§21 says it may), it is *"may it run before the host has validated anything, at CLI-parse time, in the operator's shell."*

**§26g** names what this costs that does not exist yet — the console write path is the largest item and the one deferrable without blocking the rest. **§26h** raises two open questions: where the configuration artifact lives, and whether a pack's suggested default binding is versioned with the pack.

Product Decisions 13–15 added. Regression **7351**.

### DIFF
--- a/sips/proposed/SIP-Capability-Backed-Agents.md
+++ b/sips/proposed/SIP-Capability-Backed-Agents.md
@@ -231,6 +231,7 @@ Umbrella phases; each becomes its own bounded implementation SIP. Per the roadma
 ## 20. Open Questions
 
 1. Storage layout for the workspace — logical projection over the artifact vault, DB-backed, filesystem-like, or hybrid?
+   *Related but distinct: where the PACK CONFIGURATION artifact lives is §26h Q15.*
 2. Manifest format — YAML/TOML/Python entry points/registry?
 3. Plugin discovery — Python packaging, Switchboard, config, or a dedicated registry?
    *Resolved by owner (2026-07-29): **Switchboard is the presumptive substrate** — §21.*
@@ -243,6 +244,8 @@ Umbrella phases; each becomes its own bounded implementation SIP. Per the roadma
 10. Architecture pack — part of this umbrella or a follow-on SIP?
 11. Which native capabilities migrate first after the Design pack?
 12. Do binding contracts declare required model class/context, or is model fit advisory?
+    *Advanced (2026-08-15): **requirements are data, model choice is host resolution** — §26d.
+    Buys a preflight check; the override grain in the configuration artifact remains open.*
 13. Version pinning — exact capability versions or semver ranges in roster bindings?
     *Advanced (2026-07-29): layered scheme with loud load-time enforcement — §21; the
     roster-binding pinning grain (exact vs range) remains open.*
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

Issue #922, opened 2026-08-16
1.7: three meanings of "capability" — rename the two host-internal ones before packs freeze the word


jasonLadd:
"Capability" names three different things in this codebase and the proposed capability-packs model needs a fourth meaning that is really the third one. Fixing it after packs publish against the current name is not possible — the word would be frozen in a distribution format.

## The three meanings

| Where | What it actually is | Evidence |
|---|---|---|
| `CapabilityContract.capability_id` (`capabilities/models.py:289`) | **the task type** — inputs, outputs, artifacts, acceptance checks, owner roles, trigger, timeout | `capabilities/runner.py:259` sets `task_type=contract.capability_id`; `HandlerRegistry` is keyed on it |
| `DEV_CAPABILITIES` / `dev_capability` (`capabilities/dev_capabilities.py`) | **a named settings bundle for a stack** — prompt text, expected extensions, source filter, test framework, fill-only template | selected per cycle via `--set dev_capability=nextjs_ts`; also `ScaffoldStack.dev_capability` |
| `SIP-Capability-Backed-Agents` | **bindable agent competence** — `design-system-application`, published by packs, bound by the roster | §8 binding contracts, §13 the design pack |

The SIP is careful that *"Neo (identity) ≠ architecture-review (capability)"* — but `HandlerRegistry.get(capability_id)` already exists and **looks like the seam the SIP wants while being the task-type registry**. That is the shape of collision found by someone wiring a pack into the wrong registry and having it half-work.

## Which meaning keeps the word

**The SIP's.** Externally-facing vocabulary wins the contested term: packs *publish* capabilities, operators *bind* capabilities, and it is the word in the distribution format. The other two are host-internal names that leaked outward.

## Proposed renames

**1. `CapabilityContract` / `capability_id` → task contract / task type.** Internal only — `models.py`, `runner.py`, `handler_registry.py`, `handler_executor.py`, `exceptions.py`, handlers. Mechanically wide but semantically free: the concept has *already* been collapsed (`task_type=contract.capability_id`); only the name lags.

**2. `DEV_CAPABILITIES` / `dev_capability` → development profile** — *or collapse it entirely, see the decision below.* "Profile" is this repo's consistent word for a named settings bundle selected by id (squad profile, request profile, build profile, probe profile), so this conforms to convention rather than adding to the pile.

## The one decision inside this

[...]

PR #923, opened 2026-08-16
docs(roadmap): #922 joins 1.7's vocabulary-leak pool

jasonLadd:
Docs only. Adds **#922** to 1.7's boundary/vocabulary-leaks pool, beside #559's residual `task_type ==` sites — same axis.

## The item

"Capability" names three different things:

| Where | What it is |
|---|---|
| `CapabilityContract.capability_id` | **the task type** — `runner.py:259` literally sets `task_type=contract.capability_id` |
| `DEV_CAPABILITIES` / `dev_capability` | a named settings bundle for a stack |
| `SIP-Capability-Backed-Agents` | bindable agent competence, published by packs |

`HandlerRegistry.get(capability_id)` already exists and **looks like the seam the packs SIP wants while being the task-type registry** — the shape of collision found by wiring a pack into the wrong registry and having it half-work.

## Why pool-listed rather than left as a loose issue

The sequencing constraint. It must land **before capability packs publish declarations against `capability_id`**, because at that point the word is frozen in a distribution format and all three meanings become permanent. An issue mentioning 1.7 doesn't carry that; the pool entry does.

It also fits the lane precisely: 1.7's named claim is *"every port is actually a port"* and its pool is explicitly boundary and vocabulary leaks. This is the vocabulary half, and a feature-free release is the right home for a wide mechanical rename.

## What #922 leaves open

One decision, called out in the issue rather than pre-empted here: whether `dev_capability` is **renamed** to `development_profile` or **collapsed** into the stack entirely. `ScaffoldStack.dev_capability`'s own docstring records that a cycle declares its stack twice — the same literal on adjacent CRP lines, bound by convention alone (#832). Collapse removes a word instead of replacing one, and is the larger change. Either way it is a user-visible config break, since `--set dev_capability=…` is a required override on `validated-fullstack`.

Regression **7351**.


jasonLadd (commented):
reviewed

PR #925, opened 2026-08-16
fix(llm): stop paying for a reasoning channel the adapter discards, and record what was emitted

### LINKED ISSUE #924: The qa author's reasoning channel eats the completion budget and is discarded unread — and no emission is recorded to prove it
Two failures, one root: the qa author's reasoning channel consumes the completion budget and is then **discarded unread**, and nothing records what the author emitted — so the failure can only be diagnosed by guessing.

## The measurement

Same model (`qwen3.6:27b`), same deployed fill-mode brief, same eight SIP-0104 scaffold slots:

| | reasoning on | reasoning off |
|---|---|---|
| completion tokens | **5,727** | **413** |
| response | 1,411 chars | 1,417 chars |
| fill fences emitted | 8 | **8** |
| envelope path | `body.error.code` ✓ | `body.error.code` ✓ |

**13.9× the tokens for the same output** — and the no-reasoning run additionally asserted a field the reasoning run missed.

At a 2,500-token budget the model produced **19,179 characters of thinking and an empty response**. Its reasoning showed it correctly enumerating all eight slots; it simply never reached the emission.

## Why the tokens are wasted rather than merely spent

`adapters/llm/ollama.py` says so in its own docstring:

> `thinking_tokens` is False deliberately… qwen3-family models emit `message.thinking` through Ollama by default, but this adapter never sends a `think` parameter and **reads only `message.content`**.

So the channel is billed against `num_predict` and thrown away. The adapter declares `THINKING_TOKENS: False` — it cannot use them.

## The observed production shape

SIP-0104 window rolls 3 and 5 both ended with **all eight scaffold slots unfilled**. Roll 5 ran `completion_tokens=6866` against an **8,192 clamp** for `qwen3.6:27b`, on a prompt of 14,832 tokens — 3× the probe's, and a larger prompt produces more reasoning, not less.

## The second half, which is why this took hours

**Nothing records what the author emitted.** Fills are parsed and stripped before extraction, so a successful fill leaves no artifact and a failed one leaves no trace at all. Three diagnoses were made from the *result* rather than the emission across rolls 3 and 5, and two were wrong — the third only landed by reproducing the raw completion by hand against the live model.

A protocol whose failures cannot be inspected can only be repaired by guessing.

## Fix

1. **A per-call `thinking` parameter on the LLM port**, mapped to Ollama's `think`, opted into by the structured-emission handlers (`qa.test`, `qa.test_repair`) whose output is fenced blocks rather than an argument. Deliberately **not** adapter-wide: one task was measured, and reasoning plausibly earns its cost where the output *is* an argument — manifest authoring, plan authoring, failure analysis.
2. **An emission-shape log at the completion seam** — length, fence counts by kind, bounded head sample. Unconditional, and outside the observability gate so it does not go silent exactly where it is needed. Enough to separate "emitted nothing", "emitted the wrong fence" and "emitted fills the parser rejected" at a glance.

The second is what makes the first verifiable: if the next roll logs `chars=0 completion_tokens≈6866`, the diagnosis is confirmed outright; if it logs fills, the cause is elsewhere and the data says where.

Related: #910/#911/#914 (the two earlier, insufficient diagnoses of the same symptom), #908 (correctly terminated roll 5 as a true repeat), SIP-0104 §10.2.

jasonLadd:
Closes #924.

**Measured, not inferred.** Same model, same deployed brief, same eight scaffold slots:

| | reasoning on | reasoning off |
|---|---|---|
| completion tokens | **5,727** | **413** |
| response | 1,411 chars | 1,417 chars |
| fill fences | 8 | **8** |
| envelope path | `body.error.code` ✓ | `body.error.code` ✓ |

13.9× the tokens for the same output — and the no-reasoning run additionally asserted a field the reasoning run missed. At a 2,500-token budget the model produced **19,179 characters of thinking and an empty response**: its reasoning correctly enumerated all eight slots and it never reached the emission.

The waste is documented in the adapter's own docstring — qwen3-family models emit `message.thinking`, this adapter never sends `think` and reads only `message.content`, and it declares `THINKING_TOKENS: False`. Billed against `num_predict`, discarded unread.

Window rolls 3 and 5 both ended with all eight slots unfilled. Roll 5 ran `completion_tokens=6866` against an **8,192 clamp** on a 14,832-token prompt — 3× the probe's, and a larger prompt produces more reasoning, not less.

## Part 2 is why this took hours

**Nothing recorded what the author emitted.** Fills are parsed and stripped before extraction, so a success leaves no artifact and a failure leaves no trace. Three diagnoses were made from the *result* rather than the emission across rolls 3 and 5 — #910, #911, #914 — and **two of them were wrong**. The third only landed by reproducing the raw completion by hand against the live model.

A protocol whose failures cannot be inspected can only be repaired by guessing. That is the actual defect behind the last two days.

## What ships

**A per-call `thinking` parameter** on the LLM port → Ollama's `think`, threaded through the router, accepted-and-ignored by vLLM (the port states it is a request, not a guarantee; dropping it from the signature would turn a caller's keyword into a runtime `TypeError`). Opted into by `qa.test` and `qa.test_repair` only.

**Deliberately not adapter-wide.** One task was measured. Reasoning plausibly earns its cost where the output *is* an argument — manifest authoring, plan authoring, failure analysis — and a silent quality regression there would cost more than the tokens saved.

**An emission-shape log** at the completion seam: length, fence counts by kind, bounded head sample. Unconditional, and **outside the observability gate** — that block is guarded on `llm_obs and correlation_context`, so a capture placed there would go silent in exactly the deployments where an unexplained emission is hardest to diagnose. I made that placement mistake while writing this; the wiring test now kills it.

The three diagnoses it separates, which were previously indistinguishable:

```
emission shape: chars=58   completion_tokens=413  fences={'fill': 2, ...}
emission shape: chars=0    completion_tokens=6866 fences={'fill': 0, ...}   ← the failure
emission shape: chars=47   completion_tokens=900  fences={'path': 1, 'fill': 0}
```

## Verification

**Five mutations killed**: adapter drops the parameter · router swallows it · qa handler opts out · **capture unwired** · **capture moved inside the observability gate**. The last two matter most — the first version of these tests exercised the helper as a pure function, so deleting its call site passed everything.

The repo's test-quality lint also caught one of my tests asserting class attributes without exercising behavior. Rewritten to assert the kwargs each handler actually builds, which is the stronger property: a flag set on a handler whose kwargs never carry it is the inert-fix shape.

Regression **7362**.

## What this does not claim

That the reasoning budget is *proven* to be roll 5's cause. It is strongly supported and it is measured waste regardless — but the emission itself was never recorded, so it cannot be confirmed retrospectively. Part 2 is what makes the next roll decisive: `chars=0` with a full token count confirms it outright, and anything else names the real cause immediately instead of prompting a fourth guess.

### DIFF
--- a/adapters/llm/ollama.py
+++ b/adapters/llm/ollama.py
@@ -177,13 +177,24 @@ class OllamaAdapter(LLMPort):
         temperature: float | None,
         *,
         stream: bool = False,
+        thinking: bool | None = None,
     ) -> dict[str, Any]:
-        """Build the Ollama /api/chat request payload."""
+        """Build the Ollama /api/chat request payload.
+
+        ``thinking=False`` sends Ollama's ``think: false``. This adapter reads only
+        ``message.content`` and declares ``THINKING_TOKENS: False``, so a reasoning
+        channel it cannot read is billed against ``num_predict`` and then discarded —
+        measured at 5,727 tokens versus 413 for identical output on a scaffold-fill
+        emission (#924). ``None`` sends nothing and leaves the model's default alone,
+        so every existing call is byte-identical.
+        """
         payload: dict[str, Any] = {
             "model": model,
             "messages": [{"role": m.role, "content": m.content} for m in messages],
             "stream": stream,
         }
+        if thinking is not None:
+            payload["think"] = thinking
 
         options: dict[str, Any] = {}
         if max_tokens is not None:
@@ -275,12 +286,16 @@ class OllamaAdapter(LLMPort):
         max_tokens: int | None = None,
         temperature: float | None = None,
         timeout_seconds: float | None = None,
+        thinking: bool | None = None,
     ) -> ChatMessage:
         """Stream chat internally for connection liveness, return complete ChatMessage with usage.
 
         Uses streaming transport to keep the connection alive during long-running
         inference, but returns only the final assembled response. Captures token
         usage metadata from Ollama's final `done: true` chunk.
+
+        ``thinking=False`` stops paying for a reasoning channel this adapter discards
+        unread — see ``_build_chat_payload``.
         """
         client = await self._get_client()
         resolved_model = model or self._default_model
@@ -291,6 +306,7 @@ class OllamaAdapter(LLMPort):
             max_tokens,
             temperature,
             stream=True,
+            thinking=thinking,
         )
 
         try:
--- a/adapters/llm/vllm.py
+++ b/adapters/llm/vllm.py
@@ -360,12 +360,19 @@ class VLLMAdapter(LLMPort):
         max_tokens: int | None = None,
         temperature: float | None = None,
         timeout_seconds: float | None = None,
+        thinking: bool | None = None,
     ) -> ChatMessage:
         """Stream for connection liveness, return one assembled message.
 
         The final frame carries usage when the server honors
         ``stream_options.include_usage``. A server that does not send one leaves
         the counts ``None`` — never zero, which would read as a free call.
+
+        ``thinking`` is accepted and **deliberately ignored**: the OpenAI-compatible
+        surface this speaks has no equivalent control, and the port states the
+        parameter is a request rather than a guarantee (#924). Accepting-and-ignoring
+        keeps the signature conformant; silently omitting it from the signature would
+        make a caller's keyword argument a TypeError at runtime instead.
         """
         resolved = model or self._default_model
         started = time.monotonic()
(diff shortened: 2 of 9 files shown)


jasonLadd:
Superseded. Split per review — the reasoning-budget half was a per-handler boolean switch plus a provider-shaped `think` flag on a provider-neutral port: a workaround, not a seam.

The clean half ships as the emission-capture PR. The reasoning control is filed for proper design (contract-declared, resolved through the chain that already handles temperature and max_tokens, provider-neutral level rather than an Ollama bool).

Sequencing matters here: the instrument is what determines whether the reasoning budget is even roll 5's cause. Shipping the fix and the instrument together — on one measurement — was the same mistake that produced two wrong diagnoses before it.

PR #926, opened 2026-08-16
feat(observability): record what an LLM call actually emitted

jasonLadd:
Instrument only. No behavior change, no provider coupling. Supersedes the capture half of #925.

## Why

SIP-0104 window rolls 3 and 5 both ended with **every scaffold slot unfilled**, and nothing recorded what the qa author emitted. Fills are parsed and stripped before extraction, so a success leaves no artifact and a failure leaves no trace. Three diagnoses were made from the *result* rather than the emission — #910, #911, #914 — and **two were wrong**.

Roll 5 makes the cost concrete: after hours of archaeology its cause is **still undetermined**. The shells are stored unfilled, the pre-existing zero-extraction guard stayed silent (so *something* parsed), and the expected test artifact is absent. Three candidates remain and cannot be separated from stored state:

- the author emitted nothing test-shaped
- it emitted a path-addressed file that did not survive
- **it emitted fills that were rejected**

P3 renders a rejected fill as the same failing state as a missing one, so the most likely candidate is also the least visible.

## What ships

**Emission shape, at the completion seam** — length, fence counts by kind, bounded head sample:

```
emission shape: chars=58   completion_tokens=413  fences={'fill': 2, ...}
emission shape: chars=0    completion_tokens=6866 fences={'fill': 0, ...}
emission shape: chars=47   completion_tokens=900  fences={'path': 1, 'fill': 0}
```

Unconditional, and deliberately **outside the observability block** — that block is gated on `llm_obs and correlation_context`, so a capture placed there goes silent in exactly the deployments where an unexplained emission is hardest to diagnose. I made that placement mistake while writing this; a test now kills it.

**Parse outcome, at the fill seam** — fills parsed, duplicate slots rejected, files extracted, expected artifacts, scaffold bound. This is the one that answers roll 5's open question, and it must sit at the **parse** site: a capture after the merge reports the merged result, which is precisely what loses the distinction.

Complements `_log_no_fenced_blocks` (#130) rather than duplicating it — that one fires only when *both* extraction and fills are empty. Its silence on roll 5 was itself evidence, but silence is a poor instrument.

## Verification

Three mutations killed: shape capture unwired · shape capture moved inside the observability gate · fill-seam capture removed. The first two matter most — an earlier version of these tests exercised the helper as a pure function, so deleting its call site passed everything.

Regression **7356**.

## Deliberately not included

The reasoning-budget change measured alongside this — 5,727 completion tokens with reasoning versus **413 without**, for the same eight fill blocks. That measurement stands and is filed, but its implementation was a per-handler boolean switch plus a provider-shaped `think` flag on a provider-neutral port: a workaround, not a seam.

It is also the wrong order. **This instrument is what determines whether the reasoning budget is even the cause.** Shipping a fix and its instrument together, on one measurement, is the same mistake that produced the two wrong diagnoses before it.

### DIFF
--- a/src/squadops/capabilities/handlers/cycle/base.py
+++ b/src/squadops/capabilities/handlers/cycle/base.py
@@ -45,6 +45,52 @@ from squadops.capabilities.handlers.cycle.validation import (
 logger = logging.getLogger(__name__)
 
 
+#: Fence kinds an emission is judged by. Counting them is the difference between
+#: "the author ignored the instruction" and "the author emitted the wrong shape" —
+#: two diagnoses with opposite fixes that were indistinguishable from the outside.
+_EMISSION_FENCES: tuple[tuple[str, str], ...] = (
+    ("fill", "```fill:slot-"),
+    ("path", "```typescript:"),
+    ("py", "```python:"),
+    ("plain", "```"),
+)
+
+
+def _log_emission_shape(handler_name: str, content: object, completion_tokens: object) -> None:
+    """Record what an LLM call actually emitted, in one greppable line.
+
+    **Why this exists (#924).** SIP-0104 window rolls 3 and 5 both ended with every
+    scaffold slot unfilled, and nothing recorded what the qa author emitted: fills are
+    parsed and stripped before extraction, so a successful fill leaves no artifact and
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #928, opened 2026-08-16
fix(observability): capture every LLM seam, not the one that had a helper

jasonLadd:
Instrument only. No behavior change. Stacked on #926.

## The defect

#926 placed the emission capture in `handlers/cycle/base.py` on the assumption that handlers share one LLM seam. **They do not.** Thirteen production call sites exist across nine files, each with its own `chat_stream_with_usage` and its own `response.content` read. The capture covered exactly one — and not the qa author's, which is the single emission the SIP-0104 window is blocked on.

## How it was found

Roll 6 was launched and the capture verified *loaded* in eve and runtime-api. Then a completed `data.research_context` call — 27b, 4,829 completion tokens, `handler_succeeded` — produced **no shape line at all**.

Without that check, roll 6 would have run its full ~90 minutes and produced exactly the same unexplained result as rolls 3 and 5, with an instrument that was never on the path.

## What changed

- **`emission_log.py` owns the concern.** `cycle/base.py` was unreachable from `planning/` and `impl/` without importing a sibling handler's base class — so the natural fix for a missing capture there was to write a second copy.
- **All 13 seams call it**, including the *second* (self-eval) call in `develop` and `qa_test`, whose emission can overwrite the first's artifacts.
- **`test_every_llm_seam_captures_what_it_emitted`** counts calls vs captures **per file** and fails on any shortfall, so a new seam cannot be added dark. Per-file counts rather than a boolean: instrumenting only the primary call of a pair leaves the rewrite path unrecorded while a boolean test reads green.
- Tests moved out of `test_test_runner.py`, which was never their home — the misfiling is part of how this went unnoticed.

## Two homes deliberately rejected

| Candidate | Why not |
|---|---|
| `llm/router.py` | Pure pass-through and **not on the production path** — the entrypoint injects the provider adapter directly (`bootstrap/system.py` ← `adapters/llm/factory.py`). Capturing there logs nothing today and would double-log if ever wired. Exempted in the test *with the reason recorded*, not silently skipped. |
| `adapters/llm/ollama.py` | Fence vocabulary (` ```fill:slot- `) is domain knowledge, not transport, and it would duplicate across every provider adapter. |

The remaining per-seam duplication is real; the coverage test polices it. A port-level decorator is the structural fix and is filed separately rather than done here mid-window.

## Verification

Three mutations killed:
1. primary capture removed → fails
2. **self-eval capture removed → fails** (a boolean check survives this one)
3. package import removed → fails

Regression **7358 passed**, 1 skipped.


jasonLadd (commented):
reviewed

Issue #953, opened 2026-08-17
UI data-path check probes every call site with GET and classifies the resulting 405 as PAGE_NOT_API — fails correct apps


jasonLadd:
## What happens

The UI data-path check issues a **GET to every call site**, whatever verb the UI actually uses (`scripts/dev/audit_delivered_app.py:135`):

```python
resp = await client.get(call.request_path)
```

A POST-only route answers that GET with **405 and no content-type**. `classify_ui_response` then reads it as a page collision (`src/squadops/capabilities/ui_data_path.py:140-147`):

```python
is_json = "json" in content_type.lower()
if status == 404 and not is_json:
    return ROUTE_MISSING
if via_seam and not is_json:
    return PAGE_NOT_API
return SERVED
```

405 is not 404, the content-type is empty, so it falls to `PAGE_NOT_API` and the audit reports:

> serves a PAGE, not the API — the response is HTML, so the parsed body is empty and the view renders blank with no error

Every clause of that message is false for a 405. The response is not HTML, no page was served, and the route unambiguously exists.

**A 405 is the strongest available evidence that the route file exists and matched.** The check turns it into a failure.

## Reproduction — window roll 3, a correct app failing the audit

`cyc_0e07836b5baa` / `run_e336e3bd4304`. The audit reported two failures:

```
FAIL ui-data-path app/runs/[run_id]/page.tsx:48: api('/api/runs/${runId}/join') ... serves a PAGE, not the API
FAIL ui-data-path app/runs/[run_id]/page.tsx:75: api('/api/runs/${runId}/leave') ... serves a PAGE, not the API
```

Booting the delivered workspace and exercising it directly:

```
POST /api/runs/{id}/join   → 200  application/json   participants: ["alice"]
POST /api/runs/{id}/leave  → 200
GET  /api/runs/{id}/join   → 405  (empty content-type)
```

The route files are present (`app/api/runs/[run_id]/join/route.ts`, `.../leave/route.ts`), the page calls them with `method: 'POST'` and a JSON body, and all five contract probes — which use the correct verbs — passed, including `vc-probe-api-runs-join`, `-join-duplicate` and `-leave`.

The app is correct. The check is wrong.

## Proposed fix — classify 405 as served, and do not send other verbs

The cheapest correct fix is a classification change, not a request change:

```python
if status == 405:
    return SERVED   # the route matched; it just does not take GET
```

[...]

Issue #952, opened 2026-08-17
UI data-path extractor scans line-by-line and silently drops wrapped call sites — roll 1's join/leave were never probed


jasonLadd:
## What happens

`extract_ui_calls` scans **one line at a time** (`src/squadops/capabilities/ui_data_path.py:84-85`):

```python
for lineno, line in enumerate(content.split("\n"), start=1):
    for match in _CALL_RE.finditer(line):
```

A call whose path argument wraps to the following line never matches, is never recorded, and is therefore never probed. **Nothing reports the omission** — the check emits no failure and no warning, so a file with unprobed call sites is indistinguishable from a file whose calls all passed.

## It has already hidden the exact defect this check exists for

SIP-0104 P6 window roll 1 (`cyc_24aac7c22be1` / `run_0d8ffad5f128`) **passed the boot audit**. Its detail page writes join and leave like this:

```javascript
34:    api<Run>(`/api/runs/${runId}`)
58:      const updatedRun = await api<Run>(
88:      const updatedRun = await api<Run>(
```

Line 34 was probed. **Lines 58 and 88 were not** — the paths sit on the next line. Those are the join and leave calls, and the app has `app/api/runs/[run_id]/join/route.ts` and `.../leave/route.ts` to serve them.

So roll 1's audit verified one of three data paths on its most important page and reported a clean pass.

Roll 3 (`cyc_0e07836b5baa` / `run_e336e3bd4304`) wrote the same two calls inline on one line. They were extracted, probed, and produced a failure — a false one, for a separate reason (#NEXT), but the contrast is the point: **whether a call site gets verified at all depends on how the author happened to wrap the line.**

The check was built after roll 1 of the previous window shipped an application whose every UI action 404'd, through 36/36 checks, five probes, `tests_pass`, `frontend_build` and the boot audit. On the class it was built for, it has been silently sampling.

## Why this is more than a regex bug

The failure mode is **silence**. This is the same shape as #948 and #951 — an absence that reads as coverage. A check that cannot see a call site must say so; today the only observable difference between "all call sites served" and "no call sites were found" is that both print nothing.

## Proposed dispositions

1. **Extract across the whole file, not per line.** The seam call is a syntactic construct that may span lines; scanning line-wise cannot see it. Either match over the full content with the newline handling that implies, or parse.
2. **Never drop a call site silently.** When a seam function name appears and no path can be extracted from it, record it as unresolved and surface it. An unreadable call site is a finding, not a non-event.
[...]

Issue #948, opened 2026-08-16
probe coverage silently drops to zero when a manifest puts a child action in the request body instead of the path


jasonLadd:
## What happens

Behavioural probe coverage depends on how the squad happens to name its routes. A legitimate REST decomposition that puts a child action in the **request body** instead of the **path** derives zero probes for that action, silently.

Two rolls of the SIP-0104 P6 window, same request profile and **identical `resolved_config_hash` `d4d4f662…`**:

| | endpoints | join/leave expressed as | probes derived |
|---|---|---|---|
| roll 1 `cyc_24aac7c22be1` | 5 | `POST /api/runs/{run_id}/join`, `.../leave` | **5** — create, rejects-blank, join, join-duplicate, leave |
| roll 2 `cyc_2f63e2d841eb` | 4 | `POST /api/runs/{run_id}` + `ParticipantAction{action, name}` | **2** — create, rejects-blank |

Roll 2's core PRD interaction has **no independent behavioural check at all**. The only thing covering join and leave is the authored suite — the one layer with no verifier above it.

## Why

Two filters in `src/squadops/capabilities/scaffold_contract.py::_probes`, and the endpoint falls through both:

- **`:409`** — `if ep.method != "POST" or "{" in ep.path: continue`. Skipped as a create candidate because the path is parameterized.
- **`:487-492`** — the child-action match requires a literal trailing segment:
  ```python
  and re.fullmatch(re.escape(ep.path) + r"/\{(\w+)\}/\w+", child.path)
  ```
  `/api/runs/{run_id}` has no trailing segment, so it is not a child of `/api/runs`.

The docstring immediately above, at `:389`, records this exact failure being fixed once already:

> fay-3 shipped a broken join as "functional": v7 probed create + blank rejection only, so the functional bar never touched the app's core interactions.

Roll 2's contract is back to create + blank rejection only. The #651 fix works for one route-naming convention and produces silent zero coverage for another.

## The part that makes this more than a regex bug

**Relaxing the pattern would manufacture false reds, not coverage.** Roll 2's manifest declares:

```yaml
ParticipantAction:
  required: [action, name]
```

There is no way to express that `action` ∈ {`join`, `leave`}. The deriver builds probe bodies from field names via `_probe_sample_value`, so it would send `{"action": "sample", "name": "sample"}` — which a *correct* app must reject, against a probe expecting the declared success status.

[...]

PR #954, opened 2026-08-17
docs(plan): V7 FAY window pre-registration (draft, not in force) — closes 4c

jasonLadd:
Docs only. No code, no effect on the SIP-0104 P6 window currently running.

Adds `docs/plans/1-6-0-v7-fay-window-preregistration.md` — the Guard 2 pre-registration for V7, the authored-mode FAY window, plus the rulings taken while drafting it.

## Why now

V7 requires N, PRD, deploy hash and scoring fixed **before roll 1**. Writing it while the window is still blocked on V6 is the only period in which the values can be chosen honestly, because the results do not exist yet. It is deliberately committed as a **DRAFT that is not in force** — several parameters are still blank and can only be filled when the window is ready to open.

## Rulings closed here

| # | Ruling | Owner, 2026-08-16 |
|---|---|---|
| §7.2 | **N = 6, bar ≥ 4/6** | matches the 1.4 precedent |
| §7.3 | **Gate approval is a pre-declared constant (option c)** | verbatim notes text fixed in the document |
| §7.1 | **Record probe + slot count per roll, non-gating, no floor** | **closes plan step 4c** |

Step 4c required its decision to be made *in the window record*, so ruling it in the pre-registration satisfies it by construction. `1-6-0-authorship-plan.md`'s 4c row is updated to point here rather than left open — both documents move together so the record does not drift the way the version markers did at 1.1.x.

## What the P6 window contributed

Three things in this document are measurements rather than reasoning, and none of them were available when the cadence was first written.

**4c's evidence is now much stronger than the 29-vs-57-checks observation that raised it.** Four P6 rolls at an **identical** `resolved_config_hash`, same PRD, frozen deploy, produced 5 / 2 / 5 / 4 probes and 8 / 5 / 8 / 7 slots — through two distinct authoring mechanisms, neither of them machinery drift:

- roll 2 folded join and leave into one endpoint discriminated by a request-body field (#948);
- roll 4 used separate paths but declared **no error code mapping to 409**, so no duplicate-action probe derived.

The squad authors the exam it sits. Hence: record, never gate, and report the **distribution rather than a mean** — an average would conceal exactly the variation the ruling exists to expose.

**§2.a makes #952 and #953 blocking preconditions rather than cleanup.** Both sit in `audit_delivered_app.py`, which decides the functional level. P6 rolls 3 and 4 each required a human to boot the deliverable and issue POSTs by hand to establish that an audit failure was false. **A window measuring "zero manual intervention" cannot require manual intervention to score itself.**

**§7.3 exists because the metric's own wording was never settled.** P6 handled its rolls two ways — operator approval on rolls 1–3, `system:no_open_questions` on roll 4. Option (c) resolves it by making approval a constant: identical text on every gate, no roll number and no commit hash in it, `--as-agent` on the record. Four rules keep it honest, of which rule 3 is the one that will be tempting to break — if a manifest asks something the standing text does not address, the text is **still** applied verbatim and the mismatch is logged as an observation. The moment the answer varies with the question, approval is a judgment again and the number means nothing.

## Still open, deliberately

- **The three outcome-band claim sentences (§7.4).** Setting the bar at ≥4/6 created a middle band: a 2/6 or 3/6 result **clears the 1.6 cut gate** — which asks only for *authored-mode FAY repeatably > 0* — while **missing this window's own bar**. That is legitimate rather than contradictory (the gate asks whether the capability exists, the bar asks whether it is reliable), but which one the release claims must be written while the number is unknown, or it will be argued afterward.
- **PRD sha, `resolved_config_hash`, deploy commit and image ids** — fillable only when the window is ready to open.

## Not in force

The window may not open until every precondition in §2 holds and every ruling in §7 is closed. Merging this does not start anything.

### DIFF
--- a/docs/plans/1-6-0-authorship-plan.md
+++ b/docs/plans/1-6-0-authorship-plan.md
@@ -316,7 +316,7 @@ until 2b exists.
 |---|---|---|
 | **4a** | **DECIDED 2026-08-15: keep the example, hold it constant, narrow the claim.** Pinned by test | closed; see below |
 | **4b** | **V6** viability run — **served by the SIP-0104 P6 window's counting rolls from roll 3 onward**, not separately scheduled (owner ruling 2026-08-15) | gates V7 |
-| **4c** | Decide **contract size per roll** in the window record | must precede pre-registration |
+| **4c** | **DECIDED 2026-08-16: record probe count and scaffold slot count per roll, non-gating, no floor.** Ruled in the window record — see `1-6-0-v7-fay-window-preregistration.md` §7.1 | closed |
 | **4d** | Pre-registration committed (N, PRD, deploy hash, scoring) → **freeze** | Guard 2 |
 | **4e** | **V7** — the FAY window, unfiltered, **fix nothing until it closes** | the evidence |
 | **4f** | **V8** confirmation shakedown | green |
--- a/docs/plans/1-6-0-v7-fay-window-preregistration.md
+++ b/docs/plans/1-6-0-v7-fay-window-preregistration.md
@@ -1,0 +1,223 @@
+# V7 — Authored-Mode FAY Window: Pre-Registration
+
+**Status: DRAFT. Not in force.** The window may not open until every item in §2 holds and
+every ruling in §7 is closed. Once in force, nothing in this document changes until the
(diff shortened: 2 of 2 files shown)


jasonLadd (commented):
reviewed

Issue #951, opened 2026-08-16
scaffold covers a derived subset of declared behaviours and never reports the delta — roll 2 banked green with join/leave deterministically unverified


jasonLadd:
## What happens

The verification scaffold covers a **derived subset** of the behaviours a manifest declares, and the difference between the two is never recorded anywhere. Silence reads as coverage.

Window roll 2 (`cyc_2f63e2d841eb`) banked green with **join and leave — the entire point of the application — untouched by every deterministic layer**:

| layer | covers join/leave? |
|---|---|
| scaffold slots (5) | **no** — create, blank-rejection, list, detail, detail-404 |
| contract probes (2) | **no** — both are `POST /api/runs` |
| boot audit | **no** — it fires the contract's probes, and there are two |
| authored additive suite | yes — 12 tests |

The suite that carries it, `__tests__/participants.test.ts`, is a good one: 12 real tests including duplicate rejection, case-insensitivity, unknown `run_id`, and no mocking. That is the qa author's choice, not a guarantee. **Nothing required it and nothing would have reported its absence** — and it is precisely the layer #915 shows can mock itself green undetected.

So the app's core feature was verified only by the least-guaranteed layer, in a roll that passed 26/26 checks, 9/9 criteria and the boot audit.

## Why — the scaffold's behaviour set has exactly two sources

`stack_nextjs_ts_tests.py:394-412`:

```python
behaviors = _behaviors_from_probes(classified)
behaviors += _minted_read_behaviors(manifest, creates_by_path)
```

- **probe twins** — and probes only derive for `POST` at a path with no parameter (`scaffold_contract.py:409`), plus child actions matching a literal trailing segment (#948);
- **minted reads** — `GET` only (`stack_nextjs_ts_tests.py:254`).

Everything else falls in the hole, deterministically and silently:

- `POST` at a parameterized path — roll 2's join/leave
- `PUT`, `PATCH`, `DELETE` — no source of scaffold behaviours at all, for any manifest
- child actions whose path shape the #948 regex does not match

The scaffold's coverage is therefore defined by **what it can derive**, not by **what the manifest declares** — which is a reasonable engineering constraint. The defect is that the delta is invisible.

## This is broader than #948

#948 is one *cause*: the probe deriver misses body-discriminated child actions. Fixing it would restore coverage for that particular shape, because probe twins are one of the two scaffold sources.

[...]

Issue #915, opened 2026-08-15
A qa additive suite can mock the subject and assert its own mock — nothing detects it, and it goes green


jasonLadd:
A qa-authored additive test file can mock the thing it is testing and assert its own mock. Nothing detects it, and unlike the stub-fallback class it parallels, **it produces green rather than red.**

## Found in window roll 3

`cyc_b20f58cc7cbc` / `run_5f42bd19360f`. The qa author emitted one additive suite, `__tests__/api/runs.test.ts`:

```ts
const mockFetch = vi.fn();
global.fetch = mockFetch as any;
...
mockFetch.mockResolvedValueOnce({
  ok: true,
  status: 200,
  json: async () => newRun,
});

const response = await fetch('/api/runs', { method: 'POST', ... });
const data = await response.json();
expect(data.error_code).toBe('validation_error');
```

`fetch` is replaced with a mock, the mock is told what to return, the mock is called, and the assertion reads what the mock returned. **The application is never invoked.** 479 lines, zero coverage.

## Why nothing catches it

- **Fill validation does not apply.** Fills are deterministically validated — no imports, no `require()`, no `fetch()`, no network. Additive files go through none of that; the rule exists for them only as prose in the brief ("the suite runs in-process with no server", #877).
- **`detect_stub_fallback_tests` cannot see it.** It matches a Python `ImportError`/`ModuleNotFoundError` guard combined with an app constructor (`stub_detection.py:74`). A TypeScript suite mocking `global.fetch` matches neither clause.
- **The runner cannot see it.** The suite collects and runs; `tests_pass` is exit-code driven.

## Why this is worse than the class it parallels

#276's stub-fallback tests fail loudly — a reconstructed app does not behave like the real one. A mocked suite **agrees with itself by construction**, so its natural state is passing. Roll 3 only went red because the scaffold slots were also unfilled (#910); had they been filled, `tests_pass` would have gone green on the strength of the fills while this file contributed nothing and read as 479 lines of verification.

## Bounding what it does and does not threaten

The scaffold shells are **not** mockable — their frozen spine imports the route handler and invokes it directly, and region enforcement rejects any edit to that. So on a roll where the slots are filled, the real verification is intact and this suite is waste rather than a lie. The acute risk is a roll like #3 where the additive suite is the author's *only* output.

## Fix direction

[...]

PR #955, opened 2026-08-17
docs(plan): V7 outcome bands ruled + fixed parameters filled where knowable

jasonLadd:
Docs only. No code, no effect on the SIP-0104 P6 window currently running.

Fills in what could honestly be filled in the V7 pre-registration, and closes its last ruling.

## The middle band, ruled

Owner ruling: **2–3 of 6** claims that a squad-authored design *can* produce a working application with no human intervention, and does **not** claim it does so reliably. The figure is banked as 1.8's baseline regardless.

Two rules now attach to every band, and they matter more than the wording:

1. **Bank the figure whatever it is.** A number discarded because it disappointed is worse than no number — the next window then has nothing to be compared against and quietly becomes the first.
2. **Never re-run the window to improve the figure.** Most tempting at exactly 3/6, where one more roll could tip it over the bar. That is the move pre-registration exists to block.

## What is drafted rather than ruled

The other three bands are written consistently with the ruling and **marked as drafted**, so the record does not report a ruling that was not made.

One of them is a judgment worth a look rather than a rubber stamp: **1/6 is split out of the 1–3 band.** The draft flagged it as "arguable" and left it inside; this separates it, on the reading that *repeatably > 0* means more than once. Grouping a single green with 2–3 would let one success satisfy a gate whose entire point is that one success is not evidence — and 1/6 is precisely where the pressure to argue the generous reading will be highest. Conservative reading taken, flagged in the text for overruling.

## Fixed parameters

| Parameter | Filled | Note |
|---|---|---|
| PRD sha256 | ✅ `f744843d…` | **filename corrected** |
| `resolved_config_hash` | ✅ `d4d4f662…` | re-verified at roll 1 |
| Deploy commit | — | not a decision; unknowable until the queue lands |
| Deploy image ids | — | not a decision; unknowable until the rebuild |

**The PRD filename was wrong.** The draft named `group_run_v0.5.md`, which is not a path that exists — the file is `examples/03_group_run/prd.md` and *declares* v0.5 in its header. A sha recorded against a wrong filename is worse than no sha: it looks verifiable and isn't.

**The config hash is recorded with its derivation, not just its value.** `compute_config_hash` (`cycles/lifecycle.py:243`) takes the request profile's `defaults` merged with the execution overrides and **nothing else** — no deploy state, no image, no commit. That is why it survives the rebuild separating this window from P6, and why the two windows stay comparable on recipe. It is re-verified immediately before roll 1 anyway, because anything in the fix queue touching the `validated-fullstack` CRP defaults would move it silently.

## Still open

**§2.3 — fix-or-declare for #951, #948 and #915.** Each bounds what a green roll means, so each needs a disposition recorded *before* roll 1. #795 is already ruled (declare). Deciding these after seeing rolls would be the same error the pre-registration exists to prevent.

## Not in force

Unchanged: the window may not open until every precondition in §2 holds. Merging this does not start anything.

### DIFF
--- a/docs/plans/1-6-0-v7-fay-window-preregistration.md
+++ b/docs/plans/1-6-0-v7-fay-window-preregistration.md
@@ -35,7 +35,7 @@ is a separate, deliberate act taken after the window closes.
 | 2.3 | Instrument defects either fixed or explicitly declared (§2.b) | open |
 | 2.4 | Deploy frozen; commit and image ids recorded in §3 | pending |
 | 2.5 | Zero open focus leases immediately before roll 1 | check at launch |
-| 2.6 | Every ruling in §7 closed | open |
+| 2.6 | Every ruling in §7 closed | 7.1–7.3 closed; 7.4's middle band owner-ruled, its other three bands drafted and awaiting a nod |
 
 ### 2.a Why the audit defects are blocking, not cosmetic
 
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #956, opened 2026-08-17
fix(audit): find wrapped UI call sites, and read 405 as proof the route exists

### LINKED ISSUE #952: UI data-path extractor scans line-by-line and silently drops wrapped call sites — roll 1's join/leave were never probed

### LINKED ISSUE #953: UI data-path check probes every call site with GET and classifies the resulting 405 as PAGE_NOT_API — fails correct apps

jasonLadd:
Two defects in the UI data-path check. They ship together because **they masked each other**, and fixing either alone makes the audit worse than it is today.

`Closes #952` · `Closes #953`

## Why one PR

- Roll 1 **passed** its audit because #952 never found its join and leave calls — a false pass.
- Rolls 3 and 4 **failed** theirs because #953 misread the answer those same calls give — a false failure.

Fix extraction alone and roll 1's newly-found calls fall straight into the 405 misread: a false pass becomes a false failure. Fix classification alone and roll 1 stays vacuous. Only both together verify anything.

## #952 — the extractor scanned line by line

`extract_ui_calls` iterated `content.split("\n")`, so a call written across two lines matched nothing:

```tsx
const updatedRun = await api<Run>(
  `/api/runs/${runId}/join`,
  { method: 'POST' }
);
```

That is the shape a formatter produces as soon as the call takes a second argument — which is to say, **every mutating call**. This is not a weaker check. It is no check, silently, while the roll still reads as audited.

**Measured against P6 roll 1's delivered source, the old expression found 2 of 5 call sites.** It dropped join, leave, *and* the detail-page read. The ledger recorded this as "join and leave were never probed"; it was worse — roll 1's audit pass covered under half its UI.

Scanning whole content makes a runaway match possible where line-scanning made it impossible, so the type-parameter and path classes now exclude newlines. Without that, one stray backtick yields a single multi-line garbage path and silently swallows every real call after it — trading a known blind spot for an unknown one. Only the whitespace between `(` and the opening quote may cross a line, which is exactly the wrap being tolerated.

The reported line is where the call **opens**, not where its path sits. A message pointing at an argument sends the reader somewhere they can't fix.

## #953 — a 405 was read as "this is a page"

Every probe is a GET whatever verb the UI uses, and that is deliberate: issuing the real verb would mutate the app under audit — a POST probe creates a record — and destroy the property that makes the audit re-runnable. **The cost is that a POST-only route answers 405**, which Next produces from the router with no content type. The not-JSON rule read that as `PAGE_NOT_API`.

This inverts the evidence. **A 405 is the strongest available signal that the route exists**: the router matched the path and rejected only the method, which is precisely the question this check asks. P6 rolls 3 and 4 were both failed on correct join and leave routes, each requiring a human to boot the deliverable and POST by hand.

A page never answers 405 — it answers 200 with HTML — so the page-collision case is untouched. A test pins that the correction does not over-reach into "any non-JSON answer is served," which is the obvious way to get this wrong.

## Verified against real workspaces, not only fixtures

| Roll | Old extraction | New extraction | Audit now |
|---|---|---|---|
| 1 (`run_0d8ffad5f128`) | **2 of 5** | 5 of 5 | see below |
| 2 (`run_c23ceeee4796`) | 5 | 5 | unchanged |
| 3 (`run_e336e3bd4304`) | 5 | 5 | see below |
| 4 (`run_ca8b621cf7b6`) | 5 | 5 | **PASS** — was a FAIL |

Roll 4's re-audit, same workspace, with these fixes:

```
PASS — delivered app installs, builds, boots, answers 4 contract probe(s),
and its UI reaches every path it requests
```

Rolls 1 and 3 are re-auditing as this opens; results posted as a comment.

## Consequence for the window ledger

**P6 roll 1's audit pass is weaker than banked** and should be annotated rather than revoked: it passed on 2 of 5 call sites, so its UI wiring for join, leave and the detail read was never verified. Whether it passes on all five is what the re-audit above answers.

This does not change any roll's banked status — no roll is re-scored on an instrument change after the fact — but the closing record should state which rolls were audited by which version of the check.

## Blocking status

These are §2.2 of the V7 pre-registration: **blocking preconditions, not cleanup.** A window measuring "zero manual intervention" cannot require manual intervention to score itself, and rolls 3 and 4 each did.

**Does not merge until the SIP-0104 P6 window closes.** The deploy is frozen at `d590f73c`; this is queued behind roll 6.

### DIFF
--- a/scripts/dev/audit_delivered_app.py
+++ b/scripts/dev/audit_delivered_app.py
@@ -114,9 +114,19 @@ async def _run_ui_data_path(files: dict[str, str], stack: str, base_url: str) ->
 
     The contract probes above prove the API answers where the CONTRACT says; this proves
     it answers where the UI ASKS. Roll 1 passed the former and failed the latter on every
-    page action, and nothing noticed. Each distinct path is requested once; only a 404
-    that no route produced (framework HTML rather than the app's JSON envelope) counts as
-    a failure, so an app correctly reporting an unknown id still passes.
+    page action, and nothing noticed. Each distinct path is requested once.
+
+    **Every probe is a GET, whatever verb the UI uses**, and deliberately so: issuing the
+    real verb would mutate the application under audit — a POST probe creates a record —
+    and the audit has to be re-runnable against the same workspace. The cost is that a
+    POST-only route answers 405, which ``classify_ui_response`` reads as SERVED (#953);
+    the alternative, sending real verbs, buys a stronger signal by destroying the
+    property that makes the audit repeatable.
+
+    A path fails only when the answer shows no route is mounted there: a 404 the framework
+    produced (HTML rather than the app's own JSON envelope), or HTML through the JSON seam,
+    which means the call landed on a page. An app correctly reporting an unknown id still
+    passes, and so does one that rejects the probe's method.
     """
     calls = extract_ui_calls(files, stack)
     if not calls:
--- a/src/squadops/capabilities/ui_data_path.py
+++ b/src/squadops/capabilities/ui_data_path.py
@@ -44,8 +44,15 @@ _SEAM_PREFIX: dict[str, str] = {
 #: Files whose calls are the UI's. Route handlers legitimately talk to the store, not HTTP.
 _UI_SUFFIXES = (".tsx", ".jsx")
 
+#: Matched against whole file content, not line by line — a call argument that wraps to the
+#: next line is still one call, and scanning per line silently dropped it (#952). The two
+#: character classes exclude newlines deliberately: a type parameter and a request path each
+#: live on one line, and letting either span lines lets an unbalanced quote swallow the rest
+#: of the file. Only the whitespace between ``(`` and the opening quote may cross a line,
+#: which is exactly the wrap this expression exists to tolerate.
 _CALL_RE = re.compile(
-    r"\b(?P<fn>api|apiFetch|fetch)\s*(?:<[^>()]*>)?\s*\(\s*(?P<quote>['\"`])(?P<path>[^'\"`]*)(?P=quote)"
+    r"\b(?P<fn>api|apiFetch|fetch)\s*(?:<[^>()\n]*>)?\s*"
+    r"\(\s*(?P<quote>['\"`])(?P<path>[^'\"`\n]*)(?P=quote)"
 )
 _TEMPLATE_EXPR_RE = re.compile(r"\$\{[^}]*\}")
 
@@ -73,6 +80,13 @@ def extract_ui_calls(files: dict[str, str], stack: str) -> list[UiCall]:
     live-server class, a different defect, and silently normalizing them would hide it.
     An unknown stack yields nothing rather than guessing a prefix — a wrong prefix
     would invent failures in an app that works.
+
+    Scans whole file content. The first version scanned line by line, so a call written
+    across two lines — ``await api<Run>(`` then the path beneath it, which is what a
+    formatter produces once the call carries a second argument — matched nothing and was
+    never probed (#952). That silence is the worst possible failure for this module: it
+    does not weaken the check, it removes it, and the roll still reads as audited. Roll 1
+    of the SIP-0104 P6 window passed with its join and leave call sites extracted as zero.
     """
     prefix = _SEAM_PREFIX.get(stack)
     if prefix is None:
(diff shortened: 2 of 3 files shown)


jasonLadd:
## Re-audit results — all four banked P6 rolls now pass

Same stored workspaces, same contracts, only these fixes changed. Replayed rather than re-run, so nothing depends on a fresh cycle.

| Roll | Impl run | Before | After |
|---|---|---|---|
| 1 | `run_0d8ffad5f128` | PASS — **vacuously**, on 2 of 5 call sites | **PASS on 5 of 5**, 5 probes |
| 2 | `run_c23ceeee4796` | PASS | unchanged (its calls were never wrapped) |
| 3 | `run_e336e3bd4304` | **FAIL** — false, human-verified by hand | **PASS**, 5 probes |
| 4 | `run_ca8b621cf7b6` | **FAIL** — false, human-verified by hand | **PASS**, 4 probes |

Roll 3 and roll 4 were the two that each cost a manual boot-and-POST to establish the failure wasn't real. Neither requires one now.

## The result that matters most is roll 1's

Roll 1 didn't change verdict — it changed from a pass that proved almost nothing to a pass that proves what it claimed. Its join, leave and detail-page reads had never been probed by anything; now all five are, and they answer correctly.

That is the harder half of this to notice, and the reason the ledger annotation is worth carrying into the closing record: **a green from a broken instrument and a green from a working one are not the same fact**, even when they print the same word.

## Consequence for plan step 1e

1e closes with "the first roll that passes the audit including the UI data-path check." Before these fixes no roll cleanly met that bar — roll 1's check was vacuous, and rolls 3 and 4 were failed by the tool and passed by a human. Under the corrected instrument the answer is unambiguous: **roll 1 of the reset window (`cyc_24aac7c22be1`) passes on all five call sites.**

Flagging rather than recording it — crediting 1e is a ruling, and the evidence changed after the ruling was written.

## What this does not do

No roll is re-scored. Banked status stands as banked, and a roll is never re-graded on an instrument change after the fact — that would be the same error in the opposite direction. The closing record should simply state which rolls were audited by which version of the check.

jasonLadd (commented):
Reviewed

PR #957, opened 2026-08-17
fix(qa): reject additive suites that mock the subject instead of invoking it

### LINKED ISSUE #915: A qa additive suite can mock the subject and assert its own mock — nothing detects it, and it goes green

jasonLadd:
A qa-authored additive suite can replace `global.fetch`, prime the mock, call the mock, and assert on what it was told to return. The application is never invoked, and nothing detects it.

`Closes #915`

## Why this is worse than the class it parallels

#276's stub-fallback suites **fail loudly** — a reconstructed app doesn't behave like the real one. A self-mocking suite **agrees with itself by construction**, so passing is its natural state.

Window roll 3 emitted 479 such lines. It only went red because its scaffold slots were *also* unfilled (#910). Had they been filled, `tests_pass` would have gone green on the strength of the fills while that file contributed nothing and read as 479 lines of verification.

## The asymmetry being closed

Fills **cannot** do this: SIP-0104's frozen spine imports the route handler and invokes it, and region enforcement rejects any edit to that surface. Additive files carried the same in-process rule (#877) **only as prose in the brief**. One half of the qa surface was enforced deterministically and the other half was asked nicely.

## Two clauses, both conservative

Mirrors the existing heuristic's structure rather than inventing a new posture.

| Clause | Trigger | Why |
|---|---|---|
| **Mocks the subject** | `vi.mock('…/app/api/…')` | Unconditional. There is no reading under which replacing the route module under test and then asserting on it verifies the route module. |
| **Mocks the network** | fetch seam replaced **and** no route module imported | Mocking is not the defect — mocking *instead of* invoking is. |

The first clause exists specifically because the second's discriminator would otherwise be satisfied by mock-then-import: mock the subject, import the mock, assert on it, every clause happy, application still never invoked.

**A suite that stubs an outbound dependency while still driving the real handler is legitimate** and is not flagged. That's the false positive that would actually hurt — it would reject correct work.

## Wired at both call sites

Retest included, for the reason #276 already gives: a repair told to make the suite pass can satisfy that instruction by mocking, and **mocking is the cheapest way to make any assertion true.** A guard that only covers the first attempt invites the correction loop to walk straight through it.

## Validated against real emissions, not only fixtures

- **35 test files across the four banked P6 rolls → zero flags.**
- Roll 2's additive suite — the model answer, importing both route modules and invoking their handlers — is clean.
- Roll 3's real emission shape is caught.

## Naming, flagged not fixed

The module is called `stub_detection` and nothing in it is stub-specific any more; it now owns "generated tests that validate something other than the deliverable," which is one defect in two languages. Renaming touches two call sites and their tests, and bundling a rename into a fix landing under a deploy freeze is how a small change becomes a risky one. Recorded in the docstring so the next reader knows the name is stale rather than the scope narrow.

## Blocking status

§2.b of the V7 pre-registration, ruled **fix** rather than declare. **Does not merge until the SIP-0104 P6 window closes** — deploy frozen at `d590f73c`.

### DIFF
--- a/src/squadops/capabilities/handlers/cycle/qa_test.py
+++ b/src/squadops/capabilities/handlers/cycle/qa_test.py
@@ -758,7 +758,10 @@ class QATestHandler(_CycleTaskHandler):
 
         # #276 guard holds on the retest path too: a repair must not smuggle a
         # stub-fallback past the correction loop.
-        from squadops.capabilities.handlers.stub_detection import detect_stub_fallback_tests
+        from squadops.capabilities.handlers.stub_detection import (
+            detect_self_mocking_tests,
+            detect_stub_fallback_tests,
+        )
 
         stub_offenders = detect_stub_fallback_tests(artifacts)
         if stub_offenders:
@@ -772,6 +775,25 @@ class QATestHandler(_CycleTaskHandler):
             )
             missing.append(f"stub_fallback_tests:{','.join(stub_offenders)}")
             summary = f"{summary}; repaired test masks the entrypoint: {', '.join(stub_offenders)}"
+
+        # #915 on the retest path for the same reason: a repair told to make the suite
+        # pass can satisfy that instruction by mocking the subject, and mocking is the
+        # cheapest way to make any assertion true.
+        mock_offenders = detect_self_mocking_tests(artifacts)
+        if mock_offenders:
+            passed = False
+            mock_paths = [path for path, _ in mock_offenders]
+            checks.append(
+                {
+                    "check": "no_self_mocking_tests",
+                    "offenders": mock_paths,
+                    "passed": False,
+                }
+            )
+            missing.append(f"self_mocking_tests:{','.join(mock_paths)}")
+            summary = (
+                f"{summary}; repaired test never invokes the application: {', '.join(mock_paths)}"
+            )
 
         passed_count = sum(1 for c in checks if c.get("passed"))
         outputs: dict[str, Any] = {
@@ -1167,6 +1189,7 @@ class QATestHandler(_CycleTaskHandler):
             # passes). Flag it so acceptance fails and the correction loop
             # regenerates the test against the real module.
             from squadops.capabilities.handlers.stub_detection import (
+                detect_self_mocking_tests,
                 detect_stub_fallback_tests,
             )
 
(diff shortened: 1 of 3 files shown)


jasonLadd:
## Validated against the original offender, not a reconstruction

Pulled the actual file from the artifact vault — `cyc_b20f58cc7cbc` / `run_5f42bd19360f`, `__tests__/api/runs.test.ts`, 243 lines — and ran the detector on it:

```
flagged: [('__tests__/api/runs.test.ts',
  'replaces the fetch seam and imports no route module, so the application is
   never invoked — the suite asserts what it told its own mock to return')]
```

So both halves are now measured on real emissions rather than fixtures:

| corpus | result |
|---|---|
| the real #915 offender | **caught**, with the correct reason |
| 35 test files across the four banked P6 rolls | **zero** flags |

The second number is the one that decides whether this can ship. A detector that fires on correct work would reject good suites into a re-author, and roll 2's additive suite — which imports both route modules and drives their handlers — is exactly the shape a careless heuristic would catch.

jasonLadd (commented):
Reviewed

PR #958, opened 2026-08-17
feat(scaffold): report the delta between declared and verified behaviours

### LINKED ISSUE #951: scaffold covers a derived subset of declared behaviours and never reports the delta — roll 2 banked green with join/leave deterministically unverified

jasonLadd:
The scaffold covers what it can **derive**, not what the manifest **declares**, and the difference was recorded nowhere. Silence read as coverage.

`Closes #951`

## What went unnoticed

Window roll 2 (`cyc_2f63e2d841eb`) banked green — 26/26 checks, 9/9 criteria, boot audit pass — while **join and leave, the entire point of the application, went untouched by every deterministic layer:**

| layer | reaches join/leave? |
|---|---|
| 5 scaffold slots | no — create, blank-rejection, list, detail, detail-404 |
| 2 contract probes | no — both `POST /api/runs` |
| boot audit | no — it fires the contract's probes, and there are two |
| authored additive suite | yes — 12 tests |

The suite carrying it was a good one. **Nothing required that and nothing would have reported its absence** — and it is precisely the layer #915 shows can mock itself green.

## What this emits

Per manifest, which declared endpoints a probe and/or a slot reaches. Run against the four banked rolls it reproduces the known facts exactly:

```
roll 1   scaffold coverage: 5/5 declared endpoints, none uncovered
roll 2   scaffold coverage: 3/4 declared endpoints;
         NO deterministic coverage for POST /api/runs/{run_id}
roll 3   scaffold coverage: 5/5 declared endpoints, none uncovered
roll 4   scaffold coverage: 5/5 declared endpoints, none uncovered
```

Roll 2's line is the defect, stated at seed time in one line instead of found by hand days later. Roll 4's detail view also shows join at **1 probe where roll 1 had 2** — the missing duplicate-join probe (#948) is now visible rather than inferred.

## A report, not a gate

Whether an uncovered endpoint should **block** is a separate ruling — SIP-0096 already has `blocked_unverified` if it should. Being *recorded* should not wait on that decision, and coupling the two would have held a cheap fix hostage to a policy question.

Read-only with respect to emission: **no shell byte changes**, so SIP-0104's Gate 1 pins and `GENERATOR_VERSION` are untouched. 4800 tests green, byte pins included.

## Three design points

**`summarize_coverage` takes routes, not a manifest.** That keeps it stack-neutral the way `_EMITTERS` keeps emission stack-neutral — a second scaffold stack derives its behaviours its own way and reports through the same function, rather than this becoming a second place that knows about stack #2.

**`derive_scaffold_behaviors` is extracted from the emitter** so the report reads the *same* derivation rather than a parallel reimplementation. A coverage report that computed its own answer would drift from the shells silently and report a delta that was itself wrong — worse than no report, because it would look authoritative.

**Prerequisites are not coverage.** A create replayed in order to reach a join is setup. Crediting it would let a manifest reach "full coverage" while asserting on one endpoint — inflating exactly the number this exists to deflate.

## The summary names routes rather than counting them

`3/4` invites "close enough." `NO deterministic coverage for POST /api/runs/{run_id}` invites the only question worth asking, which is whether the one left out is the feature. In roll 2's case it was.

## Blocking status

§2.b of the V7 pre-registration, ruled **fix** rather than declare. **Does not merge until the SIP-0104 P6 window closes** — deploy frozen at `d590f73c`.

### DIFF
--- a/src/squadops/capabilities/scaffold_coverage.py
+++ b/src/squadops/capabilities/scaffold_coverage.py
@@ -1,0 +1,194 @@
+"""What the deterministic layer actually covers, per manifest (#951).
+
+The verification scaffold covers a **derived subset** of the behaviours a manifest
+declares. That is a defensible engineering constraint — a probe cannot be synthesized
+for every shape without inventing request values a correct application would reject.
+The defect is that the difference was never recorded, so **silence read as coverage.**
+
+Window roll 2 (``cyc_2f63e2d841eb``) banked green with join and leave — the entire
+point of the application — untouched by every deterministic layer: five scaffold slots
+covering create, blank-rejection, list, detail and detail-404; two probes, both
+``POST /api/runs``; a boot audit that fires those same two probes. The feature was
+verified only by a freely-authored additive suite, which happened to be a good one.
+Nothing required that and nothing would have reported its absence.
+
+The scaffold's coverage is defined by **what it can derive**, not by what the manifest
+**declares**. This module makes that gap legible: for every declared endpoint, whether
+a probe and/or a scaffold slot reaches it.
+
+Deliberately a *report*, not a gate. Whether an uncovered endpoint should block is a
+separate ruling (SIP-0096 already has ``blocked_unverified`` if it should); being
+**recorded** should not wait on that decision. It is also read-only with respect to
+emission: no shell byte changes, so SIP-0104's Gate 1 pins and ``GENERATOR_VERSION``
+are untouched by adding it.
+"""
+
+from __future__ import annotations
+
+import re
+from collections.abc import Sequence
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:  # pragma: no cover - annotations only
+    from squadops.capabilities.scaffold import InterfaceManifest
+
+#: A probe's path carries capture placeholders (``/api/runs/{created_id}/join``) while the
+#: manifest declares its own parameter names (``/api/runs/{run_id}/join``). Coverage is a
+#: question about the route, not about whose name for the parameter won, so both sides are
+#: reduced to the same shape before comparison.
+_PARAM = re.compile(r"\{\w*\}")
+
+
+def _route_key(method: str, path: str) -> tuple[str, str]:
+    return method.upper(), _PARAM.sub("{}", path)
+
+
+@dataclass(frozen=True)
+class EndpointCoverage:
+    """One declared endpoint and the deterministic layers that reach it."""
+
+    method: str
+    path: str
+    probe_ids: tuple[str, ...] = ()
+    slot_ids: tuple[str, ...] = ()
+
+    @property
+    def covered(self) -> bool:
+        return bool(self.probe_ids or self.slot_ids)
+
+    def describe(self) -> str:
+        if not self.covered:
+            return f"{self.method} {self.path}: NO deterministic coverage"
+        layers = []
+        if self.slot_ids:
+            layers.append(f"{len(self.slot_ids)} slot(s)")
+        if self.probe_ids:
+            layers.append(f"{len(self.probe_ids)} probe(s)")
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
Reviewed

PR #959, opened 2026-08-17
feat(contract): probe body-discriminated child actions from declared values

### LINKED ISSUE #948: probe coverage silently drops to zero when a manifest puts a child action in the request body instead of the path

jasonLadd:
A manifest can express a child action two ways, and only one of them was verifiable.

`Closes #948`

| shape | derives |
|---|---|
| `POST /runs/{run_id}/join` — path segment | a behavioural probe |
| `POST /runs/{run_id}` + `{"action": "join"}` — body-discriminated | **nothing** |

Window roll 2 chose the second and bought **zero** behavioural probes. Join and leave — the application's entire purpose — shipped verified only by a freely-authored suite, in a roll that passed 26/26 checks and 9/9 criteria.

## The fix is not relaxing the regex, and that is the whole design

With only a field *name*, the deriver can synthesize `{"action": "sample"}` — which a **correct** application rejects. Probing on a guess manufactures false failures, which is strictly worse than the silence it replaces: a false red costs a correction round and teaches the squad to distrust the contract.

So the author declares the domain instead:

```yaml
request_shapes:
  ParticipantAction:
    required: [action, name]
    values:
      action: [join, leave]
```

**Measured on roll 2's real manifest: 2 probes before, 4 after.**

```
vc-probe-runs                 POST /api/runs         {"title": ..., ...}        -> 201
vc-probe-runs-rejects-blank   POST /api/runs         {"title": "", ...}         -> 400
vc-probe-runs-join            POST /api/runs/{run_id} {"action":"join","name":…} -> 200
vc-probe-runs-leave           POST /api/runs/{run_id} {"action":"leave","name":…}-> 200
```

The create captures the run id into the path parameter exactly as the path-segment form does, and the contract lints clean.

## Conservative by construction

| case | behaviour | why |
|---|---|---|
| no `values` declared | derives nothing | unchanged — this is what makes it a schema change, not a regex change |
| **two** discriminators | read as **none** | a cross product whose size the author never sees; which combinations are legal is not something a shape can state |
| `values` on an optional field | not a discriminator | a probe omitting the field would exercise an undeclared default |

**No duplicate-conflict probe is emitted for a folded endpoint**, deliberately. One endpoint carries every action, so a declared 409 cannot be attributed to a particular value — repeating `leave` and expecting 409 would fail a correct app. Guessing which action conflicts is exactly the class of invention this fix exists to avoid, so that case stays uncovered and (with #951) visible.

## The hash question

`values` belongs in the structural projection: it changes the derived contract, and **a manifest whose contract differs must not share a hash with one whose contract does not.**

But emitting it unconditionally as an empty list would move the hash of every manifest ever written — invalidating each stored contract binding in order to record the absence of a field none of them use. So it is emitted **only when declared**. Existing manifests hash exactly as before; a manifest that adopts the field hashes differently, which is correct.

The reference manifest still hashes `bb472e26…`, pinned as a literal in the test. Comparing the manifest to itself would prove nothing no matter what the projection did.

## Prompt asset

The authoring rules appendix gains `declare-the-choices-a-body-carries`, with its warrant and a YAML snippet, in the same voice as the neighbouring rules. Prose lives in the managed fragment, not in a Python literal (#448).

**The worked example in the manifest template is deliberately unchanged.** Plan step 4a ruled the example is held constant, and it is the strongest prior the author has — adding an action endpoint there would nudge every future manifest toward inventing one.

## Scope note

This is the largest of the §2.b items and the only one that **changes what the squad authors**. V7's manifests will be written under a schema P6's were not, so the two windows stay comparable on recipe — the config hash does not move — but **not on probe counts.** P6's 5 / 2 / 5 / 4 / 5 distribution is the motivation for this fix and must not be reported as a baseline the V7 numbers improve on.

7786 unit tests green, byte pins and both Guard 1 halves included.

**Does not merge until the SIP-0104 P6 window closes** — deploy frozen at `d590f73c`.

### DIFF
--- a/src/squadops/capabilities/scaffold.py
+++ b/src/squadops/capabilities/scaffold.py
@@ -78,6 +78,34 @@ class RequestShape:
     name: str
     required: tuple[str, ...] = ()
     optional: tuple[str, ...] = ()
+    #: Declared candidate values per field, for fields whose value *selects a behaviour*
+    #: rather than carrying data (#948). A shape that says ``{"action": ("join", "leave")}``
+    #: is stating that those two strings are the whole domain of ``action``.
+    #:
+    #: Why the manifest has to say it rather than the deriver guessing: without declared
+    #: values a synthesized body is ``{"action": "sample"}``, which a CORRECT application
+    #: rejects — so relaxing derivation to probe these endpoints anyway would manufacture
+    #: false failures. Roll 2 folded join and leave into one body-discriminated endpoint
+    #: and bought **zero** behavioural probes as a result; the fix is to let the author
+    #: declare the domain, not to let the deriver invent it.
+    values: tuple[tuple[str, tuple[str, ...]], ...] = ()
+
+    def declared_values(self, field: str) -> tuple[str, ...]:
+        """The declared domain of ``field``, or empty when it carries data not a choice."""
+        return dict(self.values).get(field, ())
+
+    @property
+    def discriminator(self) -> str | None:
+        """The single required field whose value selects a behaviour, if there is one.
+
+        Exactly one, deliberately. Two discriminators would make the probe set a cross
+        product whose size the author never sees, and "which combinations are legal" is a
+        question the shape cannot answer — so a shape declaring two is treated as
+        declaring none, and the endpoint reports as uncovered (#951) rather than being
+        probed on a guess.
+        """
+        declared = [name for name, options in self.values if options and name in self.required]
+        return declared[0] if len(declared) == 1 else None
 
 
 @dataclass(frozen=True)
@@ -308,8 +336,25 @@ class InterfaceManifest:
             ],
             "api": {
                 "base_path": self.api.base_path,
+                # ``values`` is emitted ONLY when declared (#948). It belongs in the
+                # projection — it changes the derived contract's probe set, and a manifest
+                # whose contract differs must not share a hash with one whose contract does
+                # not. But emitting it unconditionally as an empty list would move the hash
+                # of every manifest ever written, invalidating each stored contract binding
+                # to record the absence of a field none of them use. Present-when-declared
+                # keeps both properties: existing manifests hash exactly as before, and a
+                # manifest that adopts the field hashes differently, which is correct.
                 "request_shapes": [
-                    {"name": s.name, "required": list(s.required), "optional": list(s.optional)}
+                    {
+                        "name": s.name,
+                        "required": list(s.required),
+                        "optional": list(s.optional),
+                        **(
+                            {"values": [[f, list(opts)] for f, opts in s.values]}
+                            if s.values
+                            else {}
+                        ),
+                    }
                     for s in self.api.request_shapes
                 ],
                 "endpoints": [
@@ -454,6 +499,11 @@ def _parse_api(raw: dict[str, Any]) -> Api:
             name=str(name),
             required=tuple(str(x) for x in (spec or {}).get("required", [])),
             optional=tuple(str(x) for x in (spec or {}).get("optional", [])),
+            values=tuple(
+                (str(field), tuple(str(v) for v in options))
+                for field, options in ((spec or {}).get("values", {}) or {}).items()
+                if options
+            ),
         )
         for name, spec in (raw.get("request_shapes", {}) or {}).items()
     )
--- a/src/squadops/capabilities/scaffold_contract.py
+++ b/src/squadops/capabilities/scaffold_contract.py
@@ -523,6 +523,56 @@ def _probes(manifest: InterfaceManifest, pack: CriteriaPack) -> list[dict[str, A
                             "expect": {"status": 409, "error_code": conflict_codes[0]},
                         }
                     )
+
+        # #948: the same child actions, expressed the OTHER legal way — one POST at the
+        # parameterized path, discriminated by a request-body field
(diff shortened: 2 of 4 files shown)


jasonLadd:
## End-to-end on roll 2's real manifest: 5 shells → 7

Probes are only half the chain — the shells derive from them, so the question that matters is whether roll 2 would actually have received deterministic tests for join and leave. It would:

```
BEFORE (roll 2 as authored)            AFTER (values declared)
  vc-probe-api-runs                      vc-probe-api-runs
  vc-probe-api-runs-rejects-blank        vc-probe-api-runs-rejects-blank
  vs-get-api-runs                        vc-probe-api-runs-join        <—
  vs-get-api-runs-run-id                 vc-probe-api-runs-leave       <—
  vs-get-api-runs-run-id-not-found       vs-get-api-runs
                                         vs-get-api-runs-run-id
                                         vs-get-api-runs-run-id-not-found
```

And the emitted join shell is a real in-process twin, not a placeholder — it creates a run, captures the id from the response, and drives the actual route handler with the declared action:

```ts
const createRes = await (routeApiRuns.POST as Handler)(
  new Request('http://test/api/runs', { method: 'POST', ... }),
)
const created: any = await createRes.json().catch(() => ({}))
const res = await (routeApiRunsRunId.POST as Handler)(
  new Request(`http://test/api/runs/${created.id}`, {
    method: 'POST',
    body: JSON.stringify({"action": "join", "name": "sample"}),
  }),
  { params: { run_id: created.id } },
)
expect(res.status).toBe(200)
const body: any = await res.json().catch(() => ({}))
// [scaffold-slot:begin slot-vc-probe-api-runs-join]
```

The chain holds end to end without any change to the shell generator: `_classify_probes` reads the parameterized path structurally and routes it down the existing child-with-capture path. **That is the argument for declaring values rather than relaxing the regex** — the shape was always derivable once a legal request body existed; only the body was missing.

Roll 2 would have gone from **zero** deterministic coverage of its central feature to two shells with fill slots, in the roll that banked green without them.

jasonLadd (commented):
Reviewed

Issue #947, opened 2026-08-16
qa self-eval pass is fill-blind: emits a whole file the scaffold guard discards, 68% of the qa task's wall clock


jasonLadd:
## What happens

The qa self-evaluation pass does not know the fill protocol exists. It asks the model for whole path-addressed files, the model complies, and the scaffold guard then throws the entire response away — by design, because slots are the only legal way to write a shell.

The prompt is the shared one at `src/squadops/capabilities/handlers/cycle/base.py:561-578`, and it says, verbatim:

> Please produce ONLY the missing files. Use the same fenced code block format (```language:path/to/file```). Do not reproduce files you already generated.

That is precisely the emission shape `qa_test.py:1115-1117` drops when the path is a merged shell.

## The cost, measured

Roll 1 of the SIP-0104 P6 window, `cyc_24aac7c22be1` / `run_0d8ffad5f128`:

| | tokens | fences | outcome |
|---|---|---|---|
| primary fill emission | 1,117 | `fill: 8, path: 0` | merged into the shell |
| self-eval emission | 3,574 | `fill: 0, path: 1` | **discarded** |

The self-eval call ran 20:12:30 → 20:18:11 UTC: **341 seconds of a 501-second task — 68% of the qa task's wall clock, for 3.2× the useful output, entirely thrown away.**

It cannot ever help in this shape. The only file it was asked to produce is the one file it is forbidden to write.

## Why the clean rolls are the ones that pay

The trigger inverts with quality. Today's earlier diagnostic (`cyc_588579756706`) emitted 5 fills **plus** one real additive file (`extracted_files=1`) and did not self-eval. Roll 1 emitted 8 fills and no file (`extracted_files=0`) and did. The more completely a task stays inside the fill protocol, the likelier it is to burn a wasted pass.

Note also that the prompt lists `**Files You Already Produced:**` from the post-merge artifact names — which *includes* `__tests__/runs.test.ts` — and instructs "do not reproduce files you already generated." The model reproduced it anyway, because the validation summary told it something was missing.

## The class

This is the seventh instance of one recurring shape: **a protocol taught to the primary path while a parallel path silently keeps the old behaviour.** Prior instances this window were the emission instrument, the fill-only appendix on the repair path, the generation record, and the build check's stale sources. The question that catches these at review time is *which other path does this concern have?*

## Options

[...]

PR #960, opened 2026-08-17
fix(qa): log which validation check opened the self-eval branch

### LINKED ISSUE #946: qa self-eval trigger is unlogged: no record of which validation check failed
## What happens

When a qa task's output validation fails, the handler runs a self-evaluation pass. Nothing records **which check failed**, so the trigger is invisible after the fact — the only surviving signal is that a second model call happened.

Roll 1 of the SIP-0104 P6 window (`cyc_24aac7c22be1` / `run_0d8ffad5f128`, qa task 16:10:19–16:18:40 ET) is the case. The task emitted 8 clean slot fills, ran a self-eval pass, and finished with the run's verification summary showing **29 verified, 0 failed, 0 unverified, verdict accepted**. So validation failed somewhere in the middle and the final state gives no hint what it was.

Reconstructing it from logs is not possible. `qa_test.py` logs the emission shape and the fill parse, but the `ValidationResult` — its summary and its per-check rows — is never logged, and the run-level summary only stores the roll-up after the self-eval and merge have already repaired whatever was wrong.

## Why it matters

This blocks diagnosing #NEXT (the self-eval pass emits a whole file that the scaffold guard discards by design). That defect's real fix depends on knowing whether the validation failure was genuine or an artifact-naming artefact of the fill merge, and right now neither can be distinguished.

More generally: a retry loop whose trigger condition is unlogged cannot be tuned. Every roll that runs a self-eval pass spends several minutes of GPU time for a reason no one can name afterwards.

## Evidence

- `src/squadops/capabilities/handlers/cycle/qa_test.py:1075` — `if not validation.passed:` is the sole trigger; no logging on this branch.
- `src/squadops/capabilities/handlers/cycle/qa_test.py:1120` — revalidation after the merge, same silence.
- Container log for the roll-1 qa task shows the fill emission at 20:12:30 UTC and the self-eval emission at 20:18:11 UTC with nothing in between.

## Proposed change

Log at INFO when the self-eval branch is taken, carrying `validation.summary`, `validation.missing_components`, and the `check` name of each failing row. Log the same on each revalidation so a multi-pass loop is readable.

This is instrumentation only — no behaviour change, and it should ship **alone**, ahead of any fix, so the next roll answers the question rather than a guess doing it (the standing "instrument before fixing" rule).

## Sequencing

Post-window. The P6 deploy is frozen; this is reporting-only until it closes.

jasonLadd:
`Closes #946`

The self-eval branch is the **sole** trigger for a second model call, and nothing recorded what opened it.

Window roll 1 spent **3,574 completion tokens** there — 68% of the qa task's wall clock, 3.2× its useful output — and the run summary afterwards showed 29/29 checks accepted. The failure was unreadable from stored state. The only way to learn the cause was to reconstruct it by hand from artifacts, which is how #947's mechanism came to be inferred **wrongly** the first time: roll 4 later reproduced roll 1's exact setup and produced the opposite emission, refuting the chain.

## What it logs

The failing checks **by name**, with the missing components and the summary. `1 check failed` sends a reader back to the artifacts, which is the state this exists to end.

Evidence-gap rows are excluded. They are honest non-passes the validator already refuses to fail the task on — a correction cannot repair an evaluator limitation — so naming one would send a reader chasing a check that caused nothing.

## Instrument only

No behaviour changes. No check added or removed. The branch condition is untouched.

That is deliberate, and it is the point: **#947's fix depends on how often `expected_artifacts` is the trigger**, and that turned out to be variance rather than determinism. Frequency is a measurement, not a deduction — this is what makes it measurable.

## Not required by the window

Unlike #948/#951/#915, this has no bearing on what a green roll *means*. Per §2.c of the pre-registration it should **wait until after V7** rather than enlarge the deploy boundary immediately before an evidentiary window. Opened now because the work is cheap and done; merging it is a separate call.

### DIFF
--- a/src/squadops/capabilities/handlers/cycle/qa_test.py
+++ b/src/squadops/capabilities/handlers/cycle/qa_test.py
@@ -43,6 +43,21 @@ from squadops.capabilities.handlers.emission_log import log_emission_shape
 logger = logging.getLogger(__name__)
 
 
+def failing_check_names(checks: list[dict]) -> list[str]:
+    """Which checks actually opened the self-eval branch (#946).
+
+    Evidence-gap rows are excluded deliberately. They are honest non-passes that
+    ``_validate_output`` already refuses to fail the task on — a correction cannot repair
+    an evaluator limitation — so naming one as the trigger would send a reader chasing a
+    check that did not cause anything.
+    """
+    return [
+        str(c.get("check", "unnamed"))
+        for c in checks
+        if not c.get("passed", True) and not c.get("evidence_gap", False)
+    ]
+
+
 def _frontend_skip_reason(error: str) -> str:
     """Map a frontend-build skip error to a §7 not-executed reason (#407).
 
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
Reviewed

PR #961, opened 2026-08-17
docs(plan): Guard 1a and 1b are built and green — the plan said otherwise

jasonLadd:
Both guards landed **2026-08-15 in `07edfcc0`** and are green — ten tests. The 1.6 plan still recorded them as **NOT BUILT** in three places:

| location | said |
|---|---|
| Stage 3, steps 3a / 3b | listed as work to do |
| "what replaces the seeded control" table | `**NOT BUILT**` |
| **Cut-gate item 5** | "both of which must be *built* first … verified unbuilt 2026-08-09" |

A cut gate that lists satisfied work as outstanding is the same defect as one that lists outstanding work as satisfied. It just fails in the safe direction, which is why it survives longer — nobody is harmed by it until someone plans around it.

## What the guards actually assert

- **1a (structure)** — `test_authoring_mode_no_fork.py`: the authoring predicate has exactly one production call site and it is inside the framing branch. Before this, the release's architectural claim was defended by there *currently* being only one place that asks the question.
- **1b (output)** — `test_authored_mode_equivalence.py`: the reference manifest through both provenances yields a byte-identical expanded skeleton, a byte-identical derived contract, and an unmoved structural hash — **plus a negative control** proving a real structural change still moves it. Without that control the test could pass by hashing nothing.

The two are not redundant: a structural test proves nothing *branches*, not that the transformation is the same. A fork can also arrive as a field that only authored manifests carry leaking into a hash.

## The retirement argument is untouched

The seeded control's retirement never depended on Guard 1b — it is structural, and the guard's absence was a **coverage caveat** rather than a hole in the reasoning. That caveat is now closed, so the release no longer has to make the argument while carrying it.

## Evidence matrix

Adds the row, under the standing rule that a PR correcting a premise updates the matrix in the same PR. Its correction column records the staleness itself: the plan carried the claim for a day after the code landed — exactly the propagation failure the matrix exists to make visible.

Docs only. No code. Version-sync guards green.

### DIFF
--- a/docs/plans/1-6-0-authorship-plan.md
+++ b/docs/plans/1-6-0-authorship-plan.md
@@ -304,8 +304,8 @@ until 2b exists.
 
 | Step | Work |
 |---|---|
-| **3a** | **Guard 1a** architecture test — nothing branches on authoring mode below framing. The invariant holds today at one call site (`task_plan.py`); nothing pins it |
-| **3b** | **Guard 1b** — reference manifest through authored mode, byte-identical downstream artifacts |
+| **3a** | **Guard 1a** architecture test — nothing branches on authoring mode below framing. **DONE 2026-08-15** (`07edfcc0`, `tests/unit/cycles/test_authoring_mode_no_fork.py`) — the predicate's single production call site is now pinned inside the framing branch, where before nothing stopped a second one appearing |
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
Reviewed

PR #962, opened 2026-08-17
fix(qa): record what the test runner actually did

### LINKED ISSUE #935: qa.test leaves no INFO record of whether the suite actually ran
-

jasonLadd:
`Closes #935`

Whether the suite actually **ran** was recoverable only from an artifact a human had to open.

`tests_pass` is exit-code driven, so *"did not execute"* and *"executed and passed"* sit one keystroke apart in the stored record — and neither left a log line. When a window roll needs triage, the first question is what the runner did, and the answer should not require archaeology.

## What it logs

Framework, `executed`, exit code, `tests_passed`, both file counts, the uncollected list, and the error. Same class as #926's emission capture, which is what ended the last round of artifact archaeology.

**`uncollected` is logged beside the counts deliberately.** A file the runner never collected verifies nothing while the collected ones read green — SIP-0104 roll 1 shipped exactly that — and a count of test *files* cannot distinguish the two. The count and the exception have to appear together, or the count lies.

## Instrument only

No behaviour changes. A test pins that the `test_report.md` artifact a human opens is unmoved.

## Not required by the window

Like #946, this has no bearing on what a green roll *means*. Per §2.c of the pre-registration it should **wait until after V7** rather than widen the deploy boundary before an evidentiary window. Opened now because the work is done; merging it is a separate call.

### DIFF
--- a/src/squadops/capabilities/handlers/cycle/qa_test.py
+++ b/src/squadops/capabilities/handlers/cycle/qa_test.py
@@ -335,6 +335,29 @@ class QATestHandler(_CycleTaskHandler):
             timeout_seconds=capability.test_timeout_seconds,
         )
 
+        # #935: whether the suite actually RAN was recoverable only from an artifact a
+        # human had to open. `tests_pass` is exit-code driven, so "did not execute" and
+        # "executed and passed" are one keystroke apart in the record and neither leaves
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
Reviewed

PR #963, opened 2026-08-17
docs(plan): the SIP-0104 P6 window record

jasonLadd:
V6's evidence has lived in issue comments, PR bodies and session memory. This is the durable artifact: baseline, per-roll numbers, what the window established, and what it did not.

## Three things it states that would otherwise be inferred generously

**28 of 28 slots filled first try, zero corrections** means the scaffold holds *when the author gets it right*. It does **not** mean the system recovers when the author gets it wrong — the repair path (#942/#943) is unexercised in production, and six clean rolls leave that exactly as untested as four do. Written down so "no corrections needed" cannot later read as "corrections work."

**Coverage varied 5 / 2 / 5 / 4 / 5 at an identical recipe hash**, through two distinct authoring mechanisms, neither of them machinery drift — roll 2 folded join and leave into one body-discriminated endpoint (#948); roll 4 declared no error code mapping to 409, so no duplicate probe derived from a perfectly derivable manifest. **The squad authors the exam it sits.** That is why V7 records contract size per roll and reports a distribution rather than a mean.

**Gate handling was not uniform.** Rolls 1–3 carry an operator approval, roll 4 auto-approved via `system:no_open_questions`, rolls 5–6 carry the pre-registered constant. The closing claim says so rather than implying one policy. (Note the irony on roll 4: "no unresolved decisions" was true only of *declared* uncertainty — what it left unspecified was the duplicate-join rule, whose absence cost it a probe.)

## Two corrections to the ledger

- **Roll 1's audit pass was weaker than banked** — #952 meant it passed on 2 of 5 UI call sites. Under the corrected instrument it passes on 5 of 5.
- **My first diagnosis of #947 was wrong** and is corrected on the issue; roll 4 reproduced roll 1's setup exactly and produced the opposite emission.

**No roll is re-scored.** A roll is never re-graded on an instrument change after the fact — that is the same error pointing the other way. The record names which rolls were audited by which version of the check.

## Leaves 1e open, deliberately

1e credits "the first roll that passes the audit including the UI data-path check." Before #952/#953, no roll cleanly met that bar — roll 1's check was vacuous, rolls 3 and 4 were failed by the tool and passed by a human. Under the corrected instrument roll 1 passes on all five. **The evidence moved after the rule was written, so crediting it is a ruling, not bookkeeping.**

Docs only. Rows for rolls 5 and 6 fill as they bank.

### DIFF
--- a/docs/plans/sip-0104-p6-window-record.md
+++ b/docs/plans/sip-0104-p6-window-record.md
@@ -1,0 +1,126 @@
+# SIP-0104 P6 window — the record
+
+**Purpose.** V6, the viability run that gates V7 (plan step 4b). Six rolls of authored-mode
+`group_run` cycles on stack `nextjs_ts`, against a frozen deploy, measuring whether the
+deterministic verification scaffolding holds in production.
+
+**This is not V7.** P6 rolls cannot be promoted into the FAY window: designating a window
+after its results are visible is exactly what pre-registration exists to prevent. See
+`1-6-0-v7-fay-window-preregistration.md` §9.
+
+---
+
+## Baseline
+
+| | |
+|---|---|
+| Deploy — main commit | `d590f73c`, **frozen for the window's duration** |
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
Reviewed

PR #964, opened 2026-08-17
fix(verification): name the contract criteria a run did not verify

### LINKED ISSUE #945: Two contract criteria were neither verified nor disclosed — the roll-up lost them silently
-

jasonLadd:
`Closes #945`

SIP-0104 window roll 1 reported **"Contract criteria: 12/14"** with an **empty** unverified list. The two it lost — `vc-compiles-app-api-runs-run-id-route` and its join sibling — were recoverable only by diffing two stored lists by hand.

The count was honest. The *identity* of the gap was available nowhere.

That is the defect: a reader seeing a shortfall cannot tell a compile criterion for a file that shipped fine from the one behaviour the product exists for. On roll 1 the two lost criteria were on exactly the route the app's core feature lives on.

## The invariant

`criteria_verified` and `criteria_unverified` together account for **every** entry of `criteria_total`, on both the per-run summary and the cycle roll-up.

**Derived, not stored** — for the same reason `required_unmet` already is. A stored copy can disagree with the lists it summarises, and this field's entire job is to prove nothing was dropped between them.

## Surfaced where it gets read

- **The run report** names the missing criteria beneath the count. `12/14` alone sends a reader to the artifacts.
- **`CycleOutcome.to_dict`** carries them, so a consumer reading only the wire shape doesn't have to rediscover the shortfall by set arithmetic — which is precisely what nobody did on roll 1.

## Replayed against the real stored summary

Not a fixture exercise — this is roll 1's actual row, read out of Postgres and run through the new derivation:

```
roll 1 stored: 12/14
NOT verified : ('vc-compiles-app-api-runs-run-id-join-route',
                'vc-compiles-app-api-runs-run-id-route')
```

## How often it bites

Measured across all five banked rolls: **once in five.** Rolls 2–5 account for every criterion. Rare, silent, and on the roll it hit it took the app's central route with it — which is the argument for the invariant rather than for chasing the specific cause.

7786 unit tests green.

## Window status

This bounds what a green roll *means* — a roll can lose criteria and still read complete — so unlike #946/#935 it has a claim on being in the pre-V7 boundary. Your call. **Does not merge until the P6 window closes** regardless; deploy frozen at `d590f73c`.

### DIFF
--- a/src/squadops/cycles/run_report_builder.py
+++ b/src/squadops/cycles/run_report_builder.py
@@ -148,6 +148,15 @@ def _build_verification_lines(summary: RunVerificationSummary) -> list[str]:
     if summary.criteria_total:
         n, m = summary.criteria_coverage
         lines.append(f"Contract criteria: {n}/{m} executed-and-passed")
+        # #945: name the shortfall. Window roll 1 rendered "12/14" with an empty
+        # unverified list, and the two missing criteria were recoverable only by
+        # diffing the lists by hand. A reader seeing a gap cannot otherwise tell a
+        # compile criterion for a file that shipped fine from the one behaviour the
+        # product exists for.
+        if summary.criteria_unverified:
+            lines.append(
+                f"Contract criteria NOT verified: {', '.join(summary.criteria_unverified)}"
+            )
     if summary.failed:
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
Reviewed

PR #965, opened 2026-08-17
fix(cli): print data as itself, never as Rich markup

### LINKED ISSUE #931: CLI artifact table silently deletes bracketed path segments (Next.js dynamic routes)
-

jasonLadd:
`Closes #931`

Rich reads `[...]` in a string as a style tag and **removes** it. Every value the CLI prints is data, so a Next.js dynamic route arrived on screen with its parameter segment silently deleted.

**Same command, same run, before and after:**

```
before   app/api/runs//route.ts        app/api/runs//leave/ro…
after    app/api/runs/[run_id]/…       app/api/runs/[run_id]/…
```

## Why this is worse than a display bug

The artifact table is the **first surface a person reads when triaging a cycle**, and it was showing a filename that does not exist — while looking authoritative. Nothing indicated a path had been rewritten between the store and the terminal.

Stack #2 puts every parameterised route behind a bracketed directory, so this hit precisely the routes the application's core behaviour lives on, during the window where those routes were under investigation. Anything bracketed had the same exposure: a JSON list of artifact ids, a pasted log line, a regex.

## Two surfaces, one cause

| surface | how it leaked |
|---|---|
| `print_table` | cells interpolated straight into `add_row` |
| `print_detail` | the **value** interpolated into the same f-string as its own `[bold]` label |

Both now render values through `rich.text.Text`. The detail label keeps its emphasis — the label is ours and may carry markup; the value is data and never may.

## Tests worth naming

- **Quiet mode is pinned unchanged.** It never went through Rich, so it never had the bug, and it must not acquire chrome now.
- **A value that *is* `[bold red]`** must render as those characters rather than becoming a style and vanishing. That is the dangerous half — the failure mode is silent deletion, so the test has to assert the characters survive rather than that nothing crashed.

228 CLI tests green.

**Does not merge until the P6 window closes** — deploy frozen at `d590f73c`. This one is display-only and touches no cycle machinery, so it is a candidate for landing early if you want the triage surface fixed before the next window.

### DIFF
--- a/src/squadops/cli/output.py
+++ b/src/squadops/cli/output.py
@@ -11,9 +11,26 @@ from typing import Any
 
 from rich.console import Console
 from rich.table import Table
+from rich.text import Text
 
 _console = Console()
 _err_console = Console(stderr=True)
+
+
+def _literal(value: Any) -> Text:
+    """Render a data value as itself, never as Rich markup (#931).
+
+    Rich reads ``[...]`` in a string as a style tag and **removes** it. Every value the
+    CLI prints is data, not markup, so a Next.js dynamic route arrived on screen as
+    ``app/runs//page.tsx`` — the segment silently deleted, with nothing to indicate a
+    path had been rewritten between the store and the terminal.
+
+    That is worse than a display bug. The artifact table is the first surface a person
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
Reviewed

PR #966, opened 2026-08-17
docs(plan): V7 2.b dispositions built; record the deploy-boundary risk (2.c)

jasonLadd:
One commit stranded when #955 merged. It was pushed to that branch *after* the merge, so it has been sitting on a closed PR with nothing tracking it.

Adds **§2.c** — the deploy-boundary risk the 2.b ruling created and nothing recorded — and marks the three ruled dispositions as built (#957, #958, #959).

## §2.c, in short

Ten PRs now sit ready for the single rebuild separating the P6 window from V7. Each is justified and five are preconditions this document itself imposes. Together they are a **large** boundary immediately before an evidentiary window, and §5.1 resets that window on a new mechanical failure attributable to the harness. A defect introduced by this batch and found at roll 4 costs the whole window.

Two consequences, neither an argument against the fixes:

1. **Merge what this document requires, and nothing else.** The logging fixes are real defects with no bearing on what a green roll *means*.
2. **Deploy-prove in-container before roll 1** — the loaded module, not the rebuild's exit code. It has exited zero with stale agents before.

Docs only.

### DIFF
--- a/docs/plans/1-6-0-v7-fay-window-preregistration.md
+++ b/docs/plans/1-6-0-v7-fay-window-preregistration.md
@@ -31,8 +31,8 @@ is a separate, deliberate act taken after the window closes.
 | # | Precondition | State |
 |---|---|---|
 | 2.1 | V6 complete (SIP-0104 P6 window closed) | in progress |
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
Reviewed

PR #973, opened 2026-08-17
fix(scaffold): derive and type the store's table names

### LINKED ISSUE #967: The store's table name is an undeclared literal — two agents must guess it identically, and roll 6 rejected a working application because they did not
The in-memory store's table name is an **undeclared free string**, and two independently-authored artifacts have to guess it identically. When they disagree, a **working application is rejected**.

Found in SIP-0104 P6 window roll 6 (`cyc_085f992ad60d` / impl `run_dd4b9b23fd78`), which failed after three correction rounds and burned its whole budget.

## The application was fine

Measured on the delivered workspace with `audit_delivered_app.py`:

```
assembled 27 files for stack nextjs_ts (acceptance-aware last-wins)
PASS — delivered app installs, builds, boots, answers 5 contract probe(s),
and its UI reaches every path it requests
```

All five contract probes answer over real HTTP — create, blank-rejection, join, duplicate-join, leave. The UI reaches every path it requests.

## The defect

`lib/store.ts` is scaffold-owned, frozen, and **generic**:

```ts
const tables: Record<string, Record<string, unknown>[]> = {}
export function all(table: string) { return tables[table] ?? [] }
export function insert(table: string, row: ...) { (tables[table] ??= []).push(row); return row }
```

The table name is a parameter. **Nothing declares it:**

- the interface manifest says `persistence: in_memory` and nothing further;
- the verification contract freezes `lib/store.ts` and binds no key.

Roll 6's application chose `'run_store'` and used it consistently in every route — `insert('run_store', run)`, `all('run_store')`, `find('run_store', params.run_id)`. Its test suite asserted store state under a different name, read `[]`, and reported `expected [] to have a length of 1 but got +0`.

**Rolls 1–5 guessed the same name.** That is the whole reason this had not been seen: it is a coin flip, and it had come up heads five times.

## Why this one is different from the rest of the family

Third sighting after the `join` / `leave` path literals a qa author invented from prose and agreed with dev by luck, and after #913. But every other instrument defect we track bounds a result by letting something through **unverified**. This one bounds it in the opposite direction:

**It fails applications that work.** It is the mirror of the false-green class this entire verification line exists to eliminate, and it can *depress* the Functional App Yield figure rather than inflate it.

Related but not duplicates: #772 (an undeclared `success_status` producing an unwinnable contract — same shape, different field), #948 (undeclared request-shape values), #913.

## Fix direction

The same shape as #948's resolution: **let the declaration carry the name.** Options, in order of how much they change:

1. **Derive it.** The entity name is already in the manifest; make the table name a function of it and teach both the expander and the fill brief the same derivation. Cheapest, and removes the guess entirely rather than moving it.
2. **Declare it.** An explicit field on the entity or the persistence block, rendered into both the app's brief and the qa brief.
3. **Bind it.** A contract criterion pinning the table name, so a mismatch fails as a contract violation naming the key rather than as an empty array.

(1) is preferred: an undeclared string that two agents must match is a defect class, not a configuration gap, and the fix that removes the choice beats the fix that documents it.

## Bearing on V7

This has a claim on being a **blocking precondition** rather than post-window cleanup, on the same argument §2.a of the pre-registration makes for #952/#953: a window measuring whether the system produces working software must not fail working software for reasons internal to the verifier. Owner ruling wanted.

jasonLadd:
`Closes #967`

The store took a free `string`. The application named its table one thing, its test suite named another, and the disagreement surfaced as an **empty array at assertion time** — indistinguishable from a handler that never persisted.

SIP-0104 window roll 6 spent three correction rounds and its entire budget on this. All three failure analyses blamed the application. **The application worked** — it installed, built, booted, and answered all five contract probes over real HTTP.

## The fix removes the choice rather than documenting it

```ts
export const TABLES = { Participant: 'participant', Run: 'run' } as const
export type Table = (typeof TABLES)[keyof typeof TABLES]
export function all(table: Table): Record<string, unknown>[]
```

**Proven against real tsc in the sandbox image, not asserted:**

```
probe_wrong.ts(2,22): error TS2345: Argument of type '"run_store"' is not
                      assignable to parameter of type 'Table'.
```

`run_store` is roll 6's actual key. The `TABLES.Run` form produces no error. And `next build` runs tsc with `ignoreBuildErrors: false`, so a mismatch now fails **`frontend_build`, before the suite runs** — a loud build failure naming the legal values, instead of an assertion failure that reads as a broken handler.

## Details worth a reviewer's attention

**`Table` derives from `TABLES` rather than restating it.** A literal union beside the constant is two declarations that must agree, and the next author updates one. Deriving makes them incapable of disagreeing.

**The harness no longer needs an escape hatch.** It used `'__probe'`, which a typed union rejects. Reserving a member for it was the tempting fix and is exactly the hatch this removes — a reserved name is still a legal string the application could pick. It now addresses a real table through `TABLES`, which doubles as a worked example of the accessor the briefs point authors at.

**No entities → open signature.** `never` would make the store unusable rather than safe, and with no entities there is nothing to disagree about. The harness falls back with it, since emitting `TABLES.x` against a store that exports none would not compile.

## The frozen-surface index needed no change

It already derives from source, and now publishes:

```
`lib/store.ts` — defines `Table`; exports TABLES = { Participant: 'participant', RunEvent: 'run_event', };
functions `reset(): void`, `all(table: Table): …`
```

Two inventory tests pinned `table: string`; they're updated **and strengthened**. Publishing `table: Table` without the members would be roll 8's defect one level down — a type the author can see and cannot resolve. The tests now require the mapping to appear literally.

## The pins, and what they proved

`GENERATOR_VERSION` 2 → 3 with the fixture regenerated, per SIP-0104's documented one-deliberate-act procedure.

**Both pinned hashes are unchanged**, and that is the informative part: the shells are byte-identical. The only fixture byte that moved is the manifest's `expanded_tree_hash`, which the byte test caught. So the pins distinguished *"the tree changed"* from *"shell emission drifted"* — a distinction a single combined hash could not express.

## Verification

- 13 new tests, including the derivation pinned per entity and the no-entity fallback.
- Regression green: **7478**; full unit **7851**.
- Real tsc run against an expanded roll-6 workspace, both probe directions.

## Deploy note

This changes the expanded tree, so it needs a rebuild before the next window opens — and it is the precondition ruled ahead of V7.

### DIFF
--- a/src/squadops/capabilities/stack_nextjs_ts.py
+++ b/src/squadops/capabilities/stack_nextjs_ts.py
@@ -246,7 +246,13 @@ export default defineConfig({
 })
 """
 
-_HARNESS_TEST = """// Scaffold-owned harness (SIP-0100). QA suites consume the seams this proves, and must
+_HARNESS_TEST_HEAD = """// Scaffold-owned harness (SIP-0100). QA suites consume the seams this proves, and must
+// not reach past them into the app entry — the boundary that killed pf-25/26 on stack #1.
+import { describe, it, expect } from 'vitest'
+import { reset, insert, all, TABLES } from '@/lib/store'
+"""
+
+_HARNESS_TEST_NO_ENTITIES = """// Scaffold-owned harness (SIP-0100). QA suites consume the seams this proves, and must
 // not reach past them into the app entry — the boundary that killed pf-25/26 on stack #1.
 import { describe, it, expect } from 'vitest'
 import { reset, insert, all } from '@/lib/store'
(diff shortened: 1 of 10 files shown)


jasonLadd (commented):
Reviewed

PR #974, opened 2026-08-17
test(2c): the blueprint falsification pass, as a standing gate

jasonLadd:
Stage 2c, the step the plan calls *"most likely to bite"*. It bit.

S3's exit criterion: **corrupt each blueprint field and confirm something breaks in at least one stack; a field whose removal breaks nothing is decorative and should be deleted before the schema freezes.**

Executable and derived over the six registries, for 2a's reason — a read-and-classify pass fails the way reading fails, and 1a overclaimed three times out of four bounded items. A *new* field with no consumer fails here rather than entering the schema unnoticed.

### Three have no consumer of any kind

Traced twice — an AST reader search and plain grep — including `getattr` access and prompt assets.

| field | evidence | what the schema draft says |
|---|---|---|
| `BuildProfile.artifact_output_mode` | declared on all five profiles, **zero reads** | Tier 3, hoisted as a core default |
| `BuildProfile.validation_rules` | real content on every profile, **zero reads** | Tier 1 "demonstrated" |
| `DevelopmentCapability.expected_extensions` | **zero reads** — and **two docstrings assert it is "what a dev agent is given"** (`scaffold.py:1899`, `preflight.py:216`) | Tier 1 "demonstrated" |

The third is the sharpest: not merely unread, but *documented as read*.

**All three are fields the draft calls demonstrated or hoistable.** Recorded in the test rather than deleted in it — deletion is a code change with its own review, and the plan assigns it to before the freeze, i.e. before 2g. A pinning test fails if one gains a consumer, so the list cannot decay into a stale exemption.

### One has a consumer and no data

`BuildProfile.default_task_tags` is `{}` on all five profiles, and its reader merges it with `experiment_context`, which supplies every tag in practice. Tier 3 calls it "identical across both stacks" — true, and identical **because empty**, which is a different fact. Under S5's admission rule read subtractively, a field zero stacks demonstrate must not enter the blueprint.

### `check_stack` is falsified, at the acceptance layer

Worth recording how it was nearly mis-called: the first attempt used `undefined_names`, which ignores `stack` entirely and returns the same verdict for the declared value, a corrupted one, and `None`. **Four evaluators do read it**, and against `endpoint_defined` a corrupted or unset value turns `passed` into `skipped: unsupported_stack_or_syntax` — unverified under SIP-0096. Picking the wrong evaluator would have falsified the draft's claim instead of the field.

## Four harness defects, each of which changed the answer

The most important part of this PR, and the reason the write-up carries them:

1. **Corrupting an already-empty value** proves nothing — the first run reported six decorative candidates, five of them this artifact.
2. **`name` fields are registry keys** — patching the stored copy changes no lookup unless the observation resolves the key through the declaration.
3. **Observing the declarations instead of what consumers derive.** The battery serialised each dataclass's own fields, so any change trivially "differed". This reported **98 of 104 falsified** and let a deliberately decorative probe field pass. That number was meaningless.
4. **An attribute-only reader search** cannot see `getattr(cap, "build_support_files", ())`, and reported that live field as decorative.

A **negative control** now injects a decorative field and requires the gate to catch it. Before defect 3 was fixed, it did not.

## What the pass does not exercise, stated rather than implied

Fields whose only consumer is a runtime path — boot, sandboxing, prompt rendering — are falsified by that consumer, each with a cited production read, and the test reports the set. The honest claim is *"every field is either falsified offline or has a reader that can be cited"*, not *"every field is falsified"*.

## Verification

109 tests here; **7973 unit / 7587 regression green.** Independent of #973 — no overlap, passes on main alone.

### DIFF
--- a/docs/plans/stack-blueprint-schema-draft.md
+++ b/docs/plans/stack-blueprint-schema-draft.md
@@ -176,6 +176,44 @@ section as the declared successor design and the pack-combo test as its trigger.
 
 ---
 
+## 4d. 2c — the falsification pass, run 2026-08-17
+
+**Where it lives.** `tests/unit/capabilities/test_stack_blueprint_falsification.py`, executable and derived over the six registries — the same reasoning 2a used, for the same reason: a read-and-classify pass fails the way reading fails, and 1a overclaimed three times out of four bounded items.
+
+**Result: 101 of 107 field-checks falsify.** Corrupting the field either makes a real path raise or moves a derived artifact. The remaining six are four distinct fields, and they are 2c's product.
+
+### Three fields have no consumer of any kind — decorative, per S3's own definition
+
+| field | evidence | what this draft says about it |
+|---|---|---|
+| `BuildProfile.artifact_output_mode` | declared on all five profiles, **zero reads** | Tier 3 — hoisted as a core default |
+| `BuildProfile.validation_rules` | real content on every profile, **zero reads** | Tier 1 `packaging.validation_rules` — "demonstrated" |
+| `DevelopmentCapability.expected_extensions` | populated per stack, **zero reads**, and **two docstrings assert it is "what a dev agent is given"** (`scaffold.py:1899`, `preflight.py:216`) | Tier 1 `authored_extensions` — "demonstrated" |
+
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
Reviewed

PR #975, opened 2026-08-17
docs(plan): V7 preconditions met; #967 added as blocking (2.7 / 2.d)

jasonLadd:
Three preconditions move to **MET**, dated and evidenced:

| # | was | now |
|---|---|---|
| 2.1 V6 complete | in progress | **MET** — window closed 2026-08-17 03:46 ET at six rolls (5 banked, roll 6 rejected) |
| 2.2 #952/#953 | fixed, not deployed | **MET** — merged and deployed, loaded-module verified in runtime-api, eve, neo |
| 2.3 instrument defects | built, none deployed | **MET** — all three merged and deployed, verified |

`2.4` stays pending on purpose: the deploy was rebuilt again for #967, so the commit and image ids belong at the freeze rather than now.

## Adds 2.7 and 2.d — #967 as a blocking precondition

Owner-ruled after roll 6. The argument is worth stating because **it inverts every other item in §2.b:** those bound the number by letting something through *unverified*. This one **fails applications that work.**

Roll 6 installed, built, booted, and answered all five contract probes over real HTTP — and was rejected because its test suite and its application guessed different names for an undeclared store table. Rolls 1–5 guessed alike. That is a coin flip sitting under the measurement, and it can only push the figure down, never up.

Fixed in #973 by deriving the names from the manifest's entities and typing them, so a mismatch is a compile error `next build` fails on, before the suite runs.

## One consequence for the closing claim

A bare "N of 6" invites the reading that the missing rolls shipped broken software. **Roll 6 did not**, and neither would a future roll failing this way. The closing record states what each failure *was*, not only how many there were — which is precisely the category error roll 6's own three failure analyses made, three times in a row.

Docs only.

### DIFF
--- a/docs/plans/1-6-0-v7-fay-window-preregistration.md
+++ b/docs/plans/1-6-0-v7-fay-window-preregistration.md
@@ -30,10 +30,11 @@ is a separate, deliberate act taken after the window closes.
 
 | # | Precondition | State |
 |---|---|---|
-| 2.1 | V6 complete (SIP-0104 P6 window closed) | in progress |
-| 2.2 | **#952 and #953 fixed and deployed** — see §2.a, this is blocking | fixed in PR #956; **not deployed** |
-| 2.3 | Instrument defects either fixed or explicitly declared (§2.b) | **ruled 2026-08-16: fix all three**; all three built (#957/#958/#959), **none deployed** |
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
Reviewed

PR #977, opened 2026-08-17
docs(2f): S5's admission rule is enforced by the 2c gate, not just written

jasonLadd:
S5's rule — *"a new blueprint field must be demonstrated on at least two stacks before admission"* — lived **only** in `1-6-0-authorship-plan.md`. A release plan is superseded at the cut, which is the failure CLAUDE.md names and SIP-0103 §5d already paid for. The rule was one release from vanishing.

It turns out **2c's gate already enforces it**, and enforces S5's own escape valve with it.

| a new field… | gate outcome |
|---|---|
| populated on **both** stacks | falsified on each → admitted |
| populated on **one** stack | fails until a reason is recorded, *and* some stack must populate it |
| populated on **no** stack | rejected — an optional nothing populates is unevidenced, not optional |
| read by nothing | rejected outright |

**Verified rather than asserted.** Removing `harness_entry_modules`' recorded reason fails with:

> `ScaffoldStack.harness_entry_modules is unset on nextjs_ts with no recorded reason. Either record why it is optional (and confirm another stack populates it), or delete it.`

That is S5's stated consequence — a one-stack need expressed as a declared optional capability with its reason — enforced rather than requested.

Recorded in the test's docstring because *what a test enforces stays true; what discipline enforces drifts*, which is `test_docs_version_sync`'s own rationale for living where it does.

## What remains for 2f

**The statement, not the mechanism.** The rule's permanent home is the Stack Blueprint SIP, which 2g writes. Until then the enforcement outlives the plan even if the prose does not — the reverse of the usual failure, and the safer direction to be caught in.

Docs and one docstring. 109 tests green.

### DIFF
--- a/docs/plans/stack-blueprint-schema-draft.md
+++ b/docs/plans/stack-blueprint-schema-draft.md
@@ -214,6 +214,25 @@ Recorded because the numbers moved every time and the first ones were wrong: cor
 
 ---
 
+## 4e. 2f — S5's admission rule is enforced, not just written *(2026-08-17)*
+
+S5's rule lived only in `1-6-0-authorship-plan.md`. A release plan is superseded at the cut, which is the failure CLAUDE.md names and SIP-0103 §5d already paid for — so the rule was one release from vanishing.
+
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
Reviewed

Issue #939, opened 2026-08-16
undefined_names is .py-only — the nextjs_ts stack has no unresolved-name guard


jasonLadd:
-

jasonLadd:
## The title may overstate the gap — worth confirming before building anything

Read the code before scoping a TS unresolved-name checker. Stack #2 already has a static tier, and the question is how much of it is left uncovered.

**`next build` runs `tsc`, deliberately, and it is not optional:**

```js
// stack_nextjs_ts.py:225 — _NEXT_CONFIG
// Type errors fail the build deliberately: `next build` runs tsc, and it is stack #2's
// entire hygiene tier — the nine AST checks parse Python only.
const nextConfig = { typescript: { ignoreBuildErrors: false } }
```

And the emitted `tsconfig.json` (`stack_nextjs_ts.py:219`) includes **everything**, tests included:

```json
"include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
"exclude": ["node_modules"]
```

An unresolved name in TypeScript is a compile error under `strict: true`, not a call-time `NameError` — which is the whole reason #689 needed pyflakes for Python. So the defect class that motivated `undefined_names` looks structurally covered on this stack **by `frontend_build`**, not absent.

## What that leaves

Two narrower questions, and I have not settled either:

1. **Are emitted test files actually type-checked in practice?** They are inside the tsconfig `include` glob, but `next build` rewrites the tsconfig at build time — injecting `allowJs` and `skipLibCheck` (`sandbox/environment.py:172`, measured on window roll 1) — and I have not confirmed whether Next's type-check step honours the full `include` set or only the build graph. `test_runner.py:1030` already points here: *"see #939 on typechecking emitted tests."*

2. **Attribution, not detection.** A TS unresolved name fails `frontend_build`, which is a whole-project check. `undefined_names` on Python names the *file*, so the failure lands on the task that wrote it. That difference is real and matters for the correction loop — but it is a *diagnosability* gap, not a coverage one, and it should be scoped as such.

## Recommendation

Re-title toward what is actually missing once (1) is answered, and treat this as **post-V7** either way. Per the pre-registration's §2.c the deploy boundary before the FAY window should carry only what the measurement requires, and this is not on that list.

Marking this as read-from-source rather than verified by execution: I have not run a build against a workspace with a deliberately unresolved name in a test file. That experiment is the cheap way to settle (1) and I would do it before writing any code here.

jasonLadd:
[...]

PR #978, opened 2026-08-17
docs(2e): the two disclosures, and the vocabulary reconciliation is three axes

jasonLadd:
Stage 2e. Both disclosures restated as promotion must state them, and the vocabulary question answered — **differently from how it was posed.**

## The question was wrong

It was recorded as *"two stack vocabularies remain two — `fullstack_fastapi_react` vs `fastapi`. S1 removed the duplicate declaration, not the drift."* Measured, there are **three axes** and two of them are not drift:

| axis | stack #1 | stack #2 | verdict |
|---|---|---|---|
| stack identity | `fullstack_fastapi_react` | `nextjs_ts` | **one value across five registries** — S1's consolidation holds |
| probe profile | `fastapi_uvicorn` | `nextjs_next_start` | **not drift** — differs from the stack id on *both* stacks, consistently. Booting is its own axis, and two stacks could share a boot mechanism while sharing nothing else. Naming it after the stack would be the error |
| `check_stack` | `fastapi` | *(unset)* | **the real drift** |

## And `check_stack` is worse than a naming smell

It carries two different questions on one field:

| evaluator | guard | what it needs |
|---|---|---|
| `endpoint_defined` | `if stack != "fastapi"` | genuinely FastAPI — it reads `@router.get` |
| `field_present` | `if stack is None` | any declared dialect |
| `function_defined` | `if stack is None` | any declared dialect |
| `harness_boundary` | `if stack is None` | any declared dialect |

A third stack declaring `check_stack: "flask"` would skip `endpoint_defined` **correctly** and pass the other three **correctly** — *by accident*, because one field is answering "which framework's idioms?" and "is any dialect declared?" at once. **The accident holds only while exactly one framework-specific evaluator exists.**

**Recommendation for 2g to state rather than imply:** the blueprint carries the *language* dialect, and a framework-specific evaluator names its own requirement. That touches the evaluator contract, so it is a 1.7 change — 2g discloses the conflation with its evidence rather than promoting `check_stack` as though it were one axis.

## Cross-check that changes an open issue

Disclosure #6 already said `next build` runs tsc and `next.config.mjs` declines to ignore type errors — **the bundler check is the type check.** #939 was filed as *"the nextjs_ts stack has no unresolved-name guard"* and my independent read reached the same conclusion, so **#939 is largely redundant with a recorded disclosure.** Its honest residue is *attribution*: a whole-project build failure does not name the task that wrote the bad file, where a per-file AST check would.

Docs only.

### DIFF
--- a/docs/plans/stack-blueprint-schema-draft.md
+++ b/docs/plans/stack-blueprint-schema-draft.md
@@ -214,10 +214,47 @@ Recorded because the numbers moved every time and the first ones were wrong: cor
 
 ---
 
+## 4f. 2e — the disclosures and the vocabulary reconciliation *(2026-08-17)*
+
+### The two disclosures, as they must be stated at promotion
+
+**Seven whole-tree builds per acceptance pass.** `frontend_compiles` is anchored per fill slot, so stack #1 pays three builds and stack #2 pays seven, and a Next build is not the sub-second warm-cache install the check's docstring describes. Kept at parity deliberately: #641/#648 attach the check to the file under evaluation so a failure repairs *where the defect lives*, and dropping to one build for the tree would trade repair targeting for wall clock. A cost, not a concession. Revisit only with a measurement.
+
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
Reviewed

Issue #936, opened 2026-08-16
Fill appendix teaches all(), the shell never imports it, fills may not import — every documented fill fails


jasonLadd:
-

PR #979, opened 2026-08-17
fix(scaffold): the shell must import TABLES — #936's defect, recurred at #967

jasonLadd:
**Found by the pre-V7 shakedown** (`cyc_308be9dfc299`): `ReferenceError: TABLES is not defined` in six of eight shells, on a roll whose developer used `TABLES.Run` correctly throughout and whose application was sound.

## The defect

#967 taught the fill appendix to write `TABLES.<Entity>` and left the shell's frozen import at `{ reset, all }`. **A fill may not add imports** — *"Assertions only. NO imports"*, and region enforcement rejects any edit outside the slot. So every fill that *followed the instructions* could not run.

Had this reached V7, **every roll would have failed identically and the window would have been destroyed.** This is the shakedown doing the job §2.e declared it for.

## It is a documented recurrence

The comment directly above that import line — put there by **#936, for exactly this reason** — read:

> *"any store helper the fill vocabulary teaches must already be in scope here or the fill cannot run… **Keep this in step with the vocabulary the appendix teaches.**"*

It was still there, in the right place, when I changed the vocabulary and not the line. **A request is not a mechanism.**

## And the mechanism already existed

`test_fill_vocabulary_is_in_scope` is #936's fix and guards this exact property. **It was green while the defect shipped.** Its extraction was scoped to what #936 looked like — *lowercase function calls* inside ```fill:``` blocks — and #967 taught an *uppercase constant* via *member access* in *prose*. Three ways past a guard aimed at the previous defect's shape.

So that guard is **widened, not duplicated**:

- extraction is now bounded by **what `lib/store.ts` actually exports**, rather than by an identifier shape the next defect can sidestep;
- it reads the whole appendix, not only its fenced examples — a worked example in prose is still a worked example, which is precisely how #967 taught it;
- a **negative control** confirms it now catches what it slept through;
- a guard-the-guard assertion requires both symbols that have actually bitten (`all`, `TABLES`) to still be detected, so the extraction cannot narrow again silently.

I had written a parallel guard first. **Deleting it was the correction** — two tests for one property is the drift this codebase keeps paying for.

Two more copies of the same fact removed while here: the gate's tamper fixtures hardcoded the import line and broke with *"tamper did not apply"* both times it changed. They now derive it from the real emission.

## The pins discriminated, and the contrast is the point

`GENERATOR_VERSION` 3 → 4, fixture regenerated, **both** pinned hashes updated.

| version | change | spine hash | scaffold hash |
|---|---|---|---|
| 3 | typed the store | **unmoved** | **unmoved** |
| 4 | changed the shells' import line | **moved** | **moved** |

Version 3 changed the tree and not the shells; version 4 changes the shells themselves. Two versions demonstrating the pins telling those apart — a distinction a single combined hash could not express.

7973 unit, 7587 regression, green.

### DIFF
--- a/src/squadops/capabilities/stack_nextjs_ts_tests.py
+++ b/src/squadops/capabilities/stack_nextjs_ts_tests.py
@@ -339,7 +339,7 @@ def _invocation_lines(step: _Step, assign: str) -> list[str]:
     return lines
 
 
-def _shell_source(behavior: _Behavior) -> str:
+def _shell_source(behavior: _Behavior, *, store_symbols: tuple[str, ...]) -> str:
     """Render one behavior shell file: frozen spine plus one empty fill slot."""
     route_files = list(
         dict.fromkeys(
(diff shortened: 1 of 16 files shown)


jasonLadd (commented):
Reviewed

Issue #980, opened 2026-08-17
A qa.test retry can pass by retreating to weaker assertions, and the record cannot tell that from being right first time


jasonLadd:
A `qa.test` retry can make a failing suite pass by **retreating to weaker assertions**, and nothing records that it did. The roll banks as a clean success.

## Found in the pre-V7 shakedown

`cyc_308be9dfc299` / impl `run_a0884327215c`. Verdict **accepted**, 41 verified, **14 of 14 criteria**, zero unverified. Three `qa.test` attempts:

| attempt | fills | completion tokens | chars | what the fills assert |
|---|---|---|---|---|
| 1 | 8/8 | **4,896** | 15,448 | response **and store effects** — used `TABLES.<Entity>` |
| 2 (retest after repair) | — | — | — | still failing |
| 3 | 8/8 | **711** | 1,012 | `body.id`, `body.title`, `body.datetime` — **response only** |

The winning emission is a **seventh the size** and touches the store **nowhere**: no `all(...)`, no `TABLES`, no store-effect assertion of any kind.

```
```fill:slot-vc-probe-api-runs expect(body.id).toBeDefined()
expect(body.title).toBe('sample') expect(body.datetime).toBe('2026-08-01T08:00:00')
```

The fill brief asks for *"response values **and store effects** beyond the declared status"*. Attempt 3 delivered the first half and dropped the second — and the dropped half is exactly what catches a handler that does not persist, which is the defect P6 roll 6's three failure analyses spent an entire correction budget hallucinating.

## Why the record cannot see it

The per-roll record captures **fills on first attempt / slots**. That reads `8 of 8` identically whether a fill asserts store state or asserts nothing about it. Nothing captures:

- whether any fill touches the store;
- how large the fill emission was — the 7× drop between attempts was logged and unread.

So a roll that retreated is indistinguishable from a roll that was right first time, and #951's coverage report does not help: it reports which declared *endpoints* have a slot or probe, not what those slots *assert*.

## This is #969's cancellation hazard, through a different door

I filed #969 warning that teaching the repair path to write fills would let it weaken the assertions that caught a real bug. **It never needed to.** A re-authored `qa.test` did the weakening instead, with no repair involved, and the system credited the result.

## Important scoping: the retreat was INDUCED

[...]

PR #981, opened 2026-08-17
docs(plan): declare shakedown #2, and what #1 exposed about a green roll (2.f/2.g)

jasonLadd:
## §2.f — shakedown #2, declared before it runs

Same terms as §2.e, on the deploy carrying #979.

**Why a second one is not ceremony.** Shakedown #1 worked — it found a defect that would have failed *every* roll of the window. But the fix **changes the shell spine's own import line**: `GENERATOR_VERSION` 3 → 4, both pinned hashes moved. The artifact under measurement is different from the one #1 exercised, so #1's justification applies unchanged. Declining the second because the first went well would treat the shakedown as a ritual rather than a check.

**What #1 established, and what it did not.** It established that the boundary breaks loudly and where. It did **not** establish that a cycle completes on the *fixed* shells — it never ran on them.

**The bar for #2, fixed in advance and narrower than "it went green":** a `qa.test` that passes on its **first attempt**, with fills that **reference the store**. #1's eventual pass came on a third attempt from fills that had retreated to response-only assertions, so greenness alone is not the thing being checked.

## §2.g — what #1 exposed about reading a green roll (#980)

Recorded because it changes how a V7 result must be **read**, not merely what to fix.

#1 finished `accepted`, 41 verified, **14 of 14 criteria** — by retreating:

| attempt | fills | completion tokens | what they assert |
|---|---|---|---|
| 1 | 8/8 | **4,896** | response **and store effects** (used `TABLES`, hit #979) |
| 3 | 8/8 | **711** | `body.id`, `body.title`, `body.datetime` — **response only** |

The winning emission is a seventh the size and touches the store nowhere. **§4's per-roll record reads `8 of 8 fills` for both, identically.**

So a green roll can be a retreat and the record cannot tell. #980 carries the recommended split — **instrument before roll 1** (fill emission size; whether any fill references a store symbol, both derivable from data already logged), and **do not change authoring behaviour** before the window, since the retreat was *induced* by a machinery defect now fixed and attempt 1's unprompted behaviour was what we want.

**Until that instrumentation lands, the closing claim may not assert anything about assertion strength — only that the slots were filled.**

Docs only. The cycle's own notes carry the §2.f declaration, stamped at creation.

### DIFF
--- a/docs/plans/1-6-0-v7-fay-window-preregistration.md
+++ b/docs/plans/1-6-0-v7-fay-window-preregistration.md
@@ -142,6 +142,52 @@ and nothing is fixed on its account — the window opens as planned. The distinc
 one §5.1 draws, and it is the reason this is declared as a boundary check rather than a
 practice run.
 
+### 2.f Shakedown #2 — DECLARED NON-COUNTING BEFORE IT RAN *(2026-08-17)*
+
+Same terms as §2.e, on the deploy that carries #979. Not part of this window; may not be
+counted, banked, or cited as a FAY result whatever it produces.
+
+**Why a second one is not ceremony.** Shakedown #1 worked: it found a machinery defect that
+would have failed *every* roll of the window. The fix (#979) **changes the shell spine's own
+import line** — `GENERATOR_VERSION` 3 → 4, both pinned hashes moved. So the artifact under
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
Reviewed

PR #982, opened 2026-08-17
feat(qa): record what the fills assert, not just that slots were filled

### LINKED ISSUE #980: A qa.test retry can pass by retreating to weaker assertions, and the record cannot tell that from being right first time

jasonLadd:
`Closes #980`

A `qa.test` retry can make a failing suite pass by **retreating to weaker assertions**, and the record could not tell that from being right first time.

## What the shakedown did

`cyc_308be9dfc299` finished **`accepted`, 14 of 14 criteria** — by retreating:

| attempt | fills | completion tokens | what they assert |
|---|---|---|---|
| 1 | 8/8 | **4,896** | response values **and store effects** |
| 3 | 8/8 | **711** | `body.id`, `body.title`, `body.datetime` — response only |

Both were recorded identically as *"8 of 8 fills, first attempt"*. The record counted slots, never what the slots say — and the dropped half is exactly what catches a handler that does not persist, the defect P6 roll 6's three analyses spent a full correction budget hallucinating.

## What this adds

`measure_assertion_strength` lives beside the fill parser, which already carries each body. Its output is **banked into `fill_merge` evidence rather than logged** — logs rotate, and the closing record is written later.

- **`store_slots` is load-bearing, and it names the slots rather than counting them**, so a partial retreat (three rich fills, five response-only) is legible rather than averaged away.
- **`body_chars` is the cheap corroborator.** The 7× drop between the real attempts was sitting in data already logged and read by nobody.

## Measures, does not gate

An author may legitimately have nothing to say about the store for a given behaviour, so this may not reject a roll — the same disposition as #951. What it makes impossible is a closing claim that describes assertion strength the window never recorded, and §4 of the pre-registration now says so explicitly.

**An explicit `not_applicable` is counted separately, not as weakness.** It is a declared judgment the protocol asks for, and conflating the two would punish the honest disposition.

## Verification

13 new tests, built from the two real shakedown shapes — including that both filled every slot and only one asserted on state, that a partial retreat is visible, and that prose mentioning "all of the store" is not mistaken for a read.

7986 unit tests green.

## Window status

Pure instrumentation — it changes no authored artifact and no verdict — so it does not need its own shakedown. But it should land **before roll 1**, because without it a green V7 roll that retreated goes into the release as a success and nobody can tell afterwards.

### DIFF
--- a/docs/plans/1-6-0-v7-fay-window-preregistration.md
+++ b/docs/plans/1-6-0-v7-fay-window-preregistration.md
@@ -210,8 +210,19 @@ never thresholds.
 | **scaffold slot count** — non-gating | scaffold manifest |
 | correction rounds | `runtime_activities` |
 | fills on first attempt / slots | eve emission-parse log line |
+| **whether any fill asserts on the store** — non-gating (#980) | `fill_merge.assertion_strength.any_fill_touches_the_store` |
+| **fill body size, and which slots assert on state** — non-gating (#980) | `fill_merge.assertion_strength` |
 | boot audit result | `audit_delivered_app.py`, recorded **separately** |
 | gate disposition and decider | `cycle_gate_decisions.decided_by` |
+
+**Why assertion strength is recorded and not gated.** Shakedown #1 banked `accepted` with
+14 of 14 criteria after its qa author retreated — 4,896 tokens of fills asserting response
+values *and* store effects on attempt 1, 711 tokens asserting response values alone on
+attempt 3, recorded identically as "8 of 8 fills". An author may legitimately have nothing to
+say about the store for a given behaviour, so this may not reject a roll. What it makes
+impossible is a closing claim that describes assertion strength the window never recorded.
+Read from banked evidence rather than a log line, because logs rotate and the closing record
+is written later.
 
 **The two criteria are never collapsed.** SIP-0104 §13a: the mechanical result and the boot
 audit are recorded as distinct facts, and the closing claim states both rather than a single
--- a/src/squadops/capabilities/handlers/cycle/qa_test.py
+++ b/src/squadops/capabilities/handlers/cycle/qa_test.py
@@ -564,7 +564,10 @@ class QATestHandler(_CycleTaskHandler):
         inside the merged shells, attributed to the fill layer.
         """
         from squadops.capabilities.verification_scaffold import VerificationScaffoldManifest
-        from squadops.capabilities.verification_scaffold_fill import merge_fills
+        from squadops.capabilities.verification_scaffold_fill import (
+            measure_assertion_strength,
+            merge_fills,
+        )
 
         record = VerificationScaffoldManifest.from_dict(scaffold_input["manifest"])
         shell_paths = {f.path for f in record.files}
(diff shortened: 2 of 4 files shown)


jasonLadd (commented):
Reviewed

jasonLadd:
## Extended: the retreat had two halves and this covered one

Found while checking whether shakedown #1's repaired suite tripped #957's self-mocking detector. It did not — correctly. `vi` is imported and never used for mocking, and the suite drives the real store **81 times** through `TABLES`. A clean negative for that detector.

But reading the file showed my own fix was incomplete. Attempt 3 did not only weaken its fills — **it dropped that entire 81-call suite.** It ran 8 test files where the failing attempt ran 9, and `extracted_files` went 1 → 0 in a log line nobody read.

So the banked evidence now names the additive files alongside the fill measurement. Instrumenting only the fills would have recorded half a retreat — which is how a partial measurement reads as a complete one, the same error class this measurement exists to catch.

PR #983, opened 2026-08-17
docs(2g): rewrite the Stack Blueprint SIP against main — the gate is met

jasonLadd:
Stage 2g. **This SIP's acceptance gate has been met for a week and nobody noticed** — it declines its own acceptance until "the existence of a second real stack", and `nextjs_ts` landed 2026-08-10. It sat open because the stack arrived incrementally and there was no event to trigger the decision.

The rewrite is **ratification, not greenfield.** Three of the document's four worries were confirmed by the second stack, and one was answered in a way it did not consider.

## Three predictions confirmed

- **`harness_entry_modules`** shipped as `()` — a *declared fact*, not an omission.
- **`check_stack`** shipped as `""`, which forced the question *"then what verifies this language?"* — and SIP-0104's deterministic scaffold is that answer. **The empty field produced a capability rather than a gap.**
- **Criteria families** proved pack-parameterized.

## One genuine surprise

It asked whether `expand` is *"a callable, or data a generic expander consumes"*. **Neither**: two callables, two tuples, and five string names indexing registries owned by the layer that owns each vocabulary. Stack #2 derives file *location* from the route path — a computation, not a table — and expressing it as data would build a template language with worse tooling and no type checking.

## And one section is corrected against itself

The SIP proposed the blueprint as *"the natural place to collapse"* the two stack vocabularies. Stage 2e measured **three axes, two of which are not drift**: stack identity is one value across five registries, and the probe profile differs from the stack id on *both* stacks — correctly, because booting is its own axis and two stacks could share a boot mechanism while sharing nothing else. **Collapsing that would be the error.** `check_stack` is the real conflation, carrying two questions on one field.

## The stale table is kept, not rewritten away

The "problem, precisely" table describes code that no longer exists. It stays, with what actually happened recorded beneath it — **the diagnosis was right and the remedy arrived by a different route**, and rewriting it silently would erase that. The live brief is narrower than the one the document opened with: *what does a missing field mean?* was answered four times in feature PRs, and its strongest answer lives only in a docstring.

## Also carried in

- **2c's falsification** — three fields with no consumer of any kind, one with a reader and no data. None may enter the blueprint.
- **2f's admission rule**, as its permanent statement rather than a line in a release plan that is superseded at the cut.
- **The disclosures S3 must state rather than imply** — including that `nextjs_ts` is only *partially* "maximally different" (it has a bundler and shares React/TS with stack #1's frontend half), so a two-stack schema still risks a **cousin-shaped** generalisation. Waiting for stack #3 repeats the failure that produced this SIP.

## All four open questions answered

Two settled, one settled *against* the option the SIP preferred, and one — where the blueprint lives — deliberately left open **with its trigger named** (the packs/plugin-loading work, which is the durable reason to keep the document at all).

**Promotion is a maintainer act and is not done here.**

### DIFF
--- a/sips/accepted/SIP-0105-Stack-Blueprint-Contract.md
+++ b/sips/accepted/SIP-0105-Stack-Blueprint-Contract.md
@@ -1,11 +1,15 @@
 ---
-status: proposed
+status: accepted
 title: Stack Blueprint Contract
+sip_number: 105
+updated_at: '2026-08-17T18:46:44.835365Z'
 ---
-# SIP: Stack Blueprint Contract
+# SIP-0105: Stack Blueprint Contract
 
 ## Status
-Draft (proposed) — **rewritten against main 2026-08-17 (1.6 Stage 2g); acceptance gate met.**
+**Accepted 2026-08-17** — rewritten against main at 1.6 Stage 2g; the acceptance gate (a second real stack) was met on 2026-08-10.
+
+Accepted **with its unbuilt parts named** — see §"What acceptance does and does not assert". Acceptance here ran *after* implementation rather than before it, which is recorded there rather than smoothed over.
 
 **Targets:** the release that adds the **second** stack — deliberately not before. **`nextjs_ts`
 landed 2026-08-10**, so the gate is met; see §"The acceptance gate is met — what the second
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
Reviewed

PR #984, opened 2026-08-17
docs(plan): the four outstanding rulings, closed

jasonLadd:
Owner rulings 2026-08-17. **Every ruling in §7 is now closed**, so precondition §2.6 moves to **MET**.

## §7.4's remaining outcome bands

The ≥4/6 and 0/6 cells were drafted consistently with the middle band you ruled and are now ruled as written. No judgment was left in them once the middle band was fixed.

## The 1-of-6 split — ruled, and it was the contested one

**1/6 groups with 0/6, not with 2–3/6**, on the reading that the cut gate's *"repeatably > 0"* means more than once.

I flagged this as the one most open to a different reading, so it is settled deliberately **while the number is unknown** — which is the only time it can be settled honestly. At 1/6 the generous reading is at its most tempting, and the figure is banked either way, so the conservative reading costs nothing.

## 1e — credited to P6 roll 1 (`cyc_24aac7c22be1`), resting on the corrected audit

Its deliverable provably works. Hand-verified during the window — create → 201, join → 200, duplicate → 409, and a bad payload returning the squad's own error envelope — and the corrected instrument (#956) passes it on **all five UI call sites**, replayed against unchanged stored artifacts.

**The credit rests on the corrected instrument, and the record says so rather than implying otherwise.** As *run*, roll 1's UI check found **2 of its 5** call sites and silently skipped three. So *"roll 1 passed the audit"* is true of the instrument it ran under and understates what was actually checked. The re-audit is what makes the credit sound, and a reader is entitled to know which one it rests on.

The conservative alternative — roll 2, whose audit found all five contemporaneously — was weighed and not taken: 1e asks whether the stack can produce a working application, and roll 1 demonstrably did.

## V6 (4b) — complete

Rolls 3, 4 and 5 banked. Closed in the authorship plan, recorded as met in the pre-registration's §2.1 with the window record as evidence.

Docs only.

### DIFF
--- a/docs/plans/1-6-0-authorship-plan.md
+++ b/docs/plans/1-6-0-authorship-plan.md
@@ -229,6 +229,13 @@ was never in the 1a sweep) · and can **M6's taxonomy name the subsystem** if it
 **Nothing is yet known about whether the Next.js stack works.** The 2026-08-10 run never
 reached expansion for it.
 
+> **CLOSED 2026-08-17: 1e is credited to P6 roll 1, `cyc_24aac7c22be1`, resting on the
+> corrected audit (#952/#953).** Its deliverable was hand-verified during the window and the
+> corrected instrument passes it on all five UI call sites, replayed against unchanged stored
+> artifacts. **The credit rests on the corrected instrument and the record says so** — as run,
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
Reviewed

PR #985, opened 2026-08-17
docs(plan): 1e's credit in the window record — the edit that silently did not apply

jasonLadd:
#984 recorded 1e's credit in the authorship plan and **reported it recorded in the window record too. It was not.**

The replacement targeted `Before these fixes` while the file said `Before #952/#953`, so it matched nothing — and the script printed success regardless, because the print did not check whether the match had happened.

**Main therefore carried two documents disagreeing about the same ruling:** the authorship plan said closed and credited; the window record still said *"Needs a ruling."*

Caught by verifying the merge **by content** rather than by the merge succeeding. This fix asserts the anchor before writing, so a second silent no-op fails loudly instead of reporting success.

Content is unchanged from what #984 intended — 1e credited to roll 1 (`cyc_24aac7c22be1`) resting on the corrected audit, the record stating that *as run* its UI check found 2 of 5 call sites, and the roll 2 alternative recorded as weighed and not taken.

Docs only.


jasonLadd (commented):
Reviewed

Issue #988, opened 2026-08-18
A self-referential test suite's failure routes to the application, and broke a correct handler


jasonLadd:
A failing test suite routes its repair to the **application**, with nothing checking whether that suite ever calls the application. In the pre-V7 shakedown this **broke a correct app against its own contract.**

The signal needed to prevent it is already computed and not consulted.

## What happened — `cyc_6651d552e06a` / impl `run_81d7bde9d78a`

The qa author's additive suite `__tests__/runs.test.ts` **reimplements the API and tests the reimplementation.** It mocks the fetch seam, hand-rolls a `mockApiHandler`, imports no route module, and invokes the delivered application **zero times**.

Its fake is broken. The application's create:

```ts
const run = { id: nextId(), ... };
insert(TABLES.Run, run);
return NextResponse.json(run, { status: 201 });   // correct
```

The suite's fake create:

```ts
const run = insert(TABLES.Run, { ...body, participants: [] });
return { status: 201, body: run };                // no id is ever generated
```

So `body.id` is `undefined`, every later lookup 404s, and the suite fails **against itself**. (The fake also implements a `/join` sub-path this design does not declare.)

## The chain

| step | what happened |
|---|---|
| 1 | suite fails, entirely inside its own fake |
| 2 | analyzer blames the application — the fourth consecutive diagnosis contradicted by the source (#968) |
| 3 | locus falls through to the **dev** chain — *not* `own_artifact`, so this is **not** #970 |
| 4 | dev repair rewrites five files including the correct route handler |
| 5 | **the rewrite drops `status: 201`** → `Response.json(created)` returns 200 |

The manifest declares `success_status: 201`. The contract probe and the scaffold shell both assert it. **So a handler that satisfied its contract now violates it**, and the next retest failed too — the damage did not even buy convergence.

## The fix: consult the signal that already exists

`detect_self_mocking_tests` (#957) **correctly flags this suite**, with the right reason:

> `replaces the fetch seam and imports no route module, so the application is never invoked`

That detector runs on the same artifacts, in the same handler, in the same task. **The locus classifier does not look at it.**

**Proposed rule: a suite flagged as self-mocking may not route its failures to the subject.** Its failure says nothing about the application, because it never ran the application. Route to the producing role's own artifact, or refuse to route at all — but not to the app.

This is narrower than #968/#969/#970 and independent of them:

- **#968** — the analyzer asserts causes the source contradicts. True here, but a correct analysis would still need somewhere safe to route.
[...]

PR #989, opened 2026-08-18
fix(qa): bank suite-authenticity rows on pass, with what was inspected (#986)

### LINKED ISSUE #986: The self-mocking detector is deployed, correct, and never fires in a cycle
`detect_self_mocking_tests` (#957) is deployed and flags the offending file on demand. **It did not appear in the record of a cycle whose suite it flags.**

## Evidence — pre-V7 shakedown #2, `cyc_6651d552e06a` / impl `run_81d7bde9d78a`

The qa author's additive suite mocks the fetch seam, hand-rolls a `mockApiHandler`, imports no route module, and invokes the delivered application zero times. Asked directly, the **deployed** eve container flags it:

```
deployed eve flags it: True
```

The banked verification summary for that run contains `expected_artifacts`, `non_stub_files` and `tests_pass`. Its `failed` list is exactly `["tests_pass"]`. **There is no `no_self_mocking_tests` row of any kind.**

## Where it is wired, and where it is not

`detect_self_mocking_tests` is called at two sites in `handlers/cycle/qa_test.py` — the retest path, and the post-test-run validation block. **It is not called inside `_validate_output`**, which is where `expected_artifacts` and `non_stub_files` are produced — and those two reached the summary while this did not.

That is a strong hint rather than a proven mechanism, and it is stated as such: what is established is that the detector did not fire in a cycle where the deployed function flags the file. Pinning exactly why is the first task of the fix.

## Why it matters more than a missing check row

The detector's output is the signal that should stop a self-referential suite's failure routing to the application. In this same roll, that routing rewrote a correct handler and dropped a `status: 201` the manifest declares — see the companion issue on locus routing.

So this is not only an unenforced check. **It is a prerequisite for the fix that prevents the correction chain damaging correct work.**

## Also worth correcting on the record

When this suite first surfaced I described the detector as vindicated on its first live encounter. That was wrong: it catches the file when called directly and did not catch it in the cycle. The distinction between "the function works" and "the check runs" is the whole of this issue.

Related: #957 (the detector), #915 (the class it detects), #980.

jasonLadd:
## What

The two suite-authenticity detectors — #276 stub-fallback and #915 self-mocking — appended a check row **only when they found an offender**. A clean result left no record at all.

That makes an absent row ambiguous. It is equally consistent with:

- the suite was inspected and was clean, and
- the detector never ran over that suite.

## Why it matters now

`cyc_6651d552e06a` banked exactly that absence. The facts available afterwards:

- the deployed detector flags the run's offending suite when called by hand;
- the run's other qa validation rows (`expected_artifacts`, `non_stub_files`, `tests_pass`) banked normally, so qa's validation did execute;
- `no_self_mocking_tests` is absent from the banked summary.

Nothing in the stored state distinguishes the two histories, and the archaeology — checkpoints, verification summary, artifact refs — did not resolve it, because the distinguishing fact was never written down.

## The change

Both call sites (the main `handle()` path and the execute-only retest path, which each carried their own copy) now append the row unconditionally via one shared builder, carrying `inspected`: the files the detector actually read.

```json
{"check": "no_self_mocking_tests", "passed": true, "inspected": ["__tests__/runs.test.ts"]}
```

An empty `inspected` beside an executed suite now reads as the defect it is, rather than as a pass. That is the state the locus-attribution work needs in order to key on this check at all.

Two smaller items ride along:

- the check-name literals, previously duplicated across the two call sites, move to constants in the module that owns the concern;
- the two duplicated coverage-ratio recomputations on the main path fold into one unconditional pass, since the ratio now moves on a clean result too.

## Tests

Two added to the existing execute-only retest class, which drives the real handler path with no LLM:

- a clean TypeScript suite still banks `no_self_mocking_tests` with `passed: true` and `inspected` naming the file — before this change the row does not exist and the lookup raises;
- a python-only repair banks the JS row with `inspected: []` and the python row with the file it read — the "fed the wrong file set" case a bare pass cannot express.

Full regression: 7602 passed, 1 skipped.

## Scope note

This is the instrument, not yet the attribution fix. It deliberately does not change what the detectors flag, and it does not change which files each path feeds them — the retest path still sees only the repaired files. Whether that narrower input is itself a coverage hole is what the next roll's `inspected` inventory will answer, and it is left for #988 rather than guessed at here.

Closes #986

### DIFF
--- a/src/squadops/capabilities/handlers/cycle/qa_test.py
+++ b/src/squadops/capabilities/handlers/cycle/qa_test.py
@@ -72,6 +72,22 @@ def _frontend_skip_reason(error: str) -> str:
     return NotExecutedReason.SUBJECT_MISSING
 
 
+def _authenticity_row(check: str, offenders: list[str], inspected: list[str]) -> dict[str, Any]:
+    """Build a suite-authenticity check row that is banked pass OR fail (#986).
+
+    The two authenticity detectors used to append a row only when they found
+    something, so a clean suite left no trace and an absent row meant either "looked
+    and found nothing" or "never looked" — an ambiguity that cost an hour of
+    archaeology on `cyc_6651d552e06a` and still did not resolve. ``inspected`` names
+    the files the detector actually read, so an empty inventory beside an executed
+    suite reads as the defect it is instead of as a pass.
+    """
+    row: dict[str, Any] = {"check": check, "passed": not offenders, "inspected": inspected}
+    if offenders:
+        row["offenders"] = offenders
+    return row
+
+
 class QATestHandler(_CycleTaskHandler):
     """Build handler: generates test files from validation plan + source (D1).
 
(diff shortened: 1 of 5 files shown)


jasonLadd (commented):
Reviewed

PR #990, opened 2026-08-18
fix(scaffold): accept the language-prefixed fill fence, and record it (#987)

### LINKED ISSUE #987: A language-prefixed fill fence is silently reclassified as a file and dropped — the whole emission is lost
The fill appendix teaches **two fence grammars in one document**, for one emission, with incompatible rules. Blending them destroys every fill silently.

| what | form | language prefix |
|---|---|---|
| a fill | ` ```fill:slot-<id> ` | **forbidden** |
| an additive file | ` ```typescript:__tests__/<name>.test.ts ` | **required** |

## What happened — pre-V7 shakedown #2, `cyc_6651d552e06a`

The qa author emitted six fills as ` ```typescript:fill:slot-vc-probe-api-runs `:

```
emission shape: fences={'fill': 0, 'path': 6}
emission parse: fills=0 ... extracted_files=1
```

**Six well-formed fills, zero parsed.** They matched the path-fence grammar instead, landed at scaffold shell paths, and were dropped by the P3 guard. Every slot then rendered as a failing state — which is P3 working correctly, since a missing fill must fail loudly — so the suite failed on fills that were right in substance and mislabelled in form.

The roll then spent a correction round that could not fix it: the repair is aimed at `expected_artifacts` and cannot reach a fill (#970).

## Frequency, stated honestly

**One occurrence.** The next `qa.test` attempt used the correct fence unprompted, with no feedback about what had gone wrong. So this is an occasional trap, not a reliable one — the initial hypothesis that the confusion was systematic is **not supported by the evidence**.

What stands regardless is the failure *mode*: when it happens the loss is total, silent, and unrecoverable by the correction chain.

## Fix

The fence tag contains `fill:slot-<id>` verbatim, so the intent is unambiguous — and no real file is named `fill:slot-x`, so there is no collision with a genuine path emission.

- **Minimum: fail loudly.** A fence containing `fill:slot-` that is not parsed as a fill should say so rather than vanish into a drop.
- **Preferred: also accept it**, and record that it arrived in a non-canonical form, so the frequency is measurable rather than invisible.

Silent reclassification into a destructive path is the worst of the three options and is what happens today.

Related: #969, #970, #926 (the emission capture that made this diagnosable at all).

jasonLadd:
## What

The fill-fence pattern required the exact form the brief teaches:

````
```fill:slot-vc-probe-api-runs
````

An author writing markdown all day writes this instead:

````
```typescript:fill:slot-vc-probe-api-runs
````

which matched nothing. The fence then survived `strip_fill_blocks`, and the fenced-**file** extractor — which reads `<language>:<path>` info strings — read it as a file named `fill:slot-vc-probe-api-runs`.

## Why it matters

A pre-V7 shakedown lost **all six** of its fills this way, and no finding was raised: from the protocol's side the emission had simply said nothing about any slot. The shells then rendered as the failing states a missing fill is *defined* to produce, so the run read as an authoring failure rather than a parse failure. That is the most expensive shape a defect can take here — it looks exactly like the thing the scaffold exists to detect.

## The change

- The language prefix is now optional in the fence pattern, so both forms parse identically.
- `strip_fill_blocks` removes the prefixed form too — otherwise one block would be both a fill *and* a bogus additive file.
- It is **recorded, not silently normalised**: `FillEmission.language_prefixed` carries the slot ids whose fence was prefixed, surfaced in the banked `fill_merge` evidence as `language_prefixed_fences`. A brief the author keeps departing from is worth seeing before it costs another roll.
- The prefix must be a bare language token. A path-addressed fence that merely *ends* in `fill:slot-…` (`ts:src/fill:slot-x`) is still the file extractor's and is untouched.

## Tests

Four in `TestParseFillEmission`, all of which fail before the change:

- a prefixed fence parses as a fill, alongside an unprefixed sibling, with `language_prefixed` naming only the prefixed one;
- a prefixed fence cannot leak into the file extractor, while a genuine additive-file fence beside it is left intact;
- a non-language prefix is *not* claimed as a fill (the guard against over-matching);
- the empty-emission exact-dict measurement updated for the new evidence key.

Full regression: 7603 passed, 1 skipped.

Closes #987

### DIFF
--- a/src/squadops/capabilities/verification_scaffold_fill.py
+++ b/src/squadops/capabilities/verification_scaffold_fill.py
@@ -75,8 +75,20 @@ from squadops.capabilities.verification_scaffold import (
 MAX_FILL_LINES = 120
 MAX_FILL_BYTES = 8_000
 
+#: A fill fence, with the language prefix an author reflexively adds made optional (#987).
+#:
+#: The brief teaches ```` ```fill:slot-<id> ````, and an author writing markdown all day
+#: writes ```` ```typescript:fill:slot-<id> ```` anyway. Under the exact-form pattern that
+#: cost a whole pre-V7 shakedown: the fence did not match here, `strip_fill_blocks` left it
+#: in the text, and the file extractor — which reads ``<language>:<path>`` info strings —
+#: took it for a file named ``fill:slot-<id>``. All six fills vanished, and nothing said so.
+#:
+#: The prefixed form is unambiguous, so accepting it loses nothing. It is still *recorded*
+#: (see ``language_prefixed``) rather than silently normalised, because a brief the author
+#: keeps departing from is worth knowing about.
 _FILL_FENCE_RE = re.compile(
-    r"^```fill:(?P<slot_id>slot-[a-z0-9][a-z0-9-]*)[ \t]*\n(?P<body>.*?)^```[ \t]*$",
+    r"^```(?:(?P<lang>[A-Za-z][\w+.-]*):)?"
+    r"fill:(?P<slot_id>slot-[a-z0-9][a-z0-9-]*)[ \t]*\n(?P<body>.*?)^```[ \t]*$",
     re.DOTALL | re.MULTILINE,
 )
 _NOT_APPLICABLE_RE = re.compile(r"^\s*not_applicable:\s*(?P<reason>\S.*)$")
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
Reviewed

PR #991, opened 2026-08-18
fix(routing): a suite that never invoked the app is the suite's defect (#988)

### LINKED ISSUE #988: A self-referential test suite's failure routes to the application, and broke a correct handler

jasonLadd:
> **Stacked on #989** — it uses the `no_self_mocking_tests` registry constant that PR adds. Merge #989 first; the base retargets to `main` automatically.

## What

`classify_failure_locus` scans own-artifact signals **before** consulting the `tests_pass` row. Two signals were in that first scan: missing `expected_artifacts`, and the frozen contract contradicting the suite's own assertions (#629).

The two suite-authenticity rows were not. So a self-mocking (#915) or stub-fallback (#276) suite that failed produced exit 1 and classified `SUBJECT` — routing the repair at the **application**.

## Why that is the worst available answer

A suite that mocks the route module under test, or replaces the `fetch` seam and imports no route at all, returns a verdict *about itself*. The application is not implicated by it in either direction — not as passing, not as failing.

Two runs paid for this:

- the pre-V7 shakedown exhausted its correction attempts rewriting a handler against a suite that reimplemented the API and tested its own fake (round 3 restored a `status: 201` that round 2 had removed from correct code);
- roll 6's delivered app **passed a hand boot audit** — installs, builds, boots, 5 probes — while the cycle rejected it.

## The change

Both authenticity rows join the first scan, so they outrank the `tests_pass` row they sit beside.

This is safe from the test-gaming guard on the same footing as the contract row already there. The guard exists to stop a qa re-author "fixing" a broken app by rewriting tests, so it demands that any own-artifact signal be one an app defect **cannot produce**. These are structural — read off the suite's own source — and no application behaviour can make a suite mock its own route module.

## Tests

Three in the locus class, parametrized across both checks:

- a failing authenticity row beside `tests_pass` exit 1 (which alone reads `SUBJECT`) classifies `OWN_ARTIFACT`;
- the guard's other side: a **passing** authenticity row — now banked on clean runs via #989 — leaves a genuine app failure on the dev chain.

Full regression: 7605 passed, 1 skipped.

Closes #988

### DIFF
--- a/src/squadops/cycles/failure_evidence.py
+++ b/src/squadops/cycles/failure_evidence.py
@@ -10,6 +10,10 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any
 
 from squadops.cycles.acceptance_check_spec import CHECK_CONTRACT_ASSERTIONS
+from squadops.cycles.check_registry import (
+    CHECK_NO_SELF_MOCKING_TESTS,
+    CHECK_NO_STUB_FALLBACK_TESTS,
+)
 from squadops.cycles.emission_integrity import extraction_loss_suspected
 from squadops.cycles.verification_integrity import ResultStatus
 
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
Reviewed

PR #992, opened 2026-08-18
docs(v7): freeze the deploy, close the preconditions, open the window

jasonLadd:
Records the V7 pre-registration as frozen, immediately before roll 1.

- **§3 completed** — main `4fa74525`, seven image ids from the 2026-08-18 rebuild. The PRD sha and `resolved_config_hash` were **re-verified at the deploy boundary**, not carried forward, and both are unchanged from 2026-08-16: the fix queue touched the qa handler, fill parser, check registry and locus classifier and moved no part of the recipe, so this window and the six P6 rolls stay comparable on request.
- **Preconditions** 2.4 / 2.5 / 2.7 → MET. 2.8 added for the two shakedown defect sets (no earlier revision could have listed it — neither shakedown had run).
- **§2.g restriction lifted** — #982 is deployed and loaded-module verified, so assertion strength is measurable from here on, and explicitly not retrospectively for P6.
- **§2.h added** — shakedown #2 in full: it failed its own declared bar, is not counted, and found three defects that would have corrupted the window rather than merely failing a roll.

The section also states plainly what the window opens *without*: no shakedown has met §2.f's bar (first-attempt pass with store-touching fills). Roll 1 is the first observation of a cycle on fixed shells.


jasonLadd (commented):
Reviewed

PR #993, opened 2026-08-19
docs(sip): propose Agent Embodiment Runtime (2.2-era)

jasonLadd:
New proposal in `sips/proposed/` (design review, per the SIP workflow — approve the design, not code).

**Thesis:** SquadOps owns the agent; a runtime is the agent's current embodiment; the native BaseAgent loop is one embodiment among several. Extends and amends **SIP-0090** rather than introducing a parallel concept.

Key content:
- **Two orthogonal axes**: `embodiment_runtime` (engine: `squadops-native | goose | codex | claude-code`, registry-described) ⊥ location (surface/place). `embodiment_type`+`platform` dissolve — SIP-0090's own `location_system` already carried the surface. Amended while the `embodiments` table has zero rows.
- **Native as peer**: the BaseAgent worker is `squadops-native`; SIP-0089 `agent_runtime_state` re-homes as its health record.
- **Distributed attention**: surface attachments as consumers of the SIP-0090 §7 attention budget (machinery already in `budgets.py`), with primary/ambient posture pricing and a named reconciliation contract. Answers "how many places at once" as a budget decision, not a cardinality.
- **Invariant 9**: verification independence is an *identity* boundary — the projections pattern is rejected permanently, not deferred.
- **Boundary test** (P3 exit): a non-BaseAgent process consumes a TaskEnvelope and returns SIP-0096-surviving evidence.
- **Appendix A**: the distributed-embodiment fabric, explicitly non-roadmap.

Sequencing is readiness-gated (2.2 is placement, not a date): FAY baseline banked first, memory-SIP seam service-addressable, no second live runtime before the baseline exists.

Docs-only; no code or deploy impact. Complements the frozen V7 window rather than touching it.

### DIFF
--- a/sips/proposed/SIP-Agent-Embodiment-Runtime.md
+++ b/sips/proposed/SIP-Agent-Embodiment-Runtime.md
@@ -1,0 +1,367 @@
+---
+status: proposed
+title: Agent Embodiment Runtime
+author: Jason Ladd
+created_at: '2026-08-19T00:00:00Z'
+---
+# SIP: Agent Embodiment Runtime
+
+## Status
+Draft (proposed)
+
+**Author:** Jason Ladd
+**Created:** 2026-08-19
+**Revision:** 1
+**Targets:** the **v2.2 era** (even-minor feature lane; headline candidate). Placement is a
+sequencing intent, not a date — every phase is gated by the readiness conditions in §9, and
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
Reviewed

Issue #994, opened 2026-08-19
Rewind after a successful correction repair re-dispatches develop and discards the repaired state


jasonLadd:
V7 window detection (roll 1 void, `cyc_8b569ce34074` / `run_29b53b6e3d4b`). Recorded mid-window, left unfixed per pre-registration §6.

**The sequence, from `runtime_activities` + neo logs:**

1. 01:07Z `development.develop` (subtask 2) fails — `acceptance:frontend_compiles`, a real defect: server-component `app/page.tsx` fetches `api('/api/runs')` at prerender, `next build` rejects the relative URL (#996 covers the authoring gap).
2. 01:38Z `development.correction_repair` **completes** — emits `app/page.tsx` with `export const dynamic = 'force-dynamic'` (correct fix), accepted and stored.
3. 01:48Z the rewind protocol **re-dispatches `development.develop` for the same subtask**. The re-authored emissions (7,516 tok `'use client'` variant at 02:00, 5,322 tok variant at 02:08) each fail `frontend_compiles`; the self-eval loop burns the budget; the task dies at the **1800s task timeout**; the run fails with "Rewinding to checkpoint after development.develop failure".
4. `audit_delivered_app.py` on the stored tree (acceptance-aware last-wins selects the accepted repair): **PASS** — installs, builds, boots, answers all 5 probes, UI data-path clean.

**The defect:** after a successful, accepted repair, the rewind path discards that repaired state and requires the develop task to re-author the same file from scratch — the roll then depends on the model out-competing its own accepted fix under a shrinking clock. The cycle possessed a working deliverable at 01:38 and spent the next hour destroying its claim to it.

**Failure mode class:** cycle rejects a working application (the P6 roll-6 shape at a different check).

PR #997, opened 2026-08-19
fix(authoring): teach the server-component prerender constraint (#996) [POST-WINDOW]

### LINKED ISSUE #996: Authoring surfaces teach the api() client seam but not the server-component prerender constraint
V7 window detection (roll 1 void, `cyc_8b569ce34074`). Recorded mid-window, left unfixed per pre-registration §6.

Roll 1's original `frontend_compiles` failure: the dev author wrote `app/page.tsx` as a **server component** that calls `await api('/api/runs')` at module scope. `next build` executes it during prerender in Node, where a relative URL throws `TypeError: Failed to parse URL from /api/runs` — the build fails.

The fill-only appendix teaches the client seam precisely ("api() fetches the path you pass it, verbatim") and warns about dropped prefixes — but nothing in the brief, skeleton, or appendix teaches the **prerender constraint**: a server component's body runs at build time, so a page that fetches must either be a client component or declare `export const dynamic = 'force-dynamic'`.

This is the context-completeness class (#866's family): the agent is judged against a Next.js fact it cannot see from anything we show it. The correction loop DID diagnose and fix it correctly in one round (`force-dynamic`, accepted) — the knowledge exists downstream in repair but not upstream in authoring, so every roll pays a correction round (or worse — see #994's follow-on) to relearn it.

Candidate fix (post-window): one sentence + example in the dev appendix's page-component section.

jasonLadd:
**Do not merge while the V7 window is open** (§6: no merges, no prompt-asset edits to the live instrument). Stacked per the night-shift routine.

One addition to the fill-only dev appendix, directly under the client-seam section it extends: a page that fetches is either a client component or declares `export const dynamic = 'force-dynamic'` — a bare server component awaiting `api()` fails `next build` at prerender.

Evidence: V7 roll 1 (void, `cyc_8b569ce34074`) — the exact failure, correctly repaired in one round by the correction loop, which proves the knowledge exists downstream while authoring lacks it. The follow-on cost was the whole roll (#994).

Closes #996

### DIFF
--- a/src/squadops/prompts/request_templates/request.development_develop_fill_only_appendix_nextjs_ts.md
+++ b/src/squadops/prompts/request_templates/request.development_develop_fill_only_appendix_nextjs_ts.md
@@ -52,6 +52,14 @@ If your manifest declares `POST /api/runs`, the page calls `api('/api/runs')`. D
 the prefix compiles, passes type-checking, and returns 404 at runtime for every action in
 the UI — a silent break no build or unit check catches.
 
+**When the page itself fetches, it must not do so at build time.** `next build`
+prerenders server components in Node, where a relative `api('/api/runs')` throws
+`Failed to parse URL` and **the build fails**. A page that fetches data is therefore
+either a client component (`'use client'` at the top, fetch inside `useEffect`) or a
+server component that opts out of prerendering with `export const dynamic =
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
Reviewed

PR #1001, opened 2026-08-19
fix(rollup): normalizer reads the authenticity row's own verdict (#1000) [RESET-CLASS / POST-WINDOW]

### LINKED ISSUE #1000: RESET-CLASS: normalizer treats presence of no_stub_fallback_tests as failure; #989's unconditional row falsely rejects repair-then-pass rolls
**V7 window detection — RESET-CLASS. Introduced by #989 (mine), in the frozen deploy `4fa74525`. Night shift halted on detection per the standing rule; abort/re-register is the owner's call.**

`verification_normalize.py:82` was written against the old producer contract — the `no_stub_fallback_tests` row appeared **only when stubs were found**:

```python
if cid == CHECK_NO_STUB:
    # Present only when stubs were found → the stub check executed-and-failed
    stub_detected = True
    results.append(CheckResult(check_id=cid, status=ResultStatus.FAILED))
```

#989 changed the producer to bank the row **unconditionally** (with `passed` and `inspected`) and swept the producer's tests — but not this consumer. The normalizer ignores `passed: true` and records FAILED on presence.

**Observed** (launch 3, `run_82dc31768f7d`): `failed_detail` carries `{"check_id": "no_stub_fallback_tests", "reason": "", "required": false}` — a phantom failure with an empty reason, from a clean suite. Launch 3's rejection was independently genuine (`tests_pass`, required, real), so no roll has yet been *decided* by the phantom.

**Why reset-class anyway:** the verdict rule is `elif failed: REJECTED` — any failed check rejects, required or not. A roll whose qa fails once, is repaired, and **passes on retest** banks the retest's `validation_result.checks` (which carry the passed authenticity row) → normalizer marks it FAILED on presence → the phantom is the sole failed row → **an otherwise-accepted roll is falsely rejected.** That is the most common green-roll recovery path; continuing the window would waste every roll that needs a single qa correction.

Scope check (full consumer sweep, done this time): `failure_evidence.py` reads `passed is False` — correct; `no_self_mocking_tests` takes the generic passed-row branch — correct. Exactly one consumer is broken.

**Fix** (stacked, unmerged): route `CHECK_NO_STUB` through the same passed-field reading as every other boolean row, setting `stub_detected` (the §6.6.1 is_stub taint) only when the row actually failed.

Process note for the record: this is a textbook violation of "verify sweeps broader than the edit" — #989 changed a producer's contract and audited only the producer.

jasonLadd:
**Do not merge until the V7 re-register decision** — this is the reset-class fix; deploying it is part of that decision, not a hotfix.

One consumer (`verification_normalize.py`) still assumed the pre-#989 producer contract (row present ⇒ stubs found ⇒ FAILED). Since #989 banks the row unconditionally, a clean retest normalized to a phantom FAILED row, and `elif failed:` rejects on any failed check — so **any roll whose qa fails once, is repaired, and passes on retest would be falsely REJECTED.** Launch 3 banked the phantom on a clean suite; its rejection was independently genuine (required `tests_pass`), which is the only reason this was caught before it decided a roll.

Fix: the row takes the generic passed-field path; only an actual failure sets the §6.6.1 `is_stub` taint. Full consumer sweep done this time — `failure_evidence` and the self-mocking row were already correct.

New test: a `passed: true` row normalizes PASSED, taints nothing, and `tests_pass` credits — fails before the fix. Existing failure-case tests unchanged and green.

Full regression: 7609 passed, 1 skipped.

Closes #1000

### DIFF
--- a/src/squadops/cycles/verification_normalize.py
+++ b/src/squadops/cycles/verification_normalize.py
@@ -80,10 +80,19 @@ def normalize_task_checks(
             # below (richer + present on a passing run). Skip to avoid double-record.
             continue
         if cid == CHECK_NO_STUB:
-            # Present only when stubs were found → the stub check executed-and-failed,
-            # and its presence flags the synthesized tests_pass as a stub (§6.6.1).
-            stub_detected = True
-            results.append(CheckResult(check_id=cid, status=ResultStatus.FAILED))
+            # #1000: this row used to appear ONLY when stubs were found, so presence
+            # meant executed-and-failed. Since #989 the producer banks it on every
+            # qa validation — pass or fail, with `passed` and `inspected` — so the
+            # row's own verdict must be read, exactly like any boolean row. Presence-
+            # implies-failure here turned a clean repair-then-pass retest into a
+            # phantom failed row, and `elif failed:` rejects on ANY failed check —
+            # a false REJECTED on the most common green-roll recovery path (V7
+            # launch 3 carried the phantom; a repair-then-pass roll would have been
+            # decided by it). Only an actual failure taints the synthesized
+            # tests_pass as a stub (§6.6.1).
+            if row.get("passed") is False:
+                stub_detected = True
+            results.append(_from_passed_row(cid, row))
             continue
         if "status" in row:
             # Typed-acceptance row: carries a CheckOutcome status verbatim.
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
Reviewed

PR #1005, opened 2026-08-20
fix(audit): reconstruct concatenated UI paths — false PAGE_NOT_API failed the V7 deciding roll (#1004) [OWNER RULING]

### LINKED ISSUE #1004: ui_data_path extractor truncates string-concatenated paths to the first literal — false PAGE_NOT_API failed the V7 deciding roll
**V7 attempt 2, counted slot 6 — the deciding roll. The audit false-failed a functional application, and the §5 score turns on it.**

`app/runs/[run_id]/page.tsx` builds its action paths by **string concatenation**:

```ts
await api('/api/runs/' + runId + '/join', {   // line 67
await api('/api/runs/' + runId + '/leave', {  // line 92
```

`extract_ui_calls` (deployed module, reproduced directly) truncates to the first literal:

```
UiCall(line=37, written='/api/runs/${runId}', request_path='/api/runs/ui-probe-sample')  ← template literal: correct
UiCall(line=67, written='/api/runs/', request_path='/api/runs/')                          ← concatenation: truncated
UiCall(line=92, written='/api/runs/', request_path='/api/runs/')                          ← truncated
```

The audit then probes `/api/runs/` — a path the UI never requests — receives HTML, classifies PAGE_NOT_API, and fails the app.

**The app is functional by the audit's own earlier steps**: it installed, built, booted, and answered every contract probe over real HTTP — including `POST /api/runs/{run_id}/join` and `POST /api/runs/{run_id}/leave`, the precise endpoints the truncated calls actually target. The run's cycle verdict was `accepted` 39/39 with zero corrections.

**Family history:** this is the third variant of the same extractor defect class — #952 fixed line-wrapped calls, template literals were handled, `+` concatenation was not. Rolls 1–5's authors happened to use template literals; slot 6's author used concatenation; the instrument's coverage of author styles decided a window-deciding measurement.

**Fix** (stacked, unmerged): reconstruct concatenated path expressions the same way template literals are handled — non-literal segments become path parameters (`'/api/runs/' + runId + '/join'` → `/api/runs/{param}/join`), so classification probes the path the UI actually requests.

Window-protocol disposition is the owner's ruling (recorded in the window record): score-as-measured 3/6, §5.1 reset, or an explicitly-labeled corrected-instrument re-measure of the stored artifacts.

jasonLadd:
**Merge as part of the window ruling, not before** — this fix changes the §5 scoring instrument, and what happens to the deciding roll's score is the owner's call.

`extract_ui_calls` truncated `'/api/runs/' + runId + '/join'` to `/api/runs/` (template literals were handled; `+` concatenation was not — the third variant of the #952 extractor class). The audit probed the truncated path, got HTML, and failed a functional app on the window's deciding roll — an app whose cycle verdict was accepted 39/39 with zero corrections and whose join/leave endpoints the audit itself had already proven over real HTTP.

Fix: walk the concatenation chain — literals verbatim, expressions become the same inert placeholder as template expressions, unparseable pieces end the chain conservatively, messages keep the author's exact source.

Verified against the deciding roll's real file: lines 67/92 now reconstruct to `/api/runs/ui-probe-sample/join` and `/leave`. Four new tests (roll-6 shape, line-wrapped chains, arithmetic non-capture, written-as-authored); fail before the fix. Full regression: 7,612 passed.

Closes #1004

### DIFF
--- a/src/squadops/capabilities/ui_data_path.py
+++ b/src/squadops/capabilities/ui_data_path.py
@@ -56,6 +56,18 @@ _CALL_RE = re.compile(
 )
 _TEMPLATE_EXPR_RE = re.compile(r"\$\{[^}]*\}")
 
+#: A continuation of a concatenated path argument (#1004): ``+`` followed by another
+#: string literal or a bounded expression (an identifier chain, optionally one flat
+#: call — ``runId``, ``params.run_id``, ``encodeURIComponent(id)``). Whitespace around
+#: ``+`` may cross a line, matching the wrap tolerance above. The expression bound is
+#: deliberate: a piece this cannot parse ends the chain, and the caller records the
+#: partial path it can prove rather than guessing — but the bound must stay wide enough
+#: that an author's ordinary style never re-creates the truncation this exists to fix.
+_CONCAT_PIECE_RE = re.compile(
+    r"\s*\+\s*(?:(?P<q>['\"`])(?P<lit>[^'\"`\n]*)(?P=q)"
+    r"|(?P<expr>[A-Za-z_$][\w$.]*(?:\([^()\n]*\))?))"
+)
+
 
 @dataclass(frozen=True)
 class UiCall:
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
Reviewed

Issue #998, opened 2026-08-19
8192-token completion with zero extractable chars: thinking-cap exhaustion is undetectable in emission evidence


jasonLadd:
V7 window detection (roll 1 void, `cyc_8b569ce34074`, third develop dispatch 02:18–02:30Z). Recorded mid-window, left unfixed per pre-registration §6.

```
adapters.llm.ollama: completion_tokens=8192, t/s=10.7        # exactly the generation cap
emission_log: chars=0 completion_tokens=8192 fences={fill:0, path:0, plain:0} head=''
```

The model generated **8,192 completion tokens — its full cap — and the handler extracted zero characters**: 12.7 minutes of pure reasoning tokens that never closed into content. Downstream this banks as a generic emission failure (`no_fenced_blocks`, `response_chars: 0`), indistinguishable from a model that returned nothing at all.

Two asks:

1. **Detect the signature.** `completion_tokens == cap AND response_chars == 0` is a distinct, cheaply detectable failure class — "thought budget exhausted before content" — and it should be named in the emission-failure evidence rather than lumped with empty responses. The two have opposite remedies (raise/limit the thinking budget vs retry/investigate the model), and a correction decision that can't tell them apart guesses.
2. **Consider the cap and the induced-rumination pattern.** This was attempt 3 against the same 9.3k-token prompt after two rejections — the context (task + prior failures) drove the model into runaway thinking. Whether the fix is a larger `num_predict`, a thinking limiter, or prompt-shape change on re-dispatch is a post-window design question; the detection in (1) is what makes the pattern measurable first (instrument-before-fix).

Cost in this roll: the final 12.7 minutes of a void roll, and a banked analysis whose causal guesses ("safety filter, context overload") had nothing to anchor on.

PR #1006, opened 2026-08-20
feat(profiles): full-38 comparison arm + V38 model-comparison pre-registration

jasonLadd:
Kicks off the qwen3.6 → qwen3.8 comparison per the owner's ≥4/6 gate (met by V7's amended close).

- **`full-38`**: the `full` roster verbatim on `qwen3.8:27b` — only this profile added; `full` untouched (it is the meaning of every historical record).
- **Pre-registration** (`docs/plans/1-6-0-v38-model-comparison-preregistration.md`): V7 attempt 2's recipe with exactly one variable changed. Inherits §5 scoring (corrected instrument), the §7.3(c) verbatim gate text, §5.1 validity rules, and the no-reroll rule. Declares the permissible deploy delta (only #1005 + the inert profile entry), requires a declared non-counting shakedown first, and carries **no pass/fail bar** — it measures a delta against the banked V7 arm; texture (zero-correction rate, fill rates, store assertions, repair convergence, #998-class events, tokens, wall-clock) is the real instrument at N=6.
- Machinery fix queue explicitly deferred until after the window — fixing between arms would destroy the comparison.

Preconditions still open before the window can freeze: Ollama upgrade (needs interactive sudo — in progress), incumbent-model verification, 3.8 pull + t/s measurement, rebuild at merge, shakedown.

### DIFF
--- a/config/squad-profiles.yaml
+++ b/config/squad-profiles.yaml
@@ -22,6 +22,18 @@ profiles:
       - { agent_id: eve, role: qa, model: "qwen2.5:7b", enabled: true }
       - { agent_id: data, role: data, model: "qwen2.5:7b", enabled: true }
 
+  - profile_id: full-38
+    name: "Full Squad (qwen3.8)"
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #1007, opened 2026-08-20
docs(v38): ruling (a) — stack comparison, both deltas named; preconditions 2.2/2.3 MET

jasonLadd:
Recovers two commits orphaned by a merge race on #1006 (I pushed after the merge — my miss).

- **§1/§2.5**: owner ruling (a) folded in — the window is a **stack comparison** (incumbent 3.6-on-Ollama-0.21 vs 3.8-on-0.32), both deltas named, the declined pure-model isolate recorded as a knowing choice, and the emission-shape difference (inline vs routed thinking) documented.
- **§2.2/2.3 → MET** with measurements in §3: Ollama 0.32.14, incumbent verified through the deployed adapter (12.3 t/s warm); `qwen3.8:27b` digest `22130167c4c2`, **24.0 t/s** warm generation / 104.8 prompt eval, architecture question recorded rather than guessed.

Docs-only; needed on main before the §3 freeze so the pre-registration in force is the amended one.


jasonLadd (commented):
reviewed

PR #1019, opened 2026-08-21
fix(correction): emission-side ownership veto (#1014)

### LINKED ISSUE #1014: Correction repair emissions bypass the #884 ownership veto — a dev repair stored a rewrite of the qa suite, and the retest ran it
## What happened

V38 window, slot 6 (`cyc_cac1e479a462`, impl `run_d826c8de6d97`), correction round 0, 2026-08-21.

The qa suite failed on a genuine app defect (join returned 200 where the contract requires 201). The correction chain worked as designed up to the repair: the analyzer implicated one file, the lead prescribed a targeted patch, and the #884 targeting veto fired correctly —

```
11:41:14  correction_repair_target: ownership veto (#884) — qa-owned __tests__/runs-api.test.ts removed from dev-role repair target
```

The dev's repair emission then included a **full 230-line rewrite of `__tests__/runs-api.test.ts` anyway**, and it was **stored** (`art_5157b69336e4`, 11:50:43) alongside the seven app files. The rewritten suite abandons the scaffold harness pattern the qa brief mandates (imports route handlers directly) and asserts the dev's own changed response shape — the repair had also dropped `participantCount` from the join response, so rewriting the assertions converts his own contract deviation into a green expectation.

## Mechanism

The #884 veto edits the repair **target list in the brief** (`adapters/cycles/correction_runner.py:509`). Nothing filters the **emission**:

- `adapters/cycles/correction_runner.py:1338` — repair step artifacts are collected into `repair_artifacts` with no ownership check (only the #507 path rebase).
- `adapters/cycles/dispatched_flow_executor.py:2902` — `overlay_artifacts(...)` overlays them onto the failed task's artifacts, so a same-named file **replaces the qa original** in the patched tree.
- The behavioral retest runs that tree: **the dev was graded by his own suite**, not qa's.
- On accept, the #389 swap would re-store the dev-authored suite under the qa task's own type — laundering the provenance permanently.

## Why the blast radius was contained this time

The frozen scaffold probe tests ride in `workspace_files` (`dispatched_flow_executor.py:2914`), a separate base the overlay never touches. They derive deterministically from the contract and kept the retest red regardless. On any task where the scaffold probes don't cover the miss, the suite swap goes undetected and the loop self-accepts.

## Proposed fix

Apply the same ownership rule at the emission-collection seam (`correction_runner.py:1338`): when `step_role != own_artifact_role(failed_task_type)` (`src/squadops/cycles/task_plan.py:255`), drop emitted files owned by the other role, with a loud mirror log line:

```
correction_repair_target: ownership veto (emission, #NNN) — dev-emitted qa-owned __tests__/runs-api.test.ts discarded
```

**Scope decision for review** — two candidate match rules:
1. **Own-artifact set** (narrow): drop emitted files matching the failed task's own-artifact names — exactly today's case, zero risk of over-blocking.
2. **Collection surface** (recommended): drop anything the qa collection surface would collect (the surface is already defined for the #884 "no test files found" class in `src/squadops/capabilities/verification_scaffold_gate.py`) — also covers a dev inventing a *new* test filename, which the narrow rule misses and which self-grades just as effectively.

One filter at this seam protects everything downstream: patch verification, the retest tree, the #389 re-store, and boot-audit assembly. An advisory prompt line ("test files are qa-owned; do not emit them — they will be discarded") lands with the repair-template revision in the companion issue; it saves the wasted tokens but is not the enforcement.

## Sequencing

Post-window queue (V38 §3 deploy freeze; reporting-only through the window). Evidence is fully banked: stored artifacts + Prefect log lines above; the incident is re-adjudicable offline.

## Related

- **#884 (closed)** — direct parent; its fix vetoed the *targeting* side only. This issue is the emission-side completion of the same ownership rule.
- **#969 (open)** — sibling: the *qa-role* repair path re-authors its own suite without the fill brief; together they cover both re-authoring paths for the suite.
- **#971 (open)** — evidentiary dependency: the failed qa task's emission is not persisted, which is why the overwritten original cannot be diffed from banked state.
- **#1002 (open)** — same defect class (self-serving tests): a dev suite asserting the dev's own deviation is the self-mocking family; #1002's in-cycle detector would be the second net.
- **#557 (open)** — downstream complement: post-retest governance acceptance review of repairs.
- Companion: #1015 (repair-loop focus: targeting/template/loop-context), filed together.

jasonLadd:
## Summary
- Completes #884 at the emission boundary: a repair step running under a foreign role can no longer LAND the failed task's own artifacts or anything on its test-collection surface — previously only the *brief* was vetoed, and V38 slot 6's dev repairs stored a full qa-suite rewrite on all three rounds (see #1014 for the incident and mechanism).
- Surface-based match rule as recommended in the issue: own-artifact set ∪ stack `test_file_patterns` ∪ `__tests__/` convention — catches the invented-new-filename variant the narrow rule misses. Over-matching is the safe direction (a dropped borderline file just leaves the original in the tree).
- Applied post-#507-rebase so the filtered names are the names the overlay would supersede; same-role steps (qa re-authoring its own suite on the #568 locus path) and undeclared-owner types pass through untouched; pattern derivation failure-isolated per the #870 precedent.

## Tests
- Flow (standalone runner): dev repair emits app fix + qa-suite rewrite + invented `extra.test.ts` → only the app fix reaches `repair_artifacts`. **Mutation-checked**: neutering the veto call makes this test fail.
- Non-regression: qa own-artifact repair keeps its suite emission (same-role).
- Scope guard: no-declared-owner task types untouched.
- Full `test_correction_runner.py`: 153 passed; lint/format clean.

Closes #1014

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -525,6 +525,62 @@ def _apply_ownership_veto(
     return stripped
 
 
+def _apply_emission_ownership_veto(
+    artifacts: list[dict[str, Any]],
+    failed_task_type: str,
+    step_role: str,
+    failed_own_artifacts: list[str],
+    test_file_patterns: tuple[str, ...],
+) -> list[dict[str, Any]]:
+    """#1014: the emission-side completion of #884's targeting veto.
+
+    #884 edits the repair *brief* — but the failing suite is in the step's
+    context as evidence, and a model can rewrite what it can see: V38 slot 6's
+    dev repairs emitted a full rewrite of the qa-owned suite on all three
+    rounds, storage accepted it, and the overlay handed the retest a suite the
+    dev wrote to match his own changes. This is the storage-side half: a step
+    running under a foreign role may not LAND the failed task's own artifacts
+    either, nor anything on the failed task's test-collection surface (basename
+    patterns plus the ``__tests__/`` convention — over-matching is the safe
+    direction here, same rationale as source-set exclusion; a wrongly dropped
+    borderline file merely leaves the original in the tree).
+
+    Applied AFTER the #507 rebase, so the names filtered are the names the
+    overlay would actually supersede — a wrong-directory emission that #507
+    re-homes onto a qa-owned expected path is caught by the post-rebase name.
+    Same-role steps and task types with no declared owner pass through
+    untouched, mirroring the targeting veto's scope exactly.
+    """
+    from squadops.capabilities.dev_capabilities import matches_test_file_patterns
+    from squadops.cycles.task_plan import own_artifact_role
+
+    owner = own_artifact_role(failed_task_type)
+    if owner is None or step_role == owner:
+        return artifacts
+    own = {str(a) for a in failed_own_artifacts if a}
+    kept: list[dict[str, Any]] = []
+    dropped: list[str] = []
+    for art in artifacts:
+        name = str(art.get("name") or "")
+        if (
+            name in own
+            or matches_test_file_patterns(name, test_file_patterns)
+            or ("__tests__/" in name)
+        ):
+            dropped.append(name)
+        else:
+            kept.append(art)
+    if dropped:
+        logger.info(
+            "correction_repair_emission: ownership veto (#1014) — %s-role step emitted "
+            "%s-owned %s; discarded, never stored",
+            step_role,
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
Reviewed

PR #1020, opened 2026-08-21
fix(handlers): registry completion clamp on the base chat path (#1011)

### LINKED ISSUE #1011: Correction-repair handlers bypass the per-model completion clamp — repairs run uncapped
Discovered during V38 shakedown #2, declared per §2.6 (not fixed — see symmetry note). Post-window queue.

Correction-repair handlers (`development.correction_repair`, `builder.assemble_repair`, `qa.test_repair`) ride the base `_CycleTaskHandler.handle` path, whose `_build_chat_kwargs` passes `max_tokens` **only when agent overrides carry `max_completion_tokens`** — it never consults `_resolve_model_budget`, so the registry's per-model completion clamp (8,192 for both 27b arms) does not apply to repairs. Observed: a dev correction repair emitted **12,314 completion tokens** under qwen3.8 (`cyc_bc85a4b81808`, 18:56Z). Develop tasks are clamped (they call `_resolve_model_budget` explicitly); repairs are not.

**Why declared rather than fixed mid-arc:** the gap is pre-existing and **symmetric across both V38 arms** — V7's 3.6 repairs ran the same uncapped path and merely self-limited below 8k, so the comparison is not biased by it. Fixing it during the frozen window would change the instrument between arms for no comparability gain.

**Fix direction (post-window):** route the repair mixin's emission through the same `_resolve_model_budget` clamp the develop handlers use — one seam, not per-handler caps. Also worth checking which other `_CycleTaskHandler` subclasses ride the unclamped base path (the registry comment's "capability defaults pass through unchecked" class, one level up).

jasonLadd:
## Summary
- The per-model completion clamp now applies to every handler riding `_CycleTaskHandler`'s base chat path — previously `max_tokens` was set only from an explicit `agent_config_overrides.max_completion_tokens`, so correction repairs (and every other base-path rider) ran at capability ceilings while develop/qa were clamped via `_resolve_model_budget`. Observed: 12,314-token repair under an 8,192-clamped model (V38 shakedown #2, declared §2.6); slot 6's three repairs ran 10–18.8k.
- One-seam fix per the issue's direction: no override → registry `default_max_completion`; explicit override still wins; unregistered model → unchanged behavior (no invented budget).
- This also answers the issue's "which other subclasses ride the unclamped base path" question by construction: all of them are now clamped at the same seam.

## Tests
- Clamp applies without override (registry premise asserted); override wins; unknown model keeps prior behavior; flow half asserts the clamped budget reaches `chat_stream_with_usage`.
- Full unit sweep: capabilities 2,064 passed, cycles 2,953 passed; lint/format clean.

Closes #1011

### DIFF
--- a/src/squadops/capabilities/handlers/cycle/base.py
+++ b/src/squadops/capabilities/handlers/cycle/base.py
@@ -336,7 +336,17 @@ class _CycleTaskHandler(CapabilityHandler):
         return model_name, max_tokens, context_window
 
     def _build_chat_kwargs(self, inputs: dict[str, Any]) -> dict[str, Any]:
-        """Build chat() kwargs from agent config overrides (SIP-0075 §3.2)."""
+        """Build chat() kwargs from agent config overrides (SIP-0075 §3.2).
+
+        #1011: without an explicit ``max_completion_tokens`` override, the
+        registry's per-model completion clamp applies — previously ``max_tokens``
+        was simply omitted here, so every handler riding the base path
+        (correction repairs foremost) ran at capability ceilings the clamped
+        develop/qa handlers never see (a dev repair emitted 12,314 tokens under
+        an 8,192-clamped model, V38 shakedown #2). An explicit override still
+        wins, and a model without a registry entry keeps the old behavior —
+        the clamp cannot invent a budget for a model it does not know.
+        """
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
Reviewed

PR #1026, opened 2026-08-21
feat(gates): manifest↔plan consistency + completeness gate (#1013)

### LINKED ISSUE #1013: No manifest↔plan consistency gate: a framing that teaches 200 while declaring 201 costs a counted roll instead of a void
V38 arm-B roll 1 detection (`cyc_02e9af402c82`), found answering "can a framing deficiency be a hidden root cause of implementation failure?" — yes, proven. Recorded mid-window, left unfixed per §6.

**The contradiction, verbatim from stored artifacts:**
- `interface_manifest.yaml`: `POST /api/runs → success_status: 201`
- `implementation_plan.yaml`, the dev subtask's own acceptance criterion: *"POST /runs with all required fields returns **200** and the created Run"*

Neo implemented his brief faithfully (200); the contract probe and the scaffold shells judged him by the manifest (201). The initial red of a counted roll roots in a **framing-internal contradiction** — same authoring session, same model, two artifacts disagreeing about the same endpoint. A V7 green framing checked as contrast (`cyc_4cd50a3a42a6`) declared 201 in both places.

**The harness gap:** nothing validates manifest↔plan consistency. Plan validation checks *winnability* (mechanically runnable checks — it correctly voided a roll for `node` on `.tsx`), but a plan teaching a status the manifest contradicts sails through the gate, and the roll then burns its correction budget on a defect authored upstream of the implementer. The gate had every byte it needed: both artifacts, same cycle, deterministic comparison.

**Fix direction (post-window):** a deterministic gate check — every status literal bound to a declared endpoint in plan `acceptance_criteria`/descriptions must match the manifest's `success_status`/error declarations; mismatch → `system:plan_validation` rejection (a §5.1 **void**, costing wall-clock, instead of a counted roll + 3 repair rounds). Also belongs in the retry corpus: C1's machinery-fix retry will still fail on this framing — the consistency gate is what should *refuse* it.

Texture note for the comparison record: this reattributes roll 1's status miss from "3.8 implementation discipline" to "3.8 **authoring self-consistency**" — arm A's sampled framing was internally consistent. Shakedown #2's 409 miss (taught correctly, missed anyway) remains a genuine implementation slip.

jasonLadd:
## Summary
The deterministic gate for the framing-internal class that cost both V38 counted reds and recurred on the 1.6.1 shakedown: a **contradiction** check (plan prose vs the status the contract will *enforce* — declared or derived default, so the pf-39 class counts) and an **omission** check (an enforced non-200 status must be stated in plan prose wherever the stack's skeleton doesn't pin it into frozen code). Rejection lands at the inter-workload seam → framing re-rolls for free instead of a doomed implementation run.

Design points of note:
- **New per-stack fact at the S1 seam**: `skeleton_pins_success_status` — FastAPI's decorator pins it (pf-39 fix), Next.js's TODO comment doesn't survive the fill (slot 6's proof). The omission half fires only where prose is the sole channel; the typed channel is not a substitute in either mode because status probes bind to qa, never the dev's `criteria_refs`.
- **Conservative binding**: method-aware, full-path-only, success-family tokens only, enforced-status-present ⇒ consistent. A false positive rejects a good framing; the tests pin the false-positive shapes.
- **Taught, not just gated** (#686): `statuses-must-agree-with-the-manifest` added to the authoring-rules asset (v4) and registered `AUTHOR_FACING` — both drift guards fired during development and are satisfied.

## Would-have-caught
- V38 roll 1 (`cyc_02e9af402c82`): manifest 201 / plan "returns 200" → refused at plan validation
- V38 slot 6 (`cyc_cac1e479a462`): manifest-only join 201 → omission named with the remedy
- pf-39: undeclared collection-POST 201 vs plan 200 → contradiction on the derived default

## Tests
10 new, each keyed to a banked failure or a false-positive shape; corpus C1 and C6 retries are this gate's pre-registered acceptance tests (refusal expected on both framings). Full regression 7,640 passed; lint/format clean.

Closes #1013

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -3743,6 +3743,33 @@ class DispatchedFlowExecutor(FlowExecutionPort):
         if interface_content is not None:
             errors.extend(self._validate_interface_manifest(interface_content))
 
+        # #1013: manifest↔plan consistency + completeness — the two framing-internal
+        # species that cost V38 counted rolls (roll 1's 201-vs-200 contradiction; slot
+        # 6's manifest-only 201 the plan never stated). Needs BOTH artifacts, so it
+        # runs after the collection loop; an unparseable manifest defers exactly like
+        # an unreadable plan does (this check only ever adds an earlier rejection,
+        # never a pass). Lands on THIS seam deliberately: the reject must re-roll
+        # framing for free, not fail a run at dispatch (see the ownership note above).
+        if interface_content is not None and parsed_plan is not None:
+            try:
+                from squadops.capabilities.scaffold import InterfaceManifest
+
+                parsed_manifest = InterfaceManifest.from_yaml(interface_content)
+            except Exception:
+                logger.warning(
+                    "Manifest unparseable before gate %r — #1013 consistency check "
+                    "deferred to the manifest's own validation net",
+                    gate_name,
+                    exc_info=True,
+                )
+            else:
+                errors.extend(
+                    classifier.collect(
+                        "validate_manifest_plan_consistency",
+                        parsed_plan.validate_manifest_plan_consistency(parsed_manifest),
+                    )
+                )
+
         # SIP-0098 98.3: bind-mode contract validation. A seeded contract_ref switches the
         # cycle to bind mode — the plan must bind the contract's covered-file criteria by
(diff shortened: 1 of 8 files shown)


jasonLadd (commented):
reviewed

PR #1027, opened 2026-08-21
fix(scaffold): error-envelope authority settled — nextjs on message, M2 rejects root-contradicting shapes (#795)

jasonLadd:
## Summary
Settles #795 per its own preferred resolution (the envelope is blueprint-owned) and thereby **unblocks #913** (shape pinning in shells needs one envelope truth):

- **Generator half:** nextjs `lib/errors.ts` unified on `message` — stack #1, the reference manifests, and authoring habit all said `message`; the nextjs generator's `detail` was the sole divergence (the ErrorSeam docstring carried this as an open confession). `ApiError` routes message through `super()` (no shadowing parameter property); the changed skeleton `tsc --noEmit`s clean in the real sandbox against a banked window manifest.
- **Gate half:** M2 winnability finding `error_shape_agrees` — a declared `error_contract.shape` must be `error`-rooted. Rejects exactly the V4-roll-1 class (`{"detail": "string"}`); deliberately root-only so banked green manifests (`{"error": "..."}`, vague-but-true) keep passing. Classified `AUTHOR_FACING` + taught in the manifest authoring-rules asset (v2).

## Rituals followed
- GENERATOR_VERSION 4 → 5, fixture regenerated, pins verified: shells and spine hash byte-identical, only `expanded_tree_hash` moved — the version-3 discrimination shape, noted in the pin comment.
- Both #686 drift guards (proof classification ↔ asset rules) fired during development and pass.
- Downstream pins updated deliberately: two `generator_version` literals, the #875 constructor surface.

## Tests
New: V4-class rejection, vague-shape pass, absent-shape pass. Updated: seam/envelope teaching pins to the unified truth. Full regression **7,644 passed**; sandbox tsc clean.

Closes #795

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01ApBek6CA1ibm3Cc5pq57F2

PR #1028, opened 2026-08-22
feat(scaffold): frozen-spine error-envelope pinning (#913)

jasonLadd:
## Summary
The shells now own the response-field path for error behaviors: where the derived contract pins a code, the frozen spine asserts `body.error?.code` before the fill slot. Closes the last named boilerplate category SIP-0104's §"Why now" lists (imports, call shapes, placement, **response-field paths**) — the one rolls 2/3/13/17 all died re-inventing. Unblocked by #795's envelope-authority settlement (merged as #1027).

## Verification, three layers
- **Unit**: pinned code present exactly once in rejection/duplicate shells; placement is spine-not-slot (an assertion inside the slot is deletable fill residue again); success shells pin nothing.
- **Ritual**: GENERATOR_VERSION 5→6, fixture regenerated, pins updated — the version-4 shape (shells moved, spine hash moved, and the two pinned shells now differ from siblings, so the pin discriminates per-behavior emission).
- **Behavioral, in the sandbox against banked artifacts**: fresh v6 shells vs slot 5's green app → 9/9 pass (no false rejection — deliberately envelope-second-field-agnostic, so even the detail-era app passes); `lib/errors.ts` mutated to the invented `{error_code}` form → exactly the two pinned shells fail, siblings green.

## Not in this PR
Success-body shape pinning (today's shakedown killer: participants objects-vs-strings) — feasible (entities carry typed fields) but its own derivation design; sizing next.

Full regression 7,647 passed.

Closes #913

### DIFF
--- a/src/squadops/capabilities/stack_nextjs_ts_tests.py
+++ b/src/squadops/capabilities/stack_nextjs_ts_tests.py
@@ -108,6 +108,11 @@ class _Behavior:
     final: _Step
     probe_id: str = ""
     prerequisites: tuple[_Step, ...] = field(default=())
+    #: #913: the error code the contract pins for this behavior ("" = none pinned).
+    #: When set, the frozen spine asserts `body.error?.code` — the response-field
+    #: path stops being re-invented in the fill, which is where window rolls 2/3
+    #: died (fills asserting `body.error_code` against the real `{error: {code}}`).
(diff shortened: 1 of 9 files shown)


jasonLadd (commented):
reviewed

PR #1033, opened 2026-08-22
fix(cycles): framing_max_rerolls defaults to 2 (#1030)

jasonLadd:
## Summary
- The #522/#669 re-roll machinery is on by default (2, matching `manifest_max_attempts`'s posture); explicit `0` remains the deliberate disable. The shipped default of 0 — never overridden by any profile — left the machinery dead code until the first live #1013 rejection dead-ended a cycle after a *correct* refusal (`cyc_9bb225e19448`, the banked reproducer; the overridden relaunch converged green end-to-end).
- **Reachability test added**: the re-roll branch must fire under *default* resolved config — every existing test in the class passed the knob explicitly, which is precisely how the gap survived them. The old default-zero pin is converted to an explicit-disable pin; the rejection-record test class is isolated with explicit 0 so it keeps testing the record in isolation.

Full regression 7,648 passed.

Closes #1030


jasonLadd (commented):
Reviewed

PR #1034, opened 2026-08-22
fix(scripts): regression-gate preflight + positive completion evidence (#972)

### LINKED ISSUE #972: run_regression_tests.sh exits 0 when ruff is missing — the lint gate silently does not run without an active venv
`./scripts/dev/run_regression_tests.sh` fails with a bare tool-not-found when the virtualenv is not active, instead of saying so.

Observed 2026-08-17:

```
Running ruff lint + format check (fail-stop)...
./scripts/dev/run_regression_tests.sh: line 41: ruff: command not found
```

Exit code was **0** on the first run despite the check never executing — the message scrolled past above a wall of other output, and it took a second run with the venv active to notice the lint gate had not run at all.

That is the actual defect. A fail-stop gate that reports "command not found" and exits 0 is a gate that silently does not run, which is the class SIP-0096 exists to eliminate — the same shape as a check that looks enforced and is not.

## Fix

Two small things:

1. **Fail loudly on a missing tool.** `ruff` absent is not a skip; the script's own comment calls this section fail-stop, and CI runs it under `bash -e` where it does stop. Locally it does not.
2. **Name the likely cause.** `ruff not found — activate the virtualenv (source .venv/bin/activate)` costs one line and removes the guess.

Minor, but it is a dev-tooling gate that can pass without running, and it is the script contributors are told to trust before committing.

jasonLadd:
## Summary
Adjudication first: the script already exits 127 on a missing tool (`set -e`) — the observed exit-0 runs (the issue's and the one at the v1.6.0 cut) were **caller-side pipe masking**. So the fix targets evidence, not exit codes:
- **Preflight**: `ruff`/`pytest`/`python` checked first; a miss is one loud stderr line naming the fix (`source .venv/bin/activate`), before any gate banner.
- **Positive completion line**: `ALL GATES PASSED (…)` as the script's final act — masked exit codes can fake nothing when the *absence* of the line is visible in scrollback.

## Tests
Missing-tool run → 127 + hint + no gate banner ever printed; the completion line is asserted to be declared after the final gate. Full regression green (now ending with the line).

Closes #972

### DIFF
--- a/scripts/dev/run_regression_tests.sh
+++ b/scripts/dev/run_regression_tests.sh
@@ -11,6 +11,18 @@
 
 set -euo pipefail
 
+# --- #972: tool preflight — fail LOUDLY with the likely cause, first thing. ---
+# A bare "ruff: command not found" buried mid-scroll reads as noise, and a
+# caller piping output (| tail) masks the exit code entirely — at the v1.6.0
+# cut that combination produced a green-looking run in which no gate executed.
+# One clear line, before anything else runs.
+for _tool in ruff pytest python; do
+  if ! command -v "$_tool" >/dev/null 2>&1; then
+    echo "ERROR: '$_tool' not found — activate the virtualenv first: source .venv/bin/activate" >&2
+    exit 127
+  fi
+done
+
 REGRESSION_DIRS=(
     "tests/unit/api/"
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
Reviewed

PR #1035, opened 2026-08-22
fix(cycles): retry after a failed run — typed workload-position resolution (#880)

jasonLadd:
## Summary
Fixes all three interlocking refusals from #880 (roll 14's stuck-queued retry):
1. **Executor position map**: a run's own `workload_type` is authoritative (`_typed_workload_position` — ordinal-disambiguated, clamped so a failed run and its retry share the slot); untyped legacy runs keep the positional map byte-for-byte.
2. **Forwarding rebuild**: prior = the last **completed** run at the preceding *position* (typed-first), so the retry receives the framing's promoted plan instead of degrading on the failed sibling (the logged RC-4 shape).
3. **CLI resolver**: sequence overflow with a typed latest run now resolves to that type — the natural retry — instead of 422ing with advice #522 refuses by design; the remaining legacy raise names `workload_type` or `runs resume`.

## Why it matters beyond retries
This also unblocks the fix-validation retry program (`fix-retry-validation-corpus.md`): C1–C5 all have FAILED implementation runs and were retry-dead before this.

## Tests
Roll-14 shape end to end at all three layers; legacy-positional and failed-prior-degrade pins keep the old contracts where they should hold. Full regression 7,651 passed.

Closes #880

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -984,21 +984,66 @@ class DispatchedFlowExecutor(FlowExecutionPort):
             )
             i += 1
 
-    async def _starting_workload_index(self, cycle_id: str, first_run_id: str) -> int:
-        """Workload-sequence index that ``first_run_id`` occupies (#257).
+    @staticmethod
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
Reviewed

PR #1036, opened 2026-08-22
fix(cycles): carry the inspected-set reference to the record — a clean detector verdict is no longer indistinguishable from one that never looked

### LINKED ISSUE #1002: Detector-flaggable self-mocking file never flagged in-cycle; 'inspected' dies at the normalize seam so the record can't say why
V7 attempt 2 detection (roll 1, `cyc_6495d9870587` / `run_44c68d150155`). Recorded mid-window, left unfixed per §6. Not reset-class: the roll failed on genuine grounds (`tests_pass`), and a hypothetical false-green from this gap cannot corrupt the FAY headline (the boot audit is the functional oracle) — it corrupts suite evidentiary value (SIP-0096's concern).

**The reproduced event.** A stored artifact from a *completed* repair round (`art_2bd40853c575`, `__tests__/ui-flows.test.ts`) contains `vi.spyOn(global, 'fetch').mockResolvedValueOnce(...)` ×6 with **no route-module import**. The deployed detector flags it when handed the file directly:

```
offenders: [('__tests__/ui-flows.test.ts', 'replaces the fetch seam and imports no route module…')]
```

Yet `no_self_mocking_tests` appears once in the run's **verified** list, zero times in failed, and eve's logs contain zero detector firings across the whole implementation (3 qa attempts + 3 repairs). A detector-flaggable file existed in-cycle; no attempt flagged it.

**Why the record cannot resolve it (the actual defect).** #986 added `inspected` — the files each detector actually read — precisely for this question. But:
- `normalize_task_checks` maps the row to a `CheckResult`, which has no field for `inspected` — the list dies at the normalize seam;
- the failing final attempt's raw outputs aren't banked (failed-task artifacts, #971's family);
- passing attempts' `execution_evidence` isn't persisted either (#999).

So the instrument measures, and the measurement is unreachable exactly when needed — the third member of a family with #995 and #999: **evidence computed at the handler, dropped before the record.**

**Asks:**
1. Carry `inspected` (or a hash/count of it) through normalization — `CheckResult.provenance` already exists for exactly this kind of fact (it carries `exit_code` today).
2. With that landed, the open question becomes answerable from any run: whether the per-attempt artifact set fed to the detectors excludes prior-round additive files (the leading hypothesis: each attempt inspects only its own emission's files, so a spyOn file from round 2 is invisible to round 3's detector — same artifact-set-scoping class as #994's rewind behavior).

jasonLadd:
## What

`no_self_mocking_tests` reporting clean and a detector that never received the file were the same row in the record. #986 built the fix for that — an `inspected` inventory naming the files each detector actually read — and it was dropped one hop later: `normalize_task_checks` maps a producer row onto `CheckResult`, which had nowhere to put it.

This carries it to the persisted record.

## The event

`cyc_6495d9870587` / `run_44c68d150155`. A stored artifact from a *completed* repair round (`art_2bd40853c575`, `__tests__/ui-flows.test.ts`) contains `vi.spyOn(global, 'fetch').mockResolvedValueOnce(...)` six times with no route-module import. Handed that file directly, the deployed detector flags it. Across the whole implementation — three qa attempts, three repairs — the detector fired zero times and `no_self_mocking_tests` appears once in the run's **verified** list.

The instrument measured. The measurement was unreachable exactly when it was needed.

## Bounded, not a payload copy

SIP-0096 §7 admits identifiers, hashes, counts and exit metadata onto `CheckProvenance`, and excludes payload copies. A path list is a payload, so the reduction happens at the normalization boundary — the producer adapter that already owns shape translation — rather than asking every producer to pre-digest:

- `subject_ref` — sha256 over the deduplicated, **sorted** members. Sorted so the digest identifies the *set*, not the traversal order; two attempts reading the same files have to compare equal or the disclosure answers a different question than the one asked.
- `subject_count` (new) — the set's cardinality. The digest proves two attempts differed; the count says how.

That keeps the §7 contract intact, so no SIP amendment is involved.

## The whole chain, because a link computed-and-dropped is the bug being fixed

| Hop | Change |
|---|---|
| Normalize seam | `_inspection_provenance` on both producer shapes (boolean rows *and* typed-acceptance rows) and on the not-executed path |
| Aggregation | one `CheckInspection` per `(check_id, subject)`, collected **after** §6.5 final-state resolution, so the disclosure describes the same attempt the verdict does |
| Persistence | serialized both directions by the Postgres registry — answerable from any stored run, not only a live one |
| Run report | `### Inspected subjects` → `**no_self_mocking_tests** (task task-qa-3): inspected 0 files` |

Two edges that decide whether this works at all:

- An **empty** inventory survives as `subject_count == 0`. A detector that inspected nothing is the strongest form of the signal, and a truthiness guard is exactly what would erase it.
- An **absent** `inspected` key stays `None`. A producer that declares nothing must never be recorded as having inspected zero — that invents the defect the field exists to report.

`tests_pass` carries exit-code provenance and inspects no file set; it is not reported as an inspection.

## What this makes answerable

The leading hypothesis from the issue: each attempt inspects only its own emission's files, so a `spyOn` file added in round 2 is invisible to round 3's detector — the same artifact-set-scoping class as #994's rewind behavior. With per-`(check_id, subject)` counts in the stored summary, that reads off any run instead of requiring a replay.

## Verification

- 4 test files, 19 new tests. Every test names the bug it catches.
- **15 mutations run against them, all killed** — including order-sensitive digest, falsy-count guard dropping zero inventories, provenance attached only on the executed path, typed-acceptance branch unwired, collection before §6.5 resolution, and both serialization directions.
- Full regression: **7,674 passed**, 1 skipped.

## One pre-existing failure, not from this branch

`tests/unit/capabilities/test_stack_blueprint_falsification.py::test_the_pass_reports_what_it_did_not_exercise` fails on clean `main`, and at `v1.6.1`. It is a cross-test global-state dependency: `_RUNTIME_ONLY_OBSERVED` is written only inside `test_every_declared_field_is_accounted_for` (line 354), never inside `_classify`, so the disclosure test — which calls `_classify` itself but not that surrounding branch — passes only when a sibling parametrized case reached line 354 **in the same process**. xdist distribution decides that, so the gate can go red or green on identical code. Filed separately.

Closes #1002

### DIFF
--- a/adapters/cycles/postgres_cycle_registry.py
+++ b/adapters/cycles/postgres_cycle_registry.py
@@ -37,6 +37,7 @@ from squadops.cycles.models import (
 )
 from squadops.cycles.pulse_models import PulseVerificationRecord
 from squadops.cycles.verification_integrity import (
+    CheckInspection,
     FailedCheck,
     RunVerdict,
     RunVerificationSummary,
@@ -648,6 +649,18 @@ def _verification_summary_to_dict(summary: RunVerificationSummary) -> dict:
         # Postgres (rather than the in-memory summary) undercounted (#500).
         "criteria_verified": list(summary.criteria_verified),
         "criteria_total": list(summary.criteria_total),
+        # #1002: the bounded inspected-set references. Stored so the question
+        # "did this detector ever see the file?" is answerable from any run
+        # after the fact — which was the whole point of computing it.
+        "inspections": [
+            {
+                "check_id": i.check_id,
+                "subject": i.subject,
+                "subject_ref": i.subject_ref,
+                "subject_count": i.subject_count,
+            }
+            for i in summary.inspections
+        ],
     }
 
 
@@ -675,4 +688,13 @@ def _verification_summary_from_dict(d: dict) -> RunVerificationSummary:
         passed_count=int(d.get("passed_count", 0)),
         criteria_verified=tuple(d.get("criteria_verified", [])),
         criteria_total=tuple(d.get("criteria_total", [])),
+        inspections=tuple(
+            CheckInspection(
+                check_id=i["check_id"],
+                subject=i.get("subject"),
+                subject_ref=i.get("subject_ref"),
+                subject_count=(None if i.get("subject_count") is None else int(i["subject_count"])),
+            )
+            for i in d.get("inspections", [])
+        ),
     )
--- a/src/squadops/cycles/run_report_builder.py
+++ b/src/squadops/cycles/run_report_builder.py
@@ -170,6 +170,20 @@ def _build_verification_lines(summary: RunVerificationSummary) -> list[str]:
         for u in summary.unverified:
             tag = " [required]" if u.required else ""
             lines.append(f"- **{u.check_id}**: {u.reason}{tag}")
+    if summary.inspections:
+        # #1002: a clean detector verdict and a detector that never received the
+        # file are the same line in this report without these counts. `cyc_6495d9870587`
+        # carried a flaggable self-mocking file through three attempts with
+        # `no_self_mocking_tests` reporting clean each time, and nothing recorded
+        # could say which of the two had happened.
+        lines.append("")
+        lines.append("### Inspected subjects")
+        for i in summary.inspections:
+            n = i.subject_count
+            files = "1 file" if n == 1 else f"{n} files"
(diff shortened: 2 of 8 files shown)


jasonLadd (commented):
reviewed

PR #1037, opened 2026-08-22
test(2c): the blueprint falsification gate stops depending on xdist's work distribution

jasonLadd:
## The defect

The blueprint falsification suite's coverage disclosure, `test_the_pass_reports_what_it_did_not_exercise`, read a module-level dict that only its parametrized sibling ever wrote to. `_classify` computes a field's outcome; the *recording* lived in the branch after it, inside `test_every_declared_field_is_accounted_for` (line 354). The disclosure calls `_classify` itself but not that surrounding branch — so it passed only when a sibling case had run **in the same process**.

Under `pytest -n auto` that is a question about xdist's work distribution, not about the code.

## Measured, not inferred

Same commit, `main`, regression script:

| Workers | Result |
|---|---|
| `-n 2` | green |
| `-n 4` | green |
| `-n 20` | **red**, 3/3 |

CI runs the identical script on a 2–4 core runner, so it has never seen this and structurally cannot. A 20-core box fails it deterministically.

And it is new. Bisecting the gate (not the test in isolation — run alone it fails on *every* commit, for the same structural reason, which is what initially made this look like older debt):

| Commit | Regression |
|---|---|
| `v1.6.0` (7,615) | green |
| `v1.6.1` (7,647) | green |
| #1033, #1034, #1035 | green |
| **#1036** (7,674) | **red** |

#1036 added 19 tests to an unrelated module. That shifted the split enough to land the disclosure on a worker holding none of its siblings. Nothing about the subject under test changed — the gate's verdict was tracking the machine it ran on.

## The fix: derive, don't accumulate

`_disposition(registry, stack, field, baseline)` returns one field's verdict plus its citations, and both the gate and the disclosure call it. The mutable global is deleted, so there is no state left for two tests to share. Pure function, same answer in any order and any process.

## Two things found while proving it

Both are branches this PR rewrites, so both are in scope.

**The gate's rejecting assertions had no negative control.** Mutating either to a no-op left the whole suite green — a gate nobody had ever watched fire, in a file whose entire claim is that a new decorative field fails *here* rather than entering the frozen schema unnoticed. Added, driven through the real path with `_classify`/`_reader_citations` faked, plus the accepting half so the control cannot pass by failing everything. (The live declarations are all correctly accounted for, so there is no genuine decorative field left to trip it — which is precisely why the wiring needed its own proof.)

**The reader search re-parsed the tree on every call.** Harmless when it ran once per parametrized case on separate workers; calling it per field in the disclosure took that test from 2s to 27s. Trees are parsed once and the search is memoized — 7s. It returns a tuple now, since a cached mutable list lets one caller's edit rewrite another's evidence.

## The guard that matters going forward

`test_the_coverage_disclosure_stands_alone_in_a_fresh_process` runs the disclosure alone in a clean interpreter — the exact condition that failed. A future reintroduction of cross-test state fails there, deterministically, instead of on whichever machine happens to have enough cores.

## Verification

- Regression green at `-n 2`, `-n 4`, `-n 20`, `-n auto`; three consecutive `-n auto` runs, **7,679 passed**.
- 5 mutations run, 4 killed — including the order dependency reintroduced, and each gate assertion disabled. The survivor is the tuple return, which is defensive and has no observable behaviour; not claimed as covered.

### DIFF
--- a/tests/unit/capabilities/test_stack_blueprint_falsification.py
+++ b/tests/unit/capabilities/test_stack_blueprint_falsification.py
@@ -67,9 +67,12 @@ was wrong.
 
 from __future__ import annotations
 
+import ast
 import dataclasses as dc
+import functools
 import json
 import pathlib
+import sys
 from contextlib import contextmanager
 
 import pytest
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #1040, opened 2026-08-23
feat(scaffold): pin the success body's declared floor in the frozen shell spine

jasonLadd:
Three phases, one closed loop: **derive once, teach the developer, pin the shell.**

## The gap

The shells pin statuses, and since #913 the error envelope. Nothing pinned what a **success** body contains — so response-field paths were re-invented independently on both sides of the contract, and the disagreement surfaced as burned correction budget or a dead roll. Four samples in three days; the last green roll spent its *entire* correction budget (three qa rounds) on one class of it, all "expected undefined to be 'sample'".

The facts were already declared and simply unread: `entities[].fields` are typed and carry `required`, endpoints declare `response: Run` / `list[Run]`.

## The pinning budget

A floor, not a schema. Every exclusion below is a false-positive source rather than an oversight:

| Pinned | Not pinned, and why |
|---|---|
| declared-**required** fields are present | never an exact field set — a serializer that adds a field is making a legitimate choice |
| a `list[X]` field's elements match X's declared kind | nothing below the first level; a nested list is declined rather than guessed |
| generated fields present | never their value — the id exists, what it equals is the app's business |
| — | optional fields, ever: absent is a valid rendering of optional |
| — | ordering, or anything about extra fields |

An empty collection satisfies everything: no elements, no kind to violate.

Grounded in the banked corpus rather than argued. The green roll declares `participants: list[string]` and its app returns `string[]` — the floor passes it, so the pin costs that roll nothing. The 1.6.1 shakedown declares `list[Participant]` where `Participant` is `{name: string}`, and two successive dev attempts returned bare strings.

> Worth noting: this is the **reverse** of #1029's narrative, which has the app emitting objects and the suite expecting strings. The artifacts say the manifest declared objects and the app built strings — which matters, because it decides which side the teaching half has to reach.

## Replayed before it gates anything

In the qa container, against real vitest, with the green roll's 52 banked artifacts reconstructed to their final versions:

- delivered tree + pins → **27/27 pass**;
- inject a required field's absence into that same working tree → `expected undefined not to be undefined`, that roll's own loss class;
- inject the shakedown's defect → `expected 'object' to be 'string'`.

**Honest negative:** replaying the shakedown tree does *not* exercise the floor. That roll dies earlier on the request-side 400, and its collection is empty wherever the pin runs. The teeth above are proven by controlled injection on a live behavior, not by that replay. Empty-collection short-circuiting is deliberate, but it does mean the element rule only bites on behaviors that populate the collection — join/leave here.

## GENERATOR_VERSION 6 → 7

Per the procedure `test_verification_scaffold_reference.py` documents: bump, regenerate the fixture, update both anti-tautology hashes, say why. **Five of the eight reference shells move; the three error shells are byte-identical** — a 4xx body is the envelope's business, and that is the pin showing this emission is per-behavior rather than a blanket edit.

Three other tests moved with it:

- Two pinned the literal `6` while testing transport and reporting. A legitimate emission bump failed them, so they now assert against the constant; the deliberate value pin stays in the reference test where it belongs.
- The adversarial-producer attack *"moving a slot boundary above the assertion"* searched for the status assertion immediately followed by the slot marker. The floor now sits between them, so the mutation silently became a no-op — caught by that test's own "the mutation must change something" guard. Re-aimed at the new spine.

## Verification

- Regression **7,703 passed**, gate green at 20 workers.
- 7 mutations against phase 2, all killed. One survived the first pass: the primitive element-kind branch, which the reference manifest never exercises because its participants are entities. Dropping it emitted `for (const k of [])` — a loop over nothing that passes every app while reading like a check. Covered now by a variant manifest.
- 7 mutations against phase 1's derivation, all killed.

## Phase 3 — the developer is shown the floor it is judged against

Pinning the shell alone only converts burned rounds into a deterministic red: the gate gets stricter while the author it judges still cannot see the target. `development.develop` received the error contract, the model surface, the testids and the frozen index — and nothing about response shape.

This is the 1.6 defect class exactly: correct machinery wired to fewer paths than it claims, so an agent is judged against a fact it was never shown. A new `response_surface` renders per-endpoint lines from the **same** derivation the spine asserts:

> `POST /api/runs/{run_id}/join` returns `Run` — the response body carries `id`, `title`, `dateTime`, `location`; `participants` is an array of strings

On the 1.6.1 shakedown's manifest that line reads *"each `participants` element carries `name`"* — precisely what both dev attempts got wrong.

The producer lives in `response_shape.py`, the module that owns the derivation, not in `scaffold.py`: the floor already has one home and a second copy would be a second answer. Prose in the appendix asset, Python renders data only (#448).

**The invariant is a test.** Every field the emitted shells assert must be named in that endpoint's brief line — two independent renderings of "what the response must contain" would reintroduce #1029 inside the fix for it.

The context-assembly golden is regenerated in the same commit, per that harness's stated policy. Pure addition: 14 lines, only the new key, on the two dev-facing envelopes.

## A pre-existing gap this closed on the way past

A surface can be declared on the contract, derived, threaded onto the envelope, turned into a section by its own handler method — and then never added to the template's variables. Every step passes its own test and the agent still sees nothing. **No test covered that last step for any of the five dev surfaces.** Closed generically rather than for the new one, and confirmed by dropping a pre-existing surface (`frozen_surface`) and watching the new test fail.

## Deliberately not here

The probe `expect.json_has` producer and the `audit_delivered_app.py` parity fix it requires. That moves the **contract** hash on top of this branch's generator hash, and keeping them apart is what makes a regression attributable. The qa-side brief note (telling fill authors the spine already checks presence and kind, so fills should assert values) follows with it.

**Do not feed a cycle from this without rebuilding** — the deploy is well behind main.

Closes #1029

### DIFF
--- a/src/squadops/capabilities/context_assembly.py
+++ b/src/squadops/capabilities/context_assembly.py
@@ -51,6 +51,13 @@ SURFACE_DOM_TESTID = "dom_testid_surface"
 #: `reset, all, insert, find, nextId`, then repaired to the same invented name because the
 #: repair path was blind identically. A name the author cannot see is a name it will invent.
 SURFACE_FROZEN = "frozen_surface"
+#: #1029: what each endpoint's SUCCESS body must carry. The developer received the error
+#: contract, the model surface, the testids and the frozen index — and nothing about the
+#: shape of a successful response, so the suite derived one from the manifest and the app
+#: decided another. The last green roll spent its entire correction budget reconciling
+#: them; the 1.6.1 shakedown died on it. Pinning the shell without this only makes the
+#: gate stricter than the author it judges can see.
+SURFACE_RESPONSE = "response_surface"
 
 
 @dataclass(frozen=True)
@@ -131,7 +138,13 @@ ACCEPTANCE_WORKSPACE_FILTER = ArtifactFilter(
     by_type_fallback=("document",),
 )
 
-_DEV_SURFACES = (SURFACE_ERROR_CONTRACT, SURFACE_MODEL, SURFACE_TESTID, SURFACE_FROZEN)
+_DEV_SURFACES = (
+    SURFACE_ERROR_CONTRACT,
+    SURFACE_MODEL,
+    SURFACE_TESTID,
+    SURFACE_FROZEN,
+    SURFACE_RESPONSE,
+)
 
 CONTEXT_CONTRACTS: dict[str, ContextAssemblyContract] = {
     # --- build tasks (D3): curated prompt context + acceptance workspace ----
@@ -415,6 +428,9 @@ def manifest_surface_fragments(
     """
     if not contract.manifest_surfaces:
         return {}
+    # #1029 lives in the module that owns the derivation, not in scaffold.py: the
+    # response floor already has one home and a second copy would be a second answer.
+    from squadops.capabilities.response_shape import response_surface_instructions
     from squadops.capabilities.scaffold import (
         error_seam_instructions,
         frozen_surface_index_lines,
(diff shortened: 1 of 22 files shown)


jasonLadd (commented):
reviewed

PR #1043, opened 2026-08-23
fix(prompts): the developer is told the declared success status, not left to plan prose

jasonLadd:
## The gap

The manifest declares `success_status: 201`. On `nextjs_ts` the only channel carrying that fact to the implementer was **a sentence in plan prose** that some author had to remember to write.

`ScaffoldStack.skeleton_pins_success_status` (`scaffold.py:1935-1945`) says so directly: FastAPI lands `status_code=` in the frozen route decorator, so the implementer inherits it mechanically. The Next.js skeleton writes it as a TODO comment **inside the fill body, which the fill replaces** (`stack_nextjs_ts.py:481`). And `development.develop` binds no behavioral surface, so nothing else carried it at all.

`framing_consistency.py:134` compensates by firing its omission check only where that flag is False — a gate policing an author's prose in place of a missing derivation.

## Two instances of the same fact failing to travel

1. **V38 slot 6** — shipped 200 against a declared 201. The comment did not survive the fill; it is the reason that flag exists.
2. **`cyc_e3912098c0cf` run 1**, tonight — rejected by `system:plan_validation`:

   > manifest↔plan omission on POST /api/runs: the contract will enforce success 201 but no plan line states it for this endpoint — the implementer will default to 200

   The author chose the **right** convention. Only the restatement was missing. That roll was cancelled to fix this rather than continue.

## Not #1031, not #772

- **#1031** is the author choosing a *bad convention* — 400 where 404 is standard. Here the convention was correct.
- **#772** is the disagreement when a status is **not declared**. Here it is.

The defect is the transport. It is #1029's class one field over: a contract-enforced fact that reaches the agent only if an author restates it, instead of being derived and threaded like the model surface, the error contract, the frozen index, and the response floor.

## The fix

Threaded on the surface that already carries the body floor rather than a new one — same endpoint, same line, one derivation:

> `POST /api/runs` returns **HTTP 201** with `Run` — the response body carries `id`, `title`, `dateTime`, `location`; `participants` is an array of strings

Four deliberate boundaries, each mutation-covered:

| Decision | Why |
|---|---|
| status rendered **only when declared** | an undeclared endpoint takes the framework default, and which default is right is **#772's** open question — emitting one here settles it silently, in the wrong place |
| status and body floor are **independent** | an endpoint whose response names nothing the manifest defines still gets its status line; dropping it because the body is underivable loses the exact fact this fixes |
| neither ⇒ **no line** | a line that names an endpoint and states nothing reads as guidance |
| rendered for **every stack**, not gated on `skeleton_pins_success_status` | the fact is true whether or not the skeleton also pins it, and a stack-dependent brief buys nothing |

**The framing gate's omission check stays.** A second channel is not a reason to remove the first — it just stops being the only thing standing between a declared 201 and a shipped 200.

## Verification

- Regression **7,717 passed**, gate green at 20 workers.
- **5 mutations, all killed** — including "an undeclared status invents 200", which is the one that would quietly answer #772.
- Golden regenerated in the same commit per that harness's stated policy: **2 lines**, the single endpoint that declares a status. Nothing else moved.

Closes #1042

### DIFF
--- a/src/squadops/capabilities/response_shape.py
+++ b/src/squadops/capabilities/response_shape.py
@@ -156,32 +156,52 @@ def response_surface_instructions(manifest: InterfaceManifest | None) -> list[st
     the gate gets stricter while the author it judges still cannot see the target. This
     is the half that lets the app be right the first time.
 
+    The line also carries the endpoint's declared ``success_status`` (#1042). That fact
+    had exactly one surviving channel to a Next.js implementer — a sentence in plan prose
+    that some author had to remember to write. The skeleton writes it as a TODO comment
+    *inside the fill body*, which the fill replaces (``stack_nextjs_ts.py``), and
+    ``development.develop`` binds no behavioral surface, so nothing else carried it.
+    V38 slot 6 shipped 200 against a declared 201, and ``cyc_e3912098c0cf`` was rejected
+    at the plan gate for the same omission on the same endpoint kind. Same class as the
+    body floor above, one field over: derived and threaded beats remembered and restated.
+
+    A status is rendered only when the manifest DECLARES one. An endpoint that leaves it
+    unset takes the framework default, and which default that should be is the open
+    question in #772 — stating one here would be this module taking a position on it by
+    side effect.
+
     Data only — the prose lives in the appendix asset (CLAUDE.md #448). Empty for a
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

Issue #1022, opened 2026-08-21
Additive-suite containment: extend deterministic emission gating beyond fill slots — every V7 counted red was an unconstrained additive test


jasonLadd:
The V7 close-out named this "the top yield lever, to be filed" — and it never got a number. Filing it now so the 1.6.2 slate is complete.

## The gap

SIP-0104's fill-mode contains the qa author inside deterministic scaffold slots, and the scaffold gate (`verification_scaffold_gate.py`) validates that surface with named findings (placement, imports, handler references, status assertions). **Additive suites — the extra test files a qa author writes beyond the scaffold — have no equivalent containment.** They are unconstrained authorship landing in the same tree, and they are where arm A's rolls died.

## What it cost (banked)

- **Every V7 attempt-2 counted red was additive-suite-side.** Three rolls rejected with sole failed check `tests_pass` (genuine, suite executed) while all nine delivered apps of the arc passed independent boot audits — the applications were fine; the additive tests were not. Fixing this class converts arm A's amended 4/6 toward the app-side truth without touching the app path.
- **`cyc_6495d9870587` (V7 attempt-2 slot 1 / corpus C3):** additive suite fetched a live server that never exists under vitest — the #877 class, *after* #877's prompt-guidance fix (PR #879, merged) — three qa repairs, no convergence. Guidance alone demonstrably does not contain it.
- **`cyc_2913ae7abd67` (V7 attempt-2 slot 4 / corpus C4):** independent second sample, same class.
- **V7 attempt-2 roll 1:** a stored mid-round additive file spy-mocked `global.fetch` with no route import — the deployed self-mocking detector flags that file when handed it directly, yet nothing flagged it in-cycle (#1002's provenance half).

## Direction (design review decides the exact shape)

Extend the deterministic-containment principle from fill slots to additive emissions — named findings at the qa emission seam, same shape as the scaffold gate's, rejecting deterministically instead of letting vitest noise spend correction rounds:

1. **Execution-model conformance:** no live-server fetches in an in-process vitest harness (the #877 class as a *gate*, not guidance).
2. **Subject-import requirement:** an additive test must import the application (route handlers or the harness client) — a suite that touches nothing can only self-mock.
3. **Wire the existing authenticity detectors (self-mocking, stub-fallback) to fire at emission with banked evidence** — today they run late (retest path) or their provenance dies at the normalize seam (#1002, which stays the complementary fix).
[...]

PR #1045, opened 2026-08-23
feat(qa): the retreat had two halves — record the dropped additive suite too

jasonLadd:
Rebased onto main from a branch whose PR (**#982**) merged one commit earlier and left this behind — found during a branch audit, six days stranded.

## The gap

#980 gave the record what the fills *assert*, rather than merely that slots were filled. That covered one half of what shakedown #1 actually did.

Its passing attempt did not only weaken the fills. It dropped an **entire additive suite** — 81 store calls through `TABLES`, no mocking, a genuinely good suite written by the repair — and ran 8 test files where the failing attempt ran 9. `extracted_files` went 1 → 0 in a log line nobody read, and the banked evidence said only that the slots were filled.

A measurement of half a retreat reads as a whole one.

## The change

`assertion_strength` now banks the additive files that survived the merge alongside what the fills assert. Both halves or neither.

Three deliberate details, each mutation-covered:

- **Dropped shell rewrites are excluded.** A rejected attempt to edit frozen structure is not an additive test — counting it would make a suite look richer at the exact moment it was refused.
- **Sorted**, so two attempts are comparable.
- **An empty list is banked, not omitted.** The absence has to be visible *as* an absence, or a total retreat reads as "not measured" rather than "measured, and everything was dropped" — the same ambiguity #986 removed from the authenticity detectors.

## The test was replaced, not just rebased

The commit as stranded carried a test that built the banked dict inside itself and asserted the literal it had just set:

```python
banked = {**measure_assertion_strength(...), "additive_files": []}
assert banked["additive_files"] == []
```

It passed with the **entire source change deleted** — confirmed by mutation before this went up — and it never touched `_merge_fill_artifacts`, despite its own docstring claiming it asserted the handler's shape.

Replaced with two tests driven through the real handler, in `test_fill_mode_transport.py` where that harness already exists.

## Verification

- Regression **7,719 passed**, gate green at 20 workers.
- **5 mutations, all killed** — including the two the old test could not see: dropped shell rewrites counted as additive, and the empty list omitted.

## Related

#980/#982 (the half this completes), #1022 (additive-suite containment, still open in 1.6.2 — this is the measurement that pass would want in place first), #986 (absence-visible-as-absence).

### DIFF
--- a/src/squadops/capabilities/handlers/cycle/qa_test.py
+++ b/src/squadops/capabilities/handlers/cycle/qa_test.py
@@ -617,7 +617,14 @@ class QATestHandler(_CycleTaskHandler):
             # banked a retreat as a clean success because the record could not tell 8 rich
             # fills from 8 that had dropped every store assertion. Banked rather than logged
             # so the per-roll record reads it from stored state.
-            "assertion_strength": measure_assertion_strength(fill_emission),
+            "assertion_strength": {
+                **measure_assertion_strength(fill_emission),
+                # The retreat had TWO halves and the fills are only one. Shakedown #1's
+                # passing attempt also dropped an entire additive suite — 81 store calls
+                # through TABLES, no mocking — and ran 8 test files where the failing
+                # attempt ran 9. `extracted_files` went 1 -> 0 in a log line nobody read.
+                "additive_files": sorted(a["name"] for a in kept),
+            },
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #1046, opened 2026-08-23
docs(site): separate the framework's verification from the project's measurement

jasonLadd:
Jason's diagnosis, verbatim: the Evidence page was *"kind of a mashup between how the squad operates and how we (Jason and Claude) operate when building a new feature with FAY or roll windows."* Correct, and worth fixing carefully because the two layers genuinely interlock.

## The two layers

| | Verification | Measurement |
|---|---|---|
| **What** | Checks derived from the design, executed inside every cycle | Windows that score whole cycles, run by the maintainers |
| **Who gets it** | Anyone running a cycle — it is a framework feature | Nobody. It is how *this project* tests whether the framework works |
| **Produces** | A cycle verdict and an evidence roll-up | The numbers quoted on the site |

Measurement **consumes** verification — a window's scoring rule reads the cycle verdict as one input — which is exactly why they blur. But pre-registered windows, frozen deploy hashes and instrument-correction rulings are research methodology **about** the product, not a capability **of** it.

An evaluator reading the old page would reasonably have concluded that pre-registered measurement was something SquadOps does for them. It isn't.

## Changes

**Evidence** opens with the table above, states plainly that nothing on the page is something you receive, and says what it is instead: the reason to believe the claims made elsewhere on the site. Also now explicit that the boot audit and zero-intervention conditions are the maintainers checking the framework's verdict *independently* — a framework grading its own homework is not evidence.

**Overview** had the same defect in smaller form: *"the claims are measured, not demonstrated"* sat as the third of four things that "separate it", beside three genuine framework features. Now three framework claims, plus a separate **"Why believe any of it"** section that says outright it is not a feature you receive.

**Two leaks** where layer-2 vocabulary was propping up layer-1 explanations:

- `key-concepts.md` justified `resolved_config_hash` by saying it lets *a measurement window* claim a frozen deployment. It now says what it actually does for any user: a run carries proof of what it executed under, so a difference in outcome is attributable to the change you made rather than to drift.
- `architecture.md` said throughput *"bounds every measurement window"*. It now says throughput sets the wall-clock ceiling on every cycle.

Both properties stand on their own. Leaning on the measurement programme to justify them was the tell.

## Root cause

The maintainer vocabulary is what recent sessions have been *about*, so it leaks into product description by default. Recorded as a memory so the next session does not re-muddy it — the check is: which layer does this claim belong to?

Grepped the remaining product pages for `pre-registered` / `measurement window` / `re-roll` / `frozen deploy` — clean.

`mkdocs build --strict` passes; every changed page reviewed in a browser.

### DIFF
--- a/site/content/architecture.md
+++ b/site/content/architecture.md
@@ -65,9 +65,11 @@ model on an NVIDIA DGX Spark; development runs smaller models locally. No
 per-token cost, no rate limits, no source or requirement text leaving the
 machine.
 
-The cost is throughput, and throughput bounds every measurement window — which
-is why the inference engine sits behind a port with a conformance suite, and why
-alternatives are evaluated against a recorded comparison rather than a vibe.
+The cost is throughput. A local 27B model sets the wall-clock ceiling on every
+cycle, which is why the inference engine sits behind a port with a conformance
+suite every adapter must pass — swapping engines is a configuration change, and
+a candidate engine has to prove it honours the contract before it is trusted
+with a run.
 
 ## Design specs
 
--- a/site/content/evidence.md
+++ b/site/content/evidence.md
@@ -1,15 +1,32 @@
 # Evidence
 
-Agent frameworks are usually demonstrated. This one is measured, and this page
-is the whole record — method, results, and what the results do not support.
+**This page is about how the project evaluates itself — not about what the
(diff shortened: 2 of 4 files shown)


jasonLadd (commented):
reviewed

PR #1047, opened 2026-08-23
fix(correction): the tests_pass signature stops collapsing when the runner emits no machine report

jasonLadd:
1 of 4 in the 1.6.2 batch. Independent of the other three.

## The gap — and the half of it that was already closed

A4's progress-aware termination compares a normalized failure signature round over round. #761 reported it as maximally coarse — `repeated_signature: ["tests_pass||failed"]` — so a chain that fixed one failure and broke another (SHIFTED, real progress) was indistinguishable from one failing identically twice (REPEAT). A4 terminates the second.

**The issue is partly stale and does not say so**, because it was filed before both fixes it needed:

- **#626** added the `suite_broken` verdict.
- **#878** split the element per failing test wherever a machine report exists.

What remained is the runner that emits none — **pytest** — whose entire suite still reduced to one element carrying nothing but the word `failed`.

## What now joins the token

All three are already on the row (`failed_tests_pass_row`) and were simply unread. None is prose, so the module's standing rule — evidence text never alters a signature — is intact.

| Token | Separates |
|---|---|
| `exit` | "test failures" (1) from "no tests collected" (5) and "usage error" (4) — three defects, one signature |
| `runner` | a pytest failure from a vitest one; different suites, previously comparable |
| `cat` | the round's derived category — an app that **errored** vs assertions that merely failed, which no row field records |

## Two decisions worth review

**`cat` is emitted only when the category is not the ordinary `executed_and_failed`.** My first cut appended it unconditionally and broke three existing tests pinning *"a row without the newer fields renders byte-identically to its legacy form"*. That invariant is worth keeping, and omitting the baseline keeps it while losing no discrimination — the baseline is exactly the case where two rounds do **not** differ.

**`classify_failure_locus` is deliberately excluded.** It derives from largely the same signals as the category, and every added token raises the odds a genuine repeat reads as a shift — whose failure mode is that A4 never fires and the run burns its whole correction budget. Over-discrimination is the expensive direction here, so a marginal token has to earn its place.

## The invariant that had to survive

The category is computed **once per round**, not per row, so a partial fix remains a strict subset and still classifies as `PROGRESS`. A per-row category would have converted every progress round into a shift and silently re-armed the termination these tests exist to keep selective. Mutation-covered, and shk-4's canonical true positive still fires.

`_reason_token` was extracted when the additions pushed `failure_signature` past the complexity gate. It belongs apart regardless: the append **order** is the contract, since two rounds must build the token identically or every comparison is a shift.

## Verification

- Regression **7,728 passed**, gate green at 20 workers.
- **6 mutations, all killed.** One survived the first pass — a truthiness guard on `exit_code` dropping a real `exit=0`, which on a failing row means a suite the runner calls clean while the harness rejects it. Now covered.

Closes #761

### DIFF
--- a/src/squadops/cycles/correction_signature.py
+++ b/src/squadops/cycles/correction_signature.py
@@ -22,6 +22,7 @@ from __future__ import annotations
 
 from typing import Any
 
+from squadops.cycles.failure_evidence import FailureEvidenceCategory, derive_failure_category
 from squadops.cycles.verification_integrity import ResultStatus
 
 #: The decision-handler vocabulary value meaning "no structural candidate".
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

Issue #1015, opened 2026-08-21
Repair loop prompts mandate broad re-emission and hide the loop: no minimality imperative, no attempt counter, language-wide targeting fallback


jasonLadd:
## What happened

V38 slot 6 (`cyc_cac1e479a462`, `run_d826c8de6d97`), round 0: the analyzer implicated **one file** ("the join route handler… returns 200 instead of 201") and the lead's decision prescribed a **one-line** patch. The repair the dev received nevertheless mandated re-emitting **seven files**, and the dev delivered an 18,756-token 8-file rewrite whose gratuitous changes (response shape, error strings, validation refactor) left the retest red and consumed the round. The causal chain is mostly machinery, not model:

1. **Targeting fallback expanded 1 file → 7.** The #884 veto removed the qa-owned anchor; package scoping then had nothing to scope from and fell back to *every* same-language implementation source (`adapters/cycles/correction_runner.py:138`, the #688 fallback):
   ```
   correction_repair_target: package scoping matched nothing for anchors __tests__/runs-api.test.ts — falling back to same-language implementation source app/api/runs/route.ts, app/api/runs/[run_id]/route.ts, …
   ```
2. **The template converts the target list into a mandate.** `request.cycle_repair_task` (v3) renders it under "The repair **MUST produce the following file(s) by name**" — full-file re-emission is the required output format. The lead's "one-line fix" rationale rides along as advisory prose under "Correction Decision".
3. **No minimality instruction exists anywhere in the template.** The only focus-shaped constraint is the SIP-0099 fill-only section, which guards scaffold-owned *interface*, not change size within owned files.
4. **The dev cannot see the loop.** No attempt counter, no statement that rounds are finite, and on rounds ≥1 no summary of what the prior repair changed — each round renders only the fresh failure. Slot 4's rounds 1–2 emitted same-shape repeat failures; slot 6 round 1 repeated the pattern.

qwen3.8's re-emit-everything disposition makes this acute, but every element above is model-agnostic machinery.

## Proposed changes (one surface, three parts)

**A. Target from failure analysis before the language-wide fallback** (`correction_runner.py`, `_resolve_repair_target` at :263 and the #688 fallback at :138). When package scoping comes up empty — including when the #884 veto emptied the anchors — derive the target from the analyzer's implicated files (`failure_analysis`) and only then fall back to same-language-everything. Round 0's analysis named `app/api/runs/[run_id]/join/route.ts` explicitly; the correct target list was one file.

[...]

Issue #968, opened 2026-08-17
Nothing checks the failure analyzer's claims against the source — three false diagnoses in one roll, and the correction decision inherited the first verbatim


jasonLadd:
The failure analyzer emits confidently-worded factual claims about source code, **nothing checks them against the source**, and the correction decision inherits them verbatim.

#788 recorded one instance. SIP-0104 P6 window roll 6 produced **three in a single roll**, and the subject oscillated app → tests → app on identical evidence with no convergence. The roll then exhausted its correction budget rejecting an application that works (#967).

## The three claims, each refuted by the file it is about

| round | claim | what the source says |
|---|---|---|
| 1 | handlers declare "a **local shadow store array** instead of importing the scaffold-frozen store module" | `import { all, insert, nextId } from '@/lib/store';` — line 3 |
| 2 | fill assertions "expect truthy IDs and specific titles but **receive undefined values** from route handlers" | POST returns `insert(...)`'s row: `id` from `nextId()`, plus `title`, `datetime`, `location`, `participants`. `expect(body.id).toBeTruthy()` passes |
| 3 | POST "**fails to mutate** the in-memory run store"; join/leave "**omit expected fields**"; the routes are "**unimplemented**" | `insert('run_store', run)`; join is `find(...)` → mutate `participants` → `Response.json(run)`; the routes work |

Round 1's diagnosis was carried into the correction decision **word for word**, which then instructed the squad to "correct the store imports" — imports that were already correct.

## Why round 2 is not a self-correction

It reached a better *subject* (the tests) through a false premise, and round 3 swung back to the app. On identical evidence the component produced three different subjects and three false factual claims. That is not a convergence process with a bad first step; it is a process with no mechanism for being wrong.

## The structural gap

Nothing in the chain ever establishes **which side of a disagreement is at fault.** When a test asserts X and the app does Y, the analyzer picks one and asserts a cause, and the only input it is graded against is whether the repair happens to work.

Two candidate requirements, either of which would have caught all three:

1. **Cite or retract.** A claim about source must quote the source it read — `file:line` plus the line. Round 1 could not have written "declares a local shadow array" while quoting line 3.
2. **A cheap mechanical pre-check.** "Does this file import the frozen store?" and "does this handler call `insert`?" are `grep`-level questions. The expensive LLM call should be reserved for what a grep cannot answer.

## Bearing on the repair path

[...]

PR #1048, opened 2026-08-23
fix(correction): the repair is told to change the minimum, and can see the loop it is in

jasonLadd:
2 of 4 in the 1.6.2 batch. Independent of the other three.

**Parts B and C only. Part A is deferred — it is not buildable as the issue specifies it.** Reason below; `Refs`, not `Closes`.

## The sample

V38 slot 6 round 0: the analyzer implicated **one** file, the lead prescribed a **one-line** patch, the repair received **seven** files under *"The repair MUST produce the following file(s) by name"*, and the dev delivered an **18,756-token 8-file rewrite** whose gratuitous changes — response shape, error strings, a validation refactor — left the retest red and spent the round.

## B — the template states minimality

The only focus-shaped constraint in the template was the SIP-0099 fill-only section, which guards scaffold-owned *interface*, not change size inside files the repair legitimately owns. Nothing said "do not rewrite."

Template v4 adds: this is a repair, not a rewrite; the file list is what you **may** emit, not a list of things needing change; re-emit untouched files byte-identical; unrelated changes are how a round is lost. Plus the advisory ownership line — test files are qa-owned and are discarded (enforcement is #1014's half).

## C — the repair can see the loop

`correction_attempts` and `max_correction_attempts` both already existed at the dispatch site, and **neither crossed into the prompt**. The author could not tell round 1 from round 3, or know the budget was finite.

Now renders *"attempt N of M"*, and rounds after the first carry an explicit note that the failure already survived a repair. That targets #864's shape — *"diagnoses accurately, then re-emits it — twice"* — which is what a round rendering only the fresh failure invites. Prose in a managed asset, data from Python (#448).

## Part A, deferred with its reason

The issue says to target from *"the analyzer's implicated files (`failure_analysis`)"*. **The analyzer emits no such field.** Its declared output contract is `classification`, `analysis_summary`, `contributing_factors` — prose. Round 0 "named the join route explicitly" inside a sentence.

Deriving a repair target from prose is the same anti-pattern `correction_signature` bans for signatures, and the issue's own citations argue against it: **#968** (nothing verifies analyzer claims against the source) and **#788** (a repair followed an impossible analyzer recommendation into a worse failure).

Part A needs the analyzer to emit **structured** implicated files first — a schema change whose natural companion is #968's verification slice, currently 1.6.3. Recorded on the issue rather than half-built.

## Verification

- Regression **7,738 passed**, gate green at 20 workers.
- Goldens regenerated in the same commit per that harness's policy: **6 lines**, the two new keys on three repair scenarios, nothing else moved.
- **8 mutations, all killed.** Three survived the first pass and all three were **transport**: the section rendered but not injected, the variable computed but never placed in the template dict, the counter never put on the envelope. Same class #1040 found across all five dev surfaces, in the other handler — every step had a passing test while the fact reached nobody.
- One of my own tests also passed vacuously by patching the mixin's `handle` — the method under test. Now patched at the base.

Refs #1015

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -1358,6 +1358,13 @@ class CorrectionRunner:
                     "subtask_description": repair_description,
                     "expected_artifacts": step_expected_artifacts,
                     "acceptance_criteria": failed_inputs.get("acceptance_criteria", []),
+                    # #1015 part C: the loop's position. Both values were already here
+                    # and simply never crossed into the prompt, so the repair author
+                    # could not tell round 1 from round 3 or know the budget was finite.
+                    "correction_attempt": correction_attempts + 1,
+                    "max_correction_attempts": int(
+                        cycle.resolved_config().get("max_correction_attempts", 3)
+                    ),
                 }
                 # #667: fay-14's first fill complied with the manifest
                 # convention and every repair regenerated the view blind,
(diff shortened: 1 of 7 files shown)


jasonLadd (commented):
reviewed

PR #1050, opened 2026-08-23
fix(gates): the framing omission check reads BOTH channels — #1042 made its premise false

jasonLadd:
Filed and fixed tonight, because it is **actively blocking the validation rolls** — not a cleanup.

## The premise the check states is now false

`framing_consistency`'s completeness half rejects a framing when the plan omits a line restating the manifest's declared success status. Its rejection text names the harm it prevents:

> the contract will enforce success 201 but no plan line states it for this endpoint — **the implementer will default to 200 and the probe will reject a faithful build**

**#1042 threads the declared status onto the developer's brief.** The brief now reads ``POST /api/runs` returns HTTP 201 with `Run` — …`. The implementer does not default to 200, because it is told.

## What it costs, measured tonight

**Five identical rejections across three cycles**, all this check:

| Cycle | Rejections |
|---|---|
| `cyc_e3912098c0cf` | 1 |
| `cyc_181c9572bef2` | 2 — both re-rolls consumed |

The framing author omits this line reliably. With `framing_max_rerolls` defaulting to 2, that is a dead-ended run on a framing that was correct.

## The fix

The check already gated on **one** channel — `skeleton_pins_success_status_for`, because FastAPI pins a declared status in the frozen route decorator and prose silence is harmless there. There are now **two**, so the condition names both.

`brief_carries_success_status_for` sits beside its sibling: the dev brief carries the status only where the fill-only appendix renders, which requires a scaffoldable stack whose dev capability declares a template. A stack without one still has prose as its only channel and still blocks. The condition is *"no channel carries it"*, not *"we added a channel somewhere"*.

**The contradiction half is untouched.** Derivation fixes silence; it does not fix a document that says the wrong thing. A dev reading "returns 200" builds 200 no matter what an appendix also says, and the two channels then disagree. Softening that half with the same condition is the obvious wrong turn, so it is mutation-covered.

The stale reasoning is replaced in place rather than left to mislead — the comment reading *"the typed channel never carries this fact to the implementer in EITHER authoring mode"* was true when written and is now the part that is wrong.

## Tests

The slot-6 omission test moves to a stack with **neither** channel — now the only condition under which prose is the sole carrier, and what the gate exists for. Beside it: the same framing on `nextjs_ts` must now pass, and a contradiction on that stack must still fire.

## Verification

- Regression **7,726 passed**, gate green at 20 workers.
- **5 mutations, all killed.** One survived the first pass — `brief_carries_success_status_for` returning `True` regardless of the fill-only template. No registered stack reaches that guard today, so nothing exercised it; an unreachable guard is the kind that stops working silently when a third stack lands. Now tested directly with a stubbed capability.

Closes #1049

### DIFF
--- a/src/squadops/capabilities/scaffold.py
+++ b/src/squadops/capabilities/scaffold.py
@@ -2012,6 +2012,34 @@ def skeleton_pins_success_status_for(stack: str) -> bool:
     return bool(known and known.skeleton_pins_success_status)
 
 
+def brief_carries_success_status_for(stack: str) -> bool:
+    """Whether *stack*'s DEVELOPER BRIEF states declared success statuses (#1049).
+
+    The second deterministic channel, and the sibling of
+    ``skeleton_pins_success_status_for`` above — they are the two ways the fact can
+    reach an implementer without a human remembering to write it down, so they belong
+    side by side and are read together by the framing gate.
+
+    #1042 threads the declared status onto the dev brief's response surface, but only
+    where that surface renders: a scaffoldable stack whose dev capability declares a
+    fill-only template. A stack with no such template gets no appendix, so prose
+    remains its sole channel and the omission check must still block there.
+
+    ``False`` for unknown stacks and unresolvable capabilities — the conservative answer
+    when we cannot prove the fact travels is that it does not.
+    """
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

Issue #1021, opened 2026-08-21
Per-file compile criteria silently drop from criteria_verified — credited nowhere, failed nowhere, disclosed nowhere


jasonLadd:
An accounting gap in the SIP-0096 verification ledger: per-file compile criteria drop out of `criteria_verified` with no failure and no disclosure — a third state the evidence model isn't supposed to have.

## Observed (V38 window, three accepted rolls)

Every implementation run's `run_verification_summaries.summary` carries `criteria_total` (the contract's derived criteria) and `criteria_verified` (the credited subset). Across the window's accepted arm-B rolls the verified list came up short the same way every time:

| Roll | Cycle / impl run | Verified | Missing |
|---|---|---|---|
| Slot 2 | `cyc_410f3a9257f8` / `run_48e4c90d7efa` | 12 of 14 | `vc-compiles-app-api-runs-route`, `vc-compiles-app-api-runs-run-id-join-route` |
| Slot 4 | `cyc_60407deffa98` / `run_6387016a969c` | 10 of 15 | compile criteria for 3 API routes + `app/page.tsx` + `app/runs/new/page.tsx` |
| Slot 5 | `cyc_83577bc3052b` / `run_654b61665fed` | 12 of 14 | `vc-compiles-app-api-runs-route`, `vc-compiles-app-api-runs-run-id-join-route` |

Seven dropped criteria, **all `vc-compiles-*` for API routes/pages, never a probe, never the suite, never accompanied by a failure** — and never in the `unverified` disclosure either. The verdicts were `accepted`, the apps passed independent boot audits.

## The repair red herring, already run down

The window record's own attribution sharpened across the rolls and should not be re-litigated: slot 2 associated the drop with repair re-storage (both dropped criteria were the repaired files); slot 4 weakened it (all five dropped were repaired, but three other repaired files kept credit); **slot 5 refuted it — zero correction rounds, same two criteria dropped**. Repaired files are simply API-route files; the mechanism lives on the normal crediting path.

## Why it matters

SIP-0096's rule is *executed-and-passed credits, everything else disclosed*. Silently-absent is a third state: the coverage record under-reports, and there is no way to distinguish "check ran, credit lost in the join" from "check never ran." The 1.8 scorecard grades over these summaries; its gate ("grades never computed over fabricatable evidence") has this hole from the other side — earned-but-unrecorded evidence.

## Suggestive banked fact (unverified — for the code read, not assumed)

[...]

PR #1051, opened 2026-08-23
fix(verification): a contract criterion with no evidence stops being a silent fourth state

jasonLadd:
4 of 4 in the 1.6.2 batch. Independent of the other three.

**Read-first, per the issue's own disposition** — and the read changed what the fix is. `Refs`, not `Closes`: the accounting gap is closed, the underlying mechanism is not.

## What the banked reproducer actually shows

Slot 5 (`run_654b61665fed` — the zero-repair roll that refuted the repair hypothesis), replayed:

```
criteria_total 14 / criteria_verified 12
missing: vc-compiles-app-api-runs-route, vc-compiles-app-api-runs-run-id-join-route
unverified: []     failed: []
```

Both missing criteria produced **no row in either direction**. They are in `criteria_total` only because the contract declared them.

SIP-0096's rule is *executed-and-passed credits, everything else disclosed*. This is neither — a silent fourth state. `criteria_unverified` (#945) does name them, but names them **identically to a criterion whose check ran and failed**, which is the issue's exact complaint: no way to tell *"the check ran and credit was lost in the join"* from *"the check never ran."*

## The fix

`criteria_unevidenced` records criteria the contract declared that no row carried. `criteria_adverse` derives the complement, so the two **partition** the shortfall — every unverified criterion is one or the other, never both, never neither. That is tested as an invariant rather than as two lists that can drift apart.

Derived from the contract's declared set, **deliberately not** from `criteria_total`. That field unions observed ids in so unexpected evidence is never dropped; doing the same here would report every observed-but-undeclared criterion as absent — the inverse of the fact.

> Worth noting for the review: the obvious mutation of that expression, `(C | P | A) - P - A`, is algebraically identical to `C - P - A` and correctly survives. The test pins the wrong turn that is actually reachable — deriving from `criteria_total` and subtracting only the passed set.

Serialized both directions, and the run report now separates the two kinds of shortfall instead of printing one list the reader has to interpret.

## What this does not do

It does not explain **why** those two files' compile criteria produce no row while their siblings' do. That mechanism is still open. The banked hint from the issue stands: the runtime-api environment has no Node, so compile truth comes from the qa environment's build, and the credit mapping between those two surfaces is the first suspect.

This makes the question answerable from any stored run rather than by hand-diffing three lists — the same shape as #1002. The instrument comes first; the diagnosis it enables is the follow-up.

## Verification

- Regression **7,729 passed**, gate green at 20 workers.
- **7 mutations, all killed** — including adverse criteria mislabelled as unevidenced, which is the inverse error and the worse one: it would point a reader at the harness while the product is broken.

Refs #1021

### DIFF
--- a/adapters/cycles/postgres_cycle_registry.py
+++ b/adapters/cycles/postgres_cycle_registry.py
@@ -649,6 +649,10 @@ def _verification_summary_to_dict(summary: RunVerificationSummary) -> dict:
         # Postgres (rather than the in-memory summary) undercounted (#500).
         "criteria_verified": list(summary.criteria_verified),
         "criteria_total": list(summary.criteria_total),
+        # #1021: which shortfalls produced NO evidence at all, as opposed to evidence
+        # that went adverse. Derivable only here — the reader cannot recover it from
+        # the other lists, which is the whole complaint.
+        "criteria_unevidenced": list(summary.criteria_unevidenced),
         # #1002: the bounded inspected-set references. Stored so the question
         # "did this detector ever see the file?" is answerable from any run
         # after the fact — which was the whole point of computing it.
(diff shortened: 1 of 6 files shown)


jasonLadd (commented):
reviewed

PR #1052, opened 2026-08-23
feat(qa): additive-suite containment findings — banked, deliberately not enforced

jasonLadd:
The last of the 1.6.2 four. **Reporting-only — nothing here rejects.** `Refs`, not `Closes`.

## The gap

SIP-0104 contains the qa author inside deterministic scaffold slots, and `verification_scaffold_gate` validates that surface with named findings. **Additive suites — the files an author writes beyond the scaffold — have no equivalent.** They are where arm A's V7 rolls died: every counted red was additive-suite-side while all nine delivered applications of the arc passed independent boot audits. The apps were fine; the tests were not.

## Two checks, both on emitted bytes

| Check | Banked evidence |
|---|---|
| **Execution model** — an absolute-URL fetch cannot work in-process under vitest, where nothing listens | `cyc_6495d9870587` (corpus C3): three qa repairs, no convergence — **after** #877's prompt guidance shipped. Guidance does not contain this class. |
| **Subject import** — a test importing no application module asserts only against what it defined | The precondition for self-mocking (#915). V7 attempt-2 roll 1 banked exactly such a file and nothing flagged it in-cycle. |

## Why reporting-only, and why that is the disposition rather than a shortcut

#1022 defers the gate's shape to design review. These findings are banked so that review can argue from **what a gate would have flagged across real rolls**, instead of from what seems reasonable in advance.

Tonight is the live argument for that order. **#1049** — filed and fixed six hours ago — is a rejection gate whose premise stopped being true, and it cost two re-rolls per cycle and dead-ended a correct framing before anyone noticed. I am not adding a second rejection gate to the same path on the same night I removed one.

Promotion to blocking is a separate deliberate call, with the evidence in hand.

## The over-rejection tests are the load-bearing ones

My first draft's regex was case-insensitive and included `request`, so it flagged:

```ts
new Request('http://test/api/runs')   // the CORRECT in-process form the shells emit
```

The control test caught it before this went anywhere. A check that fires on the right answer is worse than no check — so prose URLs, fixture URLs, relative helper imports, and the scaffold shells themselves are all pinned as non-findings.

## Verification

- Regression **7,738 passed**, gate green at 20 workers.
- **9 mutations, all killed.** One survived the first pass: findings computed and never banked — the **third** instance tonight of one shape, a fact derived a line from where it is recorded and never placed there (#1040's five dev surfaces, #1048's loop state, this one).

## Related

SIP-0104 (the fill-mode containment this extends), #877/#879 (guidance half, merged, demonstrated insufficient), #1002 (detector provenance — the complementary fix), #915, #1049 (why this is not a gate yet).

Refs #1022

### DIFF
--- a/src/squadops/capabilities/additive_containment.py
+++ b/src/squadops/capabilities/additive_containment.py
@@ -1,0 +1,96 @@
+"""Containment findings for additive test suites — #1022, reporting-only.
+
+SIP-0104 contains the qa author inside deterministic scaffold slots, and
+``verification_scaffold_gate`` validates that surface with named findings. **Additive
+suites — the extra files a qa author writes beyond the scaffold — have no equivalent.**
+They are unconstrained authorship landing in the same tree, and they are where arm A's
+V7 rolls died: every counted red was additive-suite-side while all nine delivered
+applications of the arc passed independent boot audits. The apps were fine; the tests
+were not.
+
+Two checks, both reading the EMITTED BYTES rather than any declaration — the same
+posture as the scaffold gate, and for the same reason: a defect is precisely a gap
+between what the author meant and what it wrote.
+
+- **Execution model.** The suite runs in-process under vitest; there is no server
+  listening. A suite that fetches an absolute URL can only hang or throw.
+  `cyc_6495d9870587` (corpus C3) did exactly this and burned three qa repairs without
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

Issue #1054, opened 2026-08-23
The lead named dev task types and all three repairs dispatched to qa — arm A never touched the app


jasonLadd:
**Read before fixing.** The locus classifier is conservative *by design* — only explicit machine signals produce `OWN_ARTIFACT`, specifically so a qa repair can never "fix" an app bug by rewriting the tests. This issue may be that guard misfiring rather than a missing one, and the wrong fix here re-opens the test-gaming hole.

## What happened

Arm A of the 2026-08-23 paired validation (`cyc_181c9572bef2` / `run_da240b949e12`) exhausted its budget without the application ever being repaired.

Round 0's correction decision was unambiguous:

> "The test failures stem from route handlers instantiating a **local shadow store** instead of importing the scaffold-provided global in-memory store, which breaks beforeEach isolation hooks."
>
> `affected_task_types: ["backend_route_implementation", "store_module_integration"]`

Every repair dispatched was `qa.test_repair`:

```
07:13:07  qa.test_repair  __tests__/api_runs.test.ts
07:38:37  qa.test_repair  __tests__/api_runs.test.ts
08:05:40  qa.test_repair  __tests__/api_runs.test.ts
```

The route handlers were never emitted. The suite was rewritten three times. The shadow store — a real, correctly diagnosed application defect — survived the entire correction arc.

Contrast arm B on the identical deploy, where round 0 dispatched `development.correction_repair` and emitted seven route/page files.

## Why this is the class the machinery already knows about

`_resolve_repair_target`'s own docstring states it:

> "#531: the target defaults to the *failed task's own* artifacts — but the `tests_pass` check lives on the qa test task, so anchoring there regenerates the tests (the symptom) while the drifted source (the cause) is never rewritten and the loop can't converge."

The failing check was `tests_pass`, which is emitted by qa.test. So the symptom's owner became the repair's target, which is precisely what #531/#532/RC2/#688 were built to prevent.

## Compounding detail

Rounds 1 and 2's diagnoses drifted to the tests — "test state leakage and incorrect assertions rather than core application logic", "unpatched placeholder assertions". That drift is *downstream* of the misroute: the qa repair had destroyed the fills (round 1 shows 0 passing shells where round 0 had 5), so by round 1 the tests genuinely were broken and the lead was reading a true symptom of the previous repair. The original defect was out of frame by then.

## The read this needs

[...]

PR #1056, opened 2026-08-23
fix(correction): an emission containing nothing is not an attempt at the fix

jasonLadd:
## What arm B actually did

It exhausted its correction budget **while holding a correct and stable diagnosis across all three rounds**. It did not fail to fix the bug — it never got another attempt at it.

```
round 0  →  7 route/page files          (a real repair)
round 1  →  repair_output.md   0 bytes
round 2  →  repair_output.md   0 bytes
```

Both empty files were banked under the handler's generic fallback name, and both were billed as spent attempts. So `Max correction attempts (3) exhausted` described a loop that had actually tried **once**.

The diagnosis never drifted, either — #1029's frozen-spine floor pointed straight at the defect and every round named the same two handlers:

> **round 0** — "the join endpoint handler returns a response missing at least one required Run field … and fails to correctly register the participant"
> **round 2** — "three specific, well-scoped implementation bugs in the join and leave endpoint handlers"

This is the best case for the correction loop — a stable, specific, machine-pointed diagnosis — and two of three chances to act on it went to files containing nothing.

## The fix

A zero-byte artifact is an emission failure, not a repair attempt. The runner judges the emission on **content** rather than artifact count — a zero-byte file is still a file, which is exactly how these rounds got counted — and the executor hands the pre-incremented attempt back so the round is re-taken.

Three boundaries, each mutation-covered:

| Boundary | Why |
|---|---|
| `repair_steps_ran` gates it | a rewind/continue path legitimately emits nothing and must never be refunded |
| whitespace-only is empty | judging on length lets the same wasted round through under a marginally different emitter |
| the allowance is **bounded**, in its own counter key | taking it from the pool the empty emission is failing to consume is unbounded by construction; a producer that never emits must still terminate |

Disclosed on the `CORRECTION_COMPLETED` event rather than only in a log line — *"converged in 3"* and *"converged in 3 after two empty emissions"* must not read the same. That is the evidence-integrity rule this line keeps re-learning (#1002, #1021).

`refund_empty_emission_attempt` is module-level so the executor and its test drive the **same** rule. My first version of that test re-implemented the arithmetic, which would have kept passing while the real rule drifted.

## Verification

- Regression **7,728 passed**, gate green at 20 workers.
- **7 mutations, all killed.** Two things worth recording from that run:
  - the unbounded-refund mutation **hangs** rather than failing — correct behaviour, and it left a mutation in the working tree when the harness timed out. The harness now restores on timeout and counts a hang as a kill.
  - two anchors missed after `ruff format` reflowed the expression; re-run against the formatted text rather than reported as skips.

## Related

#566 (zero-extraction marker vocabulary), #431/#528 (extraction loss — the adjacent, distinct class: prose the parser rejected keeps its current handling), #1015/#1048 (repair envelope quality), #414 (budget allocation — note this argues the budget was never the constraint), #1054 (why arm A never reached this path at all).

Closes #1053

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -593,6 +593,15 @@ class CorrectionProtocolResult:
 
     correction_path: str
     repair_artifacts: list[dict[str, Any]] = field(default_factory=list)
+    #: #1053: repair steps ran and produced no content at all. Distinct from "produced
+    #: a bad repair" and from "no repair step ran": arm B of the 2026-08-23 pair banked
+    #: `repair_output.md` at ZERO bytes on two of its three rounds, under the handler's
+    #: generic fallback name, while holding a correct and stable diagnosis. Each was
+    #: counted as a spent attempt, so `Max correction attempts (3) exhausted` described
+    #: a loop that had actually tried once. An emission containing nothing is an
+    #: emission failure (`FailureEvidenceCategory.EMISSION_ABSENT`'s shape), not an
+    #: attempt at the fix.
+    emission_empty: bool = False
 
 
 class CorrectionRunner:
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

Issue #999, opened 2026-08-19
fill_merge assertion-strength evidence (#982) is computed then never persisted — execution_evidence has no home


jasonLadd:
V7 window detection (launch 2, `run_591987c9df8d` — the first roll with #982 live). Recorded mid-window, left unfixed per §6.

#982 computes `measure_assertion_strength` and places it in `evidence_extra["fill_merge"]["assertion_strength"]`, which flows into `HandlerEvidence.create(metadata=evidence_extra)` → `TaskResult.execution_evidence` — **and stops. Nothing persists `execution_evidence`.** It is not in `run_verification_summaries`, not in checkpoints (`prior_outputs` keeps the compact role summary only), not in any artifact, not in `pulse_verification_records`.

So the instrumentation added *because* "the 7x drop was visible in data already logged and read by nobody" measures correctly at runtime and then evaporates — the same failure mode one layer up. The window's §4 field ("read from banked evidence rather than a log line") is unsatisfiable as deployed; for launch 2 the values were **recomputed from the stored merged shells** (2 of 5 fills assert on the store via `all(TABLES.Run)`, 16 expect() lines in fill regions, 2 additive suites) and the method is noted in the window record.

Fix direction (post-window): persist the qa task's `fill_merge` evidence somewhere queryable per run — candidates: a keyed section in `run_verification_summaries.summary`, or a small evidence artifact stored beside `test_report.md`. Whichever seam owns it, the closing-claim reader must be able to reach it without container logs.

PR #1057, opened 2026-08-23
fix(qa): the containment findings actually reach the record — #1052 dropped them

jasonLadd:
**A defect in work merged a few hours ago, found while auditing PR-to-issue crediting.**

#1052 computed the additive-suite containment findings and placed them in `evidence_extra["fill_merge"]` → `TaskResult.execution_evidence`. **#999 establishes that nothing persists that** — not the verification summary, not checkpoints, not an artifact.

So the findings that PR described as *"banked on every qa emission"* were computed and dropped. The PR body said it. My comment on #1022 said it. Neither was true.

## This is the fifth instance of one pattern tonight

And the fourth was named by the change that then committed the fifth:

| | |
|---|---|
| #1040 | five dev surfaces rendered into sections, never added to the template variables |
| #1048 | loop-state block rendered by the handler, never injected into inputs |
| #1052 | findings computed at the merge seam, never banked — **the PR called this pattern out** |
| here | findings landed only in evidence nothing persists |
| and | the handler still never passed them to the summary builder — caught by mutation |

Every layer had a passing test. The fact reached nobody at every step.

## The fix

The findings land on `ScaffoldEvidenceSummary`, which serializes into `outputs["scaffold_evidence"]`. That dict is read by `build_failure_evidence`, so the findings reach the analyzer and the locus classifier.

**Scoped claim this time:** full persistence for a *passing* run is still #999's territory and this does not claim it. Being precise about which surface this lands on is the whole lesson of the overclaim it fixes.

## Verification

- Regression **7,723 passed**, gate green at 20 workers.
- **4 mutations, all killed.** The fourth survived the first pass — the handler not passing the findings to the builder, with every other layer green. Only an assertion on `outputs` can see that, so that is where the test asserts.

Refs #1022, #999

### DIFF
--- a/src/squadops/capabilities/handlers/cycle/qa_test.py
+++ b/src/squadops/capabilities/handlers/cycle/qa_test.py
@@ -694,6 +694,12 @@ class QATestHandler(_CycleTaskHandler):
             correlations,
             additive_count,
             tuple(getattr(test_result, "uncollected_test_files", ()) or ()),
+            # The containment findings' landing. Computed at the merge seam and, until
+            # now, placed only in `execution_evidence` — which nothing persists (#999),
+            # so "banked on every qa emission" was not true of the change that claimed
+            # it. `scaffold_evidence` is in `outputs`, which reaches `failure_evidence`
+            # and therefore the analyzer and the locus classifier.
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #1058, opened 2026-08-23
docs(site): factual tone, the framing task sequence, and brand assets

jasonLadd:
Three commits.

## 1. `rewrite for factual tone` — a re-land

**Pushed to #1046 but never landed**: that PR merged carrying only its first commit, so this work never reached main. Cherry-picked unchanged, message intact.

Removes three habits across every page:

- **Defining by negation** — "X, not Y", "this is not a feature you receive". These clustered around passages rewritten after a correction, preserving the wrong version as a warning in the product's voice. Now positive statements or deleted. Surviving negations are factual: `unverified` means checks that did not execute, which is the definition of the field.
- **Sermonizing** — "the load-bearing decision in the system", "cannot tell you anything true about its own output", "both are lies". Replaced with mechanism and values: the three verdicts as a table, the not-executed reason taxonomy, the ten-cycle inert lookback.
- **Comparison to unnamed others** — "most frameworks orchestrate implementation against a design a human supplied". Unfalsifiable; removed.

Verified by grep, not by eye — zero occurrences of the sermon phrases, every surviving ` not ` reviewed individually.

## 2. `the framing sequence, by role`

Overview described framing as *"the squad authors an interface manifest"* and skipped the eight other tasks that produce it. Overview and Key concepts now carry the ordered sequence read from `task_plan.py`:

| # | Task | Role |
|---|---|---|
| 1 | `data.research_context` | data |
| 2 | `strategy.frame_objective` | strategy |
| 3 | `development.design_plan` | dev |
| 4 | `development.author_manifest` | dev |
| 5 | `qa.define_test_strategy` | qa |
| 6 | `governance.prepare_plan_authoring_brief` | lead |
| 7 | `development.propose_plan_tasks` · `qa.propose_plan_tasks` · `strategy.propose_plan_guidance` | dev · qa · strategy |
| 8 | `governance.merge_plan` | lead |
| 9 | `governance.review_plan` | lead |

Two facts the summary hid. **Step 4's position is load-bearing** — `build_planning_steps` inserts it after the technical design and before the test strategy, so dev authors while best-informed and QA writes its strategy against the interface implementation will be held to (#791). **Step 7 is configurable** — `plan_authoring_contributors` selects which roles propose, and an empty list puts the merger in sole-author mode, the default for every request profile except `validation-multirole`.

Implementation and wrap-up sequences added for the same reason.

## 3. `logo, favicon, and header lockup`

A mark for the project: a cycle split into two arrows with a play symbol at the centre. The gap at 3 and 9 o'clock is the gate; play reads as run.

- **Installed as a Material custom icon**, not `theme.logo`. `theme.logo` renders an ``, where `currentColor` has nothing to inherit from; the icon path inlines the SVG so the mark follows the palette.
- **Favicon is a solid black rounded tile.** Transparent marks depend on `prefers-color-scheme`, which tab chrome applies inconsistently and can land dark-on-dark. Mark at 0.82, measured 12% edge padding — inside the 10% floor maskable icons need, while staying as large as possible for 16px.
- **Header lockup**: Material's default left 32px between logo and title, wider than the 24px mark itself. Now 15px. Two causes, not one — the title margin tied on specificity with Material's rule, *and* the logo button's own padding contributed the rest regardless.
- **Config bug fixed**: the theme block had grown two `icon:` keys. YAML keeps the last, so the logo declaration was being silently discarded by the `repo:` block.

Rationale for the tile and the scale is recorded in the SVG comments.

---

`mkdocs build --strict` clean, `ruff` clean repo-wide, every change reviewed in a browser.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -17,6 +17,8 @@ site_dir: site/_build
 
 theme:
   name: material
+  custom_dir: site/overrides
+  favicon: assets/favicon.svg
   font: false          # no Google Fonts fetch — system stack instead
   # NOTE ON THIRD-PARTY REQUESTS. Setting font:false removes the Google Fonts
   # fetch, but Material lazy-loads mermaid from unpkg.com at runtime on pages
@@ -27,6 +29,8 @@ theme:
   # worth it for a docs site (#1039). Recorded so the next person does not
   # repeat the measurement, and so the claim is not overstated elsewhere.
   icon:
+    # Inlined as SVG (not an <img>), so currentColor follows the active palette.
+    logo: squadops/logo
     repo: fontawesome/brands/github
   features:
     # navigation.instant is DELIBERATELY off: its client-side page swaps wipe
--- a/site/content/architecture.md
+++ b/site/content/architecture.md
@@ -34,7 +34,7 @@ observability, a content-hashed filesystem vault for artifacts.
 
 Two properties fall out of this that matter more than the pattern itself:
 
-**Capabilities are declared, not inferred.** An adapter states what it can do —
+**Capabilities are declared.** An adapter states what it can do —
 model listing, model management, streaming usage — and callers ask the port
 rather than checking which class they were handed. A declared capability that
 does not work fails a conformance suite every adapter must pass.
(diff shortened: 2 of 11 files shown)


jasonLadd (commented):
reviewed

PR #1059, opened 2026-08-23
feat(scaffold): insert-as-update findings on emitted route handlers — banked, not enforced

jasonLadd:
The **finding half** of #1055. The store's missing update seam is the root fix and is a separate, owner-ruled change (moves `GENERATOR_VERSION` and the Gate 1 fixture). `Refs`, not `Closes`.

## The issue was filed on a wrong diagnosis, and this module found it

Arm A's lead reported *"route handlers instantiating a local shadow store instead of importing the scaffold-provided global"*. I filed #1055 from that prose **without checking it**. The first version of this detector looked for exactly that and came back clean — because every route file in that roll imports the frozen store and holds no module state.

That is the #968 class — an analyzer claim nobody verifies against the source — and I repeated it in the same session where I cited #968 as grounds to distrust analyzer prose. The issue is corrected and retitled.

## The real defect

```ts
const run = find(TABLES.Run, params.run_id)
participants.push(name)
insert(TABLES.Run, {...run, participants})   // appends a SECOND row
```

`insert` is `push`. Join adds a row, leave adds another — the observed `length of 1 but got 2`, then `got 3`. A repair telling the developer to import the frozen store could never have fixed it, because the handler already does.

## Prevalence, measured with the detector

My first hand-grep got this wrong by anchoring `insert(` to line start, which misses `const updated = insert(...)`. Re-measured properly across three banked rolls:

| Roll | join | leave | create | read |
|---|---|---|---|---|
| arm A · 3.6 | **FLAG** | **FLAG** | ok | ok |
| arm B · 3.8 | **FLAG** | **FLAG** | ok | ok |
| green roll | ok | ok | ok | ok |

Both failing rolls. Neither create nor read handler. Never the roll that went green. Two independent authorings on two different models reached for the same wrong verb at exactly the two operations that *update* rather than *create* — which is why the root cause is the API, not the author.

## Matched as a pair, deliberately

`find` then `insert` on the **same table** — never on `insert` alone. A create legitimately inserts; reading a Run to validate before creating a Participant is ordinary. Both are pinned as non-findings, as is **the correct update form** (mutating what `find` returned) — the finding must never fire on the very thing its message tells the author to do.

## Reporting-only, and banked where it is read

`outputs["source_containment"]`, carried explicitly by `build_failure_evidence` so the analyzer sees it. Arm A's lead had to invent a mechanism because nothing told it what the handlers actually did wrong.

Landing **verified rather than assumed** — #1052 put its findings in `execution_evidence`, which nothing persists (#999), and called them banked.

## Verification

- Regression **7,811 passed**, gate green at 20 workers.
- **7 mutations, all killed.** Two survived the first pass: non-route files being judged (the predicate was tested, its *use* was not), and the handler never assigning the result into `outputs` — the **sixth** instance of that shape in one session, so it now has a structural test of its own.

Refs #1055, #968

### DIFF
--- a/src/squadops/capabilities/handlers/cycle/develop.py
+++ b/src/squadops/capabilities/handlers/cycle/develop.py
@@ -698,6 +698,18 @@ class DevelopmentDevelopHandler(_CycleTaskHandler):
 
         outputs["emission_stats"] = emission_stats(len(content), artifacts)
 
+        # #1055: insert-as-update findings over the emitted route handlers. Banked in
+        # `outputs`, NOT as a validation check row — a failing row there would reject
+        # the task, and this is reporting-only until a gate's shape is argued from what
+        # it flags on real rolls (#1049's lesson, #1052's posture).
+        #
+        # Landing verified rather than assumed: #1052 put its findings in
+        # `execution_evidence`, which nothing persists (#999), and called them "banked".
+        # `outputs` reaches `build_failure_evidence`, which carries this key explicitly.
+        from squadops.capabilities.source_containment import assess_source_containment
+
+        outputs["source_containment"] = assess_source_containment(artifacts)
+
         duration_ms = (time.perf_counter() - start_time) * 1000
         evidence = HandlerEvidence.create(
             handler_name=self._handler_name,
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #1062, opened 2026-08-23
feat(release): publish the GitHub Release from the tag, not from memory

jasonLadd:
Closes #1061

## Why automate rather than document

Across v1.4.0 → v1.6.1, **zero releases were published at cut time.** Every one was backfilled later, in two batch sweeps:

| Tag | Tagged | Release published |
|---|---|---|
| v1.4.0 – v1.5.0 | Jul 31 – Aug 7 | **2026-08-08** (the #789 sweep) |
| v1.6.0, v1.6.1 | Aug 21, Aug 22 | **2026-08-23** |

The step was written into the cut checklist on **2026-08-10** and still missed at both subsequent cuts. A documented manual step with a 0-for-8 record needs removing, not restating.

**Why it gets missed is structural.** Pushing a tag reads as completion — the version is stamped, the tag is on the remote, git reports success. Nothing in that finished state signals that a separate artifact on a different system is outstanding.

## What this does

`.github/workflows/release.yml` fires on a `v*` tag push and publishes the Release from the version's CHANGELOG section.

The body is not unreviewed content: it is written deliberately, reviewed in the release PR, and guaranteed to exist by rule 4 of `test_docs_version_sync.py` before a tag can be cut.

Four decisions worth surfacing for review:

- **Checks out the tag, not main.** The CHANGELOG must be read as it was *at the cut*; reading main would let a later rotation rewrite history.
- **`--latest` only for plain semver.** A pre-release or oddly named tag can never displace the current release.
- **An existing Release is left untouched**, so a re-run or a manual publish is never clobbered.
- **A missing or empty section fails the job.** The only outcome worse than no release is a public empty one.

`workflow_dispatch` with a tag input is included for backfills.

## Parsing is testable

Logic lives in `scripts/maintainer/release_notes.py` rather than inline YAML, so a heading-format change breaks a test instead of a release. **11 tests** cover:

- both dash styles present in the file's history (`—` and `-`)
- section-boundary bleed — the bug that would ship one release carrying the previous release's notes
- `[Unreleased]` never matching as a version
- empty sections raising rather than publishing
- the transient "Current" marker stripped from ROADMAP-derived titles
- the **real** CHANGELOG and ROADMAP, not only a fixture

## Validated against the live artifacts

Regenerating v1.6.0, v1.6.1 and v1.5.0 reproduces the published bodies byte-for-byte apart from a trailing newline GitHub appends, and reproduces both titles I created by hand. (v1.5.0's title differs by a `(stabilization)` parenthetical the ROADMAP carries and the hand-written title dropped.)

## Follow-on

CLAUDE.md's cut checklist step 6 collapses to "tag and push". Steps 4, 5 and 7 remain unguarded and stay written down.

`docs/ROADMAP.md`'s hand-maintained "Current" marker is the same class of unguarded restatement and is left for separate work.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,70 @@
+# Publishes a GitHub Release when a v* tag is pushed (#1061).
+#
+# Why this is automated rather than a checklist item: across v1.4.0–v1.6.1,
+# ZERO releases were published at cut time — every one was backfilled later, in
+# two batch sweeps. The step was written into the cut checklist on 2026-08-10
+# and still missed at both subsequent cuts. Pushing a tag reads as completion:
+# the version is stamped, the tag is on the remote, git reports success, and
+# nothing signals that a separate artifact on a different system is outstanding.
+#
+# The body is the version's CHANGELOG section, which is written deliberately,
+# reviewed in the release PR, and guaranteed to exist by rule 4 of
+# tests/unit/architecture/test_docs_version_sync.py before a tag can be cut.
+# So this publishes reviewed content at a later moment, not unread content.
+name: Release
+
+on:
+  push:
+    tags: ['v*']
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Tag to publish (e.g. v1.6.2)'
+        required: true
+
+permissions:
+  contents: write
+
+jobs:
+  publish:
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #1063, opened 2026-08-23
fix(correction): the repair sees every fact the initial author is judged against

jasonLadd:
**The declaration was three surfaces short, not one.**

## What the audit found

`_render_fill_only_section` in the repair mixin already builds an error-contract block and a model-surface block, under a comment claiming *"parity with develop's four surfaces."* Neither key was ever added to `REPAIR_CONTEXT_CONTRACT`, so neither was threaded onto a repair envelope — **both have rendered empty since the day they were written.** The renderer was added and the data never was.

Measured, not inferred:

```
develop threads : error_contract  frozen_surface  model_surface  response_surface  testid_surface
repair  threads : dom_testid_surface  frozen_surface  testid_surface
MISSING         : error_contract  model_surface  response_surface
```

## The third one cost a roll

`cyc_87c12c7f199e`. The manifest declares `participants: list[Participant]`. #1029's frozen-spine floor reported the shape defect on **all four rounds**:

```
× vc-probe-api-runs-join → expected undefined not to be undefined
  ❯ expectShape …/vc-probe-api-runs-join.scaffold.test.ts
```

And every repair re-emitted:

```ts
const participants: string[] = Array.isArray(run.participants) ? run.participants : [];
participants.push(name);
```

The agent given a second attempt at a shape defect was the only agent in the chain never shown the shape. That is #861's sentence — *"a name the author cannot see is a name it will invent"* — one contract entry over, and I introduced this instance myself: #1042 updated the developer's contract and not this one.

## The fix is the test, not the three entries

Every surface the repair carries was added retroactively **after a roll died on it**: #667 the anchors, roll 7 the frozen index, #902 the stack appendix. This one makes that pattern a test failure instead:

- `test_repair_threads_every_surface_the_developer_gets` derives the expectation from the **developer's** set rather than restating a list, so a surface added there and forgotten here fails immediately.
- A sibling asserts every threaded surface actually reaches the prompt's **variables** — because a surface declared and never rendered is the same defect facing the other way, and that step has now been missed six times in this codebase.

## One test had to be narrowed

`test_repair_contract_threads_the_same_anchor_inventory_under_both_keys` pinned the repair's exact surface set as a side effect of testing something else — and **that assertion is what silently blocked adding the three.** Narrowed to its actual subject (both testid keys carry one derivation); which surfaces the repair receives is now owned in one place.

## Verification

- Regression **7,798 passed**, gate green at 20 workers.
- **5 mutations, all killed** — including each of the three surfaces removed individually, and the renderer built but never placed in variables.
- Golden regenerated in the same commit per its policy: 36 lines, the three new keys on the two repair scenarios, nothing else moved.

Closes #1060

### DIFF
--- a/src/squadops/capabilities/context_assembly.py
+++ b/src/squadops/capabilities/context_assembly.py
@@ -345,8 +345,29 @@ DECLARED_NO_CONTEXT: frozenset[str] = frozenset(
 #: blind identically" and the chain terminated as plan_defect on a defect no repair
 #: could see. This threads the key; the repair mixin renders it (dev: inside the
 #: fill-only appendix; qa: its own appendix — ``_RepairPromptMixin``).
+#: #1060: the declaration was THREE surfaces short of what the repair mixin already
+#: renders. `_render_fill_only_section` builds an error-contract block and a
+#: model-surface block from `inputs["error_contract"]` / `inputs["model_surface"]`, and
+#: its comment claims "parity with develop's four surfaces" — but neither key was ever
+#: threaded here, so both blocks have rendered EMPTY since they were written. The
+#: renderer was added and the data never was.
+#:
+#: `response_surface` is the third, and the one that cost cyc_87c12c7f199e: the manifest
+#: declares `participants: list[Participant]`, #1029's frozen-spine floor reported the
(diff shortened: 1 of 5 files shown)


jasonLadd (commented):
reviewed

PR #1064, opened 2026-08-23
feat(scaffold): the frozen store gains an update seam — the write it never had

jasonLadd:
The **root fix** behind #1055, and the option you picked.

## The gap

The store's whole write surface was `insert`, which is `push`. *"Persist a change to an existing row"* had no correct form:

- the working answer — mutate the object `find` returned and call no write verb — is non-obvious;
- the natural-looking answer stored a duplicate.

Two independent authorings on two different models both reached for `insert`, and both **only in the handlers that update rather than create**:

| Roll | join | leave | create | read |
|---|---|---|---|---|
| arm A · 3.6 | insert-as-update | insert-as-update | ok | ok |
| arm B · 3.8 | insert-as-update | insert-as-update | ok | ok |
| green roll | ok | ok | ok | ok |

That distribution is what makes it an API gap rather than an authoring defect. Stack #1 never had the problem — it exposes `dict[str, Model]`, so `store[id] = updated` is a natural upsert.

## The change

`update(table, row)` upserts by id: replaces the stored row when one matches, appends when none does.

**`insert` stays create-only** rather than becoming an upsert. Folding them together would make the wrong call retroactively correct and a genuine double-create undetectable.

## No `GENERATOR_VERSION` bump — and I nearly got this wrong

My first pass bumped 7 → 8 out of ritual. Then I looked at what actually moved:

```
 generator_version: 7          (unchanged)
-expanded_tree_hash: 80e56320…
+expanded_tree_hash: 7ee51e60…
 aggregate_spine_hash: 1ff59445…   (unchanged)
 scaffold_hash: 6fc46f50…          (unchanged)
```

**The shells are byte-identical.** Nothing in the emitted spine references the store's write verbs, so the two anti-tautology hashes do not move. What changed is `expanded_tree_hash` — this is an **expander** change, not an emission change. `GENERATOR_VERSION` names the emission behaviour, and bumping it here would spend the signal that distinguishes a real emission change later.

Blast radius measured before writing anything: **one fixture line**, 5,198 other tests unaffected.

## Both briefs learn the verb for free

`frozen_surface_index_lines` derives from the expanded tree, so the index now advertises `update(table: Table, row: Record<string, unknown>)` with its signature — to the developer **and** the repair. #1063 landed the repair's half of that channel; without it the seam would exist and the agent most likely to need it would not know.

## The finding is re-aimed, not retired

#1059 now says *"use `update(TABLES.Run, row)`, which upserts by id"* instead of describing the pre-seam workaround. The wrong form is still writable, and it is now wrong against an API that offers the right one — which is exactly when a finding is worth reading, because it can name the fix.

## Verification

- Regression **7,816 passed**, gate green at 20 workers.
- **4 mutations, all killed** — including update replacing but never appending, matching on the wrong key, and the finding still naming the obsolete workaround.

Closes #1055


jasonLadd (commented):
reviewed

PR #1066, opened 2026-08-23
docs(site): record the accepted Safari favicon tradeoff

jasonLadd:
No behaviour change — a comment in `favicon.svg` recording a decision.

Safari draws a white frame around the favicon; Chrome renders it correctly. Safari's favicon pipeline flattens alpha to white, and the only transparency in the asset is the four rounded corners: transparent over the tab in Chrome, white in Safari, reading as an outline.

**Rounded corners cannot survive that** — the rounding *is* the transparency. The only fix is a full-bleed square with no alpha, proposed in #1065 and declined: the rounding was a deliberate choice, Chrome is the majority browser, and the corner radius is close to imperceptible at 16px either way.

Recorded in the asset itself so the Safari rendering is not mistaken for a bug and squared off later by someone who has not seen this thread.

The comment also notes the diagnostic error worth not repeating: the file was magnified and declared clean, but that magnification ran in Chrome — the browser that was already working.

### DIFF
--- a/site/content/assets/favicon.svg
+++ b/site/content/assets/favicon.svg
@@ -5,6 +5,15 @@
        inside the 10% floor a maskable icon needs, while staying as large as
        possible for 16px rendering. A maskable app icon, if ever needed, is a
        separate asset at heavier padding rather than a shrink of this one. -->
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #1068, opened 2026-08-23
fix(gates): a status that overrides the derived default must say why

jasonLadd:
The root-cause fix for the 200-vs-201 recurrence. Removes the class rather than gating it again.

## Why four fixes weren't enough

Five incidents in three weeks, four fixes: **#1013** (a gate comparing manifest to plan), **#1042** (thread the status to the developer), **#1049** (repairing that gate's stale premise), **#1031** queued to *teach* the convention. Every one locally correct. None asked why one integer needed four.

The status is **authored in three places** — manifest, plan prose, handler code — and **derivable in one**. `scaffold_contract` already computes it: collection POST → 201, child-action POST → 200. The schema then lets an author override that per endpoint with no warrant required, so every roll rolls dice on three independent authors agreeing about one integer.

## Today's roll is the clean demonstration

`POST /api/runs/{run_id}/join` is a child action, so the rule says **200**.

- the deriver computed 200
- the plan author wrote 200
- one unexplained manifest override said **201**

Declared wins — so the framing gate rejected the plan **for agreeing with the rule**.

## The change

An override is not forbidden. A child action that genuinely creates a sub-resource may well return 201, and forbidding it would trade a dice-roll for a wrong answer. It must be a **recorded judgment**: a `decisions[]` entry naming the endpoint and stating the status, which makes it challengeable downstream instead of arriving as an unexplained integer.

**Silence becomes the safe default and the common case** — say nothing, the rule decides, and there is nothing to disagree with.

## The warrant must be about the status

My first cut matched any decision naming the path, and it accepted this real entry from today's manifest:

> `join-leave-separate-endpoints` — *"distinct POST paths /api/runs/{run_id}/join and /api/runs/{run_id}/leave rather than a shared endpoint with an action dispatch field"*

A genuine judgment about **routing** that says nothing about 201. A path-only rule let the override through unexplained, which makes the whole requirement decorative. The decision must name the path **and** state the status.

## Single-sourced, deliberately

`derived_success_status` lives in one place. The rule already appears at three sites in `scaffold_contract` and once in `framing_consistency` — adding a fourth copy to police the first three would be this issue's own defect one layer up. `_path_pattern` is promoted to `path_pattern` for the same reason: reused, not duplicated.

## Validation

Run against **every banked manifest**: flags `cyc_79eebcb82205`, clean on the green roll, both failing arms, the 1.6.1 shakedown, and the reference — the artifact the 1.4 evidence was measured against.

- Regression **7,825 passed**, gate green at 20 workers.
- **7 mutations, all killed.** Two survived the first pass: the path-only warrant above, and non-POST endpoints being judged.
- The registry-completeness guard caught the missing proof classification, which is exactly its job. The rule is taught in the authoring asset as `warrant-a-status-that-breaks-convention`.

## Scope note

This narrows **#1031**: the primer is still right for conventions with no derivable answer (404-vs-400 on a failed lookup), and wrong for this one. **#772** wants the same single-source treatment for the *undeclared* case.

Closes #1067

### DIFF
--- a/src/squadops/cycles/framing_consistency.py
+++ b/src/squadops/cycles/framing_consistency.py
@@ -74,7 +74,7 @@ def _enforced_success_status(ep: Endpoint) -> int | None:
     return 201 if "{" not in ep.path else 200
 
 
-def _path_pattern(path: str) -> re.Pattern[str]:
+def path_pattern(path: str) -> re.Pattern[str]:
     """A conservative regex recognizing *path* in plan prose.
 
     ``{run_id}`` segments match any single non-space segment, so
@@ -150,7 +150,7 @@ def validate_manifest_plan_consistency(
         enforced = _enforced_success_status(ep)
         if enforced is None:
             continue
-        pattern = _path_pattern(ep.path)
+        pattern = path_pattern(ep.path)
         # Method-aware binding: a line naming HTTP methods binds only endpoints
         # whose method it names ("GET /api/runs returns 200" is about the GET,
         # not the collection POST sharing the path). Methodless pathful lines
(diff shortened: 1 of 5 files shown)


jasonLadd (commented):
reviewed

PR #1069, opened 2026-08-23
test(site): indigo favicon tile to clear Safari's contrast plate

jasonLadd:
Safari draws a light backing plate around our favicon; Chrome is fine.

**The earlier explanation was wrong.** I recorded that Safari flattens alpha to white and that the rounded corners were therefore the cause. A sampled Safari tab bar disproves it: transparent colourful favicons render clean with no plate — an orange asterisk, a blue star, GitHub's transparent white octocat — while every **dark** icon in the same bar carries a plate.

| Icon | | Plate |
|---|---|---|
| orange asterisk, blue star, green circle | colourful | no |
| GitHub octocat | white on transparent | no |
| dark glyph tiles, black circle, **our black tile** | dark | **yes** |

The trigger is darkness, not transparency. `#000000` on a near-black strip is the worst case for whatever contrast threshold Safari applies.

This colours the tile the site's indigo — the cheapest test that keeps the rounded corners you chose. If Safari still plates it, the hypothesis is wrong again and the next move is a transparent white mark, matching the octocat rendering clean two tabs away.

Merging deploys via Pages; Safari is the test. Note Safari caches favicons hard — a new tab or a cleared cache may be needed.

### DIFF
--- a/site/content/assets/favicon.svg
+++ b/site/content/assets/favicon.svg
@@ -1,20 +1,15 @@
 <svg viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
-  <!-- Squad Ops favicon. A solid tile rather than a transparent mark: tab chrome
-       applies prefers-color-scheme inconsistently, so a transparent icon can land
-       dark-on-dark. Mark scaled to 0.82, leaving 12% edge padding — comfortably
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #1072, opened 2026-08-23
fix(site): transparent favicon — Safari plates opaque icons

jasonLadd:
Safari draws a white plate behind any **fully opaque** favicon and leaves transparent ones alone.

That explains every observation:

| Favicon | | Plate |
|---|---|---|
| orange asterisk, blue star, GitHub octocat | transparent | no |
| white tiles with dark glyphs | opaque | yes — white on white, invisible |
| black circle, our black tile, our indigo tile | opaque | **yes, visible** |

**Confirmed by test, not argument.** Serving three variants on fresh origins (so Safari's favicon cache could not interfere): black tile plated, indigo tile plated, transparent did not. No tile colour avoids it — the tile itself is the trigger, so it is gone.

## Tradeoff, accepted

The mark is white, which suits a dark tab strip and disappears on a light one. Taken deliberately over a `prefers-color-scheme` variant, since inconsistent handling of that in tab chrome is what started this thread. If light-mode support is ever wanted, the fix is a **coloured** mark rather than a conditional one — a coloured mark is visible on both, which is exactly what the asterisk and star icons do.

## Note on the comments

This file accumulated two confidently-wrong explanations along the way — 'Safari flattens alpha to white', then 'Safari plates low-contrast icons'. Both are replaced rather than left sitting beside the correct one.

### DIFF
--- a/site/content/assets/favicon.svg
+++ b/site/content/assets/favicon.svg
@@ -1,20 +1,23 @@
 <svg viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
-  <!-- Squad Ops favicon.
-       Tile colour: the site's indigo, not black. Safari appears to add a light
-       backing plate to favicons it judges too low-contrast against a dark tab
-       strip — in a sampled tab bar, every dark icon carried a plate while every
-       colourful or light one (including GitHub's transparent white octocat)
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

Issue #1070, opened 2026-08-23
Collapse the authored copies of success_status — the plan's restatement (A) and the manifest's field (B)


jasonLadd:
The cheaper half of the #1067 follow-up, and the higher-value one: **remove an authored copy whose reason for existing is gone.**

## Why the plan carries a status at all

Historically, plan prose was the only channel carrying a declared success status to the implementer. `ScaffoldStack.skeleton_pins_success_status` records it: FastAPI lands `status_code=` in the frozen decorator, and the Next.js skeleton writes it only as a TODO comment inside the fill body the fill replaces. So on nextjs, a sentence in the plan was the sole surviving carrier — which is why #1013's completeness half required one.

**That channel now exists by derivation.** #1042 threads the status onto the developer's brief; #1063 does the same for the repair. #1049 already downgraded the omission check accordingly: it fires only where *neither* the skeleton nor the brief carries it.

So the plan's restatement is redundant. What it is not is harmless.

## A redundant copy that can still contradict

`cyc_79eebcb82205` was rejected **twice** at the plan-validation gate, both times as a manifest↔plan contradiction:

```
run 1: POST /api/runs/{run_id}/join          manifest 201, plan states [200]
run 2: POST /api/runs/{run_id}/participants  manifest 201, plan states [200]
```

Two framing cycles and both re-rolls, consumed by two documents disagreeing about an integer that neither of them needed to decide. #1067 addresses the manifest side of that disagreement; this issue addresses the fact that there are two sides at all.

And the reverse direction is still open after #1067: a manifest that stays silent (the new default) and a plan that volunteers "→ 201" against a derived 200 still trips the gate. The plan author has been right both times today, but nothing structurally prevents it being the outlier next time.

## Ask

1. **Stop asking the plan author for statuses.** Wherever the authoring brief or its rules invite a status in plan prose, remove the invitation — the fact is derived and delivered.
2. **Retire the completeness half of #1013** for stacks where a channel exists. #1049 already made it conditional; with the plan no longer expected to state statuses, the condition is permanently false there and the check should say so rather than lie dormant.
3. **Keep the contradiction half, narrowed.** If a plan volunteers a status anyway, contradicting the contract is still a real defect — a dev reading "returns 200" builds 200 whatever an appendix says. But it should read as *"the plan should not be stating this at all"*, not as *"align the two documents"*, because there is nothing to align to any more.

## Acceptance sketch

[...]

PR #1073, opened 2026-08-23
fix(plan): the plan stops restating success statuses — part A of #1070

jasonLadd:
Part A of #1070. **Removes an authored copy whose reason for existing is gone.** `Refs`, not `Closes` — part B stays open.

## The instruction that created the copy

The plan-authoring rule said it outright:

> an enforced non-200 status **must be STATED** in the owning task's description or acceptance criteria — a status that lives only in the manifest never reaches the developer, who will default to 200.

True when written. **#1042** threads the derived status onto the developer's brief and **#1063** onto the repair's; **#1049** already recognised the shift by making the omission check conditional on a channel existing.

What was left was an instruction to write a second copy of a derived fact. The only thing a second copy can add is a disagreement.

## It did, twice in one roll

```
run 1: POST /api/runs/{run_id}/join          manifest 201, plan states [200]
run 2: POST /api/runs/{run_id}/participants  manifest 201, plan states [200]
```

`cyc_79eebcb82205` — two framing cycles and both re-rolls, spent on two documents disagreeing about an integer neither of them needed to decide. The author was told exactly what was wrong after run 1 and reproduced it on a renamed endpoint.

## The change

The rule becomes **`do-not-restate-success-statuses`**: write the behaviour, not the code — *"joining a run adds the participant and returns the updated run"*, never *"returns 200"*.

A status that **agrees** is still tolerated. Redundant is not wrong, and rejecting it would punish authors mid-migration where there is no disagreement.

The contradiction finding's advice changes with it. *"Align them"* asks the author to keep two sources in step; the correct move is to delete the restatement, because there is nothing to keep in step with. The check itself stays — a plan volunteering a wrong status still teaches the developer to build something the contract rejects.

## The apparent conflict with the omission half is deliberate

The omission check still demands a stated status — but only where **neither** the skeleton nor the brief carries it, and there prose genuinely is the sole carrier, so its specific instruction correctly overrides the general rule. No registered stack reaches it today (fastapi pins structurally, nextjs carries the appendix). It is there for the third stack, and its message says what to do. Documented in place so the next reader does not read it as a contradiction.

## Verification

- Regression **7,820 passed**, gate green at 20 workers.
- **4 mutations, all killed** — including the message reverting to "align them" and the classification pointing at the retired rule id.
- Two of my first-pass mutations were bad rather than survivors: a `pass` before an `append` is a no-op, and one anchor had been reflowed by the formatter. Both re-run properly against the real text.

Refs #1070

### DIFF
--- a/src/squadops/cycles/framing_consistency.py
+++ b/src/squadops/cycles/framing_consistency.py
@@ -177,12 +177,24 @@ def validate_manifest_plan_consistency(
                 f"will enforce success {enforced} "
                 f"({'declared' if ep.success_status is not None else 'derived default'}) "
                 f'but the plan states {sorted(tokens)} — "{ln[:120]}". The implementer '
-                f"builds the plan; the contract judges the manifest. Align them."
+                f"builds the plan; the contract judges the manifest. REMOVE the status "
+                f"from the plan: the developer's brief already carries the derived "
+                f"status, so the plan restating it adds only the chance of this "
+                f"disagreement (#1070)."
             )
 
-        # Completeness: an enforced non-200 success is exactly the fact the dev
-        # will not default to — slot 6's roll died on it. It must be STATED in
-        # the plan's prose near the endpoint, or the implementer never sees it.
+        # Completeness: an enforced non-200 success is exactly the fact the dev will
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #1075, opened 2026-08-23
chore(release): 1.6.2 — the information-flow patch line

jasonLadd:
## 🟢 The first green roll of the 1.6 line, with an independent oracle agreeing

```
cyc_79eebcb82205 / run_9c879ff5458e        qwen3.8:27b

verdict      accepted
failed       none
probes       5/5 contract probes verified
BOOT AUDIT   PASS — the delivered app installs, builds, boots, answers every
             contract probe over real HTTP, and its UI reaches every path it requests
```

Coverage reads **12/14 with the shortfall named rather than silent** — `criteria_unevidenced` carrying the two `vc-compiles-*` criteria that produce no row. That is #1021's own instrument reporting from production, on the roll that cut it.

## The line has one shape

Every fix here is **a fact the system already holds, not reaching the agent judged against it**:

| | |
|---|---|
| #1029 | the shell spine pins the success body's declared floor |
| #1042 | the developer is told the declared success status |
| #1060 | the repair receives **every** manifest surface the author does — three were missing, two with renderers producing empty output since the day they were written |
| #1002 | the detector's inspected-file inventory reaches the record |
| #1053 | an emission containing nothing stops being billed as an attempt |
| #1021 | a criterion with no result row stops being a silent fourth state |
| #1049 | the framing gate stops enforcing a premise #1042 made false |
| #1055 | the frozen store gains the `update` seam it never had |
| #1067 + #1070A | the success status stops being authored three times |

Plus #761, #1030, #880, #972, #1015 B/C, #1022, and the blueprint gate's xdist dependency.

## Not exercised by the cut evidence

The green roll ran on `98eb805e`. Three changes merged after it launched and ride this tag:

| | exercised? | |
|---|---|---|
| **#1064** store `update` seam | no | additive — a new export, no existing behaviour altered, shells byte-identical |
| **#1067** status-override warrant | no | **authoring-facing behaviour** |
| **#1070 part A** plan stops restating | no | **authoring-facing behaviour** |

They are the right fixes and green in CI. The boot-audit evidence does not cover them, and the first roll of 1.6.3 is what will. Recorded because a release whose headline is *"validated by a green roll"* has to be exact about what the roll validated.

**#1067 and #1070A ship together deliberately** — #1067 constrains the manifest's copy of the status, #1070A removes the plan's, and shipping either alone leaves the disagreement reachable from the other side.

## Procedure (CLAUDE.md "Release cut")

| Step | Status |
|---|---|
| 1. `version_cli.py bump 1.6.2` | done — the only sanctioned path |
| 2. markers synced (CLAUDE.md, README, ROADMAP) | done — guard 16/16 |
| 3. CHANGELOG rotated, fresh `[Unreleased]` | done |
| 4. ROADMAP timeline entry | done |
| 5. SIP promotion sweep | **no promotions**, gaps named below |
| 6–7. tag, GitHub Release, release package | after merge — yours |

**Step 5, stated rather than skipped.** SIP-0104's promotion is a standing owner call — its measurement program is complete, but the ruling is not mine. SIP-0105 stays `accepted` because its own acceptance names its unbuilt parts (the SIP-0102-at-1.4 precedent).

## Still out

- **#1054** — a repair routed to qa while the lead named dev task types. 3.6 only, never fired on a 3.8 roll. Read-first: the locus classifier is conservative by design and the wrong fix reopens the test-gaming hole.
- **#1070 part B** — the manifest's `success_status` field stays authorable where the rule decides it. Blocked on the reference-manifest question: it declares `201` on its collection POST, agreeing with the rule, so a hard reject would fail the artifact the 1.4 evidence was measured against.
- **#1021's underlying mechanism** — why those two compile criteria produce no row. The accounting is fixed; the cause is not.

## Verification

Regression **7,829 passed**, gate green at 20 workers. Version-sync guard 16/16.

## After merge

Tag `v1.6.2` on the merge commit → GitHub Release from the CHANGELOG section, marked **Latest** → `build_release_package.py 1.6.2 --write --cycle cyc_79eebcb82205`, with screenshots into that release's `assets/` first (the script reads the tag range, so it must follow the tag).


jasonLadd (commented):
reviewed

PR #1077, opened 2026-08-23
fix(release): the package captures cycle evidence instead of reporting that it did

jasonLadd:
Found running **step 7 of the 1.6.2 cut**. The command printed `--- 29 PRs, 0 SIP moves, 1 cycles, 0 screenshots ---` and wrote a roll-up of nulls.

## Four defects, and the fourth hid the other three

| | |
|---|---|
| route | `/api/v1/cycles/{id}` — the real one is `/api/v1/projects/{project}/cycles/{id}` |
| auth | no `Authorization` header; the API requires one |
| field | reads `outcome`; the API returns `cycle_outcome` |
| **guard** | caught only `JSONDecodeError`, so `{"detail": "Not Found"}` took the **success** path |

`cycle_evidence`'s own docstring promises:

> Absence is disclosed, never silently omitted — an unreachable API and a cycle that genuinely produced nothing must not look the same later.

They looked identical. Valid JSON is valid JSON, so every capture reported `captured: True` with null fields, and nobody had reason to question the route, the header, or the field name.

## Why this is worse than a normal tooling bug

Step 7 is **capture, not query** — cycle evidence lives in a running deploy and is unrecoverable once it moves. A hollow capture is worse than no capture, because it looks like the evidence was taken. v1.6.1's package has an empty `cycles` list, so no false record has shipped; **1.6.2 would have been the first.**

## The fix

Project-scoped route, authorized from the CLI's cached token (the same store `squadops login` writes — no second credential path), reading `cycle_outcome`, and **gated on the roll-up actually being present**.

Absence is recorded *with its reason*, including `run \`squadops login\`` when no token was found — an expired login and a genuinely absent cycle want different responses from the maintainer.

Verified against the real 1.6.2 evidence:

```
cyc_79eebcb82205 → accepted · 4 runs · 15 verified checks named
```

where it previously captured nothing and said it had.

## Verification

- Regression **7,829 passed**, gate green at 20 workers.
- **6 mutations, all killed** — including each of the four original defects reintroduced individually, so a regression to any one of them fails.

## Related

The same shape as this line's other findings — an instrument reporting success while measuring nothing (#1002, #1021, #1052/#1057). Here the instrument was the release record itself.

Closes #1076

### DIFF
--- a/scripts/maintainer/build_release_package.py
+++ b/scripts/maintainer/build_release_package.py
@@ -130,38 +130,91 @@ def changelog_section(version: str) -> str:
     return match.group(1).strip() if match else ""
 
 
-def cycle_evidence(cycle_ids: list[str], api: str) -> list[dict]:
+def _absent(cycle_id: str, reason: str) -> dict:
+    """A cycle whose evidence could not be captured, WITH why — never a silent gap."""
+    return {"cycle_id": cycle_id, "captured": False, "reason": reason}
+
+
+def _bearer_token() -> str:
+    """The CLI's cached access token, or "" — the API requires one (#1076).
+
+    Read from the same store `squadops login` writes, so a maintainer who can drive
+    the CLI can capture a package without a second credential path.
+    """
+    try:
+        from squadops.cli.auth import load_cached_token
+
+        cached = load_cached_token()
+        return cached.access_token if cached else ""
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

Issue #1079, opened 2026-08-23
json_has has a consumer, tests and a spec but no producer — and the functional oracle would skip it if one existed


jasonLadd:
Two coupled halves of one gap, split off from #1029 where they were deliberately kept off PR #1040 (that PR moved the generator hash; adding the contract hash on top would have made a regression unattributable).

## Half 1 — `json_has` is dead capacity

The probe expectation kind is implemented, unit-tested and specified, and **nothing produces it**.

- Consumed: `src/squadops/capabilities/handlers/probe_runner.py:521-525` — reads `expect["json_has"]`, fails the probe on any missing key
- Implemented: `probe_runner.py:565-572` — `_has_key`, top-level key on an object, or present in every element of a list response
- Tested: `tests/unit/capabilities/test_probe_runner.py:182-188` — object, list, and non-JSON cases
- Specified: `SIP-0098` line 129 shows it in the contract's own example — `expect: {status: 200, json_has: [id, participants]}`
- Referenced as intended-to-be-authored: `SIP-0103` line 218 — "probe *payloads* and `json_has` fields carry product intent"

**No producer.** Searched the whole tree: outside the implementation, its tests, and those SIP references, `json_has` appears exactly once — `tests/unit/cycles/goldens/plan_context.json:405`, a hand-authored fixture feeding a synthetic contract into plan-context rendering. And **zero of the banked artifacts under `data/artifacts/` carry one**, across the entire corpus. No contract derivation, no authoring surface, and no reference contract has ever emitted a `json_has` expectation.

So every contract probe currently asserts status and error code only. Response *shape* is asserted by the scaffold shells (#1029's floor) and nowhere in the live probe path.

## Half 2 — the functional oracle would skip it if a producer existed

`scripts/dev/audit_delivered_app.py` re-implements probe checking inline rather than calling the runner, and implements **two of the three** expectation kinds:

- `audit_delivered_app.py:188-193` — `status`
- `audit_delivered_app.py:194-204` — `error_code` (and inlines its own envelope read rather than reusing `probe_runner._error_code_of`)
- **no `json_has` branch**

A contract carrying `json_has` would be checked by the in-cycle probe runner and silently under-checked by the audit script. The two would disagree, and the audit — the independent oracle — would be the permissive one.

## Why this matters now, not eventually

[...]

Issue #1017, opened 2026-08-21
Failed retest discards its test_report — the evidence that rejects a repair is built and dropped


jasonLadd:
A failed retest is exactly the moment the correction loop needs evidence: it decides whether a stored repair is rejected, and rejection sends the loop back around a round. Today that decision's evidence is **built and then discarded**.

## Mechanism

- `qa_test.py::_handle_retest` (execute-only #456 path) runs the repaired suite and builds a `test_report.md` artifact carrying the runner's stdout/stderr — the lines that name exactly which tests failed — and appends it to the result's artifacts.
- `_dispatch_protocol_step` (`adapters/cycles/correction_runner.py`, the `result.status == "SUCCEEDED"` gate before `_store_correction_task_artifacts`) persists step artifacts **only on success**. A red retest returns FAILED → TASK_FAILED path → nothing stored.
- Compounding: the failed qa task's own additive suite files are also never persisted (#971), so after the run there is no way to learn *which assertion* rejected the repair.

## What this cost, concretely

#1012 stood unadjudicable across two windows — three byte-verified correct repair candidates in one roll (V38 slot 6) rejected by retests whose failure detail no longer existed. Adjudication ultimately required a full offline sandbox replay of the stored trees. The dropped `test_report.md` would have answered it in seconds. Adjudication record: #1012 (closed 2026-08-21).

## Proposed fix

Persist the retest's `test_report.md` (and the retest's own `test_result` summary) on **both** outcomes — minimally: in the FAILED branch of `_dispatch_protocol_step`, store the report artifact (type `test_report`, task_id = the retest id) before emitting TASK_FAILED. The failing-test names then survive in the vault next to the rejected candidates they rejected.

Instrument-first per the standing rule: this is the capture that makes every future instance of the #1012 signature readable from banked state. Small, no behavior change to the loop itself.

## Related

- **#1012 (closed)** — the adjudication this issue is the actionable outcome of.
- **#971 (open)** — the general failed-emission-not-persisted class; this is its highest-leverage single instance.
- **#761 (open)** — failure-signature subject/reason; complementary (signature vs full report).
- **#1015 (open)** — the prevention side (repair minimality); this issue is the evidence side.
- 1.6.1 batch item alongside #1014 and #1011.

PR #1080, opened 2026-08-23
docs(plan): pre-register the 1.6.3 repeatability set — 8 rolls, frozen config

jasonLadd:
Pre-registers the 1.6.3 repeatability set: 8 rolls, frozen config, frozen deploy.

## Why a measurement rather than more fixes

1.6.2 merged roughly twenty fixes against a corpus of observed loss modes and **none of them has been measured**. The line's entire evidence is one green roll (`cyc_79eebcb82205` — verdict accepted *and* boot audit PASS, the first time verdict and oracle have ever agreed), and one green roll is not a rate.

A census of the seventeen rejected implementation runs from 08-14 to 08-23 makes the case sharper than "we should measure":

| Loss class | N | Fix status at 1.6.2 |
|---|---|---|
| Unfilled scaffold slots | 5 | fix landed 08-17; last instance 08-18 |
| Error envelope | 2 | #795 closed 08-22 — both instances predate it |
| Status 200/201 | 2 | #1067 / #1042 / #1070A |
| Store mutation | 2 | #1064 `update()` seam |
| Truncated emission | 2 | **#998 open, no fix** |
| Repair-induced regression | ≥1 | #994 open |

**Eleven of seventeen belong to classes already fixed.** The corpus describes a system that no longer exists — it records what was killed, not what kills the next roll. That is why the stack shrinks to three items rather than the nine the 08-21 sweep scoped, and why the deliverable is the measurement.

## What the set can and cannot claim

N=8, with the limits stated in §1.3 rather than implied. Headline precision barely moves across this range (95% CI width 62 → 57 → 53 points for N = 6 → 8 → 10), and **the set cannot distinguish a 50% system from a 65% one** — that needs N in the hundreds. What the two extra rolls over N=6 buy is a better floor on a clean sweep (61% → 68%) and texture coverage: a defect class firing at 25% goes unobserved entirely in 18% of six-roll sets versus 10% at eight. **The texture fields (§4) are the instrument at this N; the headline is not.**

There is also no prior worth beating. On the profile this set uses the prior is **1 of 3** — so the set establishes a baseline, and 1.6.4's set compares against it.

## What is inherited and what is new

Inherits V7's roll-validity rules (§5.1), gate-policy-as-a-constant (§7.3c) and freeze discipline (§6) verbatim rather than re-deriving them. Two things are new:

- **The metric is verdict AND boot audit AND zero intervention.** Verdict `accepted` and *working* have coincided exactly once in this line's history; conflating them is the error the set exists to correct.
- **#1079's parity half is a blocking precondition**, under V7 §2.a's rule that *a set measuring zero manual intervention cannot require manual intervention to score itself*. #952/#953 were ruled blocking for exactly this reason. #1079's producer half is deliberately excluded — it moves the contract hash and would void the freeze mid-set.

## Recorded caveat

§8 carries the census's own limitation rather than burying it: **only 4 of the 17 rejections have a banked test report.** #1017 closed 08-21, so the other 13 rest on the analyzer's round-0 claim — and #968 exists precisely because nothing verifies those claims against the source. That is why #971 leads the stack.

## State

DRAFT — §2 preconditions pending. §2.c stack ruled 2026-08-23: #971 (instrumentation, first), #998 + #939 (emission guards), #1079 parity. Shakeout `cyc_3ba9dc0f67da` in flight and declared non-counting before launch.

Refs #1079, #971, #998, #939, #947, #936

### DIFF
--- a/docs/plans/1-6-3-repeatability-set-preregistration.md
+++ b/docs/plans/1-6-3-repeatability-set-preregistration.md
@@ -1,0 +1,263 @@
+# 1.6.3 — Repeatability Set: Pre-Registration
+
+**Status: DRAFT — not in force until every §2 precondition holds.** Once in force, nothing in
+this document changes until the set closes.
+
+**Owner:** maintainer · **Question:** on a frozen config and a frozen deploy, at what rate does
+the post-1.6.2 system produce a **working** application? · **Prior:** none usable — see §1.2.
+
+---
+
+## 1. Design
+
+### 1.1 Why this set exists
+
+1.6.2 merged roughly twenty fixes against a corpus of observed loss modes. **None of them has
+been measured.** The line's evidence is one green roll (`cyc_79eebcb82205`, verdict accepted +
+boot audit PASS) — the first time verdict and oracle have ever agreed — and a single green roll
+is not a rate.
+
+A loss-mode census over the seventeen rejected implementation runs from 2026-08-14 to 08-23
+makes the case sharper. Classified by round-0 analysis and, where banked, by the test report:
+
+| Loss class | N | Fix status at 1.6.2 |
+|---|---|---|
+| Unfilled scaffold slots | 5 | shell-imports-TABLES fix landed 08-17; last instance 08-18 |
+| Error envelope disagreement | 2 | #795 closed 08-22 — both instances predate it |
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #1081, opened 2026-08-23
fix(evidence): a failed task's emission is banked for triage, and used for nothing

### LINKED ISSUE #971: A failed task's emission is not persisted — the artifact that caused the failure cannot be read afterwards
When a task fails, the artifacts it emitted are not persisted. **The artifact that caused the failure cannot be read afterwards** — only whatever a subsequent repair produced.

## How it bit

SIP-0104 P6 window roll 6 failed on a test suite. The only stored `__tests__/runs.test.ts` in the run's vault is the one `qa.test_repair` wrote:

```
__tests__/runs.test.ts | source | qa.test_repair
```

The suite that actually failed does not exist anywhere. Consequences during triage:

- the root cause (#967) had to be established by reading the **application's** source and inferring what the test must have asserted, rather than by reading the test;
- the specific mismatched literal remains an **inference** rather than a measurement, which is the difference between a filed defect and a proven one;
- checking whether the failing assertion sat inside a fill region — the question that decides whether #970 was load-bearing — is **permanently unanswerable** for this roll.

## Why it matters beyond convenience

Every layer of the SIP-0096 / SIP-0104 work rests on being able to say *what was verified and what was not*. A failed task is exactly the case where that question is most valuable, and it is the one case where the evidence is discarded.

It also compounds #968: the failure analyzer is asked to diagnose from evidence, and a human auditing the analyzer's claim afterwards has less to work with than the analyzer did.

## Fix direction

Store the failed emission, marked as failed. Two properties matter more than the mechanism:

1. **It must not enter the deliverable.** The acceptance-aware last-wins selection (`producing_task_type` not in `REPAIR_TASK_TYPES`) already models exactly this distinction — a rejected candidate is stored and excluded from assembly. The same shape applies: persist under a type the workspace selection skips.
2. **It must survive the repair that replaces it.** The point is the *pair* — what failed and what replaced it — so a repair overwriting the record destroys the comparison that makes it useful.

## Same family

#946 (nothing logs which validation check opened the self-eval branch) and #935 (no record of whether the suite executed) are the same class — both now fixed, both fixed *because* a window roll needed triage and the state was unreadable. This is the third and largest instance: not a missing log line, a missing artifact.

Evidence: `docs/plans/sip-0104-p6-window-record.md`, roll 6 section.
Related: #946, #935, #968, #967, #926.

jasonLadd:
Closes #971

First item of the 1.6.3 stack (plan: PR #1080 §2.c). It leads the stack because the repeatability set cannot diagnose a red roll without it.

## The gap

The one artifact guaranteed absent from a run's vault was the one that **caused** the failure. The sequential path raises or `continue`s before `_collect_artifacts_and_checkpoint` (`dispatched_flow_executor.py:3589`), so the only surviving copy of a file was whatever a later repair produced — and a repair's output is precisely not the evidence you want when asking why the original failed.

The bill is measurable. Of the 17 rejected implementation runs between 08-14 and 08-23, **only 4 have a banked test report**; the other 13 can be classified only from the analyzer's own round-0 claim, and #968 exists because nothing verifies those claims against the source.

## Capture

Hooked in `_route_outcome`, which fires on **every** failed attempt rather than only the last — so what lands is the pair the issue asks for: what failed and what replaced it.

Best-effort by construction: banking evidence must never alter the control flow of the failure it records. #1017 made the same call for the failed retest's `test_report`.

## Exclusion — the half that carries the risk

Triage only. A failed emission never feeds a workspace, a repair, or the deliverable, and that gets **three independent guarantees** rather than one:

1. the marked artifact never enters the run's `stored_artifacts` list;
2. `_resolve_artifact_contents` filters it **unconditionally**, with no flag. The repair-candidate exclusion beside it takes one because a candidate is merely unaccepted and the correction loop legitimately reads its own accumulation. A failed emission is known-bad to every caller;
3. `audit_delivered_app._select_deliverable` skips it. That script reads the vault directly, latest-per-filename, bypassing the executor entirely. **Its trap is the narrower and worse one:** a failed emission is often the *only* copy of its file — nothing re-emitted it before the run died — so the auditor would have reported on an application the run never produced. It is also the oracle the repeatability set scores on.

## On the naming

`emission_status` is provenance, not a type: `producing_task_type` stays truthful, because a failed `qa.test` emission really was produced by `qa.test` and every post-hoc reader wants to know that. This is the mark #1017 lacked — it could store *evidence only* precisely because it had no way to render workspace files non-selectable.

## Tests

9 on the executor seam — capture, the `SUCCEEDED` and no-artifact guards, vault failure swallowed, exclusion under both flag settings, exclusion when the failed emission is the only copy, and the pair banked with only the good half composing — plus 2 on the auditor's selection.

Regression suite: **7838 passed, 1 skipped, all gates** (ruff check, ruff format, test-quality lint, pytest).

### DIFF
--- a/adapters/cycles/dispatched_flow_executor.py
+++ b/adapters/cycles/dispatched_flow_executor.py
@@ -1305,6 +1305,19 @@ class DispatchedFlowExecutor(FlowExecutionPort):
         if not filter_spec:
             return {}
 
+        # #971: a FAILED emission is never workspace input, for any caller. Unlike the
+        # repair-candidate exclusion below this one is unconditional and takes no flag —
+        # a candidate is merely unaccepted (the correction loop legitimately reads its
+        # own accumulation), whereas an emission that failed its checks is known-bad.
+        # Storing it is a triage instrument (#971); letting it compose a workspace would
+        # feed known-bad bytes to the next task and, on the acceptance workspace, into
+        # the assembled deliverable.
+        stored_artifacts = [
+            (art_id, ref)
+            for art_id, ref in stored_artifacts
+            if ref.metadata.get("emission_status") != "failed"
+        ]
+
         if not include_repair_candidates:
             from squadops.cycles.task_plan import REPAIR_TASK_TYPES
 
@@ -1594,6 +1607,14 @@ class DispatchedFlowExecutor(FlowExecutionPort):
                 _holder=_last_failed_result,
             ):
                 _holder["result"] = result
+                # #971: bank THIS attempt's emission before the correction loop
+                # decides anything. The callback fires on every failed attempt, so
+                # what lands is the pair the issue asks for — what failed and what
+                # replaced it — rather than only the last failure. Stored marked and
+                # excluded everywhere; see _store_failed_emission.
+                await self._store_failed_emission(
+                    result, _envelope, cycle, run_id, all_artifact_refs
+                )
                 action = await self._handle_task_outcome(
                     result=result,
                     envelope=_envelope,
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #1083, opened 2026-08-23
fix(evidence): catch an emission that stopped mid-construct, at the task that wrote it

jasonLadd:
Closes #1082

Second item of the 1.6.3 stack (plan: PR #1080 §2.c). It targets the one loss class in the 08-14→08-23 census that is still live and unguarded.

## The gap

A completion that runs out of budget mid-file is stored as if it were whole. `cyc_87c12c7f199e` banked a 407-byte route ending on `throw new` with three unclosed braces. Nothing questioned it — the bytes look like code — until the suite tried to transform it, and then the failure belonged to somebody else: `tests_pass` red across **two whole test files**, one of which merely imported the module, correction budget already spent, and the analyzer diagnosing from suite output rather than from the emission.

`frontend_compiles` and `tests_pass` do reject such a file. Both run at acceptance, *after* the producing task is finished. This check is injected on the producer's **own** artifacts (the #689 seam), so a truncation is charged to the task that wrote it while that task still owns the round.

## Deliberately narrow

Not "this file is valid" — only "this file ends inside an unclosed construct". Python gets the exact answer from `compile`, filtered to the EOF-shaped errors; the brace languages get delimiter balance and nothing more. Widening it would make this the general syntax gate, which it cannot be on languages it does not parse.

That narrowness is also what makes it viable where a real TypeScript check is not: **delimiter balance needs no parser and therefore no Node**, which is #939's blocker and #1021's suspected mechanism.

## Validation found more than it confirmed

Swept across **all 4,513 scannable source artifacts** in the banked corpus: **8 flags, every one a genuine truncation on inspection, zero false positives.** Three of the eight had been masked by an over-conservative early return.

Two false positives surfaced during the sweep, and each drove a real fix rather than a tuned threshold:

- **JSX punctuation read as a regex opener.** `<Route ... />` made the scanner treat `/>` as a regex, swallowing text to the next `/` — which sat inside `path="/"` — and unbalancing the whole file.
- **Parens counted inside JSX text.** `Participants ({' '}` renders a literal `(` whose `)` lives in a sibling `{')'}` expression. Those never balance, so `(`/`[` are untracked in `.tsx`/`.jsx`; `{` stays, being JSX's own always-matched delimiter. Four of the eight corpus truncations are `.ts` test files that only an unclosed `(` reveals, so the tradeoff is scoped by extension rather than surrendered.

A third false positive came from the test suite rather than the corpus: an earlier Python fallback used position alone ("error on the last line"), which flags every single-line file with any syntax error. The message must independently say something was left open. All three are pinned as tests.

## Out of scope, and said so

Prose. The census's second instance is a `qa_handoff.md` that stopped mid-paragraph; markdown has no delimiter invariant and "ends mid-sentence" false-positives on documents that legitimately end in a list item. Silence beats a guess.

Worth noting: that second instance is attested only by the analyzer's round-0 account — **the truncated bytes were discarded**, which is #971's defect illustrating itself.

## Incidental

`_framework_injected_criteria` now derives its check list from the registry (framework-injected, sole param `file`) rather than naming one check inline, so the menu and the injector cannot drift.

## Tests

31 on the scanner, split between catching the shapes the corpus produced and never flagging healthy ones — regex holding braces, JSX text, nested template interpolation, division-not-regex, escaped quotes, f-strings. Regression suite: **7860 passed, 1 skipped, all gates**.

### DIFF
--- a/docs/architecture/typed-check-menu.md
+++ b/docs/architecture/typed-check-menu.md
@@ -22,6 +22,7 @@ Regenerate: `UPDATE_CHECK_MENU=1 pytest tests/unit/cycles/test_check_governance.
 | `module_imports` | authored | product | yes | yes | yes | yes | error |
 | `regex_match` | authored | product | yes | yes | yes | yes | error |
 | `undefined_names` | injected | product | yes | yes | yes | yes | error |
+| `unterminated_source` | injected | product | yes | yes | yes | yes | error |
 
 `command_exit_zero` ownership is per-command in truth. The forms it may take
 are inventoried below — this replaces the standing caveat that called the
--- a/src/squadops/capabilities/handlers/cycle/base.py
+++ b/src/squadops/capabilities/handlers/cycle/base.py
@@ -16,7 +16,7 @@ from squadops.capabilities.handlers.base import (
     HandlerResult,
 )
 from squadops.cycles.acceptance_check_spec import (
-    CHECK_UNDEFINED_NAMES,
+    CHECK_SPECS,
     CONFIG_OFF_SKIP_REASONS,
     is_check_applicable,
 )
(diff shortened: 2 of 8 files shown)


jasonLadd (commented):
reviewed

PR #1084, opened 2026-08-23
fix(audit): the oracle and the runner judge a contract with the same code

### LINKED ISSUE #1079: json_has has a consumer, tests and a spec but no producer — and the functional oracle would skip it if one existed

jasonLadd:
Closes #1079 (parity half)

Third and last item of the 1.6.3 stack (plan: PR #1080 §2.c), and a **blocking precondition** for the repeatability set under pre-registration §2.b.

## The gap

`audit_delivered_app` decides whether a delivered app actually works, independently of the run's own verdict. It re-implemented the probe expectation block inline and carried **two of the three kinds** — no `json_has` branch — and inlined its own envelope read rather than using `probe_runner._error_code_of`.

Two judges of one contract, and the one whose verdict carries the most weight was the more permissive.

Nothing had caught it because `json_has` still has no producer — the other half of #1079, deliberately left out here because it moves the contract hash and would void the set's freeze. So this is precisely the window in which it bites: a producer lands, and the oracle silently under-checks until someone notices.

## The fix

The duplication was the defect; the missing branch was its symptom. So one judgment is shared rather than a third kind added to a second copy.

`evaluate_expectations(expect, status_code, payload)` is extracted from `_run_one`, and both callers use it. Transport stays with each — the runner is sync against a sandbox URL, the auditor async against its own client — because only the *judgment* must not differ.

## Why it blocks the set

Inheriting V7 §2.a: *a set measuring zero manual intervention cannot require manual intervention to score itself.* #952/#953 were ruled blocking for exactly this reason in the V7 window. The 1.6.3 set's headline metric is boot-audit PASS, so an oracle that skips an expectation kind would make a green set unfalsifiable in the direction least worth trusting.

## Tests

Parity is asserted as **function identity**, not as agreeing outputs — matching behaviour on sampled inputs is exactly what the previous arrangement had, right up until a third expectation kind was added to one side.

Plus the judgment's own cases: list responses requiring the key in every element, the empty-collection boundary #1029 settled, non-JSON bodies, and status being judged before shape (a 500 should report its status, not list the keys its error body lacks).

Regression suite: **7869 passed, 1 skipped, all gates**.

### DIFF
--- a/scripts/dev/audit_delivered_app.py
+++ b/scripts/dev/audit_delivered_app.py
@@ -49,6 +49,7 @@ import httpx
 
 from adapters.sandbox.container_backend import ContainerBackend
 from adapters.tools.docker import DockerAdapter
+from squadops.capabilities.handlers.probe_runner import evaluate_expectations
 from squadops.capabilities.ui_data_path import (
     LIVE_SERVER,
     SERVED,
@@ -193,22 +194,14 @@ async def _run_probes(contract: VerificationContract, base_url: str) -> list[str
                 payload = resp.json()
             except ValueError:
                 payload = None
-            expected = probe.expect.get("status")
-            if expected is not None and resp.status_code != expected:
-                failures.append(
-                    f"probe {probe.id}: status {resp.status_code} != expected {expected}"
-                )
+            # #1079: the SAME judgment the in-cycle runner applies. This block used
+            # to be a second implementation carrying two of the three expectation
+            # kinds — it had no `json_has` branch — so the oracle was quietly the
+            # more permissive of the two judges of one contract.
+            failure = evaluate_expectations(probe.expect, resp.status_code, payload)
+            if failure is not None:
+                failures.append(f"probe {probe.id}: {failure}")
                 continue
-            error_code = probe.expect.get("error_code")
-            if error_code is not None:
-                actual = (
-                    (payload.get("error") or {}).get("code") if isinstance(payload, dict) else None
-                )
-                if actual != error_code:
-                    failures.append(
-                        f"probe {probe.id}: error_code {actual!r} != expected {error_code!r}"
-                    )
-                    continue
             if probe.capture:
                 captured, missing_key = capture_probe_values(probe, payload)
                 if missing_key is not None:
--- a/src/squadops/capabilities/handlers/probe_runner.py
+++ b/src/squadops/capabilities/handlers/probe_runner.py
@@ -33,6 +33,7 @@ import subprocess
 import sys
 import tempfile
 import time
+from collections.abc import Mapping
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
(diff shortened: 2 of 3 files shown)


jasonLadd (commented):
reviewed

PR #1086, opened 2026-08-24
docs(plan): the 1.6.3 set's preconditions are met — freeze recorded, both shakeouts read out

jasonLadd:
Every §2 precondition flips to MET, so the document goes in force and **§7 starts binding**: no merges, no rebuilds, no prompt or plan-asset edits until the set closes. This is the last thing that can land before roll 1.

## What is actually frozen

§3 records the deploy commit `5c6c64f7`, but the thing the driver asserts at **every** launch is the seven **image ids**. A rebuild changes those even when the commit does not, and "rebuild exits 0 with stale agents" is a known failure on this box — identity beats provenance.

HEAD may legitimately sit ahead of the deploy commit for a docs-only change made before roll 1. #1085 did exactly that, and an earlier draft of the driver's preflight wrongly called it a freeze violation.

§2.3 is met by **runtime** verification rather than grep: the injected-check tuple resolves to `('undefined_names', 'unterminated_source')` and the shared probe judge answers, in all seven agents plus runtime-api.

## §2.d — both shakeouts, and what the second one does not prove

| | shakeout 1 | shakeout 2 |
|---|---|---|
| deploy | pre-stack | frozen, full stack |
| verdict / audit | accepted / PASS | accepted / **PASS** |
| framing re-rolls · corrections | 0 · 0 | 0 · 0 |
| wall clock | 52 min | 62 min |

Shakeout 2 was required because #1082 is the only stack item that changes squad-facing behaviour — a check that can fail a task.

**It establishes** that the guard does not reject healthy work in flight, which is the risk the 4,513-file corpus sweep could not close (a sweep reads stored bytes; a roll produces new ones).

**It does not establish** that the guard catches anything. Nothing in either shakeout was truncated, so it was never asked. That rests on the sweep and the unit tests, and the closing claim may not imply otherwise.

## §2.e — the #947/#936 read corrects the census

The "unfilled scaffold slots" class — 5 of the census's 17, its largest — was **never the qa author omitting fills**. #987 (`52ed447f`, 08-18) is the cause: a language-prefixed fill fence the exact-form parser rejected, after which the shells rendered their missing-fill failing states and **the analyzer wrote a parse rejection up as an authoring failure**. §1.1's census inherited that wholesale.

Dead on the numbers: 4 of 17 before 08-19, **0 of 25 on and after**.

The generalisable lesson is kept in the document: a census built on analyzer prose inherits the analyzer's misattributions. That is what #968 is about.

## §4 — the criteria ratio is marked confounded

Two shakeouts, one frozen deploy, same config: **12/12 with nothing unevidenced, and 9/14 with five** — all five `vc-compiles-*`, the #1021 signature. The ratio moves with #1021's unexplained variability as much as with anything the squad did, so it is reported as a signal *about #1021* rather than as a quality measure.

Eight rolls on a frozen deploy will be the largest same-configuration sample that question has ever had.

Refs #1080, #1021, #1082, #987

### DIFF
--- a/docs/plans/1-6-3-repeatability-set-preregistration.md
+++ b/docs/plans/1-6-3-repeatability-set-preregistration.md
@@ -88,13 +88,13 @@ sequential speed against the memory-bandwidth ceiling, so concurrency does not s
 
 | # | Precondition | State |
 |---|---|---|
-| 2.1 | **Shakeout roll read out** — `cyc_3ba9dc0f67da`, declared NON-COUNTING before it launched (§2.a). Findings fixed or declared before roll 1 | pending |
-| 2.2 | **#1079 parity half landed** — `audit_delivered_app.py` implements every probe expectation kind, or calls `probe_runner` instead of re-implementing it. **Blocking, per §2.b** | pending |
-| 2.3 | Stack landed and each fix verified LOADED in-container, not merely built | pending — see §2.c |
-| 2.4 | **#947 / #936 read complete** — is the unfilled-slot class dead or lucky? If alive it enters the stack and this set re-registers | pending |
-| 2.5 | Deploy frozen; commit + all seven image ids recorded in §3 | pending |
-| 2.6 | Zero unreleased focus leases; nothing in flight | check at each launch |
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

Issue #1087, opened 2026-08-24
The frozen store hands the qa author a table for every declared entity — including shapes and projections no correct app writes


jasonLadd:
The frozen store exports a table handle for **every** declared entity, including entities that exist only as embedded shapes or response projections. Nothing tells the qa fill author which tables a correct application actually writes. When the author asserts on one of the others, a **working** application is rejected.

## The separation, on one frozen deploy

Five rolls of the 1.6.3 repeatability set, identical deploy `5c6c64f7`, identical config `d4d4f66217d8`, identical squad:

| roll | tables the qa fills assert on | verdict | boot audit |
|---|---|---|---|
| shakeout 2 | `Run` | accepted | PASS |
| roll 1 | `Run`, **`Participant`** | **rejected** | PASS |
| roll 2 | `Run` | accepted | PASS |
| roll 3 | `Run` | accepted | PASS |
| roll 4 | `Run`, **`Participant`**, **`RunSummary`** | **rejected** | PASS |

Every roll that asserted only on `Run` was accepted. Both rolls that reached for another table were rejected. **All five delivered a working application** — every boot audit passed: installs, builds, boots, answers all five contract probes over real HTTP, UI reaches every path.

So the two rejections are false. The applications were fine.

## Mechanism

1. The manifest declares `Participant` as an entity because `Run.participants` is `list[Participant]` — it needs the *shape*. It declares `RunSummary` because `GET /api/runs` has `response: "list[RunSummary]"` — a *projection*, and its own warrant says so: *"a stored count keeps the list projection compact."*
2. The scaffold derives `TABLES` from the entity list without distinguishing root-persisted entities from shapes and projections. The frozen `lib/store.ts` therefore exports:
   ```ts
   export const TABLES = {
     Participant: 'participant',
     Run: 'run',
     RunSummary: 'run_summary',
   } as const
   ```
3. The dev stores participants embedded in the Run — which is what `list[Participant]` specifies and what the contract probes verify. Roll 4's routes write to `TABLES.Run` and nothing else, across all three handlers. That is correct.
4. Nothing on the qa author's surface says which tables the application root-persists. The author sees three table handles and a `participants` array to verify, and asserts:
   ```ts
   expect(all(TABLES.Participant)).toHaveLength(1)
   ```
5. `tests_pass` fails: `expected [] to have a length of 1 but got +0`. The table is empty because no correct implementation ever writes it.

## Why this is a framework defect, not an authoring defect

[...]

PR #1088, opened 2026-08-24
docs(plan): the 1.6.3 repeatability set closes at 5 of 8 functional

jasonLadd:
Eight counted rolls, **no voids, no resets**, on a frozen deploy with the pre-registration in force from roll 1 and unchanged throughout.

## Headline

**5 of 8 functional — 62.5%, 95% CI [30.6%, 86.3%]** (verdict AND boot audit AND zero manual intervention). Every gate was decided by `system:no_open_questions`, so zero-intervention is literally true here rather than true-by-ruling.

The interval is what §1.3 pre-registered. The set establishes the baseline 1.6.4 compares against; it cannot detect an improvement at this N and doesn't claim one.

## The loss breakdown, which required correcting the set mid-flight

**The boot audit does not verify response bodies.** Every contract probe carries only `status` and `error_code`, because `json_has` has no producer (#1079's deferred half). So "8 of 8 audits passed" means the apps **boot and answer with the right status codes** — not that they answer correctly. The set had been reporting the stronger claim for four rolls; roll 5 forced it out.

With that correction the three rejections separate:

| roll | cause | app actually correct? |
|---|---|---|
| 1 | **#1087** — fill asserted `TABLES.Participant` | **yes** — false rejection |
| 4 | **#1087** — asserted `TABLES.Participant` and `TABLES.RunSummary` | **yes** — false rejection |
| 5 | #1029's response floor caught a missing field the audit structurally cannot see | **no** — correct rejection |

**The framework wrongly rejected two working applications and correctly rejected one broken one.** That's the actionable result, more than the rate.

#1087 was held until a trend existed — an early sweep suggesting "7 of 8" turned out to be an artifact of matching the frozen `harness.test.ts`, and it was filed only after roll 4 supplied a second frozen-deploy positive.

## Measured along the way

- **The framing tax is gone.** Zero re-rolls across eight rolls; ten consecutive cycles now frame on the first attempt, against **three** framing runs for the pre-1.6.2 green roll. That is #1067 and #1070A, measured rather than asserted.
- **Corrections split with no overlap** — 0–1 rounds all accepted, 3–4 rounds all rejected. Two greens converged on round one so the loop isn't dead weight, but **no roll ever recovered past round one**, and roughly two of the set's nine hours went to rounds that never worked. Recorded as association, not cause.
- **`criteria_unevidenced` never settled** — 1 to 5 `vc-compiles-*` dropped per roll on one frozen deploy. Marked confounded before the set opened and still confounded, now with the largest same-configuration sample #1021 has ever had.
- **#971 earned its slot** — 44/48/11/48/11 failed emissions banked on the rolls that needed them. Roll 1's root cause was traced by reading artifacts that would not have existed before this release.

## What it does not claim

Not a general rate (`full-38` only; `full` remains the canonical squad), not a claim about response correctness, not evidence that #1082 *catches* truncations — nothing was truncated, so it demonstrated only zero false positives across eight rolls, which was the risk that mattered.

Closes the set opened in #1086. Refs #1087, #1079, #1021, #1067, #1070

### DIFF
--- a/docs/plans/1-6-3-repeatability-set-record.md
+++ b/docs/plans/1-6-3-repeatability-set-record.md
@@ -1,0 +1,133 @@
+# 1.6.3 — Repeatability Set: Record
+
+**Closed 2026-08-24.** Eight counted rolls, no voids, no resets. Pre-registration:
+`docs/plans/1-6-3-repeatability-set-preregistration.md`, in force from roll 1 and unchanged
+throughout.
+
+---
+
+## 1. Headline
+
+**5 of 8 functional — 62.5%, 95% CI [30.6%, 86.3%].**
+
+Functional means what §5 pre-registered: verdict `accepted` **and** boot audit passes **and**
+zero manual intervention. All three held for five rolls.
+
+**8 of 8 boot audits passed**, and §2 below is careful about what that does and does not mean.
+
+The interval is what §1.3 said it would be. This set cannot distinguish a 50% system from a
+65% one and does not claim to. It establishes the baseline that 1.6.4's set compares against.
+
+| roll | cycle | verdict | audit | corrections | wall | criteria |
+|---|---|---|---|---|---|---|
+| 1 | `cyc_a24e4619844e` | rejected | PASS | 4 | 96 min | 8/14 |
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

Issue #1089, opened 2026-08-25
The single-sourced framework version reads stale editable-install metadata — the CLI reports 1.4.0 on a 1.6.3 tree


jasonLadd:
CLAUDE.md states the framework version is *"single-sourced via `importlib.metadata.version(\"squadops\")`"*. On a working tree at 1.6.3, that call returns **1.4.0**.

```
$ .venv/bin/python -c 'import importlib.metadata as m; print(m.version("squadops"))'
1.4.0
$ grep ^version pyproject.toml
version = "1.6.3"
$ .venv/bin/squadops --version
squadops 1.4.0
```

## Cause

An editable install writes its metadata **once, at install time**. `pip install -e .` was last run when the tree was at 1.4.0, and nothing regenerates that metadata when `pyproject.toml`'s version changes. `src/squadops/__init__.py:33` reads it into `__version__`, so every consumer inherits the stale value.

Deployed containers are correct (runtime-api reports 1.6.2 for its build) because the image build reinstalls. **This is a local/dev-environment defect** — but the surface it corrupts is the one CLAUDE.md names as authoritative.

## How it surfaced

During the 1.6.3 cut. `version_cli.py bump 1.6.3` printed:

```
Version bumped: 1.4.0 -> 1.6.3
```

The edit was correct (1.6.2 → 1.6.3 in `pyproject.toml`); only the "from" was wrong, because `get_framework_version()` reads installed metadata rather than the file it is about to edit. A maintainer reading that line would reasonably conclude the tree had been sitting at 1.4.0.

## Why it belongs in the 1.7 hardening class

This is the *"the signal that says it's fine isn't wired to what ships"* family — the same shape as the rebuild that exits 0 with stale agents, and `build_release_package` recording `captured: true` having captured nothing (#1076). Here the authoritative version source silently reports a value four minor releases old, and the one tool whose entire job is version correctness prints it without noticing.

## Fix direction

Two parts, and the second matters more:

1. **`version_cli.py` should read the version it is editing** — `pyproject.toml` — not the installed metadata. Its "from → to" line is a correctness claim about a file.
2. **Decide what `__version__` should mean in an editable install and make it true.** Either read `pyproject.toml` when the distribution is editable, or have `doctor` assert that installed metadata matches the file and fail loudly when it drifts. Silently serving a four-versions-old number from the surface documented as single-sourced is the part worth fixing.

[...]

PR #1090, opened 2026-08-25
chore(release): bump framework version to 1.6.3 + sync markers, rotate CHANGELOG

### LINKED ISSUE #971: A failed task's emission is not persisted — the artifact that caused the failure cannot be read afterwards

jasonLadd:
**The measurement patch line.** 1.6.2 merged roughly twenty fixes and none had been measured. This ships three, and for the first time a **rate**.

## The evidence

Pre-registered eight-roll set on a frozen deploy (`docs/plans/1-6-3-repeatability-set-record.md`), in force from roll 1 and unchanged throughout. **No voids, no resets.** Every gate decided by `system:no_open_questions`, so zero manual intervention is literally true rather than true-by-ruling.

- **5 of 8 functional — 62.5%, 95% CI [30.6%, 86.3%]**. Wide, and pre-registered as such: it establishes the baseline 1.6.4 compares against, and does not claim an improvement it cannot detect at N=8.
- **8 of 8 applications booted.**
- **Zero framing re-rolls across all eight** — ten consecutive cycles now frame on the first attempt, against **three** framing runs for the pre-1.6.2 green roll. That's 1.6.2's success-status single-sourcing, measured.

**Two of the three failures were the framework wrongly rejecting a working application** (#1087). That outranks the rate and the notes lead with it.

## Fixed

#971 (failed emissions banked for triage) · #1082 (truncation caught at the producing task) · #1079 parity half (oracle and runner judge with the same code) · #1076 (release package captures evidence instead of reporting that it did)

## Named, not glossed

#1087 · **#1079's producer half — contract probes do not verify response bodies**, so the audit certifies *boots and answers with the right status codes*, not *answers correctly* · #1021, unsettled across eight same-configuration rolls · #1089, found during this cut

## Two procedure notes

1. **The CHANGELOG was written from the merge log, not rotated.** `[Unreleased]` said "Nothing yet" despite three merged fixes — the #789 class of unguarded cut step, flagged for the procedure.
2. **Code drift between the tagged tree and the validated deploy: zero.** The freeze held from roll 1 to the tag — the v1.6.0 standard v1.6.2 could not meet.

Scope: `full-38` (qwen3.8) with the nextjs_ts overrides on `group_run`; `full` (qwen3.6) remains the canonical squad. No SIP promotions — these are issue-driven fixes.

Regression suite **7869 passed, all gates**. Version-sync guards: 16/16.

Closes #971, closes #1082, closes #1076


jasonLadd (commented):
reviewed

PR #1091, opened 2026-08-25
docs(release): capture the v1.6.3 release package

jasonLadd:
Step 7 of the cut, run as a **preview first and read before writing**. The 1.6.2 package reported `captured: true` having captured nothing (#1076), and a hollow capture is worse than none: it looks like the evidence was taken, and the deploy it came from is gone by the time anyone looks.

The preview showed real evidence this time — `cyc_9c43f56c5cd8`, verdict `accepted`, fourteen named verified checks including **`acceptance:unterminated_source`**: #1082's guard appearing in a live verification record on the release that shipped it.

Cycle evidence is one green roll from the eight-roll set. The set's full record — including the three rejections and what each of them actually was — is `docs/plans/1-6-3-repeatability-set-record.md`.

### DIFF
--- a/site/content/releases/v1.6.3/index.md
+++ b/site/content/releases/v1.6.3/index.md
@@ -1,0 +1,99 @@
+---
+title: v1.6.3
+---
+
+# v1.6.3
+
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #1092, opened 2026-08-25
fix(version): the framework version comes from the file, not an install-time copy

### LINKED ISSUE #1089: The single-sourced framework version reads stale editable-install metadata — the CLI reports 1.4.0 on a 1.6.3 tree

jasonLadd:
Closes #1089

CLAUDE.md said the version was *"single-sourced via `importlib.metadata.version('squadops')`"*. On a tree at 1.6.3 that call returned **1.4.0**. `squadops --version` printed it. And `version_cli.py` — whose entire job is version correctness — announced "Version bumped: **1.4.0** -> 1.6.3" while correctly editing 1.6.2.

An editable install writes its metadata once, at install time, and nothing regenerates it when `pyproject.toml` changes. So the surface documented as authoritative had been serving a number four minor releases old, and the tooling repeated it without noticing.

**Deployed containers were unaffected** — the image build reinstalls — which is exactly why it survived. Every place that would have caught it was looking at a correct copy.

## The model

`pyproject.toml` is the single source; installed metadata is the **install-time copy** of it. Resolution follows: a source checkout wins because it's the thing being edited; otherwise metadata, which is correct for an installed package and is all that exists there.

- The fallback is **tested directly** — the fix must not trade a stale dev version for a broken production one.
- Guarded by `name = "squadops"` rather than by filename, so a checkout nested inside an unrelated project cannot report the neighbour's version.
- Lives in its own module rather than as a function in `__init__`: a definition between the imports there pushes every later import past E402, and *"where does the version come from"* is a concern of its own.
- `version_cli.py` now reads the file it edits — its "from → to" line is a correctness claim about `pyproject.toml`.

## One test was pinning the defect

`tests/unit/cli/test_commands_meta.py` asserted the CLI printed `importlib.metadata.version("squadops")`. On a 1.6.3 tree it therefore **demanded the output say 1.4.0**, and passed while the CLI was wrong. It now asserts the CLI agrees with the package — the invariant worth holding, with neither surface's source pinned.

## Why it matters beyond itself

This is the **1.7 hardening class in miniature** — *the signal that says it's fine isn't wired to what ships* — alongside the rebuild that exits 0 with stale agents and #1076's captured-nothing release package.

Found during the v1.6.3 cut. Regression suite **7873 passed, all gates**.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -6,7 +6,8 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 SquadOps is a multi-agent orchestration framework for software development. It uses a hexagonal architecture (ports & adapters) with dependency injection for testability.
 
-**Framework Version**: 1.6.3 (single-sourced via `importlib.metadata.version("squadops")`)
+**Framework Version**: 1.6.3 (single-sourced from `pyproject.toml`; installed metadata is the
+install-time copy and is used only when no source tree is present — #1089)
 **Python Requirement**: 3.11+ (production runs 3.11); develop and test on **Python 3.12** to match CI
 
 ## Commands
--- a/scripts/maintainer/version_cli.py
+++ b/scripts/maintainer/version_cli.py
@@ -27,18 +27,17 @@ INSTANCES_PATH = REPO_ROOT / "agents" / "instances" / "instances.yaml"
 
 
 def get_framework_version() -> str:
-    """Get framework version from squadops package."""
-    try:
-        # Try importing directly
-        sys.path.insert(0, str(REPO_ROOT / "src"))
-        from squadops import __version__
+    """Read the version from ``pyproject.toml`` — the file this tool edits.
 
-        return __version__
-    except ImportError:
-        # Fall back to parsing pyproject.toml
-        content = PYPROJECT_PATH.read_text()
-        match = re.search(r'version\s*=\s*"([^"]+)"', content)
-        return match.group(1) if match else "unknown"
+    #1089: this used to import ``squadops.__version__``, which on an editable install
+    is a copy made at install time. Bumping 1.6.2 to 1.6.3 therefore announced
+    "Version bumped: 1.4.0 -> 1.6.3" — the edit was right and the claim about the
+    prior state was four minor releases stale. A tool whose output is a correctness
+    claim about a file must read that file.
+    """
+    content = PYPROJECT_PATH.read_text()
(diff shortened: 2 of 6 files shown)


jasonLadd (commented):
reviewed

PR #1093, opened 2026-08-25
docs(plan): 1.6.4 rev 3 — the repair does not act on the diagnosis (record + ROADMAP corrected)

jasonLadd:
Every item is something the 1.6.3 repeatability set **observed on a frozen deploy**, and the verification set is designed to test whether fixing it worked.

## Revision 3 — the repair does not act on the diagnosis

Rev 2 named #1015 as the headline without dating it: **parts B and C were already on the set's deploy** (`4f631df7`, v1.6.2) and did not prevent the reds. It also said #435's lever stayed silent in roll 5; the run report shows it **terminated the run as `plan_defect`** after the empty-emission refunds. The per-round repair emission timelines name the actual mechanism:

- **Rolls 1 and 4:** the decision names the join handler; the repair emits the create route (roll 1 r2; roll 4 r2–3). → #1015 **part A** via a structured `implicated_files` output + #968's check-against-source slice.
- **Rolls 1 and 5:** 3 of 11 repair rounds emit **zero characters**. The refund works; the signature is unrecorded. → #998.
- **Roll 5:** a repair that **passed the floor** was rejected on a qa fill asserting strings on a `list[Participant]` field, excluded by Fix E, then terminated by #435. filed as **#1094**: fills must agree with the declared element kind, derived from the same `derive_response_shape` the floor uses.

Stop-early is demoted (#414 back to 1.7); P0/P3/P4 restated; round counts become texture. **`docs/ROADMAP.md`'s 1.6.3 entry carried the same misattribution and is corrected in place.** `CHANGELOG.md` §[1.6.3] (line 30), the published Release body, and PR #1091's package say it too — those are for the owner, since they are tagged/published artifacts.

## Revision 2 — the record's attribution was wrong, and the plan is re-derived

Revision 1 took the set record's "2 of 3 rejections were false" at face value. An independent re-read of the stored per-round `test_report.md` files (seven per rejected roll) shows **rolls 1 and 4 failed the frozen #1029 response floor of `vc-probe-api-runs-join` at every round, including the last.** That assertion runs *before* the qa fill slot, so the phantom-table fill cannot cause it; roll 4's final join route returns `participants` as bare strings where the manifest declares `list[Participant]` with `name`. Both rolls' round-0 analyzers named this correctly. The #1087 assertion was a **second** failing probe, never the only one.

So all three rejections are correct, and they share one loss mode: **the join response shape is wrong, the analyzer says so at round 0, and 3–4 repair rounds never land a one-field fix.** The record had inferred "the app was fine" from a boot audit that its own §2 says cannot see response bodies. This PR adds a dated **§6 Correction** to the record so the two documents agree.

## What changed in the pack

- **Headline moves to repair convergence** (#1015 A/B/C, #864's "diagnoses accurately, then re-emits it"), preceded by two reads: which dev-brief path actually rendered the response surface, and why #435's termination lever stayed silent in roll 5 despite identical signatures with non-`none` candidates across rounds 1–3.
- **#1087 + #1079's producer still ship together** as the hash-mover pair, with the corrected yield stated (zero rolls flipped on this set) and one implementation detail added: the frozen harness addresses `entities[0]` = `Participant`, so it must re-pick a root-persisted table or stop compiling.
- **Stop-early is reframed** as an amendment to #435's rule, designed after the read, and stated honestly as a time lever, not a rate lever — rolls 1, 4, 5 would have failed faster, not passed.
- **P0 added**: no roll ends on a response floor its round-0 analyzer named. P1–P3 unchanged; the per-rejection record must now name the failing assertion's origin (spine or fill) and first round.
- **#1089** corrected from "fixed at the cut" to "PR #1092 open". The sweep's leftover 1.6.4 pack (#772, #939, #668, #930, #901, #936) is dispositioned to the 1.7 re-derivation rather than dropped silently.

## Deliberately not decided here

Whether 1.6.4 is the last patch before 1.7 — that gates on the set's result. And **the 1.7 slate should be re-derived before it opens**: it carries 39 open issues from a sweep already shown stale twice.

Refs #1087, #1079, #1094, #1015, #968, #998, #1021, #414, #435, #864, #1092


jasonLadd (commented):
reviewed

Issue #1094, opened 2026-08-25
A qa fill can contradict the element kind the floor pins — a correct repair is then rejected and discarded (roll 5)


jasonLadd:
A qa fill that asserts on a `list[X]` field can contradict the element kind the manifest declares for `X` — and the frozen response floor already pins. Nothing rejects such a fill at the scaffold gate. When the dev then returns the *declared* shape, the floor passes, the fill fails, the repair candidate is rejected, and the loop discards a correct application.

## The instance — 1.6.3 repeatability set, roll 5

`cyc_421d29473f86` / `run_996e340ea0f1`, `full-38`, frozen deploy `5c6c64f7`, config `d4d4f66217d8`.

The manifest declares `Run.participants: list[Participant]` with `Participant{name, normalized}`. The join-probe shell's floor (`expectShape`, #1029) therefore checks each element for `name` and `normalized`. The qa fill in the same test asserts:

```ts
// [scaffold-slot:begin slot-vc-probe-api-runs-join]
expect(body.participants).toContain('sample')      // element kind: string
expect(body.participants).toHaveLength(1)
expect(all(TABLES.Run)).toHaveLength(1)
// [scaffold-slot:end slot-vc-probe-api-runs-join]
```

`toContain('sample')` on a list the manifest declares as objects cannot pass against a correct application. The floor and the fill contradict each other inside one `it`, and the fill wins whenever the app is right, because it runs second.

**What the loop did with that** (artifact timeline, `created_at` UTC):

| time | event |
|---|---|
| 07:02 | round 0 retest: floor fails on the dev's join response (real defect, bare strings) |
| 07:08, 07:25 | rounds 1–2 dev repairs emit **0 characters** (#998; attempts refunded per `a93240dc`) |
| 07:38 | round 3 dev re-emit: **floor passes** — `participants` are now `{name, normalized}` objects. The retest fails only on the fill: `expected [ { name: 'sample', …(1) } ] to include 'sample'` (and the additive `runs-api.test.ts` making the same assumption) |
| — | candidate rejected on a red retest; pf-31 Fix E excludes it from the next workspace |
| 07:49 | retest on the accepted tree: floor fails again |
| 07:51 | #435 terminates the run: `plan_defect: correction terminated — failure signature repeated from round 0` |

The application was correct at 07:38. The framework rejected it because the test it was judged by disagreed with the contract the test was frozen around.

## Why this is a framework defect

[...]

PR #1095, opened 2026-08-25
docs(release): correct the v1.6.3 attribution in CHANGELOG and the release package

jasonLadd:
The v1.6.3 notes say two of the three failures were the framework wrongly rejecting a working application. The stored per-round test reports say otherwise (`docs/plans/1-6-3-repeatability-set-record.md` §6, merged in #1093): all three rejected applications failed the frozen response floor of the join probe at every round; the phantom-table assertion (#1087) was a second failure in two of them, never the only one.

#1093 corrected the record and the ROADMAP. This corrects the remaining copies, **in place and dated**, the same shape as the ROADMAP fix — not a silent rewrite, and not a note under `[Unreleased]` that a reader of the 1.6.3 section never sees:

- `CHANGELOG.md` §[1.6.3]
- `site/content/releases/v1.6.3/package.yaml` — `narrative` regenerated from the corrected CHANGELOG via the script's own `changelog_section()`; the captured cycle evidence is untouched
- `site/content/releases/v1.6.3/index.md` — same paragraph

**Still outstanding after this merges:** the published GitHub Release body for v1.6.3 was generated from the CHANGELOG at tag time and carries the old paragraph. `gh release edit v1.6.3 --notes-file` with the corrected section is the follow-up; it is outward-facing, so it is done deliberately, not by this PR.

Refs #1087, #1094

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,9 +27,13 @@ true-by-ruling.
   first attempt — against **three** framing runs for the pre-1.6.2 green roll. That is 1.6.2's
   success-status single-sourcing (#1067/#1070A), measured rather than asserted.
 
-**Two of the three failures were the framework wrongly rejecting a working application** (#1087,
-filed from the set). That is the release's most actionable finding, and it matters more than the
-rate.
+**The finding that outranks the rate — corrected 2026-08-25 (record §6):** the cut-time reading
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

Issue #1096, opened 2026-08-25
nextjs_ts expander types every entity-typed field as string — the frozen models.ts contradicts the response floor on every roll


jasonLadd:
The Next.js expander types every entity-typed field as `string`. A manifest declaring `Run.participants: list[Participant]` produces a frozen `lib/models.ts` that defines `interface Participant` and then, three lines later, declares `participants: string[]`. The developer is told this file is *authoritative — import from these, never rewrite them*; the #1029 response floor, derived from the same manifest, asserts each element carries `name`. Two scaffold-owned surfaces, one manifest, two answers — and the developer who obeys the frozen one is rejected.

## Evidence — every nextjs_ts roll of the 1.6.3 repeatability set

The frozen `lib/models.ts` seeded by `scaffold.expand`, against the manifest each roll authored:

| roll | manifest `Participant` | manifest `Run.participants` | frozen models.ts `Participant` | frozen models.ts `Run.participants` |
|---|---|---|---|---|
| 1 `cyc_a24e4619844e` | `{name, joined_at}` | `list[Participant]` | `interface Participant { name: string; joined_at: string }` | **`string[]`** |
| 2 `cyc_b04cfe4b6191` (green) | `{name, joinedAt}` | `list[Participant]` | `{ name; joinedAt }` | **`string[]`** |
| 4 `cyc_a38814afc16d` | `{name}` | `list[Participant]` | `{ name }` | **`string[]`** |
| 5 `cyc_421d29473f86` | `{name, normalized}` | `list[Participant]` | `{ name; normalized }` | **`string[]`** |

The developer's prompt for roll 4 (LangFuse generation `05:07:23Z`, task `task-run_6d6b25fe-m001-development.develop`) carries it under the FROZEN FILES surface:

> `lib/models.ts` — defines `Participant(name: string)`, …, `Run(id: string, title: string, datetime: string, meeting_location: string, …, participants: string[], participant_count: number)`

while the response floor rendered for the same manifest says:

> `POST /api/runs/{run_id}/join` returns HTTP 200 with `Run` — … `participants` …; each `participants` element carries `name`

[...]

PR #1097, opened 2026-08-25
docs(plan): 1.6.4 rev 4 — the two reads done; #1096 (frozen models.ts contradicts the floor) is the headline

jasonLadd:
The two reads rev 3 required before any code are done, written to `docs/plans/1-6-4-reads.md`, and the plan is re-derived from them (rev 4).

**Read 1 — repair target lists.** From the runtime-api container's `correction_repair_target` log lines (all 18 rounds still present): on `nextjs_ts` package scoping never matches — the anchors are qa test files with no package in common with `app/` — so the #688 fallback fired 18 of 18 times and the repair target was the **entire application every round**. The join route was always in the list and never distinguished. Rev 3's guess that the list lacked it is withdrawn; #1015-A becomes a *narrowing*.

**Read 2 — which path rendered the dev brief.** The code path is wired and the manifest loaded for all three runs, but whether the response-surface block reached the prompt is unrecoverable: LangFuse caps observation text at 10,000 chars (`telemetry/models.py:134`) and every develop prompt on this stack is longer. The surviving head shows the real cause instead: the frozen `lib/models.ts` — *"authoritative — never rewrite"* in the developer's brief — declares `Run.participants: string[]` where the manifest declares `list[Participant]`, on **every** `nextjs_ts` roll, greens included. `_ts_type` (`stack_nextjs_ts.py:59`) maps every entity reference to `string`. The floor demands objects. Three developers obeyed the frozen file and were rejected; five ignored it and were accepted. **Filed as #1096**; it moves the generator hash and joins the #1087/#1079 hash-mover set.

**Plan changes:** #1096 headline; P0 restated (seeded `models.ts` agrees with the manifest — N=1, no model in the loop), old P0 → P5; sequencing puts the three hash-movers first; the 10k cap recorded as an instrumentation gap.

Refs #1096, #1015, #1087, #1079, #1094, #998

### DIFF
--- a/docs/plans/1-6-4-plan.md
+++ b/docs/plans/1-6-4-plan.md
@@ -1,5 +1,12 @@
 # 1.6.4 — plan
 
+**Revision 4, 2026-08-25.** The two reads §2.1 required are done (`1-6-4-reads.md`), and the
+second found the root cause of the set's dominant loss mode: the `nextjs_ts` expander types
+every entity-typed field as `string`, so the frozen `lib/models.ts` — labelled authoritative
+in the developer's brief — contradicts the response floor on every roll (**#1096**). It is the
+headline now. Read 1 withdrew rev 3's targeting guess: the repair target was the whole
+application every round, the named file present but never distinguished.
+
 **Revision 3, 2026-08-25.** Revision 1 took the set record's attribution of the three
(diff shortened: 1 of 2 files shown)


jasonLadd (commented):
reviewed

PR #1098, opened 2026-08-25
fix: the 1.6.4 hash-movers — #1096 frozen model, #1087 root tables, #1079 json_has producer

jasonLadd:
The three fixes the 1.6.4 plan (rev 4, §2.2) ships together so the generator and contract hashes move once. One commit per fix.

**#1096** — `_ts_type` passes declared entity/shape names through case-preserved, so the frozen `lib/models.ts` types `Run.participants` as `Participant[]` instead of `string[]`. Pinned twice: the function, and the frozen model against the response floor for every list-of-entity field.

**#1087** — `root_persisted_entities()` (manifest seam) derives which entities a correct app stores as rows of their own: those some endpoint returns as a single object, with two widening fallbacks. The `nextjs_ts` store and harness consume it (`TABLES` now carries `RunEvent` only on the reference); the fill merge rejects a fill naming a table the store does not export, deterministically, with the real tables named; the table set rides the `verification_scaffold` input as data; both briefs stop saying "one entry per declared entity". Stack #1's per-entity dicts are the same shape and are left for a follow-up.

**#1079 producer** — success probes derive `json_has` from `derive_response_shape` (the shell's own derivation); rejection/conflict probes carry no shape; a response naming no entity adds nothing. The shared judge (#1084) fails a body missing a derived key and names it.

Regression: **7,897 pass, 0 fail.** The four pins moved as one classified commit (`chore(fixtures)`), per the owner's decision on 2026-08-25:

| pin | what moved | disposition |
|---|---|---|
| scaffold reference fixture | `expanded_tree_hash` only — **no shell changed**, spine and scaffold hashes byte-identical to v7 | `GENERATOR_VERSION` 7 → 8, manifest regenerated, reason recorded in the test |
| reference contract | `json_has` on the three success probes, nothing else | **`ambiguity_removal`** (M0 taxonomy). v9 (`art_4f368ea08799`) kept unchanged as the 1.4 evidence pin; the derived form lands beside it as `contract_v10_json_has_1079.yaml`; a new test asserts v10 − `json_has` == v9, so the classification is falsifiable. No qualification of the 1.4 figure. |

Not in this PR: #1015-A/#968, #998, #1094 (the loop-side pack), the #1021 read.

Refs #1096, #1087, #1079

### DIFF
--- a/src/squadops/capabilities/handlers/cycle/qa_test.py
+++ b/src/squadops/capabilities/handlers/cycle/qa_test.py
@@ -594,7 +594,12 @@ class QATestHandler(_CycleTaskHandler):
         shell_paths = {f.path for f in record.files}
         dropped = sorted(a["name"] for a in artifacts if a["name"] in shell_paths)
         kept = [a for a in artifacts if a["name"] not in shell_paths]
-        merged = merge_fills(list(scaffold_input["files"]), record, fill_emission)
+        merged = merge_fills(
+            list(scaffold_input["files"]),
+            record,
+            fill_emission,
+            store_tables=scaffold_input.get("store_tables"),
+        )
         merged_artifacts = [
             {
                 "name": f.path,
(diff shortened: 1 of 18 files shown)


jasonLadd (commented):
reviewed

PR #1100, opened 2026-08-26
fix(correction): the repair target reaches the defect site — #1015 part A

jasonLadd:
The 1.6.4 plan's headline loop-side fix (rev 4 §2.1). Two commits: the deterministic half, then the analyzer half.

## What the reads found, and this fixes

Roll 4's rounds 2–3 (and roll 1's round 2) named the join handler in the correction decision and emitted the create route. The runtime-api log showed why: **the repair target was the entire application in 18 of 18 rounds.** Two gaps compounded:

1. **On `nextjs_ts` the contract's `endpoint_owners()` was empty** — the pack has no Python `endpoint_defined` criterion to hang ownership on, so the #688 chain from a failing probe to the file that owns its endpoint had no map to walk. The contract now carries ownership **as data** on each route slot (`endpoints: ["METHOD /path"]`, from bend 1's own `_route_groups`); presence-keyed, so the stack #1 contract v10 stays byte-identical (its pin still passes).
2. **The reds failed the join probe *inside the suite*** — a scaffold `app_contract` observation bound to `vc-probe-api-runs-join` — never as an HTTP probe row, and `_failed_probe_ids` read only the rows. It now reads those observations too (structured data; fill/generator-layer observations carry no criterion and indict nothing).

**The narrowing:** when a failing probe resolves to its owning slot, that slot *is* the target and the language-wide fallback is withheld. Drift files and the failed task's own artifacts still ride (pf-21). Roll 4's evidence replayed through the resolver now yields `app/api/runs/[run_id]/join/route.ts` (+ the qa files the #884 veto removes), where before it yielded seven files.

**The analyzer half (part A as filed, with #968's cheapest slice):** `FailureAnalysis.implicated_files` (optional; fragment v1.1.0, manifest regenerated), consulted only when no deterministic site evidence exists, kept only where the failed task's envelope knows the path. An unverifiable claim is dropped with a log line, and its presence does not remove the fallback.

## Verification

- 7,913 regression / full unit tree green; contract v10 and scaffold reference pins unchanged.
- Plan prediction **P5**: every repair round's emission includes the file the decision named.
- Not exercised here: a live cycle. Shakeout 2 follows the rest of the loop pack (#998, #1094).

Refs #1015, #968, #688

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -310,8 +310,42 @@ def _narrowed_or_scoped(
     )
 
 
+def _verified_implicated_files(
+    failure_analysis: dict[str, Any] | None, failed_inputs: dict[str, Any]
+) -> list[str]:
+    """The analyzer's ``implicated_files``, kept only where the workspace agrees (#968).
+
+    The analyzer's file claims are prose-adjacent — #968 counted three false ones in a
+    single roll, one of them carried verbatim into the correction decision. The cheapest
+    mechanical check is whether the named path is a file the failed task's envelope knows
+    about at all: an implementation artifact, one of its own expected artifacts, or a
+    contract-owned slot. A path outside that set is not a defect site the loop can act
+    on, and is dropped with a log line rather than aimed at.
+    """
+    if not isinstance(failure_analysis, dict):
+        return []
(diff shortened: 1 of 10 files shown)


jasonLadd (commented):
reviewed

PR #1101, opened 2026-08-26
feat(evidence): an empty emission says which kind of nothing it was — #998

jasonLadd:
Second loop-side item of the 1.6.4 plan (rev 4 §2.1). One commit.

**What was measured.** Three of the eleven repair rounds in the 1.6.3 set's reds emitted zero characters (roll 1 r3; roll 5 r1–r2). #1053 refunds those rounds, but the evidence could not say *which* nothing they were — and the two shapes have opposite remedies. V7's roll-1 void was the same class: 8,192 tokens, the whole completion budget, 0 characters, and a banked analysis guessing "safety filter, context overload".

**What this does.**
- `emission_integrity.classify_empty_emission(chars, completion_tokens, completion_cap)` → `cap_exhausted` / `empty` / `unextractable`, from facts the LLM call already reports (no stop-reason crosses the port, so tokens against the requested cap is the observable). The marker carries the token facts and the signature; `emission_retry_reason_line` names the remedy per signature, so a budget-exhausted model is not told to "use fences".
- Producers: `develop` and `qa.test` pass `response.completion_tokens` and `chat_kwargs["max_tokens"]`. **The generic base path every repair handler rides emitted no marker at all for an empty response** — that is where roll 5's 0-byte `repair_output.md` came from; it now attaches one, for emptiness only (prose without fences is a document, not a failure).
- The runner collects each repair step's signature; `CorrectionProtocolResult.empty_emission_signatures` and the `CORRECTION_COMPLETED` payload carry it; the executor's refund log line names it.

**Not built:** #998's ask 2 (raise the cap / limit thinking / reshape the re-dispatch prompt) — a design question this makes measurable first.

**Verification:** 7,930 regression / 8,344 full unit tree, green. Plan prediction **P4**: every zero-character repair emission carries a named signature, disclosed on the correction event.

Refs #998, #1053

### DIFF
--- a/adapters/cycles/correction_runner.py
+++ b/adapters/cycles/correction_runner.py
@@ -280,6 +280,15 @@ def _probe_owned_slots(failure_evidence: Any, failed_inputs: dict[str, Any]) -> 
     return slots
 
 
+def _empty_emission_signature(result: Any) -> list[str]:
+    """The #998 signature a repair step's handler put on its ``emission_failure`` marker,
+    as a zero-or-one-element list so the caller can ``extend`` without branching."""
+    marker = (getattr(result, "outputs", None) or {}).get("emission_failure")
+    if isinstance(marker, dict) and marker.get("signature"):
+        return [str(marker["signature"])]
+    return []
+
+
 def _narrowed_or_scoped(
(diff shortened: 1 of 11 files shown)


jasonLadd (commented):
reviewed

PR #1102, opened 2026-08-26
feat(scaffold): a fill contradicting the declared element kind is rejected at the fill gate — #1094

jasonLadd:
Third loop-side item of the 1.6.4 plan (rev 4 §2.1). One commit.

**The instance.** Roll 5 (`cyc_421d29473f86`): the manifest declares `participants: list[Participant]`; the frozen floor asserts each element carries `name`; the qa fill asserts `expect(body.participants).toContain('sample')` — strings. Round 3's repair returned the declared objects, **passed the floor and failed the fill**; the candidate was rejected, Fix E excluded it from the next workspace, and #435 terminated the run as `plan_defect`. A correct application was discarded by a test that disagreed with the contract the test was frozen around.

**What this does.**
- `response_shape.element_kinds()` serializes the floor's per-field element kind; `slot_element_kinds()` (stack-dispatched, like the emitter) derives it per slot **from the same `_Behavior` list the shells render** — one derivation, now three renderings (spine, brief, gate).
- It rides the `verification_scaffold` input as data (beside #1087's `store_tables`); `merge_fills` rejects a fill whose matcher argument or element access contradicts the declared kind, with the declared kind and the fix named. Narrow by design: only a kind readable from the text; kind-neutral assertions (`toHaveLength`, `toEqual([])`, identifiers) never fire.
- The qa brief (appendix v3) says where to read the kind off the floor.

**The false-positive guard, on the real corpus.** Replayed over every banked fill of the 1.6.3 set plus shakeout 1 — 72 slots across nine cycles: **only roll 5's join and leave fills flag.** Rolls 3 and 7 are green with the identical `toContain('sample')` line under `list[string]` and are clean — the rule keys on the declaration, never on the assertion's shape.

**Verification:** 7,949 regression / 8,363 full unit tree, green. Plan prediction **P3**: no repair candidate is rejected on fill assertions the floor contradicts.

Refs #1094, #1029, #1087

### DIFF
--- a/src/squadops/capabilities/handlers/cycle/qa_test.py
+++ b/src/squadops/capabilities/handlers/cycle/qa_test.py
@@ -599,6 +599,7 @@ class QATestHandler(_CycleTaskHandler):
             record,
             fill_emission,
             store_tables=scaffold_input.get("store_tables"),
+            slot_element_kinds=scaffold_input.get("slot_element_kinds"),
         )
         merged_artifacts = [
             {
--- a/src/squadops/capabilities/response_shape.py
+++ b/src/squadops/capabilities/response_shape.py
@@ -34,6 +34,7 @@ kind, which keeps a legitimately-empty list from reading as a shape defect.
 from __future__ import annotations
 
 from dataclasses import dataclass
+from typing import Any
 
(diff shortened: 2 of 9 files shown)


jasonLadd (commented):
reviewed

PR #1103, opened 2026-08-26
fix(verification): per-file criteria stop superseding each other in the final-state ledger — #1021

jasonLadd:
The #1021 read named the mechanism (comment on the issue, with the nine-cycle prediction table); this is the fix. One commit, one tuple.

**What was wrong.** `_resolve_final_state` (SIP-0096 §6.5, #379) keyed a result's final state on `(check_id, subject)` — the check *name* and the producing task. A typed check like `acceptance:frontend_compiles` is one name evaluated once per file, so a develop subtask that compiled three route files recorded three results under one identity; the last superseded the other two, and their criterion ids never reached `criteria_passed`. Neither verified nor failed, listed as unevidenced. Predicted set = "each subtask's compile criteria but the last"; matched by id on rolls 1–8 and shakeout 1.

**The fix.** Identity is `(check_id, subject, criterion_id)`. Pinned behaviours hold: a framework check (`criterion_id=None`) keeps the two-part identity, so a task's `tests_pass` re-run still supersedes; a criterion re-verified in the same task supersedes itself. Distinct criteria stop superseding each other — in both directions: a *failed* compile can no longer be hidden by the next file's passing one, which was a false-green path this ledger had open.

**Verification:** 7,954 regression / 8,368 full unit tree, green. Checkable at N=1 on shakeout 2: a green roll's `criteria_verified` equals `criteria_total` (no `vc-compiles-*` in `criteria_unevidenced`).

Refs #1021, #379

### DIFF
--- a/src/squadops/cycles/verification_integrity.py
+++ b/src/squadops/cycles/verification_integrity.py
@@ -551,14 +551,25 @@ def _resolve_final_state(results: Sequence[CheckResult]) -> list[CheckResult]:
     collapsed: each is independent evidence (the correct default — e.g. distinct
     develop tasks emitting ``tests_pass`` must accumulate, "any subject failed →
     rejected"; only the *same* subject re-verified supersedes).
+
+    **The identity includes the criterion (#1021).** A contract-bound typed check is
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

PR #1108, opened 2026-08-26
docs(plan): 1.6.5 rev 2 — qa-only completion budget (eve 12,288) alongside fills-first

jasonLadd:
Answers the cut question *is a higher limit needed?* with the distribution rather than two rolls.

**All ten `qa.test` primary emissions in the 1.6.4 set** (eve's log): `4418 4947 5045 5498 5743 6292 7947 7963 8192 8192` — four at the cap or within 3%. Content ~constant (7–15k chars), reasoning the variance; both cap hits would have closed under ~12k. Dev side: 79 emissions, 2 at cap, and its clamp forces decomposition by design — untouched.

**Rev 2 adds item E:** `eve` on `full-38` gets `config_overrides: {max_completion_tokens: 12288}` — the seam is allowlisted (`models.py:184`), rides the plan (`task_plan.py:899`), applies at `base.py:351`. Config, not code; the registry clamp (the V38 comparability pin) stays at 8,192. Cost ≤ ~3 min per emission, only when used. Own prediction **Q5** (no qa primary at 12,288) plus the full token distribution as texture, so the next cap decision has a number. States what E does not fix: zero-character reasoning exhaustion. Notes the config hash moves.

§2.2 rescoped from *no raise* to *no registry or dev raise*; §4 bundles E with A+B+C on one deploy boundary, attribution preserved (Q0 static, Q5 from counts).

No issue: plan revision (docs only)

### DIFF
--- a/docs/plans/1-6-5-plan.md
+++ b/docs/plans/1-6-5-plan.md
@@ -1,5 +1,12 @@
 # 1.6.5 — plan
 
+**Revision 2, 2026-08-26.** Adds a qa-only completion budget (§2.1 E) on the strength of the
+emission distribution the cut question surfaced: four of ten qa primary emissions in the 1.6.4 set
+sat at the 8,192 cap or within 3% of it, with the content roughly constant and the reasoning the
+variance. Revision 1 deferred the raise until fills-first was measured; the distribution says the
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

PR #1106, opened 2026-08-26
chore(release): v1.6.4 — the self-consistency patch line (+ 1.6.5 plan rev 1)

jasonLadd:
Release cut for **v1.6.4 — the self-consistency patch line**, per the CLAUDE.md procedure: steps 1–5 here (bump via `version_cli.py`, markers, CHANGELOG rotation, ROADMAP entry, SIP sweep — no promotions), step 6 (tag) after merge, step 7 (package capture, preview read first) as a follow-up PR.

Also opens **`docs/plans/1-6-5-plan.md` rev 1** — the ROADMAP names it and the path guard requires it to exist, so it rides this PR rather than a second one that would have to merge inside the release window.

**Nothing else merges between opening this PR and merging it** (CLAUDE.md release cut). Main after the validated deploy `5a697dfa` is docs only (#1104, #1105, this) — zero code drift, and the notes say so.

**What the notes say that the headline alone would not** (record §1.1, the owner's qualification): 8/8 is half mechanism, half luck. The correction loop repaired neither roll that entered it; both recovered by fallback; three of eight qa emissions hit the completion cap. The CHANGELOG, ROADMAP and README all carry that sentence.

**Corrections folded in:** README's v1.6.3 sentence still said "two of three failures were false rejections" — the reading the 1.6.3 record §6 withdrew; replaced with the corrected statement. `[Unreleased]` said "Nothing yet" over nine merged PRs (the #789 class, again).

**1.6.5 plan in one line:** a truncated qa emission must cost the least important content (fills first, #998 ask 2's ordering half), and what survives must be used (the suite runs on the post-self-eval file set — the roll-8 gap, to be filed; the self-eval merges fills through the gate, #947; an own-artifact qa repair can reach a fill via the 1.6.4 resolver chain, #970/#969). Predictions Q0–Q4 plus P1/P3/P5 carried; a non-counting fault-injection arm on `lite` recommended so the loop-side predictions are finally exercised.

**Issue hygiene, not done here (needs your OK):** #1096, #1087 (nextjs_ts half), #1079, #1021, #1015-A, #1094 shipped without `Closes` lines and are still open — plan §5 lists them for closure at the cut.

Regression: 7954 passed, 0 failed, all gates.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018Tj9vAj9yumEzAmBLrZnrX

PR #1114, opened 2026-08-26
ci: PR closure references are checked, not assumed — template + required check

### LINKED ISSUE #1096: nextjs_ts expander types every entity-typed field as string — the frozen models.ts contradicts the response floor on every roll

jasonLadd:
## What

The CLAUDE.md rule *every PR body carries `Closes #N`* had nothing enforcing it; six 1.6.4 fix PRs merged with `**#N**` headings and `Refs #N` (not closures to GitHub) and five issues were still open at the v1.6.4 cut — the second recurrence of #133/#205.

- `.github/PULL_REQUEST_TEMPLATE.md` — What / Closes / Evidence, with the three accepted forms spelled out. Formalizes the human path only: `gh pr create --body-file` bypasses templates.
- `scripts/dev/check_pr_closure.sh` — the check, runnable locally: `gh pr view N --json body -q .body | scripts/dev/check_pr_closure.sh`. Passes on `Closes|Fixes|Resolves #N` where `#N` is an **open issue** (closed / PR / nonexistent all fail, with the reason), or on `Refs #N — remaining: …` / `No issue: …`.
- `.github/workflows/pr-closure.yml` — runs it on `opened, edited, synchronize, reopened` (`edited` so fixing the body re-runs without a push). Read-only token; the body reaches the script through an env var, not interpolated into the shell.
- CLAUDE.md "Close issues from PRs" names the enforcement.

**After merge, one repo setting is yours:** mark `closing reference present` as a **required status check** on `main`. Until then the check is advice.

## Closes

Closes #1113

## Evidence

Local run against the script, seven bodies: `Closes #1113` → ok; `Refs #1087 — remaining: …` → ok; `No issue: release cut` → ok; `**#1096** — fixed it` → FAIL (no reference); `Closes #1096` → FAIL (already closed); `Closes #1106` → FAIL (is a PR); `Fixes #999999` → FAIL (no such issue). This PR's own body is the first live run of the workflow.


jasonLadd (commented):
reviewed

Issue #1111, opened 2026-08-26
qa test_report.md is re-stored in its failed form at run end after a passing retest (last-writer-wins)


jasonLadd:
## What happens

When a `qa.test` task fails, enters correction, and its retest passes, the task's own `test_report.md` is re-stored **at run end in its failed form** (last-writer-wins) while the passing report lives only under the retest task's id. The first artifact a triage reads for the qa task says the tests failed, on a run whose verdict is `accepted`.

## Evidence

1.6.4 verification set, both corrected rolls — roll 6 (`cyc_bbc9da03c3c1`) and roll 8 (`cyc_8d262b968f58`): the qa task's `test_report.md` in the artifact vault is the failed report; the passing one is under the retest task id. Record: `docs/plans/1-6-4-verification-set-record.md` §4 ("Record inconsistency, minor").

Cosmetic in outcome — the ledger and verdict are right — but it is the exact artifact the next triage opens first, and the 1.6.3 record's attribution error (§6 there) came from reading the wrong report.

## Fix direction

Either do not re-store the failed report at run end once a retest for the same task has passed, or store it under a name that says which attempt it was. Planned as 1.6.5 §2.3.

jasonLadd:
**1.6.5 evidence: this misdirects the correction loop, not only the record.** FastAPI+React roll 1, `cyc_b9296c255dfc` / `run_daf53fbdcc80`: at 03:15:08Z `retest-…-00-qa.test` PASSED (the backend `None → ""` repair landed) and in the same second `task-…-m004-qa.test` re-stored its FAILED `test_report.md` **and** `typed_check_evaluation_task_4.json`. The next analysis (`art_6f7eba3677f1`, 03:20:51Z) leads with the already-fixed backend 500 (`vc-probe-runs status 500`), the decision narrows the repair to `backend/routes.py`, and round 1 is spent re-fixing a fixed file while the live failure was the frontend suite. Two rounds later the roll was rejected. The last-writer-wins re-store is what the analyzer read.

PR #1115, opened 2026-08-26
fix(qa): 1.6.5 A+B+C+E — fills first, suite runs on what is stored, self-eval merges fills, qa-only budget

### LINKED ISSUE #947: qa self-eval pass is fill-blind: emits a whole file the scaffold guard discards, 68% of the qa task's wall clock

### LINKED ISSUE #998: 8192-token completion with zero extractable chars: thinking-cap exhaustion is undetectable in emission evidence

jasonLadd:
## What

The 1.6.5 plan's §4 step 2 — items **A + B + C + E** on one deploy boundary (`docs/plans/1-6-5-plan.md` rev 3, §2.1). One commit per item plus the CHANGELOG entry.

**One mechanism, three faces** (record §3): the qa author's ~17k-token prompt asks for eight fills and an additive suite; three of eight rolls hit the 8,192-token completion cap, and what happened next was recovered by fallbacks, never by the path that should have.

- **A — fills first** (appendix v4). The brief now states emission *order*, not just priority. A cut at the cap lands on the additive file (#1082 catches it, the self-eval re-emits it), never on a fill. Roll 6 lost all eight fills to an additive file written first.
- **B — the suite runs on what the task stores** (`#1109`). `_suite_files(artifacts)` derives the execution set at the moment of the run instead of the pre-self-eval `extracted`. Roll 8 failed on a truncated file whose complete replacement was already in `artifacts`; the retest passed on the same hash.
- **C — the self-eval merges fills** (`#947`). Fill-mode addendum to the self-eval prompt from a managed asset (`request.qa_test_self_eval_fill_appendix`, v1) naming the unfilled slots from the merge record; `apply_followup_fills` folds the followup's fills per slot (missing/rejected take the fill, filled keeps its own and the re-emission is recorded) and they pass the **same** merge gate (#1087, #1094). Before: extracted as files named `slot-…` and discarded by the shell guard.
- **E — qa-only completion budget.** `full-38` / `eve`: `config_overrides: {max_completion_tokens: 12288}` through the allowlisted seam (`models.py:184` → `task_plan.py:899` → `base.py:351`). Registry clamp and dev budget untouched. Moves `resolved_config_hash` — the 1.6.5 set pre-registers the new one.

**Not built, stated:** raising the registry clamp or the dev budget (§2.2); the zero-character reasoning-exhaustion class (a larger budget makes those longer, not fewer — #998's signatures and A–C handle them).

**Deploy note:** `config/` is baked into the runtime-api image (`src/squadops/api/runtime/Dockerfile:49`), so E needs the rebuild A–C need anyway. Verify *loaded* in-container: `apply_followup_fills`, `_suite_files`, the v4 appendix, and eve's override on the resolved profile.

## Closes

Closes #1109
Closes #947
Closes #998

## Evidence

- **Replay, roll 6 (`cyc_bbc9da03c3c1`), real banked artifacts:** primary `__tests__/runs.test.ts` → 0 fills, 8 missing; the eight banked `slot-*` self-eval fills folded and merged through the gate with the roll's own manifest (`store_tables=['Run']`, element kinds) → **8/8 filled**. Where the handler had discarded all eight.
- **Roll 8 (`cyc_8d262b968f58`):** the truncated bytes were never banked — the run-end re-store (#1111) wrote the task's final state, so both `api.test.ts` artifacts are the same complete 11,966-char file. B's mechanism is pinned at handle() level instead: the primary's truncated additive file is replaced by the self-eval's, and the suite call receives byte-for-byte what is stored.
- **handle()-level tests** (`test_fill_mode_transport.py`): the roll-6 shape (0 fills → 8 via self-eval → suite runs once on merged shells; no `slot-*` pseudo-artifact; the addendum rendered with 8 missing slots), the roll-8 shape (what ran == what stored, fills untouched), a re-emitted filled slot does not regress, the addendum renders from the real asset.
- `apply_followup_fills` unit tests (4), the appendix order rule, the profile override (eve only; `full` untouched).
- **Regression: 7,970 passed, 0 failed** (script exit 0, venv active, no pipe).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,27 @@ All notable changes to SquadOps are recorded here. Format loosely follows
 
 ## [Unreleased]
 
-Nothing yet.
+### 1.6.5 line — the qa emission under the completion cap (plan: `docs/plans/1-6-5-plan.md`)
+
+- **A — fills first** (#998 ask 2, ordering half). The qa fill-mode brief (appendix v4) states
+  the emission order: every fill block, then any additive file. A cut at the completion cap now
+  lands on the additive file — which #1082 detects and the self-eval re-emits — never on a fill.
+  Roll 6 of the 1.6.4 set lost all eight fills to an additive file written first.
+- **B — the suite runs on what the task stores** (#1109). `qa.test` derived its suite-execution
+  set before the self-eval loop, so a self-eval re-emission that fixed a blocking typed check was
+  stored but never run; roll 8 failed on a truncated file whose replacement was already in hand,
+  and a correction round was spent rediscovering it. The set is now derived from the artifacts at
+  the moment of the run.
+- **C — the self-eval merges fills** (#947). Under fill mode the self-eval prompt carries a
+  fill-mode addendum naming the slots still unfilled (from the merge record), and its fills fold
+  into the primary emission per slot — a missing or rejected slot takes the followup fill, an
+  already-filled slot keeps its fill and the re-emission is recorded — then pass through the
+  same merge gate (#1087 phantom tables, #1094 element kinds). Replayed on roll 6's banked
+  self-eval fills: 8/8 merged where the handler had discarded all eight.
+- **E — a qa-only completion budget** (#998 ask 2, budget half). `full-38`'s `eve` entry carries
+  `config_overrides: {max_completion_tokens: 12288}`; four of ten qa primary emissions in the
+  1.6.4 set sat at or within 3% of the 8,192 cap. The registry clamp (the V38 pin) and the dev
+  budget are untouched. Changes `resolved_config_hash`; measured by prediction Q5.
 
 ## [1.6.4] — 2026-08-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ All notable changes to SquadOps are recorded here. Format loosely follows
 - **A — fills first** (#998 ask 2, ordering half). The qa fill-mode brief (appendix v4) states
   the emission order: every fill block, then any additive file. A cut at the completion cap now
   lands on the additive file — which #1082 detects and the self-eval re-emits — never on a fill.
-  Roll 6 of the 1.6.4 set lost all eight fills to an additive file written first.
+  Roll 6 of the previous verification set (`docs/plans/1-6-4-verification-set-record.md`) lost all eight fills to an additive file written first.
 - **B — the suite runs on what the task stores** (#1109). `qa.test` derived its suite-execution
   set before the self-eval loop, so a self-eval re-emission that fixed a blocking typed check was
   stored but never run; roll 8 failed on a truncated file whose replacement was already in hand,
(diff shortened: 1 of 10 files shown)


jasonLadd (commented):
reviewed

PR #1116, opened 2026-08-26
fix(correction): 1.6.5 D — an own-artifact qa repair can reach a fill (#970, #969)

jasonLadd:
## What

The 1.6.5 plan's item **D** (`docs/plans/1-6-5-plan.md` rev 3 §2.1): an own-artifact qa repair can reach a fill. Three commits — the seam, the runner, the handler.

**The defect** (#970, observed live on roll 6 of the previous set): under fill mode the shells are merge products and never in `expected_artifacts`, so the own-artifact branch aimed the repair at the plan's declared file (`__tests__/runs.test.ts`) while every shell rendered "no fill received". #969 is what the repair then wrote: 0 fills, one whole file behind a path fence, blind to the fill protocol and the in-process execution model.

**The fix, in the shape the plan named — the 1.6.4 resolver's own pattern, not a second resolver:**
- **Targeting** (`correction_runner.py`): the target is the shell of each fill-layer observation in the scaffold evidence — the same structured data #1015 reads app-contract observations from. `_qa_scaffold_repair_inputs` threads the scaffold plus the failed task's **current** merged shells and the failing slots onto the repair envelope, presence-keyed (qa-role step, scaffold-bound task); everything else is byte-identical to before.
- **Brief** (`fill_mode_brief.py`, #969): one composition seam. `qa.test`, its self-eval and `qa.test_repair` now render the same FILL MODE section from the same asset; the repair adds `request.qa_test_repair_fill_appendix` (v1) naming the failed slots with the runner's reason. `cycle_repair_task` v5 carries the section.
- **Emission** (`QATestRepairHandler`): fill blocks are typed `fill` (not files named `slot-…`); `handle()` recovers every other slot's fill from the current shells (`recover_fills`, pinned by the round-trip property — recover then re-merge reproduces the shell byte for byte), folds the repair's fills in (`replace_filled`), passes the **same** merge gate (#1087, #1094), and emits only the touched shell at its shell path, so `overlay_artifacts` supersedes the failed shell by name and the retest runs it. Whole-shell rewrites dropped and recorded; `outputs["fill_merge"]` banks applied/rejected/dropped.

**Not exercised live** — this is the third safety net; the 1.6.4 set entered the loop twice and neither round reached a dev or fill repair. Prediction **Q4** is where it gets measured, and the plan's non-counting `lite` arm is the honest way to make it fire.

## Closes

Closes #970
Closes #969

## Evidence

- `test_qa_repair_fill_path.py`: one slot replaced and the touched shell alone emitted; a whole-shell rewrite dropped and recorded; a phantom-table repair fill rejected at the gate with the real table named; no-fill and no-current-shells paths say so; fill/file typing; `handle()` renders `qa_test_fill_mode_appendix` → `qa_test_repair_fill_appendix` → `cycle_repair_task` with the failed slot and reason in the addendum, and lands the merged shell in outputs; the addendum renders from the real asset.
- `test_correction_runner.py::TestQaRepairReachesFills`: fill observations read per (file, slot); own-artifact target = the failing slot's shell; without the scaffold the declared file as before; an app-contract failure never takes the branch; the envelope carries `current_files` (shell paths only) and `repair_slots`; presence-keyed (foreign role / no scaffold → nothing).
- `test_verification_scaffold_fill.py::TestRecoverFills`: round trip byte for byte across filled / not-applicable / missing; a repair may replace a filled slot, a self-eval may not.
- Regression after the last edit: **7,994 passed, 0 failed**, script exit 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,17 @@ All notable changes to SquadOps are recorded here. Format loosely follows
   `config_overrides: {max_completion_tokens: 12288}`; four of ten qa primary emissions in that
   set sat at or within 3% of the 8,192 cap. The registry clamp (the V38 pin) and the dev
   budget are untouched. Changes `resolved_config_hash`; measured by prediction Q5.
+- **D — an own-artifact qa repair can reach a fill** (#970, with #969's brief). Under fill mode
+  the shells are merge products, never in `expected_artifacts`, so the own-artifact repair aimed
+  at the plan's declared file and a failing fill was structurally unreachable. Now: the target is
+  the failing slot's shell, read from the scaffold evidence's fill-layer observations; the repair
+  authors under the **same** fill-mode brief as `qa.test` (one composition seam,
+  `fill_mode_brief`, plus a repair addendum naming the failed slots with the runner's reason);
+  it emits fill blocks, and the handler recovers every other slot's fill from the task's current
+  shells, folds the repair's fills in, passes the same merge gate (#1087, #1094) and emits the
+  merged shell at the shell path, so the patch overlay supersedes the failed one and the retest
+  runs it. Round-trip pinned: recovering the fills from a merged shell and re-merging reproduces
+  it byte for byte. Measured by prediction Q4.
 
 ## [1.6.4] — 2026-08-26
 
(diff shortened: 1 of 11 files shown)


jasonLadd (commented):
reviewed

PR #1117, opened 2026-08-26
tooling: promote the verification-set driver — parameters as data, stack as the cycle's fact, UTC log window

jasonLadd:
## What

The verification-set driver, promoted from the session scratchpad that drove the 1.6.3 and 1.6.4 sets (sixteen counted rolls, three shakeouts) into `scripts/dev/verification_set_driver.py` — with the assumptions the two set records logged against it moved out of the code.

- **Parameters as data.** Project, squad, request profile, overrides, frozen deploy identity, gate constant, launch notes and the "loaded, not built" checks come from a set-config YAML in `docs/plans/verification-sets/` — the pre-registration's §1 table, committed beside the pre-registration. Gate notes are copied verbatim (no substitution); `{roll}`/`{n}` are the only substitutions anywhere.
- **The stack is the cycle's fact.** Derived from the request profile's defaults plus overrides; the P0 seeded-tree check dispatches on it — `nextjs_ts` (models.ts element types, root tables, harness table) and `fullstack_fastapi_react` (models.py element types asserted; the per-entity store *recorded* as #1087's open half, `stores_beyond_roots`). An unregistered stack is refused loudly, never passed.
- **UTC log window.** `log_since()` emits RFC 3339 with `Z`; the scratchpad passed a zone-less timestamp and the 1.6.4 P4/P5 windows came back empty (record §4).
- **A failed loaded-check says why.** Against the live deploy right now it reports `ImportError: cannot import name 'apply_followup_fills'` / `No module named fill_mode_brief` — the images predate #1115 — instead of an empty string.
- Two configs committed: `1-6-5-nextjs.yaml` (pins empty until pre-registration) and `1-6-5-stack1-regression.yaml` (no overrides: `validated-fullstack`'s defaults *are* stack #1).

Three subcommands: `preflight`, `shakeout` (non-counting: records identity), `roll --roll N` (counting: asserts image ids, pins HEAD on roll 1, asserts the config hash). Still one roll per invocation — the counted/void/reset decision stays at the roll boundary with a reader.

## Closes

No issue: tooling promotion the 1.6.5 plan §4 and the 1.6.4 record §4 called for

## Evidence

- `tests/unit/cycles/test_verification_set_driver.py` (19): UTC window (naive, aware-ET, RFC 3339); stack derivation (override wins, profile default, refusal); notes not formatted beyond `{roll}`/`{n}`; set-config validation (missing key named, unknown service refused, braces survive); **both committed configs load and derive their stack**; P0 per stack — unregistered stack refused, no manifest refused, nextjs seeded tree holds, the #1096 shape (`string[]` under `list[Participant]`) falsifies it, stack #1 holds with `stores_beyond_roots == ['participant']` recorded, a `list[str]` tamper falsifies it; run-row parsing.
- Dry-run against the live deploy, clean tree: both configs preflight clean; stacks `nextjs_ts` / `fullstack_fastapi_react`; the stack #1 loaded-check returns `True list[Participant]` in-container; the nextjs loaded-checks report the ImportErrors above (expected — no rebuild yet).
- Regression after the last edit: 7,989 passed, 0 failed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,14 @@ All notable changes to SquadOps are recorded here. Format loosely follows
   merged shell at the shell path, so the patch overlay supersedes the failed one and the retest
   runs it. Round-trip pinned: recovering the fills from a merged shell and re-merging reproduces
   it byte for byte. Measured by prediction Q4.
+- **Tooling — the verification-set driver is promoted** (`scripts/dev/verification_set_driver.py`).
+  The scratchpad copies that drove sixteen counted rolls carried the stack, the deploy pins and
+  the gate constant as code constants and were hand-edited per set; the promoted driver reads
+  every fixed parameter from a set-config YAML (`docs/plans/verification-sets/`, the
+  pre-registration's §1 as data), derives the stack from the request profile plus overrides,
+  dispatches the P0 seeded-tree check per stack (an unregistered stack is refused, not passed),
+  and reads the runtime log window with an explicit UTC zone — the `docker logs --since` defect
+  the previous set's record logged against the instrument.
 
 ## [1.6.4] — 2026-08-26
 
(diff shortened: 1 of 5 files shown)


jasonLadd (commented):
reviewed

jasonLadd (commented):
reviewed

PR #1118, opened 2026-08-26
fix(scaffold): the success-status default has one home — the skeleton pins what the contract asserts (#772)

### LINKED ISSUE #772: contract/scaffold disagree on the default success status — a POST-to-collection with no declared success_status yields an unwinnable contract

jasonLadd:
## What

**#772**: the contract deriver asserts 201 for an undeclared collection POST while the stack #1 skeleton omits `status_code=` and FastAPI answers 200 — a contract no correct app can win. Gate-mitigated since 08-10 (`PROOF_STATUS_DECLARED` rejects the omission at framing) and never fixed; on stack #1, whose regression arm is next, that is a framing re-roll every time an author forgets the line.

The rule had **seven homes**: `scaffold_contract.py:470/548/608` (`or 201` / `or 200` ×2), the skeleton decorator (omission), `framing_consistency._enforced_success_status` (mirror), `verification_scaffold_gate._allowed_statuses`, `stack_nextjs_ts.py:541` (route stub told the dev "respond 200" on an undeclared collection POST the contract probes at 201), and the nextjs GET shells. `manifest_gates.derived_success_status`'s own docstring said adding a copy to police the others "would be the same defect one layer up".

**Now:** `capabilities/success_status.py` is the one home — a leaf module (`scaffold` imports the stack modules at import time and they need the rule, so neither can own it without a cycle); `scaffold` re-exports it and `manifest_gates` keeps its name. `derived_success_status()` / `success_status_for()`: declared wins, else collection POST 201 / child POST 200, else HTTP's 200. Every site calls it. The skeleton decorator pins the **derived** status too, so the seeded join route reads `status_code=200` — the old pin ("stays bare") updated with the reason.

**Fixtures:** unchanged — the reference manifests declare their statuses; contract v9/v10 and the nextjs scaffold hash hold (the reference suites pass untouched).

## Closes

Closes #772

## Evidence

- `test_success_status_seam.py` (11): the seam's cases (declared / collection / child / GET / DELETE); **the #772 property end to end on stack #1** — the reference manifest with the collection POST's status removed: contract probe expects 201 **and** the seeded decorator carries `status_code=201`; the nextjs stub tells the dev "respond 201" for the same case; **a structural guard** that greps `src/` for `success_status or 20x` / `201 if … else 200` outside the seam and fails on any hit.
- Regression after the last edit: see below; the CI skeleton gate boots the changed stack #1 skeleton.

Regression after the last edit: **7981 passed, 0 failed**, script exit 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,14 @@ All notable changes to SquadOps are recorded here. Format loosely follows
   `config_overrides: {max_completion_tokens: 12288}`; four of ten qa primary emissions in that
   set sat at or within 3% of the 8,192 cap. The registry clamp (the V38 pin) and the dev
   budget are untouched. Changes `resolved_config_hash`; measured by prediction Q5.
+- **#772 — the success-status default has one home.** The contract deriver asserted 201 for an
+  undeclared collection POST while the stack #1 skeleton omitted `status_code=` and FastAPI
+  answered 200 — an unwinnable contract, gate-mitigated since 08-10 and never fixed. The rule
+  (`capabilities/success_status.py`: declared wins, else collection POST 201 / child POST 200,
+  else HTTP's 200) had seven homes — three deriver sites, the skeleton, the framing mirror, the
+  scaffold gate's allowed set, the Next.js route stub; every one now calls the seam, the
+  decorator pins the derived status too, and a structural test fails if a copy returns.
+  Reference fixtures unchanged (they declare their statuses).
 - **D — an own-artifact qa repair can reach a fill** (#970, with #969's brief). Under fill mode
   the shells are merge products, never in `expected_artifacts`, so the own-artifact repair aimed
   at the plan's declared file and a failing fill was structurally unreachable. Now: the target is
--- a/src/squadops/capabilities/scaffold.py
+++ b/src/squadops/capabilities/scaffold.py
@@ -42,6 +42,10 @@ from squadops.capabilities.stack_nextjs_ts import expand_nextjs_ts as _expand_ne
 from squadops.capabilities.stack_nextjs_ts import (
     fill_slots_nextjs_ts as _fill_slots_nextjs_ts,
 )
+from squadops.capabilities.success_status import (  # noqa: F401 — re-exported: the seam's home
+    derived_success_status,
+    success_status_for,
+)
 
 # Schema v1 is frozen (SIP-0099 phase 99.1): the shape below is the canonical
 # interface-manifest contract the fullstack_fastapi_react expander was proven against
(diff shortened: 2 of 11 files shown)


jasonLadd (commented):
reviewed

PR #1121, opened 2026-08-26
fix(correction): the failed task's own artifact never narrows the repair target (#1120)

jasonLadd:
## What

Found by the 1.6.5 **stack #1 regression shakeout** (`cyc_3cde35fa5204`) — the first cycle on `fullstack_fastapi_react` since 08-09 — on its second identical correction round. A free-authored frontend suite (`frontend/src/__tests__/runs.test.jsx`, 5/8 failing on DOM expectations the views do not meet) fails with **no HTTP probe and no drift**, so the resolver takes the analyzer half of #1015-A (v1.6.4, #1100). The analyzer honestly implicates the suite file itself; `_verified_implicated_files` keeps it (it is the task's own expected artifact); `_narrowed_or_scoped` treats it as a defect site and withholds the language-wide surface; the #884 veto then strips the qa-owned file from the dev-role target — **empty target**, empty emission, refunded round, byte-identical next round (runtime-api log lines in #1120).

**Fix** (`correction_runner.py`, the analyzer-half block): the failed task's own artifacts are never a *narrowing* site — they already ride via `failed_artifacts` and the veto decides who may touch them. They are excluded from `analysis_files`; when nothing remains, the package-scoped surface applies exactly as before v1.6.4 (the frontend views enter the target), and the own-artifact-only case is logged for the locus classifier. A foreign implicated file narrows as before.

## Closes

Closes #1120

## Evidence

- `TestOwnArtifactNeverNarrows` (3): the replay shape — own artifact rides, views enter, backend stays out (package scoping), and **the dev-role target after the veto is non-empty**; a foreign implicated file still narrows; own + foreign → narrows on the foreign file only. The existing #1015-A analyzer tests are untouched and pass.
- Regression after the last edit: **8027 passed, 0 failed** (the first run tripped the CHANGELOG rotation guard on a version string in the entry — reworded, re-run), script exit 0. (CI will queue until the GitHub Actions incident clears.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,13 @@ All notable changes to SquadOps are recorded here. Format loosely follows
   scaffold gate's allowed set, the Next.js route stub; every one now calls the seam, the
   decorator pins the derived status too, and a structural test fails if a copy returns.
   Reference fixtures unchanged (they declare their statuses).
+- **#1120 — a qa-side failure no longer empties the dev repair target.** The analyzer half of
+  #1015-A let the failed task's *own* artifact (a free-authored frontend suite the analyzer
+  honestly implicated) act as a narrowing site: the language-wide surface was withheld, the #884
+  veto removed the qa-owned file from the dev-role target, and every round was dispatched with
+  nothing to produce and refunded — found on the first stack #1 cycle since 08-09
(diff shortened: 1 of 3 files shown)


jasonLadd (commented):
reviewed

Issue #1127, opened 2026-08-27
Frozen `frontend/src/test-setup.js` registers no `afterEach(cleanup)` under vitest `globals:false`, so RTL never unmounts between tests — any suite with `render()` in two tests fails "Found multiple elements"


jasonLadd:
## What happens

The scaffold-owned frontend test harness for `fullstack_fastapi_react` (`vite.config.js` `test: { environment: 'jsdom', setupFiles: ['src/test-setup.js'] }`, #627) does not set `globals: true`, and the frozen `src/test-setup.js` only registers jest-dom matchers. React Testing Library registers its automatic `cleanup` only when a **global** `afterEach` exists, so under vitest's default `globals: false` nothing unmounts between tests. Any free-authored suite that calls `render()` in more than one `it` fails with `TestingLibraryElementError: Found multiple elements by: [data-testid="…"]` unless the author adds `afterEach(cleanup)` themselves.

## Where

`src/squadops/capabilities/scaffold.py:1622` (the frozen `test-setup.js` source, in the fullstack_fastapi_react expander).

## Evidence — 1.6.5 FastAPI+React roll 1, `cyc_b9296c255dfc` / `run_daf53fbdcc80`

| stored `frontend/src/__tests__/runs.test.jsx` | `render()` calls | `cleanup` | result |
|---|---|---|---|
| 03:19Z `task-…-m005-qa.test` | 3 | none | red — multiple `join-name-input` |
| 03:26Z re-dispatched `qa.test` | 3 | `afterEach(cleanup)` | **green 3/3** (then discarded by the self-mocking check) |
| 03:34Z `repair-…-02-qa.test_repair` | 4 | none | red — multiple `create-run-form`, `join-name-input` |

The component (`RunDetailView.jsx`) renders `join-name-input` exactly once; the duplicates are un-unmounted trees from earlier tests.

## Fix shape

One line in the frozen setup: `import { cleanup } from '@testing-library/react'; import { afterEach } from 'vitest'; afterEach(cleanup)` (or `globals: true` in the vitest block). Frozen output changes → GENERATOR_VERSION / fixture regen are the owner's call.

jasonLadd:
[...]

PR #1136, opened 2026-08-27
fix(scaffold): an optional entity field with a declared null default freezes nullable (#1125)

### LINKED ISSUE #1125: Scaffold freezes `distance: str = None` for a manifest field declared `required: false, default: null` — non-nullable annotation with a None default; deterministic 500 on POST /runs on fullstack_fastapi_react (5 of 6 rolls in the 1.6.5 set)
## What happens

When the interface manifest declares an optional field as `{ name: distance, type: string, required: false, default: null }`, the scaffold's model emitter freezes it as `distance: str = None` — a **non-nullable annotation with a `None` default**. The request shape for the same field is emitted correctly as `str | None = None`, so the route receives `None` and forwards it into `Run(...)` → pydantic v2 `string_type` (`input_value=None`) → **500 on POST /runs** → every dependent probe (join, join-duplicate, leave) fails on an unresolved `{run_id}`. Deterministic: it is decided at framing by whether the manifest author writes the `default` key.

## Where

`src/squadops/capabilities/scaffold.py:1332-1340` (`_model_source`): `elif f.has_default: f"{f.name}: {ann} = {f.default!r}"` fires because `has_default = "default" in f` (`:486`) is true for `default: null`; the nullable `else` branch (`{ann} | None = None`) only fires when no default key is present at all. Request shapes at `:1347` are `str | None = None`. Python-only — `stack_nextjs_ts._ts_type` renders `distance?: string` and nothing validates at runtime.

## Evidence — 1.6.5 FastAPI+React set, frozen deploy 7ebdb00e, perfect separation

| manifest wrote | cycles | frozen `backend/models.py` | round-0 outcome |
|---|---|---|---|
| `required: false, default: null` | rolls 1, 2, 4, 5, 6 — `cyc_b9296c255dfc` `cyc_f84649c68646` `cyc_5ef83cc6fc2b` `cyc_0e7bc622169a` `cyc_d43c644e52d0` | `distance: str = None` (roll 5 also `capacity: int = None`) | pydantic ValidationError, 500 on POST /runs — **5 of 5** |
| `required: false` (no key) | roll 3 `cyc_184b3a1d194e`, shakeouts `cyc_3cde35fa5204` `cyc_549c040d985b` | `distance: str \| None = None` | none — **0 of 3** |

Rolls 2 and 4 recovered by a one-line dev repair (`payload.distance or ""`) and were accepted; rolls 5 and 6 were rejected after the repair was refused (see the refused-patch issue); roll 1 recovered it and then died on the frontend suite. Every round-0 `failure_analysis.md` in those five rolls names this line. Per-round reports: `data/artifacts/group_run/<cycle>/<impl run>/art_*/test_report.md`; driver records `var/verification_sets/1-6-5-fastapi-react/roll-0N-*.json`.

## Fix shape

One branch: when `f.default is None` (or the field is not required), emit `{ann} | None = None`. A generator test with `required: false, default: null` — `tests/unit/capabilities/test_scaffold.py:68` covers `has_default` only for the list case. This changes frozen output, so GENERATOR_VERSION and the fixture regeneration are the owner's call (pins are proposals).

jasonLadd:
## Summary

**1.6.6 item A (#1125).** A manifest field declared `required: false, default: null` froze as `distance: str = None` — a non-nullable annotation with a `None` default — because the declared null set `has_default` (`scaffold.py:486`) and the nullable branch only fired when no default key existed (`:1332-1340`). The request shape was correctly `str | None`, so the route forwarded the request's `None` into `Run(...)` and pydantic answered 500 on POST /runs. Five of the six FastAPI+React rolls in the 1.6.5 set opened with that failure; the three manifests without the key were clean.

**The rule (owner's ruling, narrow):** a declared null default and an optional field with no default both freeze `X | None = None`; a non-null default keeps `X = default` (never `None` after construction); a required field stays a bare annotation.

**Stack safety — checked, not asserted:** `_model_source` is reached only from `_expand_fullstack_fastapi_react` (`scaffold.py:1851`); the TS emitter renders `distance?: string` and is untouched. **Every pinned fixture is byte-identical** — the Next.js reference scaffold (`test_verification_scaffold_reference.py`), the stack-#1 contract reference (`test_contract_derivation_reference.py`, the reference manifest writes `required: false` with no default), the stack-#1 hash in `test_stack_nextjs_ts.py`, and the three context goldens. No GENERATOR_VERSION change.

## Tests

- `TestFrozenModelOptionalFields` — parametrized over the four shapes (`default: null`, no default, non-null default, required) asserting the exact emitted line, plus the failure as the app saw it: the emitted module accepts `Run(distance=None, …)`. Against the old emitter: 3 fail / 3 pass (verified by stashing the source change).
- `./scripts/dev/run_regression_tests.sh` — 8039 passed, 1 skipped, exit 0.

Item B (#1127, the harness cleanup) is on its own PR: it moves the stack-#1 contract reference and the context goldens and needs the pin classification cleared first.

## Closes

Closes #1125

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,19 @@ All notable changes to SquadOps are recorded here. Format loosely follows
 
 ## [Unreleased]
 
-Nothing yet.
+### 1.6.6 line — the React arm's round-0 defects (plan: `docs/plans/1-6-6-plan.md`)
+
+- **A — an optional entity field freezes nullable** (#1125). A manifest field declared
+  `required: false, default: null` froze as `distance: str = None` — a non-nullable annotation
+  with a `None` default — because the declared null set `has_default` and the nullable branch
+  only fired when no default key existed; the request shape was correctly `str | None`, so the
+  route forwarded the request's `None` into the entity and pydantic answered 500 on POST /runs.
+  Five of the six FastAPI+React rolls in the previous set opened with that failure; the three
+  manifests that omitted the key were clean. The rule now: a declared null default and an
+  optional field with no default both freeze `X | None = None`; a non-null default keeps
+  `X = default`. Python emitter only — the TS emitter renders `distance?: string` and is
+  untouched; every pinned fixture (the Next.js reference scaffold, the stack-#1 contract
+  reference, the context goldens) is byte-identical.
 
 ## [1.6.5] — 2026-08-27
 
--- a/src/squadops/capabilities/scaffold.py
+++ b/src/squadops/capabilities/scaffold.py
@@ -1334,9 +1334,15 @@ def _model_source(manifest: InterfaceManifest) -> str:
                 lines.append(f"    {f.name}: {ann}")
             elif f.has_default and isinstance(f.default, list):
                 lines.append(f"    {f.name}: {ann} = Field(default_factory=list)")
-            elif f.has_default:
+            elif f.has_default and f.default is not None:
                 lines.append(f"    {f.name}: {ann} = {f.default!r}")
             else:
+                # #1125: a declared ``default: null`` and an optional field with no default
+                # mean the same thing — the value may be absent — and both freeze nullable.
+                # ``default: null`` used to take the branch above and emit ``str = None``:
+                # a non-nullable annotation with a None default, which pydantic v2 rejects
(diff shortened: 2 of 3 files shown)


jasonLadd (commented):
reviewed

PR #1137, opened 2026-08-27
fix(scaffold): the frozen React test harness registers afterEach(cleanup) (#1127)

### LINKED ISSUE #1127: Frozen `frontend/src/test-setup.js` registers no `afterEach(cleanup)` under vitest `globals:false`, so RTL never unmounts between tests — any suite with `render()` in two tests fails "Found multiple elements"

jasonLadd:
## Summary

**1.6.6 item B (#1127).** The frozen `frontend/src/test-setup.js` registers `afterEach(cleanup)`. Under vitest's default `globals:false` Testing Library never auto-registers cleanup (it needs a global `afterEach`), so nothing unmounts between tests and any suite that renders in more than one `it` fails `Found multiple elements by: [data-testid=…]` — 1.6.5 FastAPI+React roll 1 (`cyc_b9296c255dfc`), three suites, the only green one had added the line itself. One frozen file, one line of behaviour; the component renders each testid once.

**Stacked on #1136 (A).** Rebase onto main after A merges; the diff against A is B alone.

## Pins moved, each per its own policy — classification: `reference_defect`

The stack-#1 contract reference records a sha256 for every frozen file, so this one file moves nine tests. Nothing was regenerated in place:

| pin | policy | what this PR does |
|---|---|---|
| `test_contract_derivation_reference.py` | no regeneration hatch — classify per the M0 divergence taxonomy (`docs/plans/1-6-0-authorship-plan.md`), land a new file with a new hash deliberately | adds `contract_v11_harness_cleanup_1127.yaml` beside v9 (history, bytes untouched) and v10 (kept as that form's record); `_CONTRACT` → v11. **Classified `reference_defect`**: the pinned harness was wrong, found by the 1.6.5 set, not by a deriver change. **Retrospective obligation met by statement:** that harness can only *reject* a working application, never pass a broken one, so the 1.4 FAY figure (6/6) carries no qualification — every app that passed under it passes under this one. The falsifiability test now names both classified deltas from v9 (#1079's `json_has`, this one `frozen` entry) and asserts nothing else moved, and that the entry's sha is the emitted file's. |
| `test_stack_nextjs_ts.py` (stack-#1 hash) | "bound to these two hashes" | prefix follows v11 |
| context goldens ×3 (`UPDATE_CONTEXT_GOLDENS=1`) | a deliberate context change regenerates the golden in the same PR | regenerated; the only diff is the `frozen_surface_index` line now naming the file's imports |

**Next.js untouched:** `test_verification_scaffold_reference.py` (the byte-identical Next.js scaffold pin) passes unchanged; `GENERATOR_VERSION` unchanged.

## Tests

- `test_setup_file_unmounts_between_tests` — both imports and the registration must be real statements, not comments.
- `test_the_current_form_differs_from_the_evidence_contract_only_as_classified` — replaces the v10-only check.
- `./scripts/dev/run_regression_tests.sh` — 8040 passed, 1 skipped, exit 0.

## Closes

Closes #1127


jasonLadd (commented):
reviewed

Issue #1131, opened 2026-08-27
Extract `stack_fastapi_react.py` from scaffold.py (the inline stack #1 expander) and add a structural guard banning stack-shaped literals outside `stack_*` — "generic" scaffold is stack #1 (1.7)


jasonLadd:
## What happens

`src/squadops/capabilities/scaffold.py` (2,181 lines) is the stack-neutral manifest model + parser + `ScaffoldStack` registry **and** the entire `fullstack_fastapi_react` expander inline (lines ~1284-1890: `_py_type`, `_model_source`, `_routes_source`, `_errors_source`, `_app_jsx`, `_view_stub`, `_store_source`, the vite harness, `_expand_fullstack_fastapi_react`, `_fill_slots_fullstack_fastapi_react`). `stack_nextjs_ts.py` registers stack #2 from its own module — the #822 comment at `scaffold.py:37` says a second inline expander "would push this file past 2000 lines", which is what stack #1 already does to it. So "generic" is stack #1, and shared checks written while working on one stack are tested only against that stack.

## Evidence that it bleeds

`detect_self_mocking_tests` (commit `1b9b93a9`, 2026-08-16, Next.js fill-mode work) defined "invokes the application" as an `app/api/` import inside the shared `handlers/stub_detection.py` and discarded a green FastAPI+React suite in the 1.6.5 set (see the self-mocking issue). The React-path defects in that set (`str = None` model branch; missing `afterEach(cleanup)`) were never bleed, but they sat unexercised in the "generic" file for six weeks because nothing distinguishes "generic" from "stack #1". Owner's question 2026-08-27: "do I have to continually remind that we need to stay stack aware?" — the answer has to be structural.

## What to do (1.7, stabilization lane)

1. **Pure move:** lift the inline expander into `stack_fastapi_react.py`, register through `ScaffoldStack` exactly as Next.js does. GENERATOR_VERSION unchanged; the frozen fixtures must pass **byte-identical** — that is the whole regression proof for the move.
2. **Structural guard**, the #772 shape: a test that no module under `handlers/` or `cycles/` carries a stack-shaped literal (`app/api/`, `.test.tsx`, `vitest`, `pytest`, `routes.py`, …) unless it is a `stack_*` module or takes the stack as a parameter. It would have failed on `1b9b93a9`.
3. Then the per-stack predicates (`check_stack_for`, `scaffold.py:2092`) become the only home for "what counts as invoking the app / a real test / a harness entry".

[...]

PR #1139, opened 2026-08-27
fix(qa): the self-mocking detector judges through the stack's own definition of invoking the app (#1126)

### LINKED ISSUE #1126: `detect_self_mocking_tests` defines "invokes the application" as an `app/api/` import (the Next.js model) and rejects a correct FastAPI+React suite that renders the real App with fetch stubbed — discarded a green in the 1.6.5 set
## What happens

`detect_self_mocking_tests` (the `no_self_mocking_tests` check, run on the primary qa path and the retest path) decides that a suite "never invokes the application" when it stubs the network seam **and** does not import a route module — where "imports a route module" is literally *an import from a path containing `app/api/`*. That is the Next.js in-process model (#877). A React SPA on `fullstack_fastapi_react` has no `app/api/` modules; its correct integration-style test renders the real `App` (which mounts the views through `<Routes>`) with `fetch` stubbed. The check rejects exactly that suite, and **passes the more self-mocking alternative** — `vi.mock('../api.js')`, which replaces the app's own API module.

## Where

`src/squadops/capabilities/handlers/stub_detection.py:120-185` — `_NETWORK_SEAM_MOCK` (`vi.stubGlobal('fetch', …)`) and `_APP_ROUTE_IMPORT` (`… from '<path>app/api/…'`); `detect_self_mocking_tests` at `:157`. Called from `src/squadops/capabilities/handlers/cycle/qa_test.py:878` (retest) and `:1387` (primary). A shared module with no stack parameter; the per-stack seam `check_stack_for` (`scaffold.py:2092`) exists and is not consulted. Origin: commit `1b9b93a9` (2026-08-16), written during Next.js fill-mode work.

## Evidence — 1.6.5 FastAPI+React roll 1, `cyc_b9296c255dfc` (impl run `run_daf53fbdcc80`)

- 03:23:07Z retest-01 failed → the executor re-dispatched `task-…-m005-qa.test`.
- 03:26:05Z eve's new `frontend/src/__tests__/runs.test.jsx` (`art_477b87f85956`: imports `App from '../App.jsx'`, `vi.stubGlobal('fetch', vi.fn())`, `afterEach(cleanup)`) ran `exit_code=0 tests_passed=True` 3/3; typed check `tests_pass` passed.
- 03:26:07Z `handler_failed`; `corr-…-02-data.analyze_failure` dispatched the same second; its analysis names `no_self_mocking_tests`.
- Reproduced offline: `detect_self_mocking_tests([{path, content}])` on that stored file → `MOCKS_THE_NETWORK`.
- Round 2 (`qa.test_repair`) re-emitted a suite without cleanup → "Found multiple elements" → attempts exhausted → **rejected**. Without this check the roll most likely accepts at round 1.
- Both accepted FastAPI+React shakeouts (`cyc_3cde35fa5204`, `cyc_549c040d985b`) got past the check with `vi.mock('../api.js')` and no `App` import — the inversion is systematic on this stack.

## Fix shape

Make the "invokes the application" predicate stack-aware through the existing seam (`check_stack_for` / a stack parameter on the detector): Next.js keeps the `app/api/` import rule; `fullstack_fastapi_react` treats a real component/`App` import as invoking the app (and `vi.mock` of the app's own `api.js` as the self-mock). Not a third regex in the shared file. Test: the stored roll-1 suite passes on `fullstack_fastapi_react` and a `vi.mock('../api.js')`-only suite is the one flagged.

jasonLadd:
## Summary

**1.6.6 item C (#1126).** `detect_self_mocking_tests` defined "invokes the application" as *an import of an `app/api/` route module* — the Next.js in-process model (#877) — inside the shared `handlers/stub_detection.py`, with no stack parameter. On the 1.6.5 FastAPI+React set it failed a **green** suite that rendered the real `App` with `fetch` stubbed (roll 1, `art_477b87f85956`, 3/3 passed at 03:26Z → `handler_failed` two seconds later → round 2 → rejected) while passing the `vi.mock('../api.js')` shape both accepted shakeouts used. This is the bleed the owner asked about on 08-27; the fix goes through the seam that already exists rather than widening the regex.

**The rule stays stack-neutral; the definition becomes the stack's.** `AppInvocation` (`src/squadops/capabilities/app_invocation.py`, a leaf module like `success_status` so stack modules and the handler package need not import each other) carries three regex sources: the import that proves the app is invoked, the mock that replaces the subject (unconditional), the mock that replaces the network seam (flagged only when nothing invokes the app). Each `ScaffoldStack` declares one — Next.js in `stack_nextjs_ts.py` (its former patterns, verbatim), stack #1 inline beside its expander until #1131 moves it — and `app_invocation_for(stack)` sits beside `check_stack_for`. The detector takes the declaration and only applies it; both qa call sites resolve it from the cycle's build profile.

On FastAPI+React: importing `App` or a `views/` component invokes the app; a `fetch` stub **or** a mock of the app's own `api.js` under a rendered component is legitimate; mocking a view or `App` module is mocking the subject. (The issue body called the `api.js` mock "the self-mock"; on the evidence it is the *network seam* — the shakeouts mocked it while rendering real views, which is a legitimate component test. Recorded in the tests.)

**An unregistered stack is judged not at all**: the detector returns nothing and the qa handler banks the `no_self_mocking_tests` row with an **empty inspected inventory** (#986's "did not look"), never as a clean pass over files it could not judge.

## Tests

- `TestSelfMockingOnFastapiReact`: the discarded roll-1 suite passes under the React definition **and is still rejected under the Next.js one** (the bug kept as a test, so the two definitions can never collapse); a view import invokes; each seam replacement without a render is flagged; `api.js` mock + rendered view is legitimate; mocking a view/`App` is the subject; unknown stack judges nothing.
- Existing Next.js cases run through that stack's declaration unchanged.
- Handler: the #986 clean-row test hands the handler the Next.js definition; a new sibling asserts the unknown-stack row carries `inspected: []`.
- `./scripts/dev/run_regression_tests.sh` — 8053 passed, 1 skipped, exit 0. Next.js reference scaffold pin and the stack-#1 contract reference are untouched.

## Closes

Closes #1126

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,17 @@ All notable changes to SquadOps are recorded here. Format loosely follows
   failure and sent a repair at a file the loop had already fixed. Evidence artifacts in the
   re-store are now replaced by the passing retest's same-named artifact or dropped; work
   product is untouched and the retest's own suite files stay under the retest id.
+- **C — "invokes the application" is the stack's own definition** (#1126). The
+  self-mocking detector defined it as an `app/api/` import — the Next.js in-process model —
+  inside a shared module, and failed a green FastAPI+React suite that rendered the real `App`
+  with `fetch` stubbed while passing the one that `vi.mock`ed the app's own API client. Each
+  `ScaffoldStack` now declares an `AppInvocation` (the import that proves the app is invoked,
+  the mock that replaces the subject, the mock that replaces the network seam) in a leaf
+  module; the detector only applies it, resolved through `app_invocation_for` beside
+  `check_stack_for`. On React a rendered component or `App` invokes the app and a `fetch`
+  stub or `api.js` mock under it is legitimate; mocking a view or `App` module is mocking
+  the subject. An unregistered stack is judged not at all and its row is banked with an
+  empty inspected inventory, never as clean. The discarded roll-1 suite is the replay test.
 
 ## [1.6.5] — 2026-08-27
 
--- a/src/squadops/capabilities/app_invocation.py
+++ b/src/squadops/capabilities/app_invocation.py
@@ -1,0 +1,59 @@
+"""How a suite on one stack reaches the application — the stack's fact (#1126).
+
+A leaf module, like ``success_status``: the stack modules declare an
+:class:`AppInvocation` and the qa-side detector applies it, and neither may import the
+other's package to do so (``handlers/__init__`` pulls in the handler tree, and the
+stack registry is imported by handlers).
+
+The self-mocking rule is stack-neutral: *a suite that replaces the seam it would reach the
+app through, and never invokes the app, verifies nothing.* What "invokes the app" looks
+like is not. Under Next.js's in-process model (#877) it is an import of an ``app/api/``
+route module; on a React SPA it is rendering a real component or ``App``, and a global
+``fetch`` stub is the *correct* way to isolate the network under it. The first definition
+was written into the shared detector as if it were universal, and on 2026-08-27 it failed
+a green FastAPI+React suite that rendered the real ``App`` with ``fetch`` stubbed — while
+passing the suite that ``vi.mock``ed the app's own API module. Each stack now declares its
+own three patterns on its ``ScaffoldStack``; the detector only applies them.
+"""
+
+from __future__ import annotations
+
+import re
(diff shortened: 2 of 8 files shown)


jasonLadd (commented):
reviewed

PR #1141, opened 2026-08-27
fix(scaffold): an endpoint's request body has one resolver; an entity-typed request gets a body model (#1128)

### LINKED ISSUE #1128: Contract generator emits an empty probe body (`json: {}`) when an endpoint's `request:` names an entity rather than a declared request shape — validation accepts either, so the contract is unsatisfiable by construction (1.6.5 FastAPI+React roll 3)
## What happens

Manifest validation accepts an endpoint whose `request:` names **either** a declared request shape **or an entity** (`scaffold.py:438-440`: "request … is not a declared request_shape or entity"). The contract generator resolves probe bodies **only** through `request_shapes` (`scaffold_contract.py:439` `shapes = {s.name: s for s in manifest.api.request_shapes}`; `:537-541` `child_shape = shapes.get(child.request or ""); child_body = {… for field in (child_shape.required if child_shape else ())}`). An entity-typed request therefore yields `"json": {}` — while the route emitter uses that same entity as the payload class (`scaffold.py:1367`, `payload: Participant` with `name: str` required). The probe expects 201 then 409 from a body the app must 422. **Unsatisfiable by construction**, and the framing gate approved it.

## Evidence — 1.6.5 FastAPI+React roll 3, `cyc_184b3a1d194e`

Manifest (`run_134b26a15f84/art_59fe7628e026/interface_manifest.yaml`): `POST /runs/{run_id}/participants  request: Participant`; `request_shapes` declares only `RunCreate`. Contract (`art_6cf5c9af40db/verification_contract.yaml`): `vc-probe-runs-participants  json {} -> 201`, `vc-probe-runs-participants-duplicate  json {} -> 409`. Boot audit: `FAIL probe vc-probe-runs-participants-duplicate: status 422 != expected 409`. Roll 2 (`cyc_f84649c68646`) declared `ParticipantIn: required: [name]` and its probes carried `{"name": "sample"}` — that is the whole difference.

The loop's verdict was right for once: repair-00 (dev, `backend/routes.py`) could not change the outcome; round 1 repeated the signature → `correction_terminated_plan_defect … candidate=tighten_acceptance`. Cost: one wasted dev repair, one rejected roll.

## Fix shape

One resolver for "request body fields of endpoint X" — `shape.required`, else the entity's required, non-generated fields — used by **both** the contract generator and the route emitter (`entity_field_names` is already computed in `scaffold_contract.py` and unused for this). Or the validator stops accepting entity-typed requests; either way the two homes must agree. Test: endpoint `request: <entity>` → probe body carries the entity's required fields.

jasonLadd:
## Summary

**1.6.6 item E (#1128).** Manifest validation accepts an endpoint `request:` that names a declared shape **or an entity** (`scaffold.py`'s `usability_errors`); the contract generator resolved probe bodies only through `request_shapes` (`scaffold_contract.py:439`, `:537-541`), so 1.6.5 FastAPI+React roll 3's `POST /runs/{run_id}/participants  request: Participant` shipped `json: {}` expecting 201 then 409 from a route whose payload class required `name`. Unsatisfiable by construction; the loop correctly terminated `plan_defect` after one wasted dev repair.

**One resolver, on the manifest.** `InterfaceManifest.request_body_fields(request)`: a declared shape's `required`; an entity's required, non-generated, undefaulted fields (a body need not carry a generated `id` or a defaulted `participants`); nothing otherwise. Every probe-body site — the create probe, the path-param child actions, the discriminated children, and the blank-input probe — reads it.

**And the route emitter agrees by construction.** Writing the "two homes agree" test on the reference manifest surfaced the second half: `Participant` has a generated `id`, so the frozen entity model *requires* it — an entity used directly as the payload class (`payload: Participant`) can never accept the body the resolver sends. `request_model_name` now maps an entity-typed request to a synthesized `{Entity}Body` model, emitted by `_entity_body_model_lines` from the same resolver (`NonBlankStr` required fields, other non-generated fields optional), and the route takes and imports that. The blank-input probe therefore applies to entity-typed requests exactly as to declared shapes.

**Pins:** the reference manifest declares shapes throughout, so its contract (v11), the stack-#1 hash, the Next.js scaffold pin and the three goldens are byte-identical. Output changes only for manifests that write an entity-typed request — the shape that produced roll 3.

## Tests

- `TestEntityTypedRequestBodies`: the resolver's four answers (shape, entity minus generated/defaulted, unknown, none); roll 3's manifest shape yields `{"name": "sample"}` on join/join-duplicate/leave; the seeded route takes `ParticipantBody`, imports it, and the frozen model accepts exactly the probe body, rejects blank input (#593), while the entity itself still requires its `id`; the synthesized model is shaped by the resolver.
- `./scripts/dev/run_regression_tests.sh` — 8044 passed, 1 skipped, exit 0.

## Closes

Closes #1128

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,17 @@ All notable changes to SquadOps are recorded here. Format loosely follows
   so a prefixed router serves the declared paths instead of failing a literal-path check,
   and the repair brief (v6) says to keep the router construction, decorators and handler
   names as they are — a rewrite is refused before it is tested.
+- **E — an endpoint's request body has one resolver** (#1128). Manifest validation accepts
+  a `request:` that names a declared shape *or an entity*; the contract generator resolved
+  probe bodies only through `request_shapes`, so an entity-typed request shipped `json: {}`
+  expecting 201 then 409 from a route whose payload class required `name` — unsatisfiable by
+  construction, and the loop rightly called it `plan_defect`. `InterfaceManifest.request_body_fields`
+  now answers for both (a shape's `required`; an entity's required, non-generated,
+  undefaulted fields) and every probe-body site reads it. The route emitter no longer takes
+  the entity class itself as the payload — its generated `id` is required, so it could never
+  accept that body — but a synthesized `{Entity}Body` model shaped by the same resolver,
+  with `NonBlankStr` required fields so the blank-input probe applies as it does to a
+  declared shape. Pinned reference output (declared shapes throughout) is byte-identical.
 
 ## [1.6.5] — 2026-08-27
 
--- a/src/squadops/capabilities/scaffold.py
+++ b/src/squadops/capabilities/scaffold.py
@@ -256,6 +256,60 @@ class InterfaceManifest:
     #: author. Excluded from ``_canonical`` for the same reason ``decisions`` is.
     provenance: Provenance | None = None
 
+    def request_body_fields(self, request: str | None) -> tuple[str, ...]:
+        """The fields a body for ``request`` must carry — the ONE answer (#1128).
+
+        ``request`` is an endpoint's ``request:`` and validation accepts either a declared
+        request shape or an entity (see ``usability_errors``). The contract generator used
+        to resolve bodies only through ``request_shapes``, so an entity-typed request got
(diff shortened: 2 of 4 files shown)


jasonLadd (commented):
reviewed

Issue #929, opened 2026-08-16
Extract the LLM call sequence — the duplication, not the thirteen call sites


jasonLadd:
**Rewritten 2026-08-16.** The original proposal here — a decorator at the LLM port — was the weaker of the two options. It would have fixed observability while leaving the duplication that caused the problem, and it loses handler attribution. Superseded by the proposal below.

## What is *not* the problem

Thirteen call sites to `chat_stream_with_usage` is not, by itself, bad design. The LLM port **is** the abstraction; handlers calling it directly is what the architecture intends. Each site builds a different prompt, passes different model kwargs, parses a different response shape and fails differently — and several of the thirteen are legitimate second calls (self-eval, frontmatter retry), not copies. A shared `call_the_llm()` taking enough parameters to serve all of them would be a config object, which is worse.

Counting call sites was the wrong frame, and #928's framing of it was wrong.

## What *is* the problem

The ~10-line **sequence** wrapped around each call is copy-pasted:

```
resolve model budget → guard prompt size → call → read .content
→ measure duration → record generation
```

`_resolve_model_budget` is defined once and shared correctly. `_record_generation` was not, and the drift is measured, not hypothetical:

| copy | token accounting | layer ids | used by |
|---|---|---|---|
| `_CycleTaskHandler.handle`, **inline** | full | hardcoded `{role}-cycle` | every handler on the default path |
| `_CycleTaskHandler._record_generation` | full | from `_prompt_layer_kind` | dev, qa, builder |
| `GovernanceReviewHandler._record_generation` | **none** | hardcoded | governance's manifest path |

The third cost governance all token accounting in LangFuse, silently, for as long as nobody diffed two files that were never meant to be compared. Fixed in #938. Same class as #918.

The first two still coexist. They are not a defect today — both record the same fields — but they disagree about layer identity, so the same conceptual event is grouped two different ways depending on which path a handler took. That is a taxonomy that will drift again.

## Why this keeps happening

A cross-cutting concern added to this sequence has to be added N times or it goes dark in N−1 places. That is exactly how #924's emission capture ended up covering one seam of thirteen, including none of the ones the SIP-0104 window depended on.

**And #928's fix makes the duplication permanent rather than removing it.** Its coverage test asserts every file with an LLM call also logs an emission — which prevents a fourteenth dark site, but blesses thirteen copies and will need updating for every future cross-cutting concern.

## Proposal

[...]

Issue #927, opened 2026-08-16
1.7: reasoning effort should be contract-declared and provider-neutral, not a boolean switch on a handler


jasonLadd:
Reasoning-capable models spend completion budget on a channel this platform never reads, and there is no way for a task or an agent to say how much reasoning it wants. The expedient fix was rejected in review as a workaround; this is the design it should have.

## The measurement that motivates it

`qwen3.6:27b`, same prompt (the deployed SIP-0104 fill-mode brief), same eight scaffold slots:

| | reasoning on | reasoning off |
|---|---|---|
| completion tokens | **5,727** | **413** |
| response | 1,411 chars | 1,417 chars |
| fill fences | 8 | 8 |

13.9× the tokens for the same output; the no-reasoning run additionally asserted a field the other missed. At a 2,500-token budget the model produced 19,179 characters of thinking and an **empty response** — it correctly enumerated all eight slots in its reasoning and never reached the emission.

`adapters/llm/ollama.py` documents why this is waste rather than cost: qwen3-family models emit `message.thinking` by default, the adapter never sends `think`, reads only `message.content`, and declares `THINKING_TOKENS: False`. The channel is billed against `num_predict` and discarded unread. **Shared budget, single harvest.**

## Why the first attempt was rejected (#925)

- **`_suppress_reasoning: bool` on a handler class** — a switch, not a seam, and it names the mitigation rather than the fact. CLAUDE.md forbids feature flags and dual paths.
- **`thinking: bool` on the port** — Ollama's shape leaking into a provider-neutral port. OpenAI expresses this as an effort level, Anthropic as a token budget. This is exactly the class 1.7's "every port is actually a port" exists to fix, and the Atlas migration forces it.
- **One fact across five touchpoints** — port signature, payload builder, router pass-through, vLLM ignore-stub, two class attributes.
- **It treats the symptom.** The defect is a shared budget with a single harvest; an off-switch works around that rather than expressing it.

## The design

**1. The capability contract declares the shape of its output.** It already declares inputs, outputs, artifacts and acceptance checks — *"is this output a transcription or an argument"* belongs there as data, alongside them. The principle that makes it decidable per handler:

> **Reason where the output is an argument. Don't where the output is a transcription.**

Filling scaffold slots is derivation — shells, contract and envelope are all in the prompt and the answer is a fixed format. Authoring an interface manifest from behavioural prose is genuinely an argument: endpoints, statuses and error semantics are being chosen, not restated.

[...]

PR #1161, opened 2026-08-28
feat(llm): reasoning is a level on the port, declared per capability, mapped by each adapter (#927, 1.7.0 Reasoning)

jasonLadd:
## What

The first item of the 1.7.0 Reasoning line (`docs/plans/1-7-0-plan.md` §2.1). The reasoning channel was paid for on every qwen3 generation, discarded unread, and invisible: #924 measured the deployed qa fill brief at **5,727** completion tokens with it on and **413** with it off — the same eight fences — and at a 2,500-token budget, 19,179 characters of thinking and an empty response. The first fix (#925) was rejected as a boolean on a handler and an Ollama `think` bool on a provider-neutral port. This is the design #927 asked for, in four commits:

**1. The port carries a level** — `reasoning: none | low | medium | high` (`ReasoningLevel`, `squadops/llm/models.py`) on `chat` / `chat_stream` / `chat_stream_with_usage` / `LLMRequest`. A request, not a guarantee; `None` sends nothing and the wire is byte-identical to 1.6.6.
- **Ollama**: `none → think:false`, any graded level → `think:true` (the dialect's only dial).
- **vLLM**: the OpenAI shape has no reasoning field; the adapter maps by the dial the *model spec* declares — `toggle` → `chat_template_kwargs.enable_thinking`, `effort` → `reasoning_effort` (`none` becomes `low`; that dial cannot switch off), no dial / unregistered → nothing.
- `LLMCapability.REASONING_CONTROL` declares whether a level reaches the wire; the conformance suite asserts the declaration against the actual request bodies for every adapter (`bare ≠ none ≠ high` iff declared).
- `ModelSpec.reasoning_control` is **required** on every registry entry (no default — an undeclared dial is a registry gap, not a model that reasons silently): qwen3.x `toggle`, qwen2.5/llama3 `none`.

**2. Every capability declares how much reasoning its output wants** — `capabilities/reasoning_policy.py`, one table: *reason where the output is an argument; don't where it is a transcription.* `qa.test` (fills slots), the repairs, verdicts and stored reports are `none`; develop/repair `medium`; the manifest, the analyses, the plan `high`. **No default**: an undeclared capability raises, and a test over `HANDLER_CONFIGS` makes a missing entry a CI failure. When #922 gives cycle capabilities a contract as data, the column moves there.

**3. Resolution is the chain every other knob uses** (SIP-0075 §3.2, #1011): declaration → `config_overrides.reasoning` (new allowed key; value validated at the profile boundary — a misspelt level would otherwise reach Ollama as `think: true`) → the model's dial (a channel-less model gets no level: Ollama answers `think: true` with a 400 for it). Wired through `_build_chat_kwargs` for every base-path and planning handler, and the three hand-rolled sites (develop, qa_test, builder) via a branchless splice — the duplication itself is #929's.

**4. `GenerationRecord.reasoning`** carries the sent level to LangFuse (the copy-everything `replace` keeps it — #793's class), so the Reasoning pack's first prediction — *the level resolves per capability* — is read per generation rather than inferred from a token count.

**Not in this PR, said plainly:** the read half (#410 — `message.thinking` captured and emitted); a CRP-level override (the #927 chain's last link — a contract-pack change, the CRP applied-defaults carry no per-agent LLM knobs today); the two shakeouts and the #924 budget re-read (§6.1 criteria 1–2, after this merges and deploys).

## Closes

Refs #927 — remaining: the CRP-level override (contract-pack change), and #410's read half stays its own issue.

## Evidence

- `run_regression_tests.sh`: **8,174 passed**, all gates (ruff check, format, test-quality lint, pytest). 50 new tests: adapter mappings, conformance for both adapters, the policy coverage guard, the chain at the base/repair/roles-handler/develop-qa-builder seams, profile validation, the telemetry field.
- **Live mechanism check through the new port**, host Ollama 0.32.14, `qwen3.6:27b`, a three-slot fill brief, `max_tokens=2500`:

  | level | completion tokens | wall | fences | response |
  |---|---|---|---|---|
  | `none` | **256** | 30.6 s | 3/3 | 922 chars, complete |
  | `high` | **2,500** (the cap) | 206.7 s | 0/3 | **empty** |

  The #924 failure shape, reproduced: with the channel on, the fill exhausts its budget thinking and emits nothing. Same adapter, same prompt, one kwarg.
- **The clamp's premise, measured**: `qwen2.5:7b` with `think: true` → `400 "qwen2.5:7b does not support thinking"`; with `think: false` → 200. Sending nothing for a channel-less model is correct; the wording in code now states exactly that.

https://claude.ai/code/session_01CbzZWzB8VgjZdbtKX9N41L

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ All notable changes to SquadOps are recorded here. Format loosely follows
   with no channel). `LLMCapability.REASONING_CONTROL` declares whether a level
   reaches the wire, and the conformance suite asserts the declaration against the
   request bodies for every adapter. No level sent ⇒ the wire is byte-identical to
-  1.6.6.
+  what it was before this change.
 - **Every capability declares how much reasoning its output wants**
   (`capabilities/reasoning_policy.py`): a transcription — filling scaffold slots,
   a repair verdict, a stored report — is `none`; an argument — the manifest, the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,32 @@ All notable changes to SquadOps are recorded here. Format loosely follows
 
 ## [Unreleased]
 
-Nothing yet.
+**1.7.0 — the Reasoning line opens.** Plan: `docs/plans/1-7-0-plan.md` (rev 3).
+
+### Added — reasoning is a level on the port (#927)
+- **The port carries `reasoning: none | low | medium | high`**, never a provider's
+  switch. Ollama maps `none → think:false` and any graded level → `think:true`; vLLM
+  maps by the dial the model spec declares (`chat_template_kwargs.enable_thinking`
+  for a toggle model, `reasoning_effort` for an effort model, nothing for a model
+  with no channel). `LLMCapability.REASONING_CONTROL` declares whether a level
+  reaches the wire, and the conformance suite asserts the declaration against the
+  request bodies for every adapter. No level sent ⇒ the wire is byte-identical to
+  1.6.6.
+- **Every capability declares how much reasoning its output wants**
+  (`capabilities/reasoning_policy.py`): a transcription — filling scaffold slots,
+  a repair verdict, a stored report — is `none`; an argument — the manifest, the
+  analyses, the plan — is `high`. #924 measured the difference on the deployed qa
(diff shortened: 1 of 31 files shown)


jasonLadd (commented):
reviewed

Issue #1159, opened 2026-08-28
Atlas adapter behind LLMPort with the #927 reasoning mapping and observed capabilities (SIP Atlas P4)


jasonLadd:
## What

`adapters/llm/atlas.py` implementing `LLMPort` per `sips/proposed/SIP-Atlas-Provider-Adapter.md` §3.5 — a sibling of `OllamaAdapter` (httpx, the same four exception translations, `chat_stream_with_usage` streaming-for-liveness), speaking Atlas's OpenAI-shaped surface (`POST /v1/chat/completions`, `GET /v1/models`, port 8888). One adapter per provider (§3.5a) — not a subclass of `VLLMAdapter`, even where the dialect matches, because Atlas is expected to diverge (duration fields) and provider conditionals in a shared base are #559's identity branching.

## Requirements the SIP fixes, dialect-independent

- **Capabilities observed, not assumed** (§3.2): `MODEL_LISTING`, `MODEL_MANAGEMENT`, `STREAMING_USAGE`, `THINKING_TOKENS` declared from what the live endpoint does (#1158 answers these).
- **Usage accounting populated or explicitly `None`**, never zero-filled; capture the duration fields (`prefill_ms` / `load_ms` / `total_ms`) if Atlas emits them, declare their absence if not (§10.2 fact 5, option 2: wall-clock derivation, inclusive of prefill and queueing — not interchangeable with Ollama's decode-only figure).
- **Error translation is the contract**: 404 → `LLMModelNotFoundError`, connect → `LLMConnectionError`, timeout → `LLMTimeoutError` — the correction loop's locus classification (#568) reads these.
- **No structured-output surface** (`response_schema` belongs to SIP-LLM-Emission-Contracts).

## Added by the 1.7 plan (rev 3)

- **Implements #927's reasoning level from the start**: the port's `none | low | medium | high` mapped by the dial kind the model spec declares, in the shape #1158's fact 6 establishes. If Atlas exposes no control, the adapter declares that and the port's "request, not guarantee" rule applies. Built after #927's port change lands, not before — otherwise it is retrofitted.
- **Reads the reasoning channel if Atlas separates it** (#410's read half) — `reasoning_content` or Atlas's equivalent — and emits it to LangFuse under the redaction rules; if thinking arrives inline in `content`, that is a preflight failure (#1145's shape), not something the fenced parser should receive.
- **Joins the characterization suite** as its third recorded subject beside Ollama and vLLM.

## Gates

[...]

Issue #154, opened 2026-06-11
Move adapter imports out of domain modules into bootstrap wiring


jasonLadd:
## Problem
Domain modules import from `adapters/` directly:
- `src/squadops/orchestration/orchestrator.py:112` — lazy-imports `NoOpLLMObservabilityAdapter` for the always-inject-NoOp fallback. The *pattern* is deliberate and good; the *import direction* (domain → adapters) violates the hexagonal boundary.
- `src/squadops/agents/entrypoint.py` — ~12 adapter factory imports (this is wiring code, but it lives inside the domain package).
- `src/squadops/config/loader.py` — imports `adapters.secrets.factory`.

## Proposed fix
Move NoOp fallback injection and factory wiring into the existing `src/squadops/bootstrap/` package (or dedicated factory functions there). The orchestrator/agents should receive the NoOp adapter from the composition root, never import it. Consider adding a forbidden-imports test (`squadops.*` minus bootstrap/entrypoint may not import `adapters`) to lock the boundary — the SIP-0089 branch already has a forbidden-imports test pattern to extend.

_Source: architectural review, 2026-06-11 (boundary cleanup)._

jasonLadd:
Precise current state (2026-07-04 sweep) — this issue can be re-scoped much tighter than filed:

`grep -rn "from adapters|import adapters" src/squadops` → 41 real imports, **all function-local/lazy; zero module-level**. 36 of 41 are sanctioned composition roots (`api/runtime/main.py` ×18, `agents/entrypoint.py` ×11, `scheduler_bootstrap.py` ×5, `deps.py` ×2). The genuine residue is **5 imports**:

- `src/squadops/config/loader.py:560` — `from adapters.secrets.factory import create_provider` (domain config loader reaches into a concrete adapter for `secret://` resolution) — the worst one
- `src/squadops/orchestration/orchestrator.py:112` — constructs `NoOpLLMObservabilityAdapter` as its default
- `src/squadops/bootstrap/setup/checks.py:698` — `ConfigSquadProfile` (self-acknowledged in a comment)
- `src/squadops/api/routes/cycles/models.py:50` + `cycles.py:53` — `isinstance(port, OllamaAdapter)` in the route layer (should be a port capability check)

Suggest: fix these five, then add an architecture test with an explicit composition-root allowlist so the count can only ratchet down. That converts this from an open-ended cleanup into a closable issue with a guardrail.

jasonLadd:
[...]

PR #1164, opened 2026-08-28
feat(config): the LLM provider is required configuration — selected by name through the factory at both composition roots (#1157)

### LINKED ISSUE #1157: LLM provider is never selected: factory defaults to Ollama, entrypoint passes none, LLMConfig has no provider field — make it required, not defaulted (SIP Atlas P2)
## What is wrong

There are two LLM adapters in the tree and no configuration can select the second one. Provider selection is not a seam today; it is a hardcoded string with three authors.

| # | Where | What it does |
|---|---|---|
| 1 | `adapters/llm/factory.py:20` | `create_llm_provider(provider: str = "ollama", …)` — the factory **defaults** the provider |
| 2 | `src/squadops/agents/entrypoint.py:362-366` | calls `create_llm_provider(base_url=…, default_model=…, timeout_seconds=…)` — **never passes `provider`**, so every agent takes the factory default |
| 3 | `src/squadops/config/schema.py:531-539` | `LLMConfig` has `url`, `model`, `use_local`, `timeout` — **no `provider` field**, so `SQUADOPS__LLM__PROVIDER` does not exist and setting it does nothing, silently |
| 4 | `src/squadops/api/runtime/main.py:433-439` | constructs `OllamaAdapter` directly, bypassing the factory (this is #301's LLM half) |
| 5 | `src/squadops/api/routes/cycles/cycles.py:61-65` | model preflight returns `None` unless `isinstance(port, OllamaAdapter)` — under any other provider the cycle-create preflight goes silently unverifiable (#154's site list) |

`adapters/llm/vllm.py` (P6, #805, 2026-08-08) and `factory.py:54`'s `provider == "vllm"` branch are therefore unreachable from any deploy surface. The factory's own comment at `factory.py:60-63` says an *unknown* provider "must fail loudly at startup rather than silently running the previous one (the standing no-masking-fallbacks rule)" — while its signature default masks a *missing* one the same way. The factory has been at this shape since SIP-0059 (`c2fd505e`, 2026-01-31); the second adapter landed seven months later without a selection path.

## The ruling — required, not defaulted (owner, 2026-08-28)

> "ensure that the provider fix does more than just pass provider but also require it; I don't want to be silently just using the parameters passed in and not require a defined provider."

This contradicts the proposed SIP as written and the SIP must be revised, not worked around:

- `sips/proposed/SIP-Atlas-Provider-Adapter.md` §3.1 designs `LLMConfig.provider` with `default="ollama"`, copying `SandboxConfig.provider = "noop"` as the "dormant posture" precedent.
- §4.1 item 1 (the dark-ship rule) *requires* that default, on the ground that the Spark lane was running 1.6 validation windows against Ollama and a changed runner mid-window would look like a squad regression.

That ground no longer exists: the 1.6 line closed at v1.6.6 (2026-08-28). What survives of §4.1 is item 3 — with `provider=ollama` the resolved adapter, requests and responses are byte-identical, proven by test.

## The fix

1. **`LLMConfig.provider: str` with no default** (`schema.py`). A config without it fails validation at load — the same class of failure as a typo, and the same rule: a misconfiguration is not a default. Config var `SQUADOPS__LLM__PROVIDER` per the existing nesting convention; no new prefix.
2. **`create_llm_provider(provider: str, …)` with no default** (`factory.py:20`). The `ValueError: Unknown LLM provider` branch stays.
3. **Every composition root passes `provider=config.llm.provider`**: the entrypoint (site 2), and `main.py:433` stops constructing `OllamaAdapter` and goes through the factory (site 4 — this is the LLM half of #301; the queue half stays with #301 under the 1.7 Composition Root design gate, per the SIP's §10.1 scope ruling).
4. **Every deploy surface sets it explicitly**, because required means required: the seven `SQUADOPS__LLM__*` blocks in `docker-compose.yml` (lines 190, 248, 302, 375, 431, 483, 539 today), the bootstrap profiles, `.env.example` (uncommented, since it is not optional), and the test config fixtures. One-click install still holds — the value is written in the files the install reads, not typed by a person.
5. **Site 5** replaces `isinstance(port, OllamaAdapter)` with the port's declared `MODEL_LISTING` capability (P0 already declares it; SIP §3.2 "site 2, replacing isinstance").
6. `.env.example:62` reads `SQUADOPS__LLM__USE__LOCAL=true` — the double underscore nests it as `llm.use.local`, not `llm.use_local`. Fix the line while touching the block, or delete `use_local` if nothing reads it (nothing in `src/` or `adapters/` does — only `tests/integration/conftest.py:540`).

## What bug the tests catch

- A config with no `provider` **fails to load** (today it loads and runs Ollama).
- `create_llm_provider()` without `provider` is a `TypeError` (today it returns an `OllamaAdapter`).
- `SQUADOPS__LLM__PROVIDER=vllm` resolves a `VLLMAdapter` at the entrypoint seam and at the runtime-api seam — the two-seam gotcha, asserted at both.
- `provider=ollama`: the request payload is byte-identical to the pre-change adapter (the dark-ship §4.1 item 3, kept).
- An unknown name raises at startup (existing, keep).

## Placement

1.7.0's CI-verified rider, beside #929/#944 — it is the missing half of what the 1.7 plan §2.1 calls "the Atlas seam", and the SIP's own Ruling 1 allows it to "ship formally dark in 1.7". The SIP §3.1 and §4.1 text is revised in the same PR (the SIP is still `proposed`, so a revision, not an amendment section). Runs on both adapters' characterization suites (#927/#410 must land on `vllm.py` too — `THINKING_TOKENS: False` at `vllm.py:150`, `content`-only read at `vllm.py:289`).

Refs #301 — remaining: the queue-factory half and the composition-root design note.
Refs #154 — site 5 is one of its four domain→adapter sites.
Related: SIP-Atlas-Provider-Adapter P2; #927, #410, #929, #944, #1145.

jasonLadd:
## What

SIP-0106's P2, built to the owner's Ruling 3: **required, never defaulted.** Two adapters were in the tree and no configuration could select the second — the factory defaulted `provider="ollama"` (`adapters/llm/factory.py`), the agent entrypoint never passed one, and `LLMConfig` had no field, so `SQUADOPS__LLM__PROVIDER` did nothing, silently.

- **`LLMConfig.provider: str`** with no default, and **`AppConfig.llm`** without its `default_factory` — a config without the provider fails validation at load (`llm: Field required` / `llm.provider: Field required`), the same class of failure as a misspelt name in the factory. Only the selector is required; `url`/`model`/`timeout` keep their defaults.
- **`create_llm_provider(provider, …)`** — no default. The `Unknown LLM provider` refusal stays.
- **Both composition roots through the factory**: the agent entrypoint passes `provider=config.llm.provider`; the runtime-api builds its LLM port with `create_llm_provider` instead of constructing `OllamaAdapter` directly (the LLM half of #301). Construction is outside the non-fatal `try` so an unknown name stops startup.
- **Site 5**: the cycle-create model preflight asks `port.supports(MODEL_LISTING)` and calls `list_available_models()` instead of `isinstance(port, OllamaAdapter)` — every non-Ollama provider was silently "unverifiable, warn and allow".
- **Every deploy surface writes the value** — `docker-compose.yml`: the seven agent blocks (after `SQUADOPS__LLM__TIMEOUT`) plus `joi`, as a **literal `ollama`**, not a `${…:-ollama}` default; `.env.example` uncommented; the two `AppConfig`-building tests; the CLI integration fixture (the runtime-api app loads config at import). **`docker-compose.yml` is modified deliberately** — the issue's fix shape, endorsed 2026-08-28; no service or container names change.
- The mis-nested `SQUADOPS__LLM__USE__LOCAL` example (nested as `llm.use.local`) and the `use_local` field nothing read are removed.

**Not here:** the Atlas adapter itself (#1159) and the queue-factory half of #301.

## Closes

Closes #1157
Refs #301 — remaining: the queue-factory half and the composition-root design note.
Refs #154 — remaining: the other three domain→adapter sites and the all-packages import guard.

## Evidence

`run_regression_tests.sh`: **8,187 passed**, all gates. New tests: `tests/unit/llm/test_factory.py` (missing provider is a `TypeError`, each name resolves its own class, vLLM receives `api_key`, unknown raises), `tests/unit/config/test_llm_provider_required.py` (`LLMConfig()` and `AppConfig` without `llm` fail; the env var nests to `llm.provider`), `tests/unit/cycles/test_model_preflight_port.py` (listing-declaring provider verified; non-listing → `None` not `[]`; unreachable → `None`). The first regression run caught the CLI integration fixture importing the app without the var — five setup errors — which is the guard working as intended; the fixture now sets it.

**Deploy note:** after merge, `rebuild_and_deploy.sh runtime-api agents` — the compose file carries the value, so no `.env` change is needed on the Spark.

https://claude.ai/code/session_01CbzZWzB8VgjZdbtKX9N41L

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -57,9 +57,11 @@ SQUADOPS_PROFILE=dev  # Options: dev, local, lab, cloud
 # SQUADOPS__PREFECT__API__URL=http://prefect-server:4200/api
 
 # LLM Configuration
+# The provider is REQUIRED — it selects the adapter (ollama | vllm) and nothing defaults
+# it (#1157). docker-compose.yml sets it per service; set it here for host-side runs.
+SQUADOPS__LLM__PROVIDER=ollama
 # SQUADOPS__LLM__URL=http://localhost:11434
 # SQUADOPS__LLM__MODEL=llama3.1:8b
-# SQUADOPS__LLM__USE__LOCAL=true
 # SQUADOPS__LLM__TIMEOUT=60
 
 # Agent Configuration
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,22 @@ All notable changes to SquadOps are recorded here. Format loosely follows
 
 **1.7.0 — the Reasoning line opens.** Plan: `docs/plans/1-7-0-plan.md` (rev 3).
 
+### Changed — the LLM provider is required configuration (#1157, SIP-0106 Ruling 3)
+- **`SQUADOPS__LLM__PROVIDER` is required and never defaulted.** Two adapters were in the
+  tree while the factory defaulted `provider="ollama"`, the agent entrypoint never passed
+  one and `LLMConfig` had no field — the vLLM adapter was unreachable by any configuration
+  since it landed. Now `LLMConfig.provider` and `create_llm_provider(provider, …)` have no
+  default; a missing value fails at config load the way an unknown one fails in the
+  factory. Every deploy surface writes the value: the eight SquadOps service blocks in
+  `docker-compose.yml`, `.env.example` (uncommented), the test fixtures. The switch to
+  another provider is one PR changing that value (SIP-0106 §5).
+- **Both composition roots go through the factory**: the runtime-api no longer constructs
+  `OllamaAdapter` directly (the LLM half of #301), and the cycle-create model preflight asks
+  the port for `MODEL_LISTING` instead of `isinstance(OllamaAdapter)` — on any other
+  provider it was silently unverifiable.
+- The mis-nested `SQUADOPS__LLM__USE__LOCAL` example and the `use_local` field it never
+  reached are gone.
+
 ### Added — reasoning is a level on the port (#927)
 - **The port carries `reasoning: none | low | medium | high`**, never a provider's
   switch. Ollama maps `none → think:false` and any graded level → `think:true`; vLLM
--- a/adapters/llm/factory.py
+++ b/adapters/llm/factory.py
@@ -17,7 +17,7 @@ if TYPE_CHECKING:
 
 
 def create_llm_provider(
-    provider: str = "ollama",
+    provider: str,
     secret_manager: SecretManager | None = None,
     base_url: str = "http://localhost:11434",
     default_model: str = "llama3.2",
@@ -27,7 +27,9 @@ def create_llm_provider(
     """Create an LLM provider.
 
     Args:
-        provider: Provider name ("ollama" or "vllm")
+        provider: Provider name ("ollama" or "vllm"). Required — no default (#1157):
+            the factory's own rule below refuses an unknown name, and a *missing*
+            one used to fall back to Ollama the same silent way.
         secret_manager: Optional secret manager for resolving secret:// refs
         base_url: Base URL for the LLM server (may be secret:// ref)
         default_model: Default model to use
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -189,6 +189,7 @@ services:
       # LLM (SIP-0066: cycle task handlers)
       SQUADOPS__LLM__MODEL: "llama3.1:8b"
       SQUADOPS__LLM__TIMEOUT: ${SQUADOPS__LLM__TIMEOUT:-180}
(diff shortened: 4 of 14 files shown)


jasonLadd (commented):
reviewed

Issue #1160, opened 2026-08-28
Atlas vs Ollama A/B artifact on the Spark — the §3.6.1 record the switch is decided on (SIP Atlas P5)


jasonLadd:
## What

The named artifact `sips/proposed/SIP-Atlas-Provider-Adapter.md` §3.6.1 specifies, produced by Appendix C's runbook against a live Atlas on the Spark, and committed — the switch decision reads this and nothing else (Ruling 2: "the story should tell itself once I run a cycle against the new adapter"; the benchmark reports, the owner judges).

## The record

One row per (prompt, model, adapter), all fields required, `None` only where the provider genuinely cannot supply it and then declared: `adapter`, `model` (resolved value verbatim — naming differs per provider), `prompt_tokens`, `completion_tokens`, **`wall_clock_ms` (the decision field)**, `decode_tokens_per_second` (the diagnostic), `prefill_ms` / `load_ms` / `total_ms` (Ollama arm always; Atlas arm if emitted — an asymmetric A/B is anticipated and stated), `artifact_validated` (the cycle's own verdict — throughput is meaningless without quality parity).

Prompts: the deployed qa fill brief #924 probed (so the reasoning-level split is read on both engines) and the 1.7.0 shakeout's framing brief; then **one full cycle per arm** on the `full` profile's model and the standard PRD, since wall clock across a six-agent box (model residency, queueing) is what the owner wants to save, not a single generation's t/s.

Measurement hygiene kept from §C.5 — cold-vs-warm, arm ordering, token-count shift, quantization mismatch (Atlas's fp8 recipe vs Ollama's quant is a *stated* difference, not a hidden one) — because these are what make an observed number real. No threshold tables (removed under Ruling 2).

## Where it lands

Committed under `docs/plans/` beside the verification-set records, named for what it is. The 1.7 plan (rev 3) places the switch as an owner decision taken on this artifact at a line boundary — never mid-set.

Depends on #1159 (the adapter) and #1158 (Atlas on the Spark).

No issue: SIP Atlas P5, filed from the 1.7 plan rev 3 placement (2026-08-28).

jasonLadd:
[...]

PR #1170, opened 2026-08-29
fix(llm): the Atlas arm's served window covers its own prompts — 32K left the prompt guard 24,576 tokens (#1160 §1.4)

jasonLadd:
## What

Raises the Atlas arm's served window from 32,768 to 65,536 — on the serve line
(`--max-seq-len`) and on the `Qwen/Qwen3.8-27B-FP8` registry entry the prompt guard
reads — before the third A/B shakeout.

## Why

The prompt guard spends `context_window - max_completion_tokens` on the prompt
(`src/squadops/capabilities/handlers/prompt_guard.py`). Over budget it does not fail: it
deletes the `## Prior Analysis from Upstream Roles` section and sends the rest, raising
`PROMPT_EXCEEDS_CONTEXT_WINDOW` only if the remainder still will not fit. At
`--max-seq-len 32768` that budget was **24,576 tokens on arm B** against **253,952 on
arm A** — the same weights, served by Ollama at its native 262,144.

Measured off the Ollama arm's own server log (1,145 prompts on `qwen3.8:27b` since
2026-08-22, `n_ctx_slot = 262144`):

| median | p90 | p99 | max | > 24,576 | > 32,768 |
|---|---|---|---|---|---|
| 9,895 | 20,212 | 31,137 | 38,210 | ~5.9% | 0.87% |

So roughly **one generation in seventeen** would have run on arm B with its upstream
analysis removed and on arm A intact — a difference in what the model was asked,
invisible in the verdict, that the record would have credited to the engine. Framing
prompts never reach the guard at all (its only callers are the dev and qa handlers) and
would have met the server's limit directly, which is how §1.1 failed.

65,536 covers the longest observed prompt (38,210) plus the 8,192 completion clamp with
headroom. The KV pool is paged and sized independently of the cap (16.2 GB ≈ 265K tokens
at ~64 KB/token BF16), so one 64K sequence costs ~4 GB of it and nothing at the observed
median. Arm A is untouched.

## Changes

- `src/squadops/llm/model_registry.py` — `context_window` 32,768 → 65,536, with the
  measurement in the comment
- `tests/unit/llm/test_model_registry.py` — the pinned window follows, plus a new guard
  that the usable prompt budget covers the longest measured real prompt
- `docs/plans/1-7-0-atlas-ab-preregistration.md` — §1.4 records the revision and the
  distribution; the §1 table's serve line and window cell follow
- `CHANGELOG.md` — the registry line names the served window it now carries
- (outside the repo) `~/atlas/scripts/atlas_serve.sh` raises `--max-seq-len` to match

## Consequence for the A/B

The registry entry ships in the images, so the agent and runtime-api rebuild on this
commit becomes the frozen deploy — superseding §1.1's, whose ids the §1 table still
named. Those ids are pinned into the four set files before the first counting roll. The
third shakeout runs on this line.

Whether Atlas rejects or silently truncates past its window is asserted in the registry
comment but has never been measured; one over-window request is probed before the
shakeout and the answer recorded in §1.4.

## Tests

`tests/unit/llm/` + `tests/unit/architecture/` — 234 passed; prompt-guard tests — 20
passed; ruff clean.

Refs #1160 — remaining: the A/B itself (shakeout #3, the A1/B1/B2/A2 and A3/B3 rolls,
and the record).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,9 @@ All notable changes to SquadOps are recorded here. Format loosely follows
   served Qwen3.8 template requires; streamed `reasoning_content` deltas never reach the
   text. Registered as the conformance suite's third case, with the measured shapes.
 - `ChatMessage` / `LLMResponse` gain `reasoning_tokens` (None when not reported, never
-  zero); `Qwen/Qwen3.8-27B-FP8` joins the model registry with the served 32K window.
+  zero); `Qwen/Qwen3.8-27B-FP8` joins the model registry with the window the A/B recipe
+  serves — 64K, raised from the first recipe's 32K because that left the prompt guard a
+  24,576-token budget and 5.9% of the arm's real prompts run longer (#1160 §1.4).
 
 ### Added — reasoning is a level on the port (#927)
 - **The port carries `reasoning: none | low | medium | high`**, never a provider's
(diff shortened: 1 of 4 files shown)


jasonLadd (commented):
reviewed

PR #1174, opened 2026-08-29
fix: the framing half of a cycle is observable and correctly budgeted — one seam for generation records, per-attempt manifest telemetry, and a completion budget that accounts for declared thinking

### LINKED ISSUE #1171: Framing generations reach LangFuse with zero token usage and no decode rate — planning/base.py builds the record without the four fields cycle/base.py passes
## What

Every generation from the planning/framing handlers lands in LangFuse with
`usage = {input: 0, output: 0, total: 0}` and no `tokens_per_second` in metadata. The
implementation-side capabilities are unaffected.

Measured over the 1.6.6 Next.js rolls (`cyc_68b0e1769526`, `cyc_06987a951236`), all
generations in the window:

| capability | n | `usage.total == 0` | no `tokens_per_second` |
|---|---|---|---|
| development.develop | 10 | 0 | 0 |
| qa.test | 4 | 0 | 0 |
| builder.assemble | 2 | 0 | 0 |
| development.correction_repair | 1 | 0 | 0 |
| **governance.review_plan** | 2 | **2** | **2** |
| **qa.define_test_strategy** | 2 | **2** | **2** |
| **development.design_plan** | 2 | **2** | **2** |
| **strategy.frame_objective** | 2 | **2** | **2** |
| **data.research_context** | 2 | **2** | **2** |

## Where

`src/squadops/capabilities/handlers/planning/base.py:290` constructs the
`GenerationRecord` with `model`, `prompt_text`, `response_text`, `latency_ms`,
`prompt_name`, `prompt_version`, `reasoning` — and omits `prompt_tokens`,
`completion_tokens`, `total_tokens` and `tokens_per_second`.

The cycle handler does it correctly at
`src/squadops/capabilities/handlers/cycle/base.py:905`, passing all four off
`chat_response`. The two records are built ~600 lines apart from the same imports with
different field sets; nothing enforces that they agree.

## Why it matters now

The 1.7.0 Atlas A/B pre-registration (#1160) reads three of its six predictions out of
LangFuse per generation — P1 (decode rate), P4 (completion tokens by capability), P5
(`reasoning_tokens = 0` on `none` generations). For every framing capability those
fields are absent on **both** arms, so those predictions are unreadable exactly where
the two engines are currently diverging. This was found while pulling the arm-A
baseline for the merge_plan failures.

It also means every framing generation has been costed at zero tokens in LangFuse for
as long as the planning handlers have existed.

## Fix sketch

Pass the four fields from the chat response, as `cycle/base.py` does. Better: give the
two call sites one shared constructor so a third handler cannot omit them again — the
divergence is the actual defect, the missing fields are the symptom.

## Evidence

LangFuse public API, `/api/public/observations?type=GENERATION`, window
`2026-08-28T01:00:00Z`–`2026-08-28T04:30:00Z`; 27 generations, 10 zero-usage, all five
zero-usage capabilities are planning-path.

Related: the merge_plan path emits no record at all (filed separately).

### LINKED ISSUE #1172: governance.merge_plan is invisible to LangFuse — the plan-authoring service calls the LLM and never emits a GenerationRecord
## What

`governance.merge_plan` produces no LangFuse generation at all. Not a sampling or flush
gap — the path has no telemetry wired.

`src/squadops/capabilities/handlers/_plan_authoring_service.py:260` calls
`context.ports.llm.chat_stream_with_usage(messages, **chat_kwargs)`, and the file
contains no reference to `llm_observability`, `GenerationRecord`, or any record
emission. Every other LLM-calling handler builds one (`planning/base.py:290`,
`cycle/base.py:729` and `:905`, `agents/entrypoint.py:677`,
`adapters/comms/chat_executor.py:279`).

Confirmed empirically: querying LangFuse across the 1.6.6 Next.js rolls returns
`review_plan`, `frame_objective`, `design_plan`, `research_context`, `develop`,
`test`, `assemble`, `correction_repair` — and no `merge_plan`, on any roll.

## Why it matters

merge_plan is the gate that decides whether a cycle can start building: it authors
`implementation_plan.yaml` and retries up to 4 times, then falls back to static task
steps. It is also, right now, the task where the Atlas arm of the #1160 A/B fails —
three shakeouts, the last one exhausting 8 attempts across two rounds.

Diagnosing that failure required Atlas's own request dump and `docker logs` on the
lead agent, because the observability layer records nothing for this capability. On
the Ollama arm there is no equivalent dump, so the same failure would currently be
undiagnosable after the fact: agent container logs do not survive a rebuild, and the
roll record's `framing_rerolls` counts whole framing re-rolls, not manifest attempts.
Eleven arm-A Next.js cycles all record `framing_rerolls: 0` — which cannot distinguish
"first attempt clean" from "third attempt clean".

## Fix sketch

Emit a `GenerationRecord` per manifest attempt, carrying the attempt number and the
validator's rejection reason. Per-attempt is the point: a roll-up hides the repair
loop, and the repair loop is the thing worth seeing (the prompt grows ~2,000 tokens per
attempt as the failed output and validator complaint are appended).

Worth pairing with: the `emission_log` INFO lines the handler already writes are the
best per-attempt evidence that exists and they live only in container logs.

Related: #1171 (framing generations record zero usage).

### LINKED ISSUE #1173: The completion clamp is one latency-derived constant, unreconciled with the reasoning level a capability declares or the size of the document it must emit — merge_plan lands 1 attempt in 3
## What

`default_max_completion` is a single flat number (8,192 for both 27B entries) applied to
every capability, whether it declares `ReasoningLevel.NONE` or `HIGH`, and whatever size
of document it is asked to produce. Thinking tokens and output tokens come out of that
same budget, and nothing reconciles the three settings that decide the outcome.

Three knobs, none aware of the others:

| setting | where | what it decides |
|---|---|---|
| `REASONING_BY_CAPABILITY` | `capabilities/reasoning_policy.py` | how much the model thinks (`merge_plan` = `HIGH`) |
| `max_build_subtasks` | cycle config (currently 15) | how large the plan document is; scales with the PRD |
| `default_max_completion` | `llm/model_registry.py` | one flat cap over both |

And the cap was never chosen as an output-size budget. The registry comment states its
basis: *"qwen3.6:27b at ~10 t/s on Spark takes ~13 min for 8K tokens"* — a **wall-clock**
guard. A latency constraint is silently acting as a document-size constraint.

## Evidence

Measured 2026-08-29 on the production Ollama arm (`qwen3.8:27b`), replaying a real stored
`governance.merge_plan` prompt (9,244 prompt tokens) three times through the handler's own
`ImplementationPlan.from_yaml` gate:

```
replay 1   thinking ~4,500 tok + plan ~3,400 tok = ~7,900   done_reason=stop     ACCEPTED (5 tasks)
replay 2   thinking ~7,600 tok + plan   ~480 tok =  8,192   done_reason=length   REJECTED: Missing required field: summary
replay 3   thinking ~6,100 tok + plan ~2,000 tok =  8,192   done_reason=length   REJECTED: Missing required field: summary
```

A complete plan needs ~3,400 tokens of visible output. Thinking at `HIGH` costs 4,500–7,600.
The budget is 8,192. It fits only when the model happens to think on the low end — so the
capability succeeds roughly **one attempt in three**, by construction.

Nothing surfaces this. The handler retries 4 times per round over two rounds, so at p≈1/3
the chance of landing a plan is ~96%, which is why all 11 recorded Next.js cycles show
`framing_rerolls: 0`. The cost is two or three wasted 5-minute generations per cycle,
invisible because the retry eventually wins and the attempts are not recorded (#1172).

## Why it gets worse, not randomly worse

The test PRD (`group_run`) produced a 5-task plan. `max_build_subtasks` allows 15. Each task
carries a description, `expected_artifacts` and `acceptance_criteria` with typed checks —
roughly 800+ characters each. At the configured ceiling the visible document alone exceeds
8,192 tokens before the model thinks at all, so for a larger PRD `merge_plan` does not fail
one time in three; it fails every time. We have not seen this because our verification PRD
is small.

## What the fix is not

Not "raise the number". The budget should be **derived** from what the capability declares —
its reasoning level, and the size of output it is asked for — rather than a constant chosen
for latency reasons and applied to a capability that reasons hard and writes a document
whose size scales with the input. #924 already measured the reasoning swing at 13.9× (5,727
completion tokens with thinking on, 413 with it off) and the registry comment says "the dial
is the budget" — but nothing acts on that.

Also not the plan document's shape. Emitting required fields before `tasks` would let a
*truncated* plan pass validation, which blinds the gate to the failure it exists to catch.

## Scope note

This is engine-independent and unrelated to the Atlas guard investigated in #1160. Of 44
Atlas plan emissions, 41 were severed by its content-loop watchdog **below** the cap; on
Ollama the cap is the only failure mode. Raising the budget does not help Atlas (measured:
`max_tokens 16384` still produced 3,190–6,110-token emissions, all guard-cut), and fixing
the guard would not help Ollama, which never trips it.

Refs #1160, #1172.

jasonLadd:
Three defects found while diagnosing why `governance.merge_plan` would not converge
(#1160). All three are engine-independent — they affect the production Ollama arm as
much as the Atlas one — and all three are in the framing half of a cycle, which is the
half the observability layer could not see.

## 1. Generation records are built at one seam — **Closes #1171**

`GenerationRecord`'s own docstring has named `build_generation_record()` as the required
construction path since SIP-0061. **The function never existed.** Five call sites
hand-rolled the record and their field sets drifted: the planning handlers omitted
`prompt_tokens`, `completion_tokens`, `total_tokens` and `tokens_per_second` entirely,
while the cycle handlers a few hundred lines away passed all four.

Measured over the 1.6.6 Next.js rolls, every generation from a planning capability —
`frame_objective`, `design_plan`, `research_context`, `define_test_strategy`,
`review_plan` — arrived at LangFuse with usage `{0,0,0}` and no decode rate, on both
engines, while `develop`, `qa.test`, `builder.assemble` and `correction_repair` were
complete. That is three of the six Atlas A/B predictions (P1, P4, P5) unreadable for
exactly the half of a cycle where the engines diverge.

The seam now exists; every site goes through it. Token fields are read off whatever
response object the caller holds, so a caller with nothing to report says so rather than
forgetting. The 10k text cap lives at the one place that knows it — two of the five sites
had not applied it. An architecture test fails the build if a sixth handler constructs a
record directly.

## 2. Every manifest attempt is a recorded generation — **Closes #1172**

`merge_plan` had no telemetry at all: `_plan_authoring_service` called the LLM and emitted
nothing. A LangFuse query across the 1.6.6 rolls returns every other capability and no
`merge_plan`, on any roll.

The cost surfaced on 2026-08-29: diagnosing a non-converging plan loop required the
inference engine's own request dump and `docker logs` on the lead agent. Neither survives
a rebuild, and the arm without a dump would have left no evidence at all. The roll
record's `framing_rerolls` counts whole framing re-rolls, not manifest attempts, so eleven
cycles recording `0` could not distinguish "landed first try" from "landed on the eighth".

One record per attempt, not a roll-up — the repair loop is the diagnostic object.
`GenerationRecord` gains `attempt` and `outcome` (the validator's own message, or
`accepted`), forwarded to LangFuse metadata. The prompt recorded is the user turn that
produced the generation, which on a repair round is the validator's complaint rather than
the original brief. Recording is best-effort: a broken port must not lose a valid plan.

## 3. The completion budget accounts for declared thinking — **Closes #1173**

`default_max_completion` budgets **output** — the registry comment records its basis as
wall-clock, not document size. But thinking is billed against the same wire budget and
every adapter discards it. `_build_chat_kwargs` computed the clamp and the reasoning level
four lines apart and never combined them, so declaring `HIGH` silently shrank the answer.

Measured on the production arm, replaying a real `merge_plan` prompt three times through
the handler's own validator:

```
thinking ~4,500 + plan ~3,400 = ~7,900   stop     ACCEPTED
thinking ~7,600 + plan   ~480 =  8,192   length   REJECTED: Missing required field: summary
thinking ~6,100 + plan ~2,000 =  8,192   length   REJECTED: Missing required field: summary
```

**One attempt in three**, hidden by a retry loop that eventually wins — at the cost of two
or three wasted five-minute generations per cycle. And it degrades with input size rather
than randomly: at the configured ceiling of 15 tasks the document alone exceeds 8,192
before the model thinks at all, so for a larger PRD `merge_plan` fails every time. Our
verification PRD is small, which is why this never surfaced.

`thinking_headroom()` adds room for the declared level on top of the output budget, at
both resolution sites. `NONE` capabilities — the majority — are unchanged. An unknown
level returns 0 rather than guessing.

**Deliberately not fixed:** reordering the plan so required fields precede `tasks`. That
would let a *truncated* plan pass validation, blinding the gate to the failure it exists
to catch.

## Tests

Full regression suite: **8,253 passed**, ruff check + format + test-quality lint clean.

New coverage: the seam's usage extraction, absent and partial usage, the text cap, id
uniqueness; an architecture test that no record is constructed outside the seam;
per-attempt recording with verdicts, usage and reasoning travelling with it, and survival
of a failing observability port; headroom monotonicity, the NONE capabilities being
unmoved, the measured worst case being covered, and an unknown level not inventing a
budget.

Two existing clamp assertions in `test_repair_handlers.py` are updated rather than
removed: they pinned the pre-#1173 contract (bare clamp), and now assert the new one
(output clamp plus declared headroom) with the rationale recorded.

Closes #1171
Closes #1172
Closes #1173

## Consequence for #1160

The rebuild on merge becomes the frozen deploy for shakeout #4. §1.5 of the A/B
pre-registration is written separately and records the 14 serve configurations, the four
inert vendor controls, and the arm-A control run.

### DIFF
--- a/adapters/comms/chat_executor.py
+++ b/adapters/comms/chat_executor.py
@@ -14,7 +14,6 @@ from __future__ import annotations
 
 import logging
 import time
-import uuid
 from typing import TYPE_CHECKING
 
 from a2a.server.agent_execution import AgentExecutor, RequestContext
@@ -265,9 +264,9 @@ class ChatAgentExecutor(AgentExecutor):
         try:
             from squadops.telemetry.models import (
                 CorrelationContext,
-                GenerationRecord,
                 PromptLayer,
                 PromptLayerMetadata,
+                build_generation_record,
             )
 
             ctx = CorrelationContext(
@@ -276,8 +275,9 @@ class ChatAgentExecutor(AgentExecutor):
                 agent_id=self._role_id,
                 agent_role=self._role_id,
             )
-            record = GenerationRecord(
-                generation_id=str(uuid.uuid4()),
+            # usage=None: this recorder is handed text and a latency by its caller,
+            # never the response object, so there are no token counts to report here.
+            record = build_generation_record(
                 model=self._ports.llm.default_model,
                 prompt_text=prompt_text,
                 response_text=response_text,
--- a/adapters/telemetry/langfuse/adapter.py
+++ b/adapters/telemetry/langfuse/adapter.py
@@ -432,6 +432,8 @@ class LangFuseAdapter(LLMObservabilityPort):
                     "latency_ms": record.latency_ms,
                     "tokens_per_second": record.tokens_per_second,
                     "reasoning": record.reasoning,
+                    "attempt": record.attempt,
+                    "outcome": record.outcome,
                     "prompt_layer_set_id": prompt_layers.prompt_layer_set_id,
                     "prompt_layers": [
                         {
--- a/src/squadops/agents/entrypoint.py
+++ b/src/squadops/agents/entrypoint.py
@@ -641,7 +641,6 @@ class AgentRunner:
             # Generate response using the LLM
             if self.system and self.system.ports.llm:
                 import time
-                import uuid
 
                 from squadops.llm.models import LLMRequest
 
@@ -663,9 +662,9 @@ class AgentRunner:
                 if llm_obs is not None:
                     from squadops.telemetry.models import (
                         CorrelationContext,
-                        GenerationRecord,
                         PromptLayer,
                         PromptLayerMetadata,
+                        build_generation_record,
                     )
 
                     ctx = CorrelationContext(
@@ -674,12 +673,12 @@ class AgentRunner:
                         agent_id=self.agent_id,
                         agent_role=self.role,
                     )
-                    record = GenerationRecord(
-                        generation_id=str(uuid.uuid4()),
+                    record = build_generation_record(
                         model=self._llm_model,
                         prompt_text=full_prompt,
                         response_text=response_text,
                         latency_ms=latency_ms,
+                        usage=response,
                     )
                     layers = PromptLayerMetadata(
                         prompt_layer_set_id=f"{self.role}-chat",
--- a/src/squadops/capabilities/handlers/_plan_authoring_service.py
+++ b/src/squadops/capabilities/handlers/_plan_authoring_service.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 import hashlib
 import logging
 import re
+import time
 from typing import TYPE_CHECKING, Any
 
 from squadops.capabilities.handlers._plan_authoring import (
@@ -256,6 +257,7 @@ async def produce_plan(
     ]
 
     for attempt in range(1, max_attempts + 1):
+        started = time.perf_counter()
         try:
             response = await context.ports.llm.chat_stream_with_usage(messages, **chat_kwargs)
         except Exception as exc:
@@ -281,6 +283,17 @@ async def produce_plan(
 
         manifest, error_msg = _validate_manifest_candidate(
             yaml_content, min_subtasks, max_subtasks, profile_roles
+        )
+        _record_manifest_attempt(
+            context,
+            handler_name=handler_name,
+            model=chat_kwargs.get("model"),
+            prompt_text=messages[-1].content,
+            response=response,
+            latency_ms=(time.perf_counter() - started) * 1000,
+            attempt=attempt,
+            outcome="accepted" if error_msg is None else error_msg,
+            reasoning=chat_kwargs.get("reasoning"),
         )
 
         if error_msg is None and manifest is not None:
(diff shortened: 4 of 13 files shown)


jasonLadd (commented):
reviewed

PR #1175, opened 2026-08-29
docs(plan): Atlas A/B §1.5 — the third shakeout, the 14-configuration replay matrix, and the arm-A control (#1160)

jasonLadd:
The pre-registration amendment for the 2026-08-29 shakeout phase. Separate from #1174 so
the code changes stay reviewable on their own; this is the record, that is the fix.

## What it records

**§1.2's conclusion is falsified.** The startup line "content-loop watchdog ENABLED" was
not a vendor log quirk — the watchdog is armed and `--content-loop-watchdog false` is not
in force. Four documented controls were tried and none disarms it: the CLI flag, the env
var (confirmed set on the container), the per-request `repetition_detection` object the
vendor's help says "still outranks this", and `--content-loop-min-repeats`.

**The mechanism.** It fires on legitimately repetitive YAML — a `- check: / file: /
severity:` block per file, fifty of them in a 15-task plan, which is the false-positive
class the vendor's own help names. The damage is the rollback+re-steer rewinding
mid-token: `depends_on: []` → `depend0]`, `frontend_compiles` → `frontend_compil`. Those
corruptions *are* the validator's "malformed YAML".

**The matrix.** One stored `merge_plan` prompt, replayed byte-identically, scored through
`ImplementationPlan.from_yaml` — the handler's own gate, not a bare parse — across 14
configurations covering both checkpoints, both KV dtypes, two completion budgets, two
reasoning tiers, a capped thinking budget, low temperature, the vendor's recipe verbatim,
and the SSM rollback ring forced on:

**44 plan emissions · 41 guard-stopped · 0 accepted** — and all 41 were stopped *below*
the completion cap, so the budget was never the binding constraint on that arm.

**The arm-A control.** Same prompt, arm A's production request shape, same validator: **1
of 3 accepted**. The arms fail for different reasons — arm A on budget exhaustion with no
guard involved, arm B on the guard below the budget — which is why 8 retries rescue one
and not the other, and why 11 of 11 recorded cycles show `framing_rerolls: 0` while the
per-attempt rate is about one in three.

**P1 falsified on this workload.** Arm A decoded at 28.8–29.2 tok/s against arm B's 12–14,
both decode-only from each engine's own reporting. The prediction was arm B ≥ 2× arm A.

## Two corrections to the record

- The kernel audit cited in §2 precondition 3 was run against the **NVFP4** checkpoint,
  not the FP8 one the A/B has been serving. R11 closes the gap on the audited weights —
  same outcome.
- The §1 frozen-deploy row no longer lists image ids. They have been superseded three
  times, and a table naming one deploy while the box runs another is worse than a
  pointer; it now points at the set files, which the driver actually asserts.

## And how §1.2 went wrong

Its "4 of 4 clean plans" gate counted tasks in parsed YAML, while the handler requires six
named fields. A check weaker than production cleared a remedy that production then
rejected — and that had never taken effect in the first place. Every row in this matrix is
scored through the production validator for that reason.

Refs #1160 — remaining: shakeout #4 on the deploy rebuilt from #1174, the pins PR, and the
A1/B1/B2/A2 and A3/B3 rolls.

### DIFF
--- a/docs/plans/1-7-0-atlas-ab-preregistration.md
+++ b/docs/plans/1-7-0-atlas-ab-preregistration.md
@@ -20,7 +20,7 @@ configuration, stated in full**. The tuning pass that chose arm B is on #1160.
 | stacks | **Next.js+TS** (`nextjs_ts` overrides) for the paired wall-clock — the 1.6.6 arm that ran 6/6, so verdict noise does not swamp the pair; then **one FastAPI+React pair** for correction texture (Appendix C.3's sleeper metric needs a stack where corrections happen — 1.6.6 ran 4/6 there) | same |
 | order | **A1, B1, B2, A2** (Next.js), then **A3, B3** (FastAPI+React) — the first pair A-first, the second B-first, so an ordering effect shows instead of being credited to one side | |
 | shakeout | one non-counting Atlas cycle before B1 (no cycle has ever run on Atlas: the model preflight over `/v1/models`, the 32K guard, the fenced parser on Atlas output, timeouts at 2.8× decode). It is also the warm-up Appendix C.2 requires, recorded and flagged | |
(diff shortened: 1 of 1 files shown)


jasonLadd (commented):
reviewed

Issue #1178, opened 2026-08-30
The Spark has no OOM containment — systemd-oomd inactive, no earlyoom, no MemoryMax: a memory runaway becomes an unreachable box instead of one killed process


jasonLadd:
The Spark has **no** memory-pressure containment of any kind. Verified on the box
(2026-08-29, after the incident in the sibling issue):

```
systemctl is-active systemd-oomd   -> inactive
systemctl is-enabled systemd-oomd  -> not-found
which earlyoom nohang              -> (neither installed)
vm.swappiness                      -> 60
vm.overcommit_memory               -> 0     (heuristic)
```

`ollama.service` carries no `MemoryMax=`/`MemoryHigh=` either — the unit sets only `PATH` and
an `OLLAMA_HOST=0.0.0.0` override.

## Why this matters

The kernel OOM killer does **not** fire during thrash, because swap is technically still
available. So a single over-large allocation does not produce one dead process and a live
box — it produces a box that pages in and out indefinitely and answers nothing. That is
exactly what happened on 2026-08-29: 95 minutes unreachable, ~800 MB free the entire time,
`pswpin` and `pswpout` both sustained at 300–400/s, ended only by a power cycle.

With any of `systemd-oomd`, `earlyoom`, or a `MemoryMax=` on the serving units, that incident
is one killed `llama-server` and a box you can still SSH into.

This is a host-containment gap, independent of what triggered it. Any future memory runaway on
this machine — an oversized model, a leak, a runaway test — degrades the same way.

## Proposed fix

Add memory containment to the `local-spark` bootstrap profile
(`config/profiles/bootstrap/local-spark.yaml`, `scripts/bootstrap/profiles/local-spark.sh`),
and a corresponding `squadops doctor local-spark` check so the gap is visible rather than
discovered during an outage. Options, roughly in order of preference:

1. **`earlyoom`** with a conservative threshold — acts on *available* memory before thrash
   begins, which is the failure mode here; `systemd-oomd` keys on PSI and may act too late for
   a unified-memory box where the GPU allocation is invisible to it.
2. **`MemoryMax=` on `ollama.service`** — bounds the specific known-large consumer. Narrower,
   but directly targets the process that has caused this once.
3. Lower `vm.swappiness` so the box prefers reclaiming to swapping. Mitigation only; it does
   not bound anything.

Scope note: **only** the `local-spark` profile. `dev-mac` and `dev-pc` are not affected by
[...]

Issue #1177, opened 2026-08-30
The Atlas A/B rig can put both arms in the Spark's unified memory at once — arm.sh enforces exclusivity, the replay scripts bypass it, and the box livelocked for 95 minutes


jasonLadd:
The Spark's 121 GiB is **unified** — CPU and GPU share one pool — so two 27B servers cannot
coexist. On 2026-08-29 both A/B arms became resident at once and the box livelocked in swap
thrash from 17:50 to the power cycle at ~19:25 EDT (95 minutes). SSH could not complete a
login because `sshd` could not fault its own pages back in.

The box did **not** crash. It logged continuously throughout. This is a distinct failure from
the four hard halts on record (Jul 4/11/15, Aug 20) and should not be counted against them.

## What happened

| Time (EDT) | Event |
|---|---|
| 16:46:52 | Atlas container starts via `atlas_serve_nvfp4.sh` at `--gpu-memory-utilization 0.90` |
| 16:56 | Atlas finishes its last request, goes idle — stays resident holding the allocation |
| 17:42 | `replay_ollama.py` (the arm-A control, R0) last modified |
| 17:44:27 | It calls `127.0.0.1:11434` directly; Ollama loads qwen3.8:27b on top of Atlas |
| 17:50:44 | Cliff: 803 MB free, swap 65.8% → 76.8%, bidirectional paging begins |
| 17:50 → 19:25 | Livelocked, unreachable |
| ~19:25 | Manual power cycle |

## Evidence

Ollama measured the collision, reported it, and loaded anyway:

```
17:44:27 msg="system memory" total="121.6 GiB" free="7.8 GiB" free_swap="8.0 GiB"
17:44:27 msg="gpu memory" id=0 library=CUDA available="7.0 GiB" free="7.5 GiB"
17:44:28 common_params_fit_impl: cannot meet free memory target of 1936 MiB,
                                 need to reduce device memory by 28802 MiB
```

It needed ~28.8 GiB more than existed. `sar -r` / `sar -W -f /var/log/sysstat/sa29` dates the
cliff to the minute — note `pswpin` and `pswpout` both high at once, which is thrash rather
than ordinary swap-out:

```
             kbmemfree   %memused  %swpused   pswpin/s  pswpout/s
17:40:07       5612160      90.87     65.76       0.69       0.00
17:50:44        803868      89.97     76.77    1233.53    2386.94   <-- cliff
18:00:06        805508      89.96     76.65     797.85     783.70
[...]

PR #1179, opened 2026-08-30
docs(plan): the Atlas A/B returns a negative — the record, the SIP amendment, and vLLM as the second arm

### LINKED ISSUE #1160: Atlas vs Ollama A/B artifact on the Spark — the §3.6.1 record the switch is decided on (SIP Atlas P5)

jasonLadd:
Arm B produced **0 accepted plans out of 44 emissions across 14 serve configurations**. 41 were stopped by Atlas's content-loop guard, every one of them *below* the completion cap, so the budget was never the binding constraint. Four documented controls are inert against that guard, and a second detector (`simhash_semantic_loop`) has no CLI flag and no env var at all. It fires on legitimately repetitive YAML and its rollback-and-re-steer rewinds mid-token — `depends_on: []` emerging as `depend0]` — which *is* the validator's "malformed YAML".

**P1 is falsified in the reverse direction**: arm A decodes at 28.8–29.2 tok/s on the framing prompt against arm B's 12–14, both decode-only from each engine's own reporting. P2–P6 are unanswerable, not missing: arm B never clears `governance.merge_plan` to reach build, so no counted cycle exists on either arm.

Owner ruled on 2026-08-29 that the A/B stops and Atlas is not adopted.

## What's here

- **`docs/plans/1-7-0-atlas-ab-record.md`** — the §4 artifact. Written as a negative-result record rather than the per-cycle table §4 describes, because the set never opened to counting rolls. Says which predictions are answered vs unanswerable and why, carries the arm-A control, and does not let "unsupported model" stand as an escape hatch: R11 reproduced the identical outcome on the NVFP4 checkpoint the §2 precondition-3 kernel audit actually covered.
- **SIP-0106 §1.2** — a post-acceptance amendment section, per the CLAUDE.md §5a rule that a disposition deliberately not built is an amendment too. §1.2a the negative and what it does *not* change; §1.2b P6 activating with vLLM; §1.2c the Appendix C.5 trap-1 incident. `updated_at` deliberately untouched.
- **The two arm-OLLAMA verification sets** — notes claimed "the first cycle ever on this engine", never true for production Ollama, and pointed at a set that no longer runs. They now declare what a shakeout on them actually validates.
- **`docker-compose.vllm.yml`** — the vLLM arm's deploy override, on the Atlas pattern, without the auth plumbing vLLM doesn't use.

## Notes for review

The merged `AtlasAdapter` stays inert in the tree under the SIP's §4 dark-ship rule — no revert required. Everything else SIP-0106 built stands and is shipped; the seam is what made rejecting the engine cheap.

vLLM as the second arm is not a new direction: §3.5a wrote P6 as the hedge for exactly this outcome. Its gating unknown is the one that section already stated — vLLM on GB10/ARM64 is unverified, and the SIP says verify before scheduling. A feasibility probe precedes the new pre-registration, so this PR adds only the deploy override, not a pre-registration.

The set never reached counted rolls, but it produced #1171, #1172 and #1173 — all engine-independent, all fixed in #1174, and #1173 is the whole of arm A's own failure rate.

Closes #1160
Refs #1177 — remaining: the rig's arm-exclusivity guard fix itself
Refs #1178 — remaining: OOM containment on the box

https://claude.ai/code/session_01EEKMmFdD43zvr9eLyggJUe

### DIFF
--- a/docker-compose.vllm.yml
+++ b/docker-compose.vllm.yml
@@ -1,0 +1,45 @@
+# vLLM arm (SIP-0106 P6, activated by §1.2b after the Atlas A/B returned a negative).
+# Pairs with ~/atlas/scripts/vllm_serve.sh. Usage:
+#   VLLM_MODEL=Qwen/Qwen3.8-27B docker compose -f docker-compose.yml -f docker-compose.vllm.yml up -d ...
+# No API key: vLLM serves unauthenticated on the loopback, unlike Atlas (--require-auth).
+services:
+  runtime-api:
+    environment:
+      SQUADOPS__LLM__PROVIDER: vllm
+      SQUADOPS__LLM__URL: http://host.docker.internal:8000
+      SQUADOPS__LLM__MODEL: "${VLLM_MODEL:?set VLLM_MODEL to the HF id vLLM serves}"
+  max:
+    environment:
+      SQUADOPS__LLM__PROVIDER: vllm
+      SQUADOPS__LLM__URL: http://host.docker.internal:8000
+      SQUADOPS__LLM__MODEL: "${VLLM_MODEL:?set VLLM_MODEL to the HF id vLLM serves}"
+  nat:
+    environment:
+      SQUADOPS__LLM__PROVIDER: vllm
+      SQUADOPS__LLM__URL: http://host.docker.internal:8000
+      SQUADOPS__LLM__MODEL: "${VLLM_MODEL:?set VLLM_MODEL to the HF id vLLM serves}"
+  neo:
+    environment:
+      SQUADOPS__LLM__PROVIDER: vllm
+      SQUADOPS__LLM__URL: http://host.docker.internal:8000
+      SQUADOPS__LLM__MODEL: "${VLLM_MODEL:?set VLLM_MODEL to the HF id vLLM serves}"
+  eve:
+    environment:
+      SQUADOPS__LLM__PROVIDER: vllm
+      SQUADOPS__LLM__URL: http://host.docker.internal:8000
+      SQUADOPS__LLM__MODEL: "${VLLM_MODEL:?set VLLM_MODEL to the HF id vLLM serves}"
+  data:
+    environment:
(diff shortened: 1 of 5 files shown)


jasonLadd (commented):
reviewed

